PR #17, opened 2021-06-03
feat(connector): receipts, modify stream dst

sentientwaffle:

- When routing a locally-bound stream packet, use a modified destination with the outgoing account id embedded.
- Stream receipts
  - Add redis dependency.
  - Track total received amount per stream connection (in redis).
  - Use the connection total received to generate a stream receipt.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/connector/src/services/core/controllers/stream.ts
```diff
@@ -1,39 +1,46 @@
...
+      return
+    }
+
+    const { connectionId } = moneyOrReply
+    const totalReceived = await redis.incrby(
+      streamReceivedKey(connectionId),
+      +request.prepare.amount
+    )
+    moneyOrReply.setTotalReceived(totalReceived)
+    if (totalReceived !== +request.prepare.amount) {
```
wilsonianb, line 40:
```suggestion
    if (totalReceived === +request.prepare.amount) {
```

packages/connector/src/services/core/controllers/stream.ts
```diff
@@ -1,39 +1,46 @@
-import * as assert from 'assert'
-import { StreamServer } from '@interledger/stream-receiver'
 import { isIlpReply } from 'ilp-packet'
 import { RafikiContext, RafikiMiddleware } from '../rafiki'
 
-interface StreamControllerOptions {
-  serverSecret: Buffer
-  serverAddress: string
-}
+const CONNECTION_EXPIRY = 60 * 60 // seconds
```
wilsonianb, line 4: fwiw, this is configurable in the receipt-verifier
https://github.com/coilhq/receipt-verifier#receipt_ttl

packages/connector/src/services/core/controllers/stream.ts
```diff
@@ -1,39 +1,46 @@
...
+
+    const { connectionId } = moneyOrReply
+    const totalReceived = await redis.incrby(
+      streamReceivedKey(connectionId),
+      +request.prepare.amount
+    )
+    moneyOrReply.setTotalReceived(totalReceived)
+    if (totalReceived !== +request.prepare.amount) {
+      // Set key expiry once, when the first packet arrives.
+      await redis.expire(connectionId, CONNECTION_EXPIRY)
```
kincaidoneil, line 42: Would it be more desirable set the key expiry on every packet, to reset/extend it as new packets come in?

sentientwaffle: Yeah, that would work. I figured this way would have fewer redis requests, but `expire`-ing after every increment would probably be more accurate.

packages/connector/src/services/core/controllers/stream.ts
```diff
@@ -1,39 +1,46 @@
...
+    const moneyOrReply = streamServer.createReply(request.prepare)
+    if (isIlpReply(moneyOrReply)) {
+      response.reply = moneyOrReply
+      return
+    }
+
+    const { connectionId } = moneyOrReply
+    const totalReceived = await redis.incrby(
+      streamReceivedKey(connectionId),
+      +request.prepare.amount
```
kincaidoneil, line 37: Should this be capped to `MAX_SAFE_INTEGER` so it's not coerced to a greater amount? Or, should we pass it as a string and overwrite the types?

I tested that `ioredis` _does_ support strings here, but the types don't reflect that. There's also a `stringNumbers` config option that can be passed to the constructor to return numbers as strings, but I'm guessing that probably also conflicts with the types.

sentientwaffle: Or the connector could enforce a max packet amount of `MAX_SAFE_INTEGER`.

packages/connector/src/services/core/controllers/stream.ts
kincaidoneil: Or pipeline it with the `incrby`?

sentientwaffle: I'm not sure what you mean by pipeline -- `incrby` doesn't have an extra `expiry parameter like some commands.

kincaidoneil: I think it'd look something like this:
```js
redis.pipeline()
  .incrby(...)
  .expire(...)
  .exec()
```
So ioredis sends both commands in a single batch instead of multiple network calls: https://github.com/luin/ioredis#pipelining

sentientwaffle: Interesting, I didn't know about that. :+1:

packages/connector/src/services/core/controllers/stream.ts
sentientwaffle: But the connector doesn't know what the receipt verifier has set for the ttl, right?

wilsonianb: Right.
I was confusing myself thinking ahead to rafiki also having an receipt verifier.
Not configuring the expiry and extending it for each new packet sounds good to me.

packages/connector/src/services/core/controllers/stream.ts
sentientwaffle: I ended up just using separate commands -- that link suggests pipelining for batches of >5 commands. Also, there's an option called `enableAutoPipelining` that we could enable eventually,  so I don't think its worth explicitly pipelining commands in the code.

packages/connector/src/connector.ts
```diff
@@ -89,7 +89,7 @@ export const gracefulShutdown = async (): Promise<void> => {
   if (server) {
     return new Promise((resolve, reject): void => {
       server.close((err?: Error) => {
-        redis.disconnect()
+        redis.quit()
```
wilsonianb, line 92: `quit` returns a promise that we might want to wait to resolve

wilsonianb (dismissed):

packages/connector/src/services/core/controllers/stream.ts
kincaidoneil: `enableAutoPipelining` -- possibly, but it looks like this might require calling all the functions in the same tick which might not work well with async/await.

It still seems like this is adding unnecessary latency. Could you at least `Promise.all` the `incrby` and `expire` calls?

And could you use the `stringNumbers` config option so `incrby` returns a string instead of requiring a sequential `get` call?

sentientwaffle: https://github.com/luin/ioredis/blob/abd9a82433ad67b91a4bddb45aff8da2e20d45e8/API.md says it batched by `event loop iteration` -- I don't think that's the same thing as a tick. Either way, I suspect it's a premature optimization.

`Promise.all` could lead to the `expire` being processed before `incrby`. If that happens on the first packet, no expiry will be set. What if it just didn't `await` the `expire` command?

:+1: for `stringNumbers`, I'll add that.

kincaidoneil: Oh, good point. Then this seems like a good use case for a transaction using `multi`? Then Redis ensures they're executed sequentially, and we don't need to wait for the response before issuing the next command

packages/connector/src/services/core/controllers/stream.ts
```diff
@@ -33,15 +33,18 @@ export function createStreamController(): RafikiMiddleware {
...
+    // This ensures that precision isn't lost when dealing with integers larger than MAX_SAFE_INTEGER.
+    const [[err, totalReceived], [err2]] = await redis
+      .multi()
+      .incrby(
+        connectionKey,
+        (request.prepare.amount.toString() as unknown) as number
+      )
+      .expire(connectionKey, CONNECTION_EXPIRY)
+      .exec()
+    if (err) throw err
```
kincaidoneil, line 46: I don't like throwing unexpected, untyped errors, especially for a non-critical function like signing STREAM receipts. (I'm going to open an issue about replacing errors with returning Rejects elsewhere, too. It seems like a source of bugs: if a downstream middleware throws, do upstream ones undo their side effects, such as balance updates, timers, etc.? What if those middlewares need to do things after downstream errors?)

I think these are typically syntax/invalid argument errors -- which shouldn't occur here. One _possible_ case is an overflow error if the connection key hits the i64 max, but that either represents a misconfiguration or something nefarious.

I'd propose either: don't call `setTotalReceived` and still fulfill the packet, or gracefully close the STREAM connection using one of the decline functions. (So, ignore the Redis errors are maybe just confirm `totalReceived` is an integer/non-null?)

packages/connector/src/services/core/middleware/stream-address.ts
```diff
@@ -6,16 +6,8 @@ export function createStreamAddressMiddleware() {
...
-    if (!accountId) {
-      await next()
-      return
-    }
-
     // To preserve sender privacy, the accountId wasn't included in the original destination address.
-    const segments = destination.split('.')
-    segments.splice(-1, 0, accountId)
-    state.streamDestination = segments.join('.')
+    state.streamDestination = streamServer.decodePaymentTag(destination)
```
kincaidoneil, line 10: Alot cleaner 👍 (still not 100% sure how this state should be shared between the middlewares, but looks good for now)

packages/connector/src/services/core/controllers/stream.ts
sentientwaffle: :+1: I like `don't call setTotalReceived and still fulfill the packet`, combined with logging the error.

That said, in an `async` function isn't `throw foo` effectively `return Promise.reject(foo)` anyway? Then the rejection get turned back into an exception by the caller's `await` -- and so on, until eventually the error-handler middleware turns it into an error response. Not a great system, but I don't see how returning rejects is different.

kincaidoneil (approved):

packages/connector/src/services/core/controllers/stream.ts
kincaidoneil: > That said, in an async function isn't throw foo effectively return Promise.reject(foo) anyway? Then the rejection get turned back into an exception by the caller's await -- and so on, until eventually the error-handler middleware turns it into an error response. Not a great system, but I don't see how returning rejects is different.

Yes, I think this is undesirable.

I'm suggesting Fulfills & Rejects would be passed as return values, rather than via mutable context: so middlewares and `next` calls return `Promise<IlpFulfill | IlpReject>`. This enforces that middlewares must call `next` and/or return, and if a middleware has logic dealing with an ILP reply, it handles all Reject cases. While middlewares could currently attach a `catch` handler when they call `next`, handling Promise rejections isn't enforced by TS, and there's no type checking on the rejected values.

Additionally:
- The middleware encountering the error has the most context for which ILP Reject should be returned, so it's preferable for it to decide over a generic error
- Throwing custom ILP errors adds indirection & generates a call stack, whereas returning a specific Reject directly  doesn't have that overhead
- Some middlewares might want to map one Reject to another Reject, or apply side effects on a particular type of Reject

sentientwaffle: Ah, I see what you mean. I was conflating Promise rejects with ILP rejects.
I agree, that's a much better plan. It also makes the real unexpected thrown errors stand out much better. :+1:

PR #18, opened 2021-06-08
feat(connector): implement accounts service with tigerbeetle

wilsonianb:

- implement accounts service with [tigerbeetle](https://github.com/coilhq/tigerbeetle)

## Context

~**NOTE**: Pending https://github.com/actions/virtual-environments/pull/3533, GitHub Actions is unable to run tests against tigerbeetle. Follow these instructions to test locally:
https://github.com/coilhq/rafiki/tree/bw-tigerbeetle/packages/connector#local-development~

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,723 @@
...
+      .select('assetCode', 'assetScale', 'balanceId')
+      .throwIfNotFound()
+    if (!sourceAccount || !destinationAccount) {
+      throw new UnknownAccountError()
+    }
+
+    const transfers: ClientTransfer[] = []
+
+    const flags = callback ? 0 | TransferFlags.two_phase_commit : 0
+    const timeout = callback ? BigInt(1000000000) : BigInt(0)
```
sentientwaffle, line 592:
```suggestion
    const timeout = callback ? BigInt(1e9) : BigInt(0)
```

To make it clear that this is 1 second (= 1e9 nanoseconds).

packages/connector/src/services/core/services/accounts.ts
```diff
@@ -27,17 +37,13 @@ export interface Transaction {
 export interface IlpAccount {
   accountId: string
   parentAccountId?: string
-  disabled: boolean // you can fetch config of disabled account but it will not process packets
+  disabled?: boolean // you can fetch config of disabled account but it will not process packets
```
sentientwaffle, line 40: Why is this optional? The migration defines it as `notNullable`.

packages/connector/src/services/core/test/controllers/ildcp-controller.test.ts
```diff
@@ -61,7 +61,7 @@ describe('ILDCP Controller', function () {
     )
   })
 
-  test('returns an ildcp response if incoming account is not a peer', async () => {
+  test.skip('returns an ildcp response if incoming account is not a peer', async () => {
```
sentientwaffle, line 64: Why skip?

packages/connector/src/services/core/services/accounts.ts
```diff
@@ -4,20 +4,30 @@ export interface AccountsService {
...
+export type Transfer = {
   sourceAccountId: string
   destinationAccountId: string
-  callback: (trx: Transaction) => Promise<void>
-}
+
+  sourceAmount?: bigint
+  destinationAmount?: bigint
+
+  callback?: (trx: Transaction) => Promise<void>
```
sentientwaffle, line 22: Why is this optional?

packages/connector/src/services/core/test/controllers/ildcp-controller.test.ts
wilsonianb: The ILDCP controller currently doesn't distinguish a peer account. It just always gets an address from `ctx.services.accounts.getAddress`.
https://github.com/coilhq/rafiki/pull/18/files/f60e5412527fe7db202f2c87bd39661fad31cddf#diff-07bf7f9a9ab11803e477e436f9379f39032dd932b4225464ea0f198bd8d096a9L20-R21
Should it compare with the server's ILP address?

packages/connector/src/services/core/services/accounts.ts
wilsonianb: I think I did so `disabled` would be optional with in `createAccount` (with `.defaultTo(false)` in the migration).
I can either just make in optional in `CreateOptions` below or require it to be specified when creating an account.

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,723 @@
...
+      )
+      if (found) {
+        const account = await Account.query().findById(found[0])
+        if (account) {
+          return toIlpAccount(account)
+        }
+      }
+      return null
+    } catch {
+      return null
```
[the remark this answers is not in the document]
wilsonianb: :thinking: Probably just anything `Account.query()` might throw:
https://vincit.github.io/objection.js/recipes/error-handling.html#error-handling
(Not `NotFoundError` since I'm not doing any `.throwIfNotFound` here)
Would it be better to catch, log, and return null or not catch at all?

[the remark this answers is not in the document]
sentientwaffle: I'm leaning towards not catching at all -- that way the caller can decide whether or not to report the error.

packages/connector/src/services/accounts/services/accounts.ts
sentientwaffle: I don't understand what the timeout here is for. Is it unrelated to the prepare's expiry?

packages/connector/src/services/core/test/controllers/ildcp-controller.test.ts
sentientwaffle: Ah, you're right; skipping is fine.

packages/connector/src/services/core/services/accounts.ts
sentientwaffle: > make it optional in `CreateOptions`

:+1:

packages/connector/src/services/accounts/errors/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class InvalidAssetError extends Error {
+  constructor(code: string, scale: number) {
+    super('Invalid asset. code=' + code + ' scale=' + scale)
+    this.name = 'InvalidAssetError'
+  }
+}
+
+export class InvalidAmountError extends Error {
+  constructor() {
```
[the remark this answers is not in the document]
sentientwaffle: No, probably not

packages/connector/src/services/accounts/services/accounts.ts
kincaidoneil: I'm not sure we should use TigerBeetle's timeout functionality at all, and instead use the balance middleware to rollback the transfer on ILP Rejects - which would roll it back when the connector expires the ILP Prepare.

I have questions about the safety of using TigerBeetle for this:
- If TigerBeetle times out transfers, what assumptions does that make about clock synchronization between each connector instance and the TigerBeetle instance, or latency between connector and TigerBeetle?
- How does the connector know if TigerBeetle times out a transfer?
- What happens if TigerBeetle automatically times out the transfer, but a Fulfill is received before the connector expired that ILP Prepare?

wilsonianb: Tigerbeetle's timeout is `required for two phase commit`
https://github.com/coilhq/tigerbeetle/blob/main/docs/DESIGN.md#events
I agree we should have the connector handle expiration instead of tigerbeetle, in which case the value could be set to max uint64? Even then we probably need to handle the connector attempting to fulfill or reject a transfer that tigerbeetle has already expired.

jorangreef: I disagree that the connector should handle expiration instead of TigerBeetle, since this would be making the connector the *system of record*, and moving critical invariants outside TigerBeetle's atomic state machine, inviting cache incoherency.

For example, if the invariants around expiration were to be moved outside the system of record, then it's possible that a two-phase commit would never timeout and never release the associated resources. That's also why timeouts are required by TigerBeetle, so that they can't accidentally be left off of a two-phase commit transfer.

This is a similar problem to locking because the two-phase commit ties up liquidity in the first-phase where resources are acquired. If the connector processes failed, or failed to agree (some may trigger the timeout, some may allow the Fulfill through to TigerBeetle... because of duplicate network packets coming in from outside the connector being routed to different processes), or were not fault-tolerant to the same degree as TigerBeetle, then the locks would never be released, or there would be disagreement as to whether they were. We would then need to consult the system of record to see what the final outcome was, but this only brings us back to the beginning. In other words, the connector processes would need to duplicate the same degree of fault-tolerance with Paxos for distributed consensus if they want to make the same guarantee that timeouts will always be activated at some point.

For these reasons, I believe that the timeout should be seen as critical data and should therefore be managed atomically within the system of record.

But correctness and safety aside, I also don't see an advantage to making the connector do more work than it needs to do. Caching millions of objects in JavaScript will put strain on the GC, and will impact the system as a whole (thinking of the system as a whole as `connector + database`), unless everything is tracked using flat buffers only, which would be necessary at high throughput to avoid pointer chasing. One would also need to do timer wheels to batch timeouts into windows and be careful about not creating thousands of closures. There's a real cost to doing this outside TigerBeetle, whereas TigerBeetle already does static allocation and can do this more efficiently with less cache misses through more control over the memory layout.

>What happens if TigerBeetle automatically times out the transfer, but a Fulfill is received before the connector expired that ILP Prepare?

The system of record provides distributed strict serializability and atomicity exactly for these kinds of decisions, so it's the perfect place to prevent this kind of TOCTOU. In fact, the [Wikipedia article on TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) has a "see also" reference to Linearizability right at the end, which is again why TigerBeetle has to have Paxos to do this right. **So long as the connector does not try to duplicate any atomic invariants outside of the state machine, this question disappears. With the system of record handling the timeout invariant, all you need to do is send in the received Fulfill, and you'll get a commit or an error, atomically** (and see also below where the follow up question is addressed, where the Fulfill is never received). There's no need to make the Connector stateful where it doesn't need to be, and again, caching timeouts outside of the system of the record not only adds expense but makes the system as a whole less safe.

> If TigerBeetle times out transfers, what assumptions does that make about clock synchronization between each connector instance and the TigerBeetle instance, or latency between connector and TigerBeetle?

No assumptions about clock sync at all between connectors and TigerBeetle (this was one of the things we really set out to solve in the beginning), because all timeouts are relative to the timestamp at which the transfer is created within the state machine, subject to the leader's wall clock. All leaders' wall clocks within TigerBeetle are always monotonically increasing, even if the system time strobes or there's DST or leap smears etc. The worst that can happen is that a new leader can jump forward in time and expire the timeout sooner, but there's no dependency at all on any other external system outside the system of record.

And if the connector does want to express that some kind of time has already elapsed before the transfer is created within the system of record, or that the monitored EWMA of RTT has suddenly drastically jumped between the connector and TigerBeetle, then we can reflect that simply by decreasing the `timeout` quantity passed in on the transfer. The `timeout` simply means: expire this `N` nanoseconds after the nanosecond timestamp at which it was recorded within the system of record.

> How does the connector know if TigerBeetle times out a transfer?

This is probably not so much a question of safety, but of interface. What would be the best interface for TigerBeetle to provide? At present, we are thinking something along the lines of a periodic poll to get the next batch of expired transfers since a reference token (cryptographic low water mark linked to the hash chain for idempotency), with pull rather than push for automatic flow control and in line with reactive streams.

kincaidoneil: > So long as the connector does not try to duplicate any atomic invariants outside of the state machine, this question disappears. With the system of record handling the timeout invariant, all you need to do is send in the received Fulfill, and you'll get a commit or an error, atomically

> since this would be making the connector the system of record, and moving critical invariants outside TigerBeetle's atomic state machine, inviting cache incoherency.

There are already many "critical" invariants outside TigerBeetle. The difference is, these invariants don't require the _manipulation_ of any global state, merely local state of the instance servicing the request:

- Enforcing a maximum packet amount, so a transfer for a particular account is limited to a particular value
- Correlating an ILP Fulfill with its ILP Prepare; instructing TigerBeetle to commit balance updates
- Correlating an ILP Reject with its ILP Prepare; instructing TigerBeetle to rollback balance updates
- SHA-256 condition/fulfillment verification (right now, yes)
- Enforcing the expirations are not too far in the future
- Enforcing that there's sufficient time left before the expiration for the connector to return the Fulfill to the previous node, and account for this - or that the packet has not already expired

Timeouts, in my view, fall in this same bucket.

> If the connector processes failed, or failed to agree (some may trigger the timeout, some may allow the Fulfill through to TigerBeetle... because of duplicate network packets coming in from outside the connector being routed to different processes)

For a pending ILP Prepare request, there's no shared state between connector instances (other than the hold on an account's balance, of course!). Only one instance handles the ILP Prepare, and that same instances either applies the Fulfill or expires the packet.

This is what I see as the role of TigerBeetle for Rafiki: super fast manipulation of that minimal shared state to mitigate contention _for in-flight requests_. Timing out a request does not depend on global state, so in my model, this shouldn't be handled by TigerBeetle.

> For example, if the invariants around expiration were to be moved outside the system of record, then it's possible that a two-phase commit would never timeout and never release the associated resources. That's also why timeouts are required by TigerBeetle, so that they can't accidentally be left off of a two-phase commit transfer.

Yes, but the current architecture of Rafiki does not persist pending ILP Prepare packets (we can talk about whether it should or not, but this a major change). So, the error case in which Rafiki crashes during a pending ILP Prepare is already pretty bad, since it may have missed an ILP Fulfill packet. Meaning, the upstream counterparty fulfilled the request and thinks the downstream nodes owes it those funds, but this is not reflected in the downstream counterparty's books. This balance inconsistency then needs to be manually reconciled in some way: either the downstream node agrees to owe it the funds, or the liability is removed.

Since these inconsistencies are possible in this architecture & require manual reconciliation, in the failure case where Rafiki crashes, I don't know if it's desirable for the transfer to be automatically rolled back. Possibly, but we need to talk about it within the context of this failure case.

> (and see also below where the follow up question is addressed, where the Fulfill is never received)

Yes...?

> But correctness and safety aside, I also don't see an advantage to making the connector do more work than it needs to do. Caching millions of objects in JavaScript will put strain on the GC, and will impact the system as a whole (thinking of the system as a whole as connector + database), unless everything is tracked using flat buffers only, which would be necessary at high throughput to avoid pointer chasing. One would also need to do timer wheels to batch timeouts into windows and be careful about not creating thousands of closures. There's a real cost to doing this outside TigerBeetle, whereas TigerBeetle already does static allocation and can do this more efficiently with less cache misses through more control over the memory layout.

We're probably already doing this? I don't understand this point: we're not writing a connector in C. You're probably correct that it would be faster for any number of reasons, but this a completely different architecture.

I'm highly skeptical that timers in Node, as compared to all the other inefficiencies of this architecture, is going to meaningfully slow down packet processing or limit throughput.

> The worst that can happen is that a new leader can jump forward in time and expire the timeout sooner, but there's no dependency at all on any other external system outside the system of record.

This is not true for ILP nodes: two Interledger peers do need (somewhat) synchronized clocks. If TigerBeetle handles expirations, this puts a dependency on TigerBeetle. (Nodes reduce the expiration time by some duration to try to account for upper bound clock skew and latency between them.)

Expiring the timestamp sooner _is_ undesirable, since it could create the dropped Fulfill scenario described earlier, when one node thinks it got it back in time, but the other doesn't. (Being lazier about expiring timestamps is probably safer for consistency between peers). However, we also don't want to expire too late, either, since that could delay applications sending payments from failing or recovering faster, affecting the user experience.

packages/connector/src/services/accounts/app.ts
```diff
@@ -39,10 +39,11 @@ export class App {
     private container: IocContract<AppServices>,
     private config: typeof Config,
     private logger: Logger,
-    private closeEmitter: EventEmitter // tigerbeetle: Client
+    private closeEmitter: EventEmitter,
+    tigerbeetle: Client
```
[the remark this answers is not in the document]
wilsonianb: Actually, all of the arguments except `tigerbeetle` are [property parameters](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties), which I don't think can be destructured.
I think I'd prefer to keep it as is vs declaring and defining those properties.

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,723 @@
...
+      .findById(accountId)
+      .select('assetCode', 'assetScale', 'balanceId')
+      .throwIfNotFound()
+    await this.createTransfers([
+      {
+        sourceBalanceId: toSettlementId(assetCode, assetScale),
+        destinationBalanceId: uuidToBigInt(balanceId),
+        amount
+      }
+    ])
```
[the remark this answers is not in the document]
wilsonianb: :upside_down_face: I forgot to add that.
I'll plan on doing it in a subsequent pr.

packages/connector/src/services/core/test/mocks/accounts-service.ts
```diff
@@ -62,19 +68,33 @@ export class MockAccountsService implements AccountsService {
...
+    } else {
+      return this.serverIlpAddress + '.' + accountId
+    }
   }
 
-  private find(predicate: (account: IlpAccount) => boolean): IlpAccount | null {
+  private find(
+    predicate: (account: MockIlpAccount) => boolean
+  ): IlpAccount | null {
     for (const [, account] of this.accounts) {
```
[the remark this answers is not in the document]
wilsonianb: Does that work with a `Map`?

wilsonianb (commented):
https://github.com/coilhq/rafiki/pull/18/commits/c958d8278ff5c5f0caf73adc4235b8eb8e3884ad addresses https://github.com/coilhq/rafiki/pull/18#discussion_r648498235
but I'm planning on changing it so that (expected) errors are returned instead of thrown as discussed https://github.com/coilhq/rafiki/pull/18#discussion_r649305823

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,723 @@
...
+          }
+        ],
+        account.asset.scale
+      )
+      const accountRow = await Account.query().insertAndFetch({
+        id: account.accountId,
+        disabled: account.disabled,
+        assetCode: account.asset.code,
+        assetScale: account.asset.scale,
+        balanceId,
```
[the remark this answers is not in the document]
wilsonianb: I might give this a try
https://defn.io/2020/04/05/postgres-bytea-to-uuid/

packages/connector/src/services/accounts/services/accounts.ts
adrianhopebailie: We're having these same discussions in the Mojaloop design sessions 😄 

Looking at the list of invariants you posted above @kincaidoneil, I don't think these are invariants of the protocol, these are business rules, evaluated against the prepare:
- Enforcing a maximum packet amount, so a transfer for a particular account is limited to a particular value
- Enforcing the expirations are not too far in the future
- Enforcing that there's sufficient time left before the expiration for the connector to return the Fulfill to the previous node, and account for this - or that the packet has not already expired

These are always evaluated against the prepare, not the fulfil, so they can be evaluated before the connector even persists the transfer in TigerBeetle. The things we care about for atomicity are things we evaluate when we receive the fulfil and are ideally not tracked in multiple places or achieving atomicity gets very complex.

These are artefacts of the wire protocol used for the bilateral communication.
- Correlating an ILP Fulfill with its ILP Prepare
- Correlating an ILP Reject with its ILP Prepare

This correlation itself is not an essential part of ensuring atomicity. 

Finding the ILP Prepare that correlates to the ILP Fulfil or Reject is a precursor to processing the fulfil/reject so that the system doing the processing has all the data it needs to accept or reject the fulfil. 

There should be only two ways a fulfil can be rejected by a connector:
 1. the fulfilment is invalid
 2. the fulfil is received after the expiry

As such, the data required from the original prepare is: 
- the timestamp it was captured and the timeout specified, to decide if the fulfil has been received in time
- the condition, to check if the fulfilment is correct

Again, doing these checks outside the system of record might feel like a good separation of concerns or an appropriate application of single responsibility principle but all it does is add complexity by requiring the system to manage consistency between the sub-system evaluating timeouts, the sub-system evaluating conditions/fulfilments and the sub-system that is the system of record for transfers.

SIDE NOTE: Unfortunately, ILP over HTTP depends on HTTP's request/response matching so responses don't contain the id of the request and we can't easily persist a prepare so that the subsequent fulfil/reject could be handled by a different process. If the HTTP server service dies and is behind an ingress/load balancer that hides this from the client then it's possible that a client sending a fulfilment will be unaware the fulfilment has been delivered to a peer that can't process it.

This last one IS an invariant of ILP and SHOULD be processed atomically with the evaluation of the timeout.
- SHA-256 condition/fulfillment verification

It is a perfectly valid design to move both condition/fulfilment checking and timeout checks outside of TigerBeetle. In effect you are using TigerBeetle as a simple ledger.

My question would be WHY?

TigerBeetle is explicitly designed to solve for this challenge of ensuring atomicity when handling two-phase commits and explicitly for commits thats have a condition/fulfil check requirement. If you don't want to use that feature then why even use two phase transfers in TigerBeetle, just use simple auto-committing transfers.

TigerBeetle has the built-in ability to handle all of ILP's invariants when processing a commit signal, timeout and fulfil validation, in such a way that it can guarantee atomicity without needing to write any code outside TigerBeetle to manage consistency.

This feels like intentionally not using the transactional capabilities of a database and trying to replicate that in application code.

packages/connector/src/services/accounts/models/account.ts
```diff
@@ -0,0 +1,53 @@
...
+        to: 'tokens.accountId'
+      }
+    }
+  }
+
+  public disabled!: boolean
+
+  public assetCode!: string
+  public assetScale!: number
+  public balanceId!: string
```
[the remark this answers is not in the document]
sentientwaffle: @wilsonianb does it compile if the fields are marked `readonly`?

packages/connector/src/services/accounts/services/accounts.ts
kincaidoneil: > This feels like intentionally not using the transactional capabilities of a database and trying to replicate that in application code.

No. I don't view the timeout and condition verification are transactional; they're invariants no different than how you categorized the others from the Prepare leg.

Why not?

1. They're stateless (not reliant on global state, only the request state).
2. The Prepare invariants must be checked in advance of balance mutations involving that ILP Prepare. I don't see how this is any different with the Fulfill: the condition and timeout are invariants that must be checked in advance of balance mutations involving that Fulfill.
3. ILP Prepare expirations are relativistic. Joran emphasized TigerBeetle's monotonically increasing sequence numbers, but that's not necessary here, since there's no need for a global ordering of expiration events. And, in this context, I'm not even sure this property is desirable: an operator may want to adjust its system clock backward, *if it lowers clock skew with its peers*. Skipping ahead is undesirable if it increases clock skew and triggers expirations sooner than they should occur.

(Aside: you made a distinction between "protocol invariants" and "business rules," but I don't agree. What if two peers agreed to disable expiration checking, or conditional execution? What if they implemented a higher-layer protocol on top of the condition? What if peers deduplicate conditions? How do IL-DCP, CCP, or peering protocols fit into this? e.g. at the end of the day, these all feel like "business rules" decided among peers — with some simply more universally enforced than others.)

> Again, doing these checks outside the system of record might feel like a good separation of concerns or an appropriate application of single responsibility principle but all it does is add complexity by requiring the system to manage consistency between the sub-system evaluating timeouts, the sub-system evaluating conditions/fulfilments and the sub-system that is the system of record for transfers.

These "subsystems" are middlewares, with a precisely defined order of operations enforced by the middleware pipeline — which are defined in the same codebase and operated in the same process. I view *this proposal* as adding complexity, by splitting some — but not all — of these invariants across distinct systems and codebases.

I think we should more precisely define where we need atomicity and where we don't. These checks do **not** require a transactional guarantee:

1. Verifying if the fulfillment is the preimage of the ILP Prepare's execution condition.
2. Checking if the current time exceeds the ILP Prepare's expiration timestamp.

Balance mutations may proceed these checks, but they don't require atomic transactions: any other state changes, events, or time can elapse in between the check and its corresponding balance mutation. Since neither checks depends on global state, there's no risk of any TOCTOU bug.

Expiring ILP Prepares does involve an atomic invariant, but it's more nuanced than the balance rollback itself: it's that only *one message* can be returned to the previous peer. That is, if you already returned a Reject (such as after an expiration), you can no longer return a Fulfill. This may seem trivial, but I discovered an [expiry bug with mini-accounts and ilp-plugin-xrp-asym-server](https://github.com/interledgerjs/ilp-plugin-xrp-asym-server/pull/66#issuecomment-442931765) that boiled down to, these business rules were split across multiple components — and it wasn't clear how they interacted or the responsibility of each component. (This is the gist of my apprehension against splitting these across the connector and TigerBeetle.)

However, I'd expect this shouldn't be the case with TigerBeetle, since it certainly enforces only a single outcome for a transfer, and will only ACK one of those outcomes.

But, it's also important to point out that this isn't an argument *to* use TigerBeetle, either — it's perfectly safe to have this logic in a connector middleware which blocks fulfilling packets after a Prepare times out, or after a Reject is returned. This is how every existing implementation works. (And, I'd argue it's easier to audit this when it's all within a single codebase).

> why even use two phase transfers in TigerBeetle, just use simple auto-committing transfers.

You could. But then the application has to create additional accounts for in-flight transfers; it's convenient for TigerBeetle to offer this natively. (But on second thought that might be nice simplification...)

kincaidoneil: @jorangreef If TigerBeetle handles expirations, would it atomically rollback linked transfers?

jorangreef: Thanks @kincaidoneil for this epic dialogue!

Just to check that I understand correctly, the context here is the `Transfer.timeout` governing a two-phase commit transfer as persisted in TigerBeetle, where the prepare is sent in with `create_transfers`, and where the fulfill is sent in with `commit_transfers`, or the transfer is expired and rolled back?

If this is the context, then the question under discussion is whether TigerBeetle should roll back these two-phase commit transfers already prepared and persisted in TigerBeetle when they expire, or whether the application layer should take on the responsibility to expire these prepares by sending in a `commit_transfers` batch with `commit.flags.reject = true`?

> No. I don't view the timeout and condition verification are transactional

A two-phase commit in TigerBeetle is "a two-phase transaction" governed by the timeout. Resources are prepared or reserved or locked in the first phase, a promise or contractual offer being made. These resources are then fulfilled or committed in the second phase, a promise or contractual offer being accepted. TigerBeetle implements the two-phase commit protocol as requiring that "there is stable storage at each node with a write-ahead log, that no node crashes forever, that the data in the write-ahead log is never lost or corrupted in a crash" and that for both prepares and commits "the record is forced to stable storage".

> They're stateless (not reliant on global state, only the request state).

If we persist two-phase commit transfers in TigerBeetle by calling `create_transfers` to do the prepare to reserve liquidity (by setting `transfer.flags.two_phase_commit = true`), and if we then want to manage the expiration for these two-phase commits externally as you propose, but then we treat these timeouts as stateless, without any replicated fault tolerance for persistence, then the invariant of the timeout governing the two-phase commit transaction will be broken: the timeout will be lost when the application layer instance crashes, and the two-phase commit transaction will never roll back. Eventually, as enough stateless timeouts are lost, accounts will start to become frozen for the user, with all their liquidity tied up in `reserved` balances that will never be released.

The principle here is that we can't persist prepares in TigerBeetle without the timeout information also being persisted to stable storage. The timeout information is state, and needs to be given at least the same degree of fault tolerance as the two-phase commit transaction it governs.

> Skipping ahead is undesirable if it increases clock skew and triggers expirations sooner than they should occur.

This issue currently only applies to transfers that span leader elections, but can also be solved by estimating the relative clock offset between the TigerBeetle leader and all followers and using this as an upper bound on the leader's wall clock. When timestamping events, the leader would then bound batch event timestamps to be greater than the latest cluster commit timestamp (which we already do to solve the problem of clocks stuck in the past), but also less than the most conservative timestamp after discarding outliers (anything before the commit timestamp). We can ride on our existing cluster heartbeats to do this, and we've already been wanting to add EWMA RTT measurements for all follower network paths for some other optimizations. This would prevent any runaway leader clock from drifting the commit timestamp into the future. CockroachDB do something similar: https://github.com/cockroachdb/cockroach/blob/master/pkg/rpc/clock_offset.go

> What if two peers agreed to disable expiration checking?

Would they then not have to agree to disable two-phase commit in lieu of something else, or accept resource leaks?

> However, I'd expect this shouldn't be the case with TigerBeetle, since it certainly enforces only a single outcome for a transfer, and will only ACK one of those outcomes.

Yes, for two-phase commit transfers.

TOCTOU race conditions come into play if you only issue one-phase transfers into TigerBeetle, treating TigerBeetle as a pure ledger of one-phase journal entries only, without any two-phase commit features, while using TigerBeetle to power a two-phase commit protocol at the application layer. For example, if you wanted to decompose an external two-phase commit transaction into separate journal entries, without using any two-phase commit features in TigerBeetle, with application layer instances initiating journal entries for expiry rollback as well as journal entries for fulfills. I believe this scenario is only safe if you can guarantee that a set of one-phase transfers is mutually exclusive to all other possible concurrent one-phase transfers that might relate to the same external transaction.

> If TigerBeetle handles expirations, would it atomically rollback linked transfers?

No, the guarantee made by Linked Events applies to any combination of one-phase or two-phase transfers, within the immediate execution context of the state machine processing the batch, so that linked event errors (if any) are returned immediately.

If we were to extend this guarantee outside the immediate execution context, to span the maximum duration of the longest running two-phase transaction within a chain of linked events, then this would:

* make the commit of the batch to the state machine asynchronous, 
* make the error path for the individual events in the chain asynchronous,
* mean that one-phase transfers could not be made visible until the last two-phase commit transfer completes or rolls back, or if one-phase transfers were made visible we would have to roll them back later, but this could not be done without violating subsequent causal dependencies (e.g. a later debit that depended on the transfer providing the initial credit).

kincaidoneil: > Just to check that I understand correctly, the context here is the Transfer.timeout governing a two-phase commit transfer as persisted in TigerBeetle, where the prepare is sent in with create_transfers, and where the fulfill is sent in with commit_transfers, or the transfer is expired and rolled back?
> 
> If this is the context, then the question under discussion is whether TigerBeetle should roll back these two-phase commit transfers already prepared and persisted in TigerBeetle when they expire, or whether the application layer should take on the responsibility to expire these prepares by sending in a commit_transfers batch with commit.flags.reject = true?

Yes! The only thing I'd add to the synopsis is in both cases, the application layer will still need functionality to rollback the transfers when it receives an ILP Reject before the timeout (so the liquidity isn't left on hold and those funds can be used for retries).

jorangreef: Yes! This can already be done by sending in a batch of `commit_transfers` setting `commit.flags.reject = true` to rollback these transfers before the timeout. One remaining feature around this is a way for the application layer to query the latest batch of expired transfers since a given `timestamp` so that it can fire off expiry events elsewhere as needed.

kincaidoneil: > > If TigerBeetle handles expirations, would it atomically rollback linked transfers?
>
> No, the guarantee made by Linked Events applies to any combination of one-phase or two-phase transfers, within the immediate execution context of the state machine processing the batch, so that linked event errors (if any) are returned immediately.

I ask this because, the application can/will enforce that all linked Prepare transfers are committed as linked transfers, or rolled back as linked transfers.

Maybe a different way to ask this is, if multiple transfers have the same expiration timestamp, will TigerBeetle roll them back atomically or is there ever a possible state in which one transfer is rolled back, but another with the same timestamp is not?

> the timeout will be lost when the application layer instance crashes, and the two-phase commit transaction will never roll back. Eventually, as enough stateless timeouts are lost, accounts will start to become frozen for the user, with all their liquidity tied up in reserved balances that will never be released.

I addressed this in my first response. Could you respond to this:

_Yes, but the current architecture of Rafiki does not persist pending ILP Prepare packets (we can talk about whether it should or not, but this a major change). So, the error case in which Rafiki crashes during a pending ILP Prepare is already pretty bad, since it may have missed an ILP Fulfill packet. Meaning, the upstream counterparty fulfilled the request and thinks the downstream nodes owes it those funds, but this is not reflected in the downstream counterparty's books. This balance inconsistency then needs to be manually reconciled in some way: either the downstream node agrees to owe it the funds, or the liability is removed._

_Since these inconsistencies are possible in this architecture & require manual reconciliation, in the failure case where Rafiki crashes, I don't know if it's desirable for the transfer to be automatically rolled back. Possibly, but we need to talk about it within the context of this failure case._

Otherwise, another option is a hybrid approach: set intentionally late expirations in TigerBeetle (add some duration) to mitigate clock synchronization requirements but still eventually rollback failed transfers, but in all cases except crashes, rely on the application to initiate the TigerBeetle rollbacks close to the actual expiration timestamp.

> This is probably not so much a question of safety, but of interface. What would be the best interface for TigerBeetle to provide? At present, we are thinking something along the lines of a periodic poll to get the next batch of expired transfers since a reference token (cryptographic low water mark linked to the hash chain for idempotency), with pull rather than push for automatic flow control and in line with reactive streams.

What do you think a reasonable poll interval would be? How much latency would that create between the expiration and the application knowing of it (to return the corresponding Reject packet)?

> This issue currently only applies to transfers that span leader elections, but can also be solved by estimating the relative clock offset between the TigerBeetle leader and all followers and using this as an upper bound on the leader's wall clock. When timestamping events, the leader would then bound batch event timestamps to be greater than the latest cluster commit timestamp (which we already do to solve the problem of clocks stuck in the past), but also less than the most conservative timestamp after discarding outliers (anything before the commit timestamp). We can ride on our existing cluster heartbeats to do this, and we've already been wanting to add EWMA RTT measurements for all follower network paths for some other optimizations. This would prevent any runaway leader clock from drifting the commit timestamp into the future. CockroachDB do something similar: https://github.com/cockroachdb/cockroach/blob/master/pkg/rpc/clock_offset.go

I'm not saying these optimizations are unreasonable (e.g. the improved RTT estimation surely has other benefits), but both these synchronization changes and the client polling behavior seems like A LOT of effort for something that doesn't require it. The bottom line is, a replicated state machine isn't necessary to decide _if_ a transfer is expired, and this is all additional overhead required by that approach.

Literally, the connector is already doing this in 10 lines of code. The application code to rollback transfers is required in either case. To me, this feels way overengineered when there's still (1) no safety reason this is beneficial to do in TigerBeetle, (2) it's really hard to justify any performance reason given the existing inefficiencies of this architecture, and (3) the benefits of maintaining these invariants in the same codebase makes the much easier to reason about.

It really only comes down to connector crashes and taking liquidity off hold -- what is the desirable behavior?

sharafian (changes requested):

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,784 @@
...
+
+  public async getAccountByDestinationAddress(
+    destinationAddress: string
+  ): Promise<IlpAccount | null> {
+    const account = await Account.query()
+      // new RegExp('^' + staticIlpAddress + '($|\\.)'))
+      .where(
+        raw('?', [destinationAddress]),
+        'like',
+        raw("?? || '%'", ['staticIlpAddress'])
```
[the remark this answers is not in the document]
wilsonianb: I think so. The query currently returns `undefined`
https://github.com/coilhq/rafiki/blob/3778c07cdb400025527e45ebde9995f716e97e09/packages/connector/src/services/accounts/tests/services/accounts.test.ts#L881

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,784 @@
...
-          ])
-          if (res.length) {
-            // throw
+    if (peerAddress) {
+      const account = await Account.query().findById(peerAddress.accountId)
+      if (account) {
+        return toIlpAccount(account)
+      }
+    }
+    if (this.config.ilpAddress) {
```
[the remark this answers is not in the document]
wilsonianb: fwiw, the connector throws if `ilpAddress` isn't defined
https://github.com/coilhq/rafiki/blob/main/packages/connector/src/connector.ts#L50
but I'll add the check to `getAddress`

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,784 @@
...
+      staticIlpAddress: accountRow.staticIlpAddress
+    }
+  }
+  return account
+}
+
+export type UpdateIlpAccountOptions = Omit<
+  CreateOptions,
+  // 'asset' | 'parentAccountId'
+  'asset'
```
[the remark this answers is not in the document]
sharafian: oh duh, that makes sense. having it in the object is fine

packages/connector/src/services/accounts/utils/balance-id.ts
```diff
@@ -0,0 +1,50 @@
+const LIQUIDITY_ID_PREFIX = '6c69717569646974793a' // 'liquidity:'
+const SETTLEMENT_ID_PREFIX = '736574746c653a' // 'settle:'
+
+const REAL_ID_CODE = '7265616c3a' // 'real:'
+// const CREDIT_ID_CODE = '637265643a' // 'cred:'
+// const LOAN_ID_CODE = '6c6f616e3a' // 'loan:'
```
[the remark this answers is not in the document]
kincaidoneil: I don't feel too strongly on this (the MAC should resolve the collision concern), but one thought is, it feels weird that these deterministic mappings are necessary to interpret TigerBeetle state, but can't be defined in any DB schema. (So, even if you have the PG schema, understand the TB model, you still need to know this logic to interpret certain state.)

> > These seem like good candidates for tigerbeetle's future querying capabilities
>
> deterministic IDs could be a lot more performant

Yeah, I don't think we'd want to add latency to query account IDs before executing the transfers (assuming that'd be a separate call), whether that's in TigerBeetle or PG. So the deterministic IDs are good there.

One idea is, since there are probably relatively few settlement & liquidity accounts, those IDs could be fetched from PG and cached locally.

packages/connector/migrations/20210517125429_create_tokens_table.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function (knex) {
+  return knex.schema.createTable('tokens', function (table) {
+    table.uuid('id').notNullable().primary()
+    table.string('token').notNullable().unique().index()
+    table.uuid('accountId').notNullable().index()
+    table.foreign('accountId').references('accounts.id').onDelete('CASCADE')
```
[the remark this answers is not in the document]
kincaidoneil: @sharafian I'll update the delete account API call for this behavior. And what do you think if hard delete was also supported by passing some flag in the request?

packages/connector/migrations/20210422194131_create_accounts_table.js
```diff
@@ -0,0 +1,34 @@
...
+
+    table.uuid('balanceId').notNullable()
+    // table.uuid('debtBalanceId').notNullable()
+    // table.uuid('trustlineBalanceId').notNullable()
+    // table.uuid('loanBalanceId').nullable()
+    // table.uuid('creditBalanceId').nullable()
+
+    // table.uuid('parentAccountId').nullable()
+
+    table.bigInteger('maxPacketAmount').notNullable()
```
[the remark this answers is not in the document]
kincaidoneil: > global default value for maxPacketAmount

How would this work? Were you thinking of per asset/scale global defaults?

packages/connector/src/services/accounts/utils/net-balance.ts
```diff
@@ -0,0 +1,7 @@
+import { Account } from 'tigerbeetle-node'
+
+export function getNetBalance(balance: Account): bigint {
```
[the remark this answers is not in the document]
kincaidoneil: Yep!

packages/connector/src/services/core/test/mocks/accounts-service.ts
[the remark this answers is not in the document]
kincaidoneil: Didn't catch that, gotcha 👍

sharafian (changes requested):

packages/backend/src/index.ts
```diff
@@ -60,7 +60,11 @@ export function initIocContainer(
       }
     })
     // node pg defaults to returning bigint as string. This ensures it parses to bigint
-    knex.client.driver.types.setTypeParser(20, 'text', BigInt)
+    knex.client.driver.types.setTypeParser(
```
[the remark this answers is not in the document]
wilsonianb: I'd prefer that happen in a separate pr.

wilsonianb (commented):
The build error should be resolved once an updated [tigerbeetle-node](https://github.com/coilhq/tigerbeetle-node/issues/12) is published.

sharafian (approved):

packages/connector/src/services/accounts/services/accounts.ts
```diff
@@ -1,117 +1,979 @@
...
+            sourceAccount.assetCode,
+            sourceAccount.assetScale
+          )
+        case CreateTransferError.exceeds_credits:
+          if (index === 1) {
+            return TransferError.InsufficientLiquidity
+          }
+          return TransferError.InsufficientBalance
+        default:
+          throw new BalanceTransferError(code)
```
kincaidoneil, line 912: What types of TigerBeetle errors does this case involve?

packages/connector/src/services/accounts/tests/helpers/asset.ts
```diff
@@ -0,0 +1,14 @@
+import { randomInt } from 'crypto'
+
+// Use unique assets as a workaround for not being able to reset
+// Tigerbeetle between tests
```
kincaidoneil, line 4: What's the blocker on resetting TigerBeetle? (or will testcontainers work resolve this?)

packages/connector/src/services/core/middleware/balance.ts
```diff
@@ -15,19 +15,20 @@ export function createBalanceMiddleware() {
...
+    await next()
 
-        if (response.fulfill) {
-          await trx.commit()
-        } else {
-          await trx.rollback()
-        }
+    if (!isTransferError(trxOrError)) {
+      if (response.fulfill) {
+        await (trxOrError as Transaction).commit()
```
kincaidoneil, line 28: Nit: can these casts be removed? I think the `!isTransferError` narrows the type to `Transaction`

kincaidoneil (approved):
Nice work 👍 Thanks for the error handling improvements.

packages/connector/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - database-data:/var/lib/postgresql/data/ # persist data even if container shuts down
+      - ./scripts/init.sh:/docker-entrypoint-initdb.d/init.sh
+    ports:
+      - "5433:5432"
+    environment:
+      POSTGRES_PASSWORD: password
+      POSTGRES_USER: postgres
+      POSTGRES_DB: testing
+  redis:
+    image: "redis"
```
sharafian, line 16: what is redis used for?

sharafian (approved):

packages/connector/docker-compose.yml
wilsonianb: STREAM receipts
https://github.com/coilhq/rafiki/pull/17

kincaidoneil (approved):

PR #19, opened 2021-06-10
feat(backend): scaffold user service and backend structure

cairin:

- Implements a base for the user service to be built upon
- Refactors the backend to follow a ddd structure.

## Context

Moves the backend project structure to follow a more domain driven development approach. 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/user/service.ts
```diff
@@ -0,0 +1,42 @@
...
+  get(id: string): Promise<User>
+  create(): Promise<User>
+}
+
+interface ServiceDependencies {
+  logger: Logger
+  knex: TransactionOrKnex
+}
+
+export async function createUserService(
```
sentientwaffle, line 17: Why not pass in a (destructured) `ServiceDependencies` instead of individual arguments?

matdehaast:
https://github.com/go-kit/kit/tree/master/examples/shipping for more context

packages/backend/src/user/service.ts
cairin: 1. It's better to be explicit - this helps with knowing what a service requires as inputs (remember this function is used to instantiate the service).
2. You don't need to structure an object when instantiating the service.
3. You often need to change/format some inputs before passing them on to children services. Thus it makes sense to structure that object within the service.

matdehaast (approved):

PR #20, opened 2021-06-11
feat(backend): spsp

sentientwaffle:

- Add SPSP endpoint.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/config/app.ts
```diff
@@ -26,6 +28,12 @@ export const Config = {
   coilApiGrpcUrl: envString('COIL_API_GRPC_URL', 'localhost:6000'),
   nonceRedisKey: envString('NONCE_REDIS_KEY', 'nonceToProject'),
 
+  accountsUrl: envString('ACCOUNTS_URL', 'http://127.0.0.1:3456'), // TODO default port?
```
sethalicious, line 31: Need a ticket for this TODO?  Or do we track TODOs differently with a lint-like script or something?

packages/backend/src/services/spsp.ts
```diff
@@ -0,0 +1,91 @@
...
+    timeout: 10_000,
+    validateStatus: (status) => status === 200 || status === 404
+  })
+  const server = new StreamServer({
+    serverSecret: config.streamSecret,
+    serverAddress: config.ilpAddress
+  })
+
+  return async function handleSPSP(ctx: AppContext): Promise<void> {
+    if (!uuid.validate(ctx.params.id) || uuid.version(ctx.params.id) !== 4) {
```
sethalicious, line 24: Should the version number be made a constant somewhere?  Will this `4` need to be updated at some point?

packages/backend/src/services/spsp.ts
```diff
@@ -0,0 +1,91 @@
...
+      ctx.throw(400, 'Failed to generate credentials: invalid user id')
+    }
+
+    if (!ctx.get('accept').includes('application/spsp4+json')) {
```
sethalicious, line 28: I see "application/spsp4+json" referenced in multiple spots.  Will this string ever change?  If so, do you think it should be moved to a constant somewhere?

sethalicious (dismissed):
Tests look awesome!  Great work!  Approval for test code.

packages/backend/src/services/spsp.ts
sentientwaffle: We always generate uuid version-4 (the "version" is really a different algorithm). But I'll remove that check. The validation here is only to ensure that the UUID is valid so that knex doesn't throw a validation error.

sethalicious: I'll defer to your judgement on that.  I'm not sure of the proper way to validate the uuid.  My only concern is having version numbers for things hidden in code versus a config file or a constants file.

sharafian (approved):

PR #21, opened 2021-06-15
feat(backend): accounts service

cairin:

- Adds `accounts` table and service.
- Creates an account and links it to a user when the user is created.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/migrations/20210615140530_add_account_id_to_users_table.js
+++ b/packages/backend/migrations/20210615140530_add_account_id_to_users_table.js
@@ -1,0 +1,11 @@
+exports.up = function (knex) {
+  return knex.schema.table('users', function (table) {
+    table.uuid('accountId').notNullable().unique()
+  })
+}
+
+exports.down = function (knex) {
+  return knex.schema.table('users', function (table) {
+    table.dropColumn('accountId')
+  })
+}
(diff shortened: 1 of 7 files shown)


packages/backend/src/account/service.ts
```diff
@@ -0,0 +1,52 @@
...
+  deps.logger.info('returns an account')
+  return Account.query(deps.knex).findById(id)
+}
+
+async function createAccount(
+  deps: ServiceDependencies,
+  scale: number,
+  currency: string
+): Promise<Account> {
+  deps.logger.info('Creates an account')
```
matdehaast, line 47: Add a TODO to say we need to call out to connector to create the underlaying account

matdehaast (approved):

matdehaast (approved):

packages/backend/migrations/20210615140530_add_account_id_to_users_table.js
```diff
@@ -0,0 +1,11 @@
+exports.up = function (knex) {
```
sentientwaffle, line 1: Migrations are great for progressively upgrading/downgrading a live database, but since there are no live Rafiki instances yet it would be simpler to just add the new fields directly to the original table setup migration.

PR #23, opened 2021-06-18
feat(backend): scaffold graphql services

cairin:

- Install an implement the Apollo graphql server.
- Install and configure codegen for graphql.
  - Allow generate on precommit.
  - Fix precommit to handle generated files.
- Refactor user, account and spsp services to use the Ioc container more effectively.
- Implements a user resolver and tests.
  - Passes the user id in the header for now until we decide how to do auth properly.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/spsp/service.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  const deps: ServiceDependencies = {
+    logger: log,
+    accountService: accountService,
+    userService: userService,
+    streamServer: streamServer
+  }
+  return {
+    GETPayEndpoint: (ctx) => getPay(deps, ctx),
+    streamServer: () => getStreamServer(deps)
```
sentientwaffle, line 40: Would it be better to make the `streamServer` a service/property attached to the container, so a getter isn't needed here? It's awkward for the SPSP service to provide access to an object that it neither creates nor modifies.

packages/backend/src/index.ts
```diff
@@ -69,6 +73,44 @@ export function initIocContainer(
...
   })
 
+  /**
+   * Add services to the container.
+   */
+  container.singleton('userService', async (deps) => {
+    const logger = await deps.use('logger')
+    const knex = await deps.use('knex')
+    const accountService = await deps.use('accountService')
+    return await createUserService({
```
sharafian, line 83: a question about our use of dependency injection: Why do we construct the dependencies in a function here and instantiate the service with those passed in, rather than passing in the deps object and allowing the service to request all its dependencies from there? That's the pattern used in reduct, which is the dependency injection library we use in some other projects.

packages/backend/src/types/UserToken.ts
```diff
@@ -0,0 +1,6 @@
+export interface UserToken {
+  iat: number
+  exp: number
+  userId: string
+  userPermanentId: string // TODO add validation this exists
```
sharafian, line 5: I'm assuming this is just imported, but we shouldn't have a distinction of userIds and permanentIds in Rafiki

sharafian (approved):
looks straightforward to me

packages/backend/package.json
```diff
@@ -5,19 +5,27 @@
...
   "devDependencies": {
+    "@apollo/client": "^3.3.20",
+    "@graphql-codegen/cli": "1.21.5",
+    "@graphql-codegen/introspection": "1.18.2",
+    "@graphql-codegen/typescript": "1.22.2",
+    "@graphql-codegen/typescript-resolvers": "1.19.3",
     "@types/ioredis": "^4.22.3",
     "@types/jest": "^26.0.22",
-    "@types/koa": "^2.13.1",
+    "@types/koa": "^2.13.3",
```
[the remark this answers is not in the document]
cairin: No, types minor versions don't necessarily need to match their corresponding packages.

I also didn't explicitly bump this, yarn did, which makes me think its a requirement in `apollo-server-koa`.

packages/backend/src/spsp/service.ts
cairin: Yes that makes sense.

packages/backend/src/index.ts
cairin: The IoC container handles registration and resolution of service dependencies for us. By doing it this way we don't have to worry about the instantiation order of interdependent services.

packages/backend/src/types/UserToken.ts
cairin: Good catch, will remove :+1:

matdehaast (approved):

sentientwaffle (approved):

PR #34, opened 2021-06-29
feat(backend): create invoice service and graphql resolvers

cairin:

- Implements Invoice service:
  - create invoice with a sub account of a users account.
  - allow fetching of individual invoices
  - allow fetching of invoice pages
- Implements Invoice resolver
  - Implements schema appropriately.
  - Uses cursor pagination following [GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm)
  - allows fetching user invoices on a per page basis
- Enforces a user Id to be passed in on all queries
- Created function to truncate tables in tests because the transaction logic doesn't work anymore.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/account/service.ts
```diff
@@ -4,6 +4,7 @@ import { BaseService } from '../shared/baseService'
 export interface AccountService {
   get(id: string): Promise<Account>
   create(scale: number, currency: string): Promise<Account>
+  createSubAccount(parentAccountId: string): Promise<Account>
```
matdehaast, line 7: Lets change all references of `parentAccountId` to `superAccountId` as per https://www.notion.so/coil/Rafiki-Account-Service-API-6d172588dd534becacfa21e178cd7407#a850539758e445b58fca7a3305249506

matdehaast (approved):
Fix failing

packages/backend/package.json
```diff
@@ -4,7 +4,7 @@
     "lint": "yarn lint:local",
     "format": "yarn format:local",
     "format:check": "yarn format:check:local",
-    "test": "yarn test:local",
+    "test": "yarn test:local --runInBand",
```
sentientwaffle, line 7: @wilsonianb this flag might help make the tests more consistent for debugging https://github.com/coilhq/rafiki/pull/44#issuecomment-875102781

packages/backend/src/account/service.ts
sentientwaffle: Why is `super` preferable to `parent`? I find the latter much more straightforward.

packages/backend/src/invoice/service.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      user = await userService.create()
+    }
+  )
+
+  afterAll(
+    async (): Promise<void> => {
+      await appContainer.shutdown()
+      await workerUtils.release()
+      await resetGraphileDb(knex)
+      await truncateTables(knex)
```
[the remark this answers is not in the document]
cairin: The tables are already rolled back and destroyed in the teardown. This is more to prevent conflicts when running multiple test suits against the same table - i.e. just ensuring each test suite has a clean table to work with.

packages/backend/src/graphql/resolvers/invoice.ts
```diff
@@ -0,0 +1,72 @@
...
+  try {
+    hasNextPageInvoices = await invoiceService.getUserInvoicesPage(
+      firstInvoice.userId,
+      {
+        after: lastEdge,
+        first: 1
+      }
+    )
+  } catch (e) {
+    hasNextPageInvoices = []
```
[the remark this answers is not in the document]
cairin: No, that's intentional. This resolver is generating the `PageInfo` so at the edges of the data set those queries should fail.

packages/backend/src/account/service.ts
```diff
@@ -45,3 +47,17 @@ async function createAccount(
     currency: currency
   })
 }
+
+async function createSubAccount(
+  deps: ServiceDependencies,
+  superAccountId: string
+): Promise<Account> {
+  deps.logger.info('Creates an account')
```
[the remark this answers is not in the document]
cairin: Yeah will just remove it :+1:

packages/backend/migrations/20210615140737_create_accounts_table.js
```diff
@@ -3,6 +3,7 @@ exports.up = function (knex) {
     table.uuid('id').notNullable().primary()
     table.integer('scale').notNullable()
     table.string('currency').notNullable()
+    table.uuid('superAccountId').nullable()
```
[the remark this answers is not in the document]
cairin: Not necessarily. It's just a parent account (which is what I initially called it) but @kincaidoneil used the term `super` in his spec. I don't feel particularly strongly about either, so I'm happy to revert it.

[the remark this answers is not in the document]
sentientwaffle: I find `parent` clearer, but don't feel too strongly either.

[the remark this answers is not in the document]
wilsonianb: >To disambiguate Interledger routing relations, nested accounts are referenced with "sub" and "super" instead of "parent" and "child."

packages/backend/src/account/service.ts
sentientwaffle: Never mind; Brandon answered my question.

packages/backend/src/invoice/service.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      expect(invoices).toHaveLength(20)
+      expect(invoices[0].id).toEqual(invoicesCreated[0].id)
+      expect(invoices[19].id).toEqual(invoicesCreated[19].id)
+      expect(invoices[20]).toBeUndefined()
+    })
+
+    test('Can paginate backwards from a cursor with a limit', async (): Promise<void> => {
+      const pagination = {
+        last: 10,
+        before: invoicesCreated[10].id
```
wilsonianb, line 177: Using a later cursor or a smaller limit would better test the limit.
As is, you'd get the same result without `last`.

packages/backend/src/invoice/service.test.ts
```diff
@@ -0,0 +1,220 @@
...
+        user.id,
+        pagination
+      )
+      expect(invoices).toHaveLength(10)
+      expect(invoices[0].id).toEqual(invoicesCreated[0].id)
+      expect(invoices[9].id).toEqual(invoicesCreated[9].id)
+      expect(invoices[10]).toBeUndefined()
+    })
+
+    test('Providing before and after results in forward pagination', async (): Promise<void> => {
```
wilsonianb, line 189: What do you think about making `after`/`first` and `before`/`last` mutually exclusive in the `Pagination` definition?

wilsonianb (commented):
>Created function to truncate tables in tests because the transaction logic doesn't work anymore.

What was the issue with the transaction logic?

cairin:
> > Created function to truncate tables in tests because the transaction logic doesn't work anymore.
> 
> What was the issue with the transaction logic?

Basically, it can't rollback the transactions because we're using the services to create items in the db. Where as in bernie we weren't doing that, so we could link all the changes to a transaction.

packages/backend/src/graphql/resolvers/invoice.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  afterAll(
+    async (): Promise<void> => {
+      await appContainer.apolloClient.stop()
+      await appContainer.shutdown()
+      await truncateTables(knex)
+    }
+  )
+
+  describe("User's invoices", (): void => {
```
sethalicious, line 48: Are tests needed for the following situations:
* no invoices, but invoices requested
* request invoices past the 50th? (I see the after = 44, first = 10.  What about after = 50, first = 10?
* negative index for pagination? (after = -1, first = -1)

packages/backend/src/invoice/service.test.ts
cairin: I'm not sure how you would achieve this? Also, `Pagination` is just mimicking the Schema input to invoices. And you wouldn't be able to change that unless you deviated from the spec.

packages/backend/src/graphql/resolvers/invoice.test.ts
cairin: > no invoices, but invoices requested

:+1: 

> request invoices past the 50th? (I see the after = 44, first = 10. What about after = 50, first = 10?

Could you clarify why this would be useful? There are only 50 dummy invoices created for testing. Also, `after` is not an index it's a cursor.

> negative index for pagination? (after = -1, first = -1)

This is tested in the service, along with other pagination mechanics and limits.

packages/backend/src/invoice/service.test.ts
cairin: This test was being overloaded to check that a backwards paginated output results in the same as a forwards pagination. So I split that out into another test too.

packages/backend/src/graphql/resolvers/invoice.test.ts
sethalicious: > request invoices past the 50th? (I see the after = 44, first = 10. What about after = 50, first = 10?

If it's already tested, no worries.  I was wondering what happens on the edge cases where we go beyond the available invoices.  Trigger any errors that might occur.  If it's already tested, no worries.

matdehaast (approved):

PR #37, opened 2021-06-30
feat(connector): rate backend, rate service

sentientwaffle:

- In the connector, use a rate service to convert currencies.
  - If no rate/price service is configured, the connector will only handle same-currency payments.
- `packages/rates/` is an example rates service that pulls rates from ECB & XRP (like `ilp-connector`'s built-in backend).
  - The name "rates service" is a little misleading -- the API actually returns prices (like `@interledger/pay`'s API) -- but I still refer to it as a rates service because of its purpose.

## Context

Fixes https://github.com/coilhq/rafiki/issues/30

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/connector/src/services/core/services/rates/service.test.ts
```diff
@@ -0,0 +1,165 @@
...
+        })
+      ).resolves.toBe(123_000n)
+      await expect(
+        service.convert({
+          sourceAmount: 123456n,
+          sourceAsset: { code: 'USD', scale: 12 },
+          destinationAsset: { code: 'USD', scale: 9 }
+        })
+      ).resolves.toBe(123n)
+      expect(requestCount).toBe(1)
```
wilsonianb, line 65: What do you think about skipping the `sharedLoad` request for same currency, different scales?

packages/connector/src/services/core/services/rates/service.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    })
+
+    it('returns the converted amount when assets are dfiferent', async () => {
```
wilsonianb, line 68:
```suggestion
    it('returns the converted amount when assets are different', async () => {
```

packages/rates/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "lib": ["ES2020"],
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
+  "include": ["src/**/*"]
```
wilsonianb, line 8: tests could be excluded

wilsonianb (commented):
The rates workspace could be added to the [README](https://github.com/coilhq/rafiki/blob/main/README.md#packages) and [pr labeler](https://github.com/coilhq/rafiki/blob/main/.github/labeler.yml)

wilsonianb (approved):

Issue #43, opened 2021-07-06
Reset/rollback tigerbeetle between accounts tests


wilsonianb:
Tigerbeetle does not currently support resetting or rolling back state: https://github.com/coilhq/rafiki/pull/6#issuecomment-832442870

Accounts services tests currently use random/unique assets per test for idempotency.

An alternative (until the functionality is supported by tigerbeetle) would be to have [`testcontainers`](https://www.npmjs.com/package/testcontainers) start a new tigerbeetle container per test.

wilsonianb:
- https://github.com/interledger/rafiki/pull/199 starts a new tigerbeetle testcontainer per asset test and creates dedicated tigerbeetle testcontainer for accounting tests

PR #44, opened 2021-07-06
feat(ci): use testcontainers for testing with databases

### LINKED ISSUE #33: Use testcontainers for tests

wilsonianb:

- Previously, running tests required starting redis and/or postgres and/or tigerbeetle depending on the combination of packages being tested. Now [`testcontainers`](https://www.npmjs.com/package/testcontainers) starts required database containers when running tests.

## Context

Resolves #33 
Does not address #43 (probably blocked by https://github.com/coilhq/rafiki/issues/46)
This implementation relies on unique database configuration environment variables between rafiki packages.
As noted in the update README instructions, `testcontainers` require configuring Docker to [run as non-root user](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user).

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/connector/jest.setup.js
```diff
@@ -1,14 +1,11 @@
...
-    await redis.disconnect()
-  } else {
-    throw new Error('expected redis at ' + REDIS_URL)
-  }
+  const redisContainer = await new GenericContainer('redis')
+    .withExposedPorts(6379)
+    .start()
+
+  global.__CONNECTOR_REDIS__ = redisContainer
+  process.env.REDIS = `redis://localhost:${redisContainer.getMappedPort(6379)}`
```
sethalicious, line 10: The other files use constant variables for the port number.  Could you do the same here to be consistent?

sethalicious (approved):

sentientwaffle (commented):
Could we keep supporting tests w/o docker/testcontainers? When I'm running the tests often, faster is better. If the tests are run with a `DATABASE_URL` or `REDIS` env variable already set, it could just connect instead of spinning up a database.

sentientwaffle:
`require('testcontainers')` should be part of the conditionals, since it fails if docker isn't running

```
Error: connect ENOENT /var/run/docker.sock
    at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
```

wilsonianb:
> `require('testcontainers')` should be part of the conditionals, since it fails if docker isn't running
> 
> ```
> Error: connect ENOENT /var/run/docker.sock
>     at PipeConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
> ```

@sentientwaffle and I determined that the error only happens if the database configuration environment variable(s) aren't defined, which is ok (we need docker to be running in that case).

wilsonianb:
That [build failure](https://github.com/coilhq/rafiki/actions/runs/1005768738) appears to be a sporadic occurrence with `testcontainers`. I got the same while testing on my fork and got it to succeed by re-running.

I was running into the same `error code 129` trying to resolve https://github.com/coilhq/rafiki/issues/43 by start a new tigerbeetle container for each test. I'm guessing there's something about how I'm starting tigerbeetle with `testcontainers` that isn't reliable.

sentientwaffle (approved):

wilsonianb:
> That [build failure](https://github.com/coilhq/rafiki/actions/runs/1005768738) appears to be a sporadic occurrence with `testcontainers`. I got the same while testing on my fork and got it to succeed by re-running.
> 
> I was running into the same `exit code 129` trying to resolve #43 by starting a new tigerbeetle container for each test. I'm guessing there's something about how I'm starting tigerbeetle with `testcontainers` that isn't reliable.

@sentientwaffle pointed out that the error seems to predate this change: https://github.com/coilhq/rafiki/runs/2954837414?check_suite_focus=true

sethalicious (dismissed):

wilsonianb:
The [build failures](https://github.com/coilhq/rafiki/pull/44#issuecomment-875097353) are now being tracked in https://github.com/coilhq/rafiki/issues/46 and appear to be tigerbeetle related.

re: https://github.com/coilhq/rafiki/pull/45
- https://github.com/coilhq/rafiki/pull/44/commits/4b55fbbfbb06497cc5eb7bed92058f18b92f3640 is probably overkill since redis appears to currently only be used in one test, but something similar could be done in the `backend` once it's using/testing with redis.
- Since `accounts` currently only has one test file (which afaik means there won't be multiple jest workers), I don't think it's currently beneficial to create `maxWorkers` postgres tables (or tigerbeetle databases) when running `accounts` tests. For https://github.com/coilhq/rafiki/issues/43, there'll likely be a new tigerbeetle container per test (but that might be blocked by https://github.com/coilhq/rafiki/issues/46)

sethalicious (approved):

kincaidoneil (approved):

cairin (approved):

PR #45, opened 2021-07-09
test(backend): configure db per jest worker

cairin:

- Creates a db per jest worker thread, to ensure that tests can run concurrently. [Inspiration](https://walrus.ai/blog/2020/04/testing-database-interactions-with-jest/)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/jest.setup.js
+++ b/packages/backend/jest.setup.js
@@ -1,15 +1,13 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const Knex = require('knex')
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const { cpus } = require('os')
-// eslint-disable-next-line @typescript-eslint/no-var-requires
 const { GenericContainer } = require('testcontainers')
 
 const POSTGRES_PORT = 5432
 const REDIS_PORT = 6379
 
-module.exports = async () => {
-  const workers = cpus().length - 1
+module.exports = async (globalConfig) => {
+  const workers = globalConfig.maxWorkers
 
   if (!process.env.DATABASE_URL) {
     const postgresContainer = await new GenericContainer('postgres')
--- a/packages/backend/jest.setup.js
+++ b/packages/backend/jest.setup.js
@@ -1,13 +1,15 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const Knex = require('knex')
 // eslint-disable-next-line @typescript-eslint/no-var-requires
+const { cpus } = require('os')
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const { GenericContainer } = require('testcontainers')
 
 const POSTGRES_PORT = 5432
 const REDIS_PORT = 6379
 
 module.exports = async () => {
-  const workers = parseInt(process.env.JEST_WORKERS || '1')
+  const workers = cpus().length - 1
 
   if (!process.env.DATABASE_URL) {
     const postgresContainer = await new GenericContainer('postgres')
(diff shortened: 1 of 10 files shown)


packages/backend/src/account/service.test.ts
```diff
@@ -22,10 +22,14 @@ describe('Account Service', (): void => {
   const mockMessageProducer = {
     send: jest.fn()
   }
+  const overrideConfig = {
```
wilsonianb, line 25: Could this instead happen in `createTestApp` like how it currently is in `main` (what I removed in the [testcontainers pr](https://github.com/coilhq/rafiki/pull/44/files#diff-189b2fcc08bab2590fbd9ef5e29200b7d7bbbe3fbf4f486fa9ae541f2e929da9L18-R97)) instead of each test file?

packages/backend/jest.setup.js
```diff
@@ -7,6 +7,8 @@ const POSTGRES_PORT = 5432
 const REDIS_PORT = 6379
 
 module.exports = async () => {
+  const workers = parseInt(process.env.JEST_WORKERS || '1')
```
wilsonianb, line 10: What happens if the number of actual workers exceeds `process.env.JEST_WORKERS`?
It'd be nice if this jest had a way for the global setup file to know the number of workers, but I'm not seeing one.

wilsonianb: [`--max-workers`](https://jestjs.io/docs/26.x/cli#--maxworkersnumstring)
>this defaults to the number of the cores available on your machine minus one for the main thread.

Could we ditch `process.env.JEST_WORKERS ` and use that default value?

```suggestion
  const workers = os.cpus().length - 1
```

packages/backend/src/account/service.test.ts
cairin: I've moved it to the config file. It's needed when we create the IoC container, which happens before `createTestApp` is run.

packages/backend/jest.setup.js
cairin: https://github.com/facebook/jest/issues/8708

I guess we could just write a script that counts how many test files there are rather.

cairin: > Could we ditch process.env.JEST_WORKERS and calculate that default value ourselves?

Didn't see that, that could work :+1:

wilsonianb: Actually, `maxWorkers` is accessible in [globalSetup](https://jestjs.io/docs/26.x/configuration#globalsetup-string):
>This function gets Jest's `globalConfig` object as a parameter.

So no computation required

cairin: > This function gets Jest's globalConfig object as a parameter.

That's not actually accessible from the function, that's just for storing global values like db configs that need to be accessed in teardown.

packages/backend/jest.setup.js
```diff
@@ -1,12 +1,16 @@
...
 // eslint-disable-next-line @typescript-eslint/no-var-requires
+const { cpus } = require('os')
+// eslint-disable-next-line @typescript-eslint/no-var-requires
 const { GenericContainer } = require('testcontainers')
 
 const POSTGRES_PORT = 5432
 const REDIS_PORT = 6379
 
 module.exports = async () => {
+  const workers = cpus().length - 1
```
wilsonianb, line 12: This appears to work (`globalConfig.maxWorkers` is `7` for me)

```suggestion
module.exports = async (globalConfig) => {
  const workers = globalConfig.maxWorkers
```

packages/backend/jest.setup.js
```diff
@@ -1,12 +1,16 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const Knex = require('knex')
 // eslint-disable-next-line @typescript-eslint/no-var-requires
+const { cpus } = require('os')
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
cairin, line 5:
```suggestion
```

wilsonianb (approved):

PR #51, opened 2021-07-19
feat(interledger): add sub-accounts with credit and debt

### LINKED ISSUE #32: Implement Interledger sub-accounts

wilsonianb:

- Add sub-accounts
- Add credit and debt to sub-accounts

## Context

Resolves #32 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


wilsonianb:
https://github.com/coilhq/tigerbeetle/issues/29 has more context for https://github.com/coilhq/rafiki/pull/51/commits/950d6b0958261add4ad70a6d1611883ad843fd9e :point_up_2: 
Also, here's what the commit looks like without `husky`/`yarn format` blowing up the diff with indentation changes: https://github.com/wilsonianb/rafiki/commit/0f1a9bb9ae6cfb40db48438cdbdb913c7617e817

packages/accounts/src/types/ilpBalance.ts
```diff
@@ -1,16 +1,11 @@
...
 export interface IlpBalance {
   id: string
+  asset: Asset
   balance: bigint
-  // children?: IlpBalanceChildren
-  // parent?: IlpBalanceParent
+  availableCredit: bigint
+  creditExtended: bigint
+  totalBorrowed: bigint
+  totalLent: bigint
```
matdehaast, line 10: Can we document what these mean please

packages/accounts/src/utils/balance-id.ts
```diff
@@ -83,6 +84,10 @@ export function toSettlementId({
 //   )
 // }
 
+export function bigIntToUuid(id: bigint): Pojo {
+  return raw('?::uuid', [id.toString(16).padStart(32, '0')])
+}
+
```
matdehaast, line 90: Ideally we want to get this not having to call postgres to do the conversion

matdehaast (commented):
Just some comments for now. Want to dive deeper into the accounts and accounts.test to get a better gripe on this

packages/accounts/src/models/ilpAccount.ts
```diff
@@ -3,12 +3,32 @@ import { IlpHttpToken } from './ilpHttpToken'
 import { uuidToBigInt } from '../utils'
 import { Model, Pojo, raw } from 'objection'
 
+function bigIntToUuid(id: bigint): Pojo {
```
sentientwaffle, line 14: Not directly related to this PR, but why ever represent UUIDs as bigints (rather than strings) at all?

packages/accounts/migrations/20210422194131_create_ilp_accounts_table.js
```diff
@@ -8,12 +8,13 @@ exports.up = function (knex) {
...
 
     table.uuid('balanceId').notNullable()
-    // table.uuid('debtBalanceId').notNullable()
-    // table.uuid('trustlineBalanceId').notNullable()
-    // table.uuid('loanBalanceId').nullable()
-    // table.uuid('creditBalanceId').nullable()
+    table.uuid('trustlineBalanceId').nullable()
+    table.uuid('creditExtendedBalanceId').nullable()
+    table.uuid('borrowedBalanceId').nullable()
+    table.uuid('lentBalanceId').nullable()
```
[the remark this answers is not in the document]
wilsonianb: https://github.com/coilhq/rafiki/pull/51/commits/8921200b22186cdeb248c791c06e78075f9e24a5

packages/accounts/src/utils/balance-id.ts
wilsonianb: https://github.com/coilhq/rafiki/pull/51/commits/7a2934ad7c09842ebe811ff6e8edf050049b110d limits this to `$formatDatabaseJson` where the typecast is added to the postgres query
https://vincit.github.io/objection.js/api/model/overview.html#model-data-lifecycle

packages/accounts/src/models/ilpAccount.ts
wilsonianb: TigerBeetle account ids are 128-bit.
[tigerbeetle-node](https://github.com/coilhq/tigerbeetle-node/blob/main/src/index.ts#L23) deals with these as `bigint`s.
UUIDs seems to be the best way to store 128-bit values in postgres (https://github.com/coilhq/rafiki/pull/18#discussion_r649345413)

packages/accounts/src/services/accounts.ts
```diff
@@ -959,4 +933,280 @@ export class AccountsService implements AccountsServiceInterface {
...
+    amount,
+    mode
+  }: {
+    accountId: string
+    amount: bigint
+    mode: AdjustTrustlineMode
+  }): Promise<void | TrustlineError> {
+    try {
+      const transfers: BalanceTransfer[] = []
+      const err = await transaction(
```
sentientwaffle, line 999: Will this function behave if it is run for overlapping account trees in parallel?

I think [repeatable read](https://www.postgresql.org/docs/9.5/transaction-iso.html) + `SELECT FOR UPDATE` would prevent parallel modifications from trampling each other.

packages/accounts/src/models/ilpAccount.ts
sentientwaffle: Ok. I wasn't referring to how the are stored in postgres -- I was talking about how they are represented on the model.

> tigerbeetle-node deals with these as bigints.

That answers my question.

packages/accounts/src/services/accounts.ts
```diff
@@ -282,68 +257,82 @@ export class AccountsService implements AccountsServiceInterface {
   }
 
   public async getAccount(accountId: string): Promise<IlpAccount | undefined> {
-    const accountRow = await IlpAccountModel.query().findById(accountId)
+    const accountRow = await IlpAccountModel.query()
+      .withGraphJoined('subAccounts')
```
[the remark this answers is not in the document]
wilsonianb: @matdehaast do you know what type of call would be most useful for the backend?

1. get account optionally with its sub-accounts
2. get account optionally with its sub-accounts' ids
3. get an account's sub-accounts
4. get an account's sub-accounts' ids

packages/accounts/src/models/ilpAccount.ts
jorangreef: Great question @sentientwaffle.

Just to add some background explanation on why `tigerbeetle-node` uses a JS BigInt rather than a JS String type:

I believe a String in V8 requires at least 24 bytes to store regardless of how small the string is, so there is more "boxing" overhead for String metadata (at least this was the case when last I checked a few years ago). It's also stored as a form of UTF-8 I think so there may be decoding overhead to get at the raw bytes... I'm not sure.

A BigInt might have a little bit of boxing overhead to store the word count but should be more efficient. However, we haven't dived into the V8 source to double-check this.

Apart from these smaller details, the main reason is to keep the fields all as first-class numbers, which is how they are represented in the Zig client (`u128`s).

packages/accounts/src/services/accounts.ts
```diff
@@ -959,4 +933,322 @@ export class AccountsService implements AccountsServiceInterface {
...
+                .patch(newBalanceIds[accountId])
+                .findById(accountId)
+            }
+          }
+        }
+      )
+      if (err) {
+        return err
+      }
+      const res = await this.client.createTransfers(
```
sentientwaffle, line 1218: How does (or can) things recover if the postgres transaction above succeeds but the tigerbeetle transaction fails (e.g. network trouble)?

packages/accounts/src/services/accounts.ts
```diff
@@ -959,4 +949,280 @@ export class AccountsService implements AccountsServiceInterface {
     }
     return trx
   }
+
+  public async extendTrustline({
```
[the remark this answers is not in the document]
wilsonianb: https://github.com/coilhq/rafiki/pull/51/commits/fa097826aef68d95b07ffb1a8a473c153bf33f34

packages/accounts/src/services/accounts.ts
wilsonianb: In that case, the ids of any new TigerBeetle accounts that were created by `adjustTrustline` will remain in postgres. Those TigerBeetle accounts will be used for any subsequent relevant trustline transfers for the Interledger account.

packages/accounts/src/services/accounts.ts
wilsonianb: Why wouldn't `SELECT FOR UPDATE` with the default [Read Committed](https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED) suffice?

>UPDATE, DELETE, SELECT FOR UPDATE, and SELECT FOR SHARE commands behave the same as SELECT in terms of searching for target rows: they will only find target rows that were committed as of the command start time. However, such a target row might have already been updated (or deleted or locked) by another concurrent transaction by the time it is found. In this case, the would-be updater will wait for the first updating transaction to commit or roll back (if it is still in progress). If the first updater rolls back, then its effects are negated and the second updater can proceed with updating the originally found row. If the first updater commits, the second updater will ignore the row if the first updater deleted it, otherwise it will attempt to apply its operation to the updated version of the row. The search condition of the command (the WHERE clause) is re-evaluated to see if the updated version of the row still matches the search condition. If so, the second updater proceeds with its operation using the updated version of the row. In the case of SELECT FOR UPDATE and SELECT FOR SHARE, this means it is the updated version of the row that is locked and returned to the client.

sentientwaffle: You're right, that should be fine. :+1:

packages/accounts/src/utils/balance-id.ts
matdehaast: Ah interesting, thats pretty neat!

packages/accounts/src/services/accounts.ts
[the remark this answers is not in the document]
matdehaast: Yeah agreed with @sentientwaffle. An account could potentially have 1000s of subaccounts. We should be very deliberate about getting them. I can't be sure of access patterns yet. Lets just leave the basic for now and add as needed

sentientwaffle:
What is the 'right' way to create a new trustline without transferring any money to it? `extendTrustline` with `autoApply:false` and a caller-generated uuid, or `createAccount` with a super account?

wilsonianb:
As it is now, it'd be:

1. `createAccount` with specified `superAccountId`
2. `extendTrustline` with `autoApply:false` and the id of the account created in step 1

`createAccount` currently accepts a caller-generated `accountId`.
As is, trustlines don't have an id. Trustline methods just take `accountId` of the sub-account to which the trustline is extended.

If we wanted, we could add a way to create sub-account + extend trustline (steps 1+2) in a single operation.

sentientwaffle:
- What is the best way to refund all money in a trustline/subaccount back to the parent account (e.g. to refund leftovers on payment completion, or on payment failure/cancellation)? Get balance then immediately `revokeTrustline`?
- If  `extendTrustline` is called multiple times in a row (with the same parameters), will it move money every time? Or just the first time?

wilsonianb:
>What is the best way to refund all money in a trustline/subaccount back to the parent account (e.g. to refund leftovers on payment completion, or on payment failure/cancellation)? Get balance then immediately `revokeTrustline`?

Yeah, you'd need to call `revokeTrustline` with the current balance.
Maybe `revokeTrustline`'s  `amount` could be optional and default to the current balance? The same could be done for `utilizeTrustline` and `settleTrustline`.
I'm also wondering if any trustline operations should happen if a sub-account is disabled.
 
>If  `extendTrustline` is called multiple times in a row (with the same parameters), will it move money every time? Or just the first time?

Every time

packages/interledger/src/accounts/service.ts
```diff
@@ -790,122 +832,101 @@ export class AccountsService implements AccountsServiceInterface {
...
+          twoPhaseCommit: true
         }
       )
     }
-    const res = await this.client.createTransfers(transfers)
-    for (const { index, code } of res) {
-      switch (code) {
+    const error = await this.createTransfers(transfers)
+    if (error) {
+      switch (error.code) {
```
sentientwaffle, line 894: Nit: The `if { switch { ` blocks (here and elsewhere in this file) could be simplified to `switch (error?.code) {` if you add a noop `null`/`undefined` `case`.

sentientwaffle: Huh, that's annoying. Would `error && error.code` work instead?

wilsonianb: It complains about the subsequent `error.index`s and `error.code`

sentientwaffle: Ok, never mind then.

sentientwaffle (approved):
LGTM. I apologize for the slow review; I thought I had marked it last week.

PR #56, opened 2021-07-21
feat(wm): add webmonetization capability into backend

### LINKED ISSUE #27: Transaction History for Web Monetization
## Public SPSP Endpoint

`GET /pay/:id`

This endpoint is described in the SPSP-STREAM project but should be upgraded in the following way:

- Instead of using the user's root ILP account for the connection tag, check its postgres DB to see if there is a currently active WM invoice
    - Invoices should be stored with metadata, expiry (optional), and a reference to the account service item representing the resource
- If there is no currently active WM invoice, make a call to the account service and a call to postgres to create one
    - If the current WM invoice is expired, it should be disabled via the account service
- Use the ILP account ID of the active invoice as the connection tag when generating SPSP credentials
- Return the destinationAccount and sharedSecret

## Balance & Transaction History

This should be implemented as part of a graphQL API on the API backend. For now we should make it accessible to the wallet operator, and in future we will make it available to the user and to authorized third parties

```graphql
query {
	user: User
}

type Account {
	id: string!
	balance: Amount!
}

type User {
	balance: Amount

	# in the future it should be possible to fetch a transaction list. for now we'll start with just a list of invoices
	invoices(count: number, skip: number): [Invoice]
}

type Amount {
	amount: string!
	currency: string!
	scale: number
}

type Invoice {
	receivedAmount: Amount!
	maximumAmount: Amount
	active: boolean!
	createdAt: Date!
	expiresAt: Date
	description: string
}
```

matdehaast:

- remove user from backend
- add web monetization invoice logic
- alter spsp to use web monetization service
- fixes #27 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/spsp/service.ts
```diff
@@ -63,9 +63,13 @@ async function getPay(
     )
   }
 
-  const user = await deps.userService.get(ctx.params.id)
-  const account = user && (await deps.accountService.get(user.accountId))
-  if (!account) {
+  const accountId = ctx.params.id
+  const invoice = await deps.wmService
+    .getCurrentInvoice(accountId)
+    .catch(() => {
```
sentientwaffle, line 69: Why not `500` on errors? (eg. a database error, not a missing record)

packages/backend/src/spsp/service.ts
```diff
@@ -76,8 +80,9 @@ async function getPay(
   }
 
   try {
+    const account = await deps.accountService.get(invoice.invoiceAccountId)
     const { ilpAddress, sharedSecret } = deps.streamServer.generateCredentials({
-      paymentTag: user.accountId,
+      paymentTag: account.id,
```
sentientwaffle, line 85: Why fetch the account instead of just `paymentTag: invoice.invoiceAccountId`?

packages/backend/src/webmonetization/model.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseModel } from '../shared/baseModel'
+
+export class WebMonetization extends BaseModel {
+  public static get modelPaths(): string[] {
```
sentientwaffle, line 4: This is the default implementation on the parent class `BaseModel` already.

packages/backend/src/webmonetization/model.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  public accountId!: string
+  public currentInvoiceId!: string
+
+  public $beforeInsert(): void {
+    this.createdAt = new Date().toISOString()
+    this.updatedAt = new Date().toISOString()
+  }
+
+  public $beforeUpdate(): void {
```
sentientwaffle, line 24: This is the default implementation on the parent class BaseModel already.

packages/backend/src/webmonetization/model.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  static get idColumn(): string {
+    return 'accountId'
+  }
+
+  public accountId!: string
+  public currentInvoiceId!: string
+
+  public $beforeInsert(): void {
```
sentientwaffle, line 19: Instead of reimplementing here, `BaseModel`'s `$beforeInsert` could generalize to `this[this.idColumn] = this[this.idColumn] || uuid()`.

packages/backend/src/webmonetization/service.ts
```diff
@@ -0,0 +1,98 @@
...
+      return wm
+    })
+
+  ok(deps.knex)
+  const expectedExpiryAt = DateTime.utc().endOf('day') //Expire Every Day
+  // Create an invoice
+  if (!wm.currentInvoiceId) {
+    return WebMonetization.transaction(deps.knex, async (trx) => {
+      const description = 'Webmonetization earnings'
+      const invoice = await deps.invoiceService.create(
```
sentientwaffle, line 64: This code is a little misleading. The invoice isn't actually created within the sql transaction.
(likewise for the `else` block below).

packages/backend/src/graphql/schema.graphql
```diff
@@ -1,11 +1,16 @@
 type Query {
-  user(userId: String!): User
+  account(id: String!): Account
 }
 
-type User {
+type Amount {
+  amount: Int!
+  currency: String!
+  scale: Int
```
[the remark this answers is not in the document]
matdehaast: Good point, it should be!

packages/backend/migrations/20210624151514_create_invoices_table.js
```diff
@@ -1,16 +1,16 @@
...
     table.boolean('active').notNullable()
     table.string('description').nullable()
     table.timestamp('expiresAt').nullable()
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
 
-    table.index('userId')
+    table.index('accountId')
     table.index(['createdAt', 'id'])
```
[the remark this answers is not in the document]
matdehaast: So
```
table.index(['createdAt', 'id']) 
```
is a compound index that greatly speeds up the pagination. Even though we first query by accountId if we did not have that index we would do a full scan of those records to paginate. We tested this on a million record dataset and the compound index improved query performance by 10x

packages/backend/src/spsp/service.ts
matdehaast: Good catch

packages/backend/src/webmonetization/model.ts
matdehaast: Yeah that seems better than this 👍

packages/backend/src/webmonetization/model.ts
matdehaast: This was to get around the stuff above as initially I wasn't extended the baseModel. Will cleanup

packages/backend/src/webmonetization/service.ts
matdehaast: Thanks for picking this up. I think I was going to add the invoice service taking in an optional trx but forget.

packages/backend/src/graphql/resolvers/account.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      appContainer = await createTestApp(deps)
+      knex = await deps.use('knex')
+    }
+  )
+
+  afterAll(
+    async (): Promise<void> => {
+      await appContainer.apolloClient.stop()
+      await appContainer.shutdown()
+      await truncateTables(knex)
```
wilsonianb, line 31: `knex.destroy()` is called in `appContainer.shutdown()`.
Does that mean `truncateTables(knex)` should happen first?
Ditto for other (some pre-existing) test files.

packages/backend/src/invoice/service.ts
```diff
@@ -50,13 +51,16 @@ async function getInvoice(
...
+  description: string,
+  expiresAt?: Date
 ): Promise<Invoice> {
-  const user = await deps.userService.get(userId)
-  const subAccount = await deps.accountService.createSubAccount(user.accountId)
+  const subAccount = await deps.accountService.createSubAccount(accountId)
   return Invoice.query(deps.knex).insertAndFetch({
-    userId: userId,
-    accountId: subAccount.id,
+    accountId: subAccount.superAccountId,
```
wilsonianb, line 60:
```suggestion
    accountId,
```
I'm guessing it would be a problem if `subAccount.superAccountId` isn't the `accountId` passed to `createInvoice`, so maybe also plan to add a check in `createSubAccount` when it's actually calling the accounts service.

packages/backend/src/webmonetization/service.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  test('Creates a new one if the old one has expired', async (): Promise<void> => {
+    jest.useFakeTimers('modern')
+    const currentWmInvoice = await wmService.getCurrentInvoice(account.id)
+    const msTomorrow = DateTime.now().plus({ day: 1 }).toMillis()
+    jest.setSystemTime(msTomorrow)
+
+    const wmInvoice = await wmService.getCurrentInvoice(account.id)
+
+    expect(wmInvoice).not.toEqual(currentWmInvoice)
```
wilsonianb, line 91: Could also add a check that there are two total invoices.

packages/backend/migrations/20210624151514_create_invoices_table.js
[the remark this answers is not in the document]
sentientwaffle: > Even though we first query by accountId if we did not have that index we would do a full scan of those records to paginate.

Maybe I don't understand your response. I'm suggesting altering the compound index, not removing it.

---

Is a compound index different from a [multicolumn index](https://www.postgresql.org/docs/10/indexes-multicolumn.html)?

> A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns. The exact rule is that equality constraints on leading columns, plus any inequality constraints on the first column that does not have an equality constraint, will be used to limit the portion of the index that is scanned.

Right now there is an `accountId` index and a `(createdAt,id)` index. But the table is queried by `(accountId,createdAt,id)` (equality then order). So why not index on `(accountId,createdAt,id)`?

(And as an aside with the triple index the single-field `accountId` index could be [removed](https://dba.stackexchange.com/questions/27481/is-a-composite-index-also-good-for-queries-on-the-first-field).)

packages/backend/src/webmonetization/service.ts
```diff
@@ -0,0 +1,98 @@
...
+  const wm = await WebMonetization.query(deps.knex)
+    .findById(account.id)
+    .then((wm) => {
+      if (!wm) {
+        return WebMonetization.query(deps.knex).insertAndFetch({
+          accountId: account.id
+        })
+      }
+      return wm
+    })
```
wilsonianb, line 56: This appears to also work
```suggestion
  const wm = await WebMonetization.query(deps.knex)
    .insertAndFetch({
      accountId: account.id
    })
    .onConflict('accountId')
    .ignore()
```

packages/backend/src/webmonetization/service.ts
wilsonianb: >likewise for the else block below

I'd vote for not duplicating the code creating the new invoice if possible.

packages/backend/src/spsp/service.ts
matdehaast: @sentientwaffle I now remember, you need to get the account as we need to pass in the asset code and scale as well.

packages/backend/src/spsp/service.ts
matdehaast: So the problem I'm trying to solve here is that if the underlying account doesn't exist to return the 404 error that was previously defined
```
ctx.status = 404
    ctx.set('Content-Type', CONTENT_TYPE_V4)
    ctx.body = JSON.stringify({
      id: 'InvalidReceiverError',
      message: 'Invalid receiver ID'
    })
```

Obviously the above is not great as it swallows all errors. I can either leave it like this or create a specific error for when an account can't be found or something else? What do you recommend @sentientwaffle

packages/backend/src/webmonetization/model.ts
matdehaast: The issue you now run into is that the id field is defined on the base model but isn't used. 

I think what may be better here is to rather just have the primary key be called `id` and put a comment that it is the `accountId`

packages/backend/src/webmonetization/service.ts
matdehaast: @wilsonianb yeah agreed

packages/backend/migrations/20210624151514_create_invoices_table.js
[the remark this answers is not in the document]
matdehaast: Thanks I understand now. That's pretty handy. Also I see that the multi-column index will still allow the access pattern for `where accountId = x`

packages/backend/src/graphql/resolvers/account.test.ts
matdehaast: Thanks I changed the ordering to ensure truncate should happen first

packages/backend/src/invoice/service.ts
matdehaast: I've added an assert now to check that the are equal in `createSubAccount`

packages/backend/src/spsp/service.ts
sentientwaffle: The account is fetched later anyway (in the `try..catch`). If that was moved up, and `accountService.get` returned `Account | undefined` (instead of `throw`-ing on an absent account) it could trigger a `404`.

matdehaast: Ok, I have moved things around to better solve that case now

packages/backend/src/invoice/service.ts
```diff
@@ -50,13 +53,18 @@ async function getInvoice(
...
+  description: string,
+  expiresAt?: Date,
+  trx?: Transaction
 ): Promise<Invoice> {
-  const user = await deps.userService.get(userId)
-  const subAccount = await deps.accountService.createSubAccount(user.accountId)
-  return Invoice.query(deps.knex).insertAndFetch({
-    userId: userId,
-    accountId: subAccount.id,
+  const subAccount = await deps.accountService.createSubAccount(accountId)
```
sentientwaffle, line 61: I think this will work fine even though `trx` isn't passed to `accountService.createSubAccount`, but it is a little awkward. (It may create a dangling account if the invoice's transaction fails.) Too bad javascript doesn't have dynamically-scoped variables :stuck_out_tongue:

matdehaast: @sentientwaffle I will rather just make the subaccount also be wrapped in the transaction. I just missed it TBH.

> Too bad javascript doesn't have dynamically-scoped variables 
You'll be waiting a long time 😆

sentientwaffle (approved):

PR #60, opened 2021-07-26
feat(interledger): add graphql server to interledger package

### LINKED ISSUE #58: Add graphQL server to account service
Allows us to proceed with #57 . This can just be a scaffolding of the graphQL server, so that the resolvers can be implemented later

cairin:

- Adds graphql server as the admin-api for the interledger package
- resolves #58

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/interledger/src/index.ts
```diff
@@ -84,3 +92,36 @@ if (!module.parent) {
...
+}
+
+async function _setupAdminApi(): Promise<ApolloServer> {
+  const tbClient = createClient({
+    cluster_id: Config.tigerbeetleClusterId,
+    replica_addresses: Config.tigerbeetleReplicaAddresses
+  })
+  const accountsService = new AccountsService(tbClient, Config, logger)
+
+  return createAdminApi({ accountsService })
```
sentientwaffle, line 126: Why not use the same `accountsService` for both the admin api and the connector?

cairin: Mainly because I didn't want to clutter the start function with creating a bunch of services. But I'm happy to change it.

sentientwaffle: Are there any drawbacks to having multiple tigerbeetle clients in the same process?

cairin: Yeah probably, might just be easier not to test it.

sentientwaffle (dismissed):

packages/interledger/codegen.yml
```diff
@@ -0,0 +1,14 @@
+overwrite: true
+schema: "./src/admin-api/schema.graphql"
+documents: null
```
matdehaast, line 3: I think this is optional, can we just remove it?

matdehaast (approved):

matdehaast (approved):

PR #61, opened 2021-07-27
feat(interledger): ilp account api

### LINKED ISSUE #57: Expose account API to backend via graphQL
Instead of using REST, we can expose the account service API to consumers (the API backend and the operator) using graphQL. This has the advantage of only fetching necessary data, especially in situations where data comes only from one of postgres or tigerbeetle

cairin:

- resolves #57 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,437 @@
...
+    originAmount: Int!
+    originAccountId: ID!
+    destinationAccountId: ID!
+    destinationAmount: Int
+    autoCommit: Boolean
+    idempotencyKey: ID!
+  ): TransferMutationResponse
+
+  "Provision trustline"
+  provisionTrustline: ProvisionTrustlineMutationResponse
```
sentientwaffle, line 54: What does this do?

packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,437 @@
...
+  id: ID!
+  createdTime: Int!
+  creditorAccountId: ID
+  debtorAccountId: ID!
+  availableCredit: Int!
+  debtBalance: Int!
+}
+
+type CreateIlpAccountMutationResponse implements MutationResponse {
+  code: String!
```
sentientwaffle, line 301: What does `code` refer to in all of these responses? Is it the string error enum?

packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,437 @@
...
+
+  "Delete Interledger account"
+  deleteIlpAccount: DeleteIlpAccountMutationResponse!
+
+  "Create Interledger sub-account"
+  createIlpSubAccount(superAccountId: ID!): CreateIlpSubAccountMutationResponse!
+
+  "Transfer between Interledger accounts"
+  transfer(
+    originAmount: Int!
```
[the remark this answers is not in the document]
cairin: That's what was in the spec doc. I'll change it to match the accounts service 👍

packages/interledger/src/admin-api/schema.graphql
cairin: It's meant to represents the status of the data transfer, like an HTTP status code.

packages/interledger/src/admin-api/schema.graphql
cairin: I'm not sure what you're referring to?

sentientwaffle: What does it mean to 'provision' a trustline? How is it different than creating a sub-account? Why does this call have no parameters?

cairin: > TODO — Provision trustline IDs for operator and/or super-account at the time of ILP account creation, but don't create TigerBeetle accounts until `extend` operation is called?

Reading it more carefully I see we don't need this at all. I left out the parameters because it was TODO in the docs.

packages/interledger/src/admin-api/schema.graphql
sharafian, line 45: graphQL ints are typically limited to 32 bits: https://graphql.org/learn/schema/#scalar-types

We should use string numbers instead, or define a custom scalar type that serializes to bigints.

packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,437 @@
...
+  cursor: String!
+}
+
+type IlpAccount {
+  id: ID!
+  enabled: Boolean!
+  superAccount: IlpAccount
+  subAccounts(
+    "Paginating forwards: the cursor before the the requested page."
+    after: String
```
sharafian, line 155: can we create an input type for Pagination so that we don't need to keep re-defining this?

sharafian (changes requested):
I think all the Ints in the schema need to be changed to allow UInt64 and up. Overall this all makes sense to me

packages/interledger/src/admin-api/schema.graphql
cairin: I think it's probably best to create a custom scalar. I'll create a `UInt64` that maps to a `BigInt`.

packages/interledger/src/admin-api/schema.graphql
cairin: We can, theoretically, but I don't think we should.

Firstly, the pagination algorithm we use ([GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm)) defines it this way. This is mostly to optimize for the clients side - making it easier to cache and fetch more data without necessarily having to program the logic each time. Of coarse this can be adapted to take in an input rather than individual arguments. [Relevant apollo docs](https://www.apollographql.com/docs/react/pagination/cursor-based/#relay-style-cursor-pagination). Example implementation: [Shopify pagination spec](https://shopify.dev/api/usage/pagination-graphql).

Secondly, it seems like standard practice to only use `input` objects for mutations, whereas queries usually have individual arguments. Example: [Github graphql](https://docs.github.com/en/graphql/reference/queries).

Lastly, a graphql input can't implement an interface. The fields on an input object type can themselves refer to input object types, but you can't mix input and output types in your schema. So you would have to redefine this anyway for each input:

```
type Query {
  ilpAccounts(input: IlpAccountsInput): IlpAccountsConnection!
}

input IlpAccountsInput {
  "Paginating forwards: the cursor before the the requested page."
  after: String
  "Paginating backwards: the cursor after the the requested page."
  before: String
  "Paginating forwards: The first **n** elements from the page."
  first: Int
  last: Int
}

```
Which just feels like unnecessary abstraction, and also more typing :keyboard:

packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,429 @@
...
+    destinationAccountId: ID!
+    destinationAmount: UInt64
+    autoCommit: Boolean
+    idempotencyKey: ID!
+  ): TransferMutationResponse
+
+  "Extend Trustline"
+  extendTrustline(
+    "The id of the trustline to extend."
+    trustlineId: ID!
```
wilsonianb, line 56: https://github.com/coilhq/rafiki/pull/51 
>As is, trustlines don't have an id. Trustline methods just take accountId of the sub-account to which the trustline is extended.

Similarly, there's no trustline resource, so the `trustline` query above isn't necessary. Trustline info is just included with (sub)account balances.

sharafian (approved):

sharafian:
@cairin  were there any places here where you wanted input on translating the Account service API to graphQL? Or have Brandon's clarifications on the credit/debt functionality covered everything?

packages/interledger/src/admin-api/schema.graphql
cairin: @wilsonianb Cool, I have updated the schema, please check you're happy with it!

packages/interledger/src/admin-api/schema.graphql
```diff
@@ -1,22 +1,407 @@
...
+}
+
+type RollbackPendingWithdrawalMutationResponse implements MutationResponse {
+  code: String!
+  success: Boolean!
+  message: String!
+}
+
+type Asset {
+  currency: String!
```
wilsonianb, line 380:
```suggestion
  code: String!
```

wilsonianb (commented):
The commit in https://github.com/interledger/rafiki/pull/69 adds `accountId` and `subAccountId` to the credit and debt mutations

packages/interledger/src/admin-api/index.ts
```diff
@@ -1,20 +1,19 @@
...
-  GraphQLFileLoader,
-  loadSchemaSync
-} from 'graphql-tools'
+import { loadSchemaSync } from '@graphql-tools/load'
+import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader'
 import { ApolloServer } from 'apollo-server'
 
 import { resolvers } from './resolvers'
-import { AccountsService } from '../accounts/service'
+// import { AccountsService } from '../accounts/service'
```
wilsonianb, line 7:
```suggestion
```

wilsonianb (dismissed):
:+1:
With the bigint serialization fix, I can pass a string as a query/mutation variable. It's received by the resolver as a bigint, and the test client receives bigint in the response. :heavy_check_mark: 

I'll be submitting subsequent pull requests with accounts service resolvers.

packages/interledger/codegen.yml
```diff
@@ -8,6 +8,8 @@ generates:
     config:
       defaultMapper: Partial<{T}>
       resolverTypeWrapperSignature: Promise<T>
+      scalars:
+        UInt64: bigint
```
matdehaast, line 12: bigint isn't an unsigned bigint. Should we keep the scalar as UInt64? 

Also I can't remember but are there any values in the account spec that could be negative?

packages/backend/src/graphql/schema.graphql
```diff
@@ -2,6 +2,17 @@ type Query {
...
+  "Paginating forwards: Are there more pages?"
+  hasNextPage: Boolean!
+  "Paginating backwards: Are there more pages?"
+  hasPreviousPage: Boolean!
+  "Paginating backwards: the cursor to continue."
+  startCursor: String
+}
+
 type Amount {
   amount: Int!
```
matdehaast, line 17: We should add an issue to update all the amount usages from Int to the UInt64?

packages/interledger/package.json
```diff
@@ -66,6 +68,7 @@
     "@types/pino": "^6.3.9",
     "@types/rosie": "^0.0.39",
     "@types/uuid": "^8.3.1",
+    "cross-fetch": "^3.1.4",
```
matdehaast, line 71: What is cross-fetch needed for?

matdehaast (approved):
Looks good @cairin . There are lots of TODO's in here. We just need to make sure those get captured correctly in the issues/project board

packages/interledger/package.json
cairin: `apollo-client` uses it in tests.

packages/backend/src/graphql/schema.graphql
cairin: removed as unnused

matdehaast (approved):

packages/interledger/codegen.yml
wilsonianb: I think the accounts service (and tigerbeetle) only does unsigned values

PR #117, opened 2021-09-07
fix(interledger): store asset in Tigerbeetle account 

### LINKED ISSUE #114: Store asset in Tigerbeetle account unit
[Tigerbeetle accounts](https://github.com/coilhq/tigerbeetle/blob/main/docs/DESIGN.md#data-structures) have a 2 byte `unit` field for storing the account's asset. Transfers between accounts with the different `unit`s are disallowed.

Add a postgres table of accounts service assets (code & scale) mapped to indexes, which can be used as Tigerbeetle account `unit`s.
Replace the asset columns in the postgres accounts table with foreign keys to the assets table.

wilsonianb:

- store ILP account asset in separate postgres table with `smallserial` used in Tigerbeetle account 2 byte `unit` field.
- store liquidity/settlement ids in assets table

## Context

fixes #114 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/interledger/src/accounts/service.ts
+++ b/packages/interledger/src/accounts/service.ts
@@ -36,8 +36,6 @@ import {
 import {
   calculateCreditBalance,
   calculateDebitBalance,
-  toLiquidityId,
-  toSettlementId,
   randomId,
   uuidToBigInt,
   validateId
@@ -160,7 +158,7 @@ export class AccountsService implements AccountsServiceInterface {
             newAccount.superAccountId = account.superAccountId
             const superAccount = await IlpAccountModel.query()
               .findById(account.superAccountId)
-              .withGraphFetched('asset')
+              .withGraphFetched('asset(withUnit)')
               .forUpdate()
               .throwIfNotFound()
             newAccount.assetId = superAccount.assetId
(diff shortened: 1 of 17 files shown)


packages/interledger/src/accounts/service.ts
```diff
@@ -102,6 +105,7 @@ function toIlpAccount(accountRow: IlpAccountModel): IlpAccount {
 interface BalanceOptions {
   id: bigint
   flags: number
+  unit: number
```
sentientwaffle, line 108: What does `unit` actually refers to? I understand it corresponds to a field in tigerbeetle... Is it a tigerbeetle-defined unique identifier of a scale/code tuple?

wilsonianb: https://github.com/coilhq/tigerbeetle/blob/main/docs/DESIGN.md#events

>unit:  2 bytes ( 16-bit) [optional, opaque unit of value, e.g. a currency code, or even something exotic like gold bars]

We're using it to represent an asset code + scale.
Tigerbeetle disallows transfers between accounts with different `unit`s

sentientwaffle: Ok, but I mean where is the actual value for it generated? Or do we just hard-code a mapping somewhere, e.g. USD:9 :: 0x1234?

wilsonianb: It's the [`smallserial`](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL) `unit` column in the assets table:
https://github.com/interledger/rafiki/pull/117/commits/8a39d388028adbc87f680a9e14cb700147f5d801#diff-311f81ba7c657dbb0ad94d4b0c226dc9649478d55578f579ed72047aa581ffebR6

sentientwaffle: Ahh, I see, so it autoincrements. :+1:

sentientwaffle (approved):

PR #120, opened 2021-09-08
fix(interledger): support more asset/amount transfer combinations

### LINKED ISSUE #118: Support same asset code, different scale transfers
The accounts service doesn't correctly handle transfers between accounts with the same asset code but different asset scales:
https://github.com/interledger/rafiki/blob/main/packages/interledger/src/accounts/service.ts#L935-L945
It currently treats them as identical assets.

### LINKED ISSUE #119: Support same asset transfers with different source and destination amounts
via that asset's liquidity account

wilsonianb:

- support same asset code, different scale transfers
- support same asset, different amounts transfers

## Context

fixes #118 
fixes #119 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/interledger/src/accounts/service.ts
+++ b/packages/interledger/src/accounts/service.ts
@@ -932,7 +932,10 @@ export class AccountsService implements AccountsServiceInterface {
 
     const transfers: TwoPhaseTransfer[] = []
 
-    if (sourceAccount.asset.code === destinationAccount.asset.code) {
+    if (
+      sourceAccount.asset.code === destinationAccount.asset.code &&
+      sourceAccount.asset.scale === destinationAccount.asset.scale
+    ) {
       if (destinationAmount && sourceAmount !== destinationAmount) {
         return TransferError.InvalidDestinationAmount
       }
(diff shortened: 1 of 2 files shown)


packages/interledger/src/accounts/service.ts
```diff
@@ -932,17 +932,39 @@ export class AccountsService implements AccountsServiceInterface {
...
+      sourceAccount.asset.code === destinationAccount.asset.code &&
+      sourceAccount.asset.scale === destinationAccount.asset.scale
+    ) {
       transfers.push({
         id: randomId(),
         sourceBalanceId: sourceAccount.balanceId,
         destinationBalanceId: destinationAccount.balanceId,
-        amount: sourceAmount,
+        amount:
+          destinationAmount && destinationAmount < sourceAmount
```
sentientwaffle, line 944: maybe `Math.min(sourceAmount, destinationAmount ?? Infinity)` would be clearer?

wilsonianb: Yeah, I had tried `Math.min` but it sounds like it won't work for `BigInt`s

sentientwaffle: ah, right, nevermind then

sentientwaffle (approved):

PR #172, opened 2021-11-01
feat(backend): add peer accounts

### LINKED ISSUE #154: Add peer accounts
Move [`PEER_ADDRESSES` config](https://github.com/interledger/rafiki/blob/main/packages/backend/src/config/app.ts#L43-L45) to peers table.

The account service will still exist, as peers and incoming/outgoing payments will have accounts.

Account mutations can be removed.

wilsonianb:

- add peers table and service
- move `http` and `staticIlpAddress` from account to peer
- remove account resolvers (accounts are now only created via invoice, outgoing payment, or peer)

## Context

- fixes #154

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/account/errors.ts
```diff
@@ -1,16 +1,12 @@
 export class UnknownAssetError extends Error {
-  constructor(public accountId: string) {
-    super('Asset not found. accountId=' + accountId)
+  constructor(public assetId?: string) {
```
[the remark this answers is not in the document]
wilsonianb: That was supposed be temporary and removed after `createAccount` no longer auto-created the asset. Will fix

packages/backend/src/connector/core/middleware/account.ts
```diff
@@ -2,23 +2,54 @@ import { Errors } from 'ilp-packet'
 import { RafikiAccount, ILPContext, ILPMiddleware } from '../rafiki'
 import { AuthState } from './auth'
 import { AccountNotFoundError } from '../errors'
+import { OutgoingHttp } from '../services/client'
+import { validateId } from '../../../shared/utils'
 
-export function createAccountMiddleware(): ILPMiddleware {
+export interface OutgoingState {
+  outgoing?: OutgoingHttp
```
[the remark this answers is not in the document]
wilsonianb: I'm in the process of also moving `maxPacketAmount` from account to peer, and removing `stream.enabled` (the connector will identify outgoing account as peer and disable stream).
I currently have this as:
```
export interface OutgoingState {
  outgoing?: {
    http?: OutgoingHttp
    stream: {
      enabled: boolean
    }
  }
}
```
But factoring in your other comments, I'm wondering if the connector should just make `RafikiAccount` like I currently have [`MockIlpAccount`](https://github.com/interledger/rafiki/blob/bw-peers/packages/backend/src/connector/core/test/mocks/accounts-service.ts#L6-L18), and hold all of the optional peer fields instead of matching the now reduced accounts model.
So we'd go back to getting this info from `ctx.accounts.outgoing` vs `ctx.state.outgoing`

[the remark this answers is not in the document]
sentientwaffle: > make RafikiAccount like I currently have MockIlpAccount, and hold all of the optional peer fields instead of matching the now reduced accounts model.
:+1: that sounds cleaner, especially if they were going to get joined into the account query anyway.

packages/backend/src/graphql/schema.graphql
```diff
@@ -146,8 +169,36 @@ input UpdateAccountInput {
...
+}
+
+input UpdatePeerInput {
+  id: String!
+  disabled: Boolean
+  maxPacketAmount: UInt64
+  http: HttpInput
+  stream: StreamInput
+  routing: RoutingInput
+}
```
matdehaast: Another option is we could potentially have a single `PeerInput` object and then for the `updatePeer` mutation take in two args. 1. the `peerId` and 2. the `PeerInput`.

Don't have strong feels about it, just an observation

packages/backend/src/account/service.ts
```diff
@@ -11,26 +11,17 @@ import {
...
 
-export type Options = {
+export interface CreateOptions {
   disabled?: boolean
-}
-
-type CreateAccountOptions = Options & {
   assetId: string
   sentBalance?: boolean
 }
```
matdehaast: I'm wondering if the enabling/disabling of an account is a common enough action that it may make sense to just make it a method on the service? ie `enabled` and `disable` that take in the accountId. The bulk patch method is nice, but also not as explicit when being used to enable and disable an account.

matdehaast (approved):

packages/backend/src/graphql/schema.graphql
wilsonianb: The single `PeerInput` would need to either require fields when updating or allow fields to be optional when creating.

packages/backend/src/account/service.ts
wilsonianb: I went ahead and removed the bulk patch since the only remaining update-able field was `disabled`.
I imagine payment and peer services can either use an `accountService.disable`/`enable` or patch the `account.disabled` directly.
For example, I'm not sure if outgoing payments already have the account available for update?
https://github.com/interledger/rafiki/blob/main/packages/backend/src/outgoing_payment/worker.ts#L51-L81

sentientwaffle (approved):

Issue #123, opened 2021-09-13
Invoice Creation API


sharafian:
- A third party should be able to create an invoice with metadata about the purpose of the invoice
- There needs to be some kind of spam protection for unpaid invoices, to prevent someone from creating messages in someone's transaction history by creating an invoice and leaving it unpaid or paying some tiny amount to keep it visible in the transaction history

PR #178, opened 2021-11-05
open payments invoices

sentientwaffle:

- Add a 'receive limit' to invoices.
- Add a new TigerBeetle account to the invoice's account to track the limit. This will (atomically) prevent transfers which would overpay the invoice. The invoice's limit is tracked by a **debit** account. The asset's counter-balance account is a **credit** account.
- Receipts aren't supported when paying to the invoice.
- If an invoice receives no payment before its expiry, it is deleted (see https://github.com/interledger/rafiki/issues/123).
- `MAX_EXPIRY` is currently set to 24 hours, but I picked that more-or-less at random.

## Context

Implement Open Payments invoices.

- Fixes https://github.com/interledger/rafiki/issues/123

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/migrations/20210422194130_create_assets_table.js
+++ b/packages/backend/migrations/20210422194130_create_assets_table.js
@@ -13,6 +13,8 @@ exports.up = function (knex) {
     table.uuid('settlementBalanceId').notNullable()
     // TigerBeetle account id tracking reserved outgoing payments balance
     table.uuid('outgoingPaymentsBalanceId').notNullable()
+    // TigerBeetle account id tracking cumulative invoice receive limit
+    table.uuid('receiveLimitBalanceId').notNullable()
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
--- a/packages/backend/migrations/20211022210203_create_invoices_table.js
+++ b/packages/backend/migrations/20211022210203_create_invoices_table.js
@@ -2,7 +2,7 @@ exports.up = function (knex) {
   return knex.schema.createTable('invoices', function (table) {
     table.uuid('id').notNullable().primary()
     table.uuid('accountId').notNullable()
-    table.foreign('accountId').references('accounts.id')
+    table.foreign('accountId').references('accounts.id').onDelete('CASCADE')
     table.uuid('paymentPointerId').notNullable()
     table.foreign('paymentPointerId').references('paymentPointers.id')
     table.boolean('active').notNullable()
--- a/packages/backend/migrations/20211022210203_create_invoices_table.js
+++ b/packages/backend/migrations/20211022210203_create_invoices_table.js
@@ -8,6 +8,7 @@ exports.up = function (knex) {
     table.boolean('active').notNullable()
     table.string('description').nullable()
     table.timestamp('expiresAt').nullable()
+    table.bigInteger('amountToReceive').nullable()
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
--- a/packages/backend/migrations/20211022210203_create_invoices_table.js
+++ b/packages/backend/migrations/20211022210203_create_invoices_table.js
@@ -14,6 +14,20 @@ exports.up = function (knex) {
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
 
     table.index(['paymentPointerId', 'createdAt', 'id'])
+
+    table.index('expiresAt')
+    /*
+    TODO: The latest version of knex supports "partial indexes", which would be more efficient for the deactivateInvoice use case. Unfortunately, the only version of 'objection' that supports this version of knex is still in alpha.
+
+    // This is a 'partial index' -- expiresAt is only indexed when active=true.
+    table.index('expiresAt', 'idx_active_expiresAt', {
+      // Knex partial indices are a very new feature in Knex and appear to be buggy.
+      //
+      // Use a 'raw' condition because "knex.where('active',true)" fails with:
+      // > migration failed with error: create index "idx_active_expiresAt" on "invoices" ("expiresAt") where "active" = $1 - there is no parameter $1
+      predicate: knex.whereRaw('active = TRUE')
+    })
+    */
   })
 }
 
--- a/packages/backend/src/account/model.ts
+++ b/packages/backend/src/account/model.ts
@@ -26,4 +26,9 @@ export class Account extends BaseModel {
   public readonly balanceId!: string
   // TigerBeetle account id tracking amount sent
   public readonly sentBalanceId?: string
+
+  // TigerBeetle account id tracking an invoice's receive limit.
+  // - "credits" is the actual limit.
+  // - "debits" is the progress towards that limit.
+  public readonly receiveLimitBalanceId?: string
 }
--- a/packages/backend/src/account/service.ts
+++ b/packages/backend/src/account/service.ts
@@ -19,6 +19,7 @@ export interface CreateOptions {
   disabled?: boolean
   assetId: string
   sentBalance?: boolean
+  receiveLimit?: bigint
 }
 
 export interface AccountTransferOptions {
@@ -102,11 +103,23 @@ async function createAccount(
         })
       ).id
     }
+
+    let receiveLimitBalanceId: string | undefined
+    if (account.receiveLimit) {
+      // TODO or create the TigerBeetle account with the limit (i.e. pass in credits_accepted)? Is that allowed or would it mess up accounting?
+      receiveLimitBalanceId = (
+        await deps.balanceService.create({
+          debitBalance: true,
+          unit: accountRow.asset.unit
+        })
+      ).id
+    }
     const newAccount = await accountRow
       .$query(acctTrx)
       .patchAndFetch({
         balanceId,
-        sentBalanceId
+        sentBalanceId,
+        receiveLimitBalanceId
       })
       .withGraphFetched('asset')
 
@@ -273,6 +286,15 @@ async function transferFunds(
       timeout
     })
   }
+  if (destinationAccount.receiveLimitBalanceId) {
+    transfers.push({
+      id: uuid(),
+      sourceBalanceId: destinationAccount.asset.receiveLimitBalanceId,
+      destinationBalanceId: destinationAccount.receiveLimitBalanceId,
+      amount: destinationAmount || sourceAmount,
+      timeout
+    })
+  }
   const error = await deps.transferService.create(transfers)
   if (error) {
     switch (error.error) {
@@ -287,6 +309,12 @@ async function transferFunds(
         }
         throw new UnknownLiquidityAccountError(sourceAccount.asset)
       case TransferError.InsufficientBalance:
+        if (
+          destinationAccount.receiveLimitBalanceId &&
+          error.index === transfers.length - 1
+        ) {
+          return AccountTransferError.InvoiceLimitExceeded
+        }
         if (error.index === 1) {
           return AccountTransferError.InsufficientLiquidity
         }
--- a/packages/backend/src/account/service.ts
+++ b/packages/backend/src/account/service.ts
@@ -126,7 +126,9 @@ async function createAccount(
           { transferError },
           'receive limit setup TigerBeetle error'
         )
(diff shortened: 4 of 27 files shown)


packages/backend/src/invoice/service.ts
```diff
@@ -100,6 +123,39 @@ async function createInvoice(
...
+      // Ensure the invoices cannot be processed concurrently by multiple workers.
+      .forUpdate()
+      // If an invoice is locked, don't wait — just come back for it later.
+      .skipLocked()
+      .where('active', true)
+      .andWhere('expiresAt', '<', now)
+    const invoice = invoices[0]
+    if (!invoice) return
+
+    // Fetch the account with a separate transaction to avoid locking it.
```
wilsonianb, line 146: What's the downside of locking the account? The only thing we'd currently patch on it is the `disabled` field.

packages/backend/src/invoice/routes.ts
```diff
@@ -0,0 +1,147 @@
...
+  )
+
+  const { body } = ctx.request
+  if (typeof body !== 'object') return ctx.throw(400, 'json body required')
+  const amountToReceive = tryParseAmount(body['amount'])
+  if (!amountToReceive) return ctx.throw(400, 'invalid amount')
+  const expiresAt = Date.parse(body['expiresAt'] as string)
+  if (!expiresAt) return ctx.throw(400, 'invalid expiresAt')
+  if (typeof body.description !== 'string')
+    return ctx.throw(400, 'invalid description')
```
wilsonianb, line 106: Do we want to make `description` optional?
https://docs.openpayments.dev/invoices#create

packages/backend/src/account/model.ts
```diff
@@ -26,4 +26,9 @@ export class Account extends BaseModel {
   public readonly balanceId!: string
   // TigerBeetle account id tracking amount sent
   public readonly sentBalanceId?: string
+
+  // TigerBeetle account id tracking an invoice's receive limit.
+  // - "credits" is the actual limit.
+  // - "debits" is the progress towards that limit.
```
wilsonianb, line 32:
```suggestion
  // - "debits" is the actual limit.
  // - "credits" is the progress towards that limit.
```

wilsonianb: This comment _should_ be correct but credits and debits are swapped elsewhere in Rafiki

packages/backend/src/invoice/model.ts
```diff
@@ -17,10 +17,12 @@ export class Invoice extends BaseModel {
...
     }
   }
+  public account!: Account
 
   public paymentPointerId!: string // Refers to payment pointer this invoice is for
   public accountId!: string // Refers to the account created for this invoice
   public active!: boolean
   public description!: string
   public expiresAt?: Date
+  public readonly amountToReceive?: bigint
```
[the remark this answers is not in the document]
sentientwaffle: When it is returned on HTTP the API it is labelled `amount`. But I prefer the (slightly more verbose) name internally for clarity. `amount` alone is ambiguous whether it is the amount paid so far, the amount remaining, or the limit.

`amountToReceive` is a mirror of `@interledger/pay`'s `amountToDeliver`. (https://github.com/interledgerjs/interledgerjs/blob/92c2155672e33f22a8e5e311d1ab24be891001cc/packages/pay/src/open-payments.ts#L54-L55)

packages/backend/src/invoice/service.ts
sentientwaffle: You mean locking instead of deleting it? The endpoint to create invoices is unauthenticated, so there's concern about spam (see: #123).

sentientwaffle: Nevermind, `locking` here refers to Postgres.

packages/backend/src/invoice/service.test.ts
```diff
@@ -109,6 +134,60 @@ describe('Invoice Service', (): void => {
...
+    test('Deletes an expired invoice with no money', async (): Promise<void> => {
+      const invoiceId = (
+        await invoiceService.create({
+          paymentPointerId,
+          description: 'Test invoice',
+          expiresAt: new Date(Date.now() - 40_000)
+        })
+      ).id
+      await expect(invoiceService.deactivateNext()).resolves.toBe(invoiceId)
+      expect(await invoiceService.get(invoiceId)).toBeUndefined()
```
wilsonianb, line 187: I tried this, and it looks like the invoice account isn't being deleted.
Adding `.onDelete('CASCADE')` [here](https://github.com/interledger/rafiki/blob/main/packages/backend/migrations/20211022210203_create_invoices_table.js#L5) didn't fix it for me.

```suggestion
      const { id: invoiceId, accountId } =
        await invoiceService.create({
          paymentPointerId,
          description: 'Test invoice',
          expiresAt: new Date(Date.now() - 40_000)
        })
      await expect(invoiceService.deactivateNext()).resolves.toBe(invoiceId)
      expect(await invoiceService.get(invoiceId)).toBeUndefined()
      const accountService = await deps.use('accountService')
      expect(await accountService.get(accountId)).toBeUndefined()
```

wilsonianb (commented):
Even if the account of a money-less expired invoice is deleted, the two tigerbeetle accounts would persist.
Would it be possible to leave the invoice account-less until it has (is about to) received funds, by putting the invoice id in the payment tag?

packages/backend/src/invoice/service.ts
```diff
@@ -98,6 +104,39 @@ async function createInvoice(
...
+    const balance = await deps.accountService.getBalance(invoice.accountId)
+    if (balance) {
+      deps.logger.trace({ invoice: invoice.id }, 'deactivating expired invoice')
+      await invoice.$query(trx).patch({ active: false })
+    } else {
+      deps.logger.debug({ invoice: invoice.id }, 'deleting expired invoice')
+      await invoice.$query(trx).delete()
```
wilsonianb, line 134: This will delete the account and invoice if you add the `.onDelete('CASCADE')` from https://github.com/interledger/rafiki/pull/178#discussion_r745036171
```suggestion
      await invoice.$relatedQuery('account', trx).delete()
```

packages/backend/src/invoice/service.ts
```diff
@@ -19,33 +22,32 @@ export interface InvoiceService {
...
     pagination?: Pagination
   ): Promise<Invoice[]>
+  deactivateNext(): Promise<string | undefined>
 }
 
 interface ServiceDependencies extends BaseService {
+  knex: TransactionOrKnex
   accountService: AccountService
   paymentPointerService: PaymentPointerService
+  transferService: TransferService
```
wilsonianb, line 32: Transfer service is no longer needed.
```suggestion
```

packages/backend/src/account/service.ts
```diff
@@ -102,11 +103,38 @@ async function createAccount(
...
+          // Allow a little extra, to be more forgiving about (favorable) exchange rate fluctuations.
+          amount: account.receiveLimit + BigInt(1)
+        }
+      ])
+      if (transferError) {
+        deps.logger.error(
+          { transferError },
+          'receive limit setup TigerBeetle error'
+        )
+        throw new Error('unable to create invoice, TigerBeetle error')
```
wilsonianb, line 129:
```suggestion
        throw new Error('unable to create receive limit balance, TigerBeetle error')
```

packages/backend/migrations/20211022210203_create_invoices_table.js
```diff
@@ -2,17 +2,32 @@ exports.up = function (knex) {
...
     table.uuid('id').notNullable().primary()
     table.uuid('accountId').notNullable()
-    table.foreign('accountId').references('accounts.id')
+    table.foreign('accountId').references('accounts.id').onDelete('CASCADE')
     table.uuid('paymentPointerId').notNullable()
     table.foreign('paymentPointerId').references('paymentPointers.id')
     table.boolean('active').notNullable()
     table.string('description').nullable()
     table.timestamp('expiresAt').nullable()
+    table.bigInteger('amountToReceive').nullable()
```
wilsonianb, line 11: I'm guessing these are `nullable` for spsp/web monetization?
What if we didn't use invoices for that?
We could put a web monetization account id in the paymentpointers table, and then make invoice `amountToReceive` and `expiresAt` notNullable

packages/backend/src/invoice/routes.test.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  afterAll(
+    async (): Promise<void> => {
+      await resetGraphileDb(knex)
+      await appContainer.shutdown()
+      await workerUtils.release()
+    }
+  )
+
+  describe('get', (): void => {
```
wilsonianb, line 89: You could also validate the response by doing a `pay.setupPayment` with the invoice url, but it looks like pay would need some fixes
https://github.com/interledgerjs/interledgerjs/issues/257

packages/backend/src/invoice/routes.ts
```diff
@@ -0,0 +1,147 @@
...
+    //      }
+    //    : undefined,
+    asset: {
+      code: invoice.account.asset.code,
+      scale: invoice.account.asset.scale
+    }
+  })
+
+  body['ilpAddress'] = ilpAddress
+  body['sharedSecret'] = sharedSecret
```
wilsonianb, line 82: Does this need to be base64 encoded like in spsp?
https://github.com/interledger/rafiki/blob/main/packages/backend/src/spsp/service.ts#L98

packages/backend/migrations/20211022210203_create_invoices_table.js
sentientwaffle: I think the idea of spsp/WM being pseudo-invoices internally is that that is how we want them to appear from the user's perspective. i.e. if we are displaying a transaction history, it would show all of their invoices and WM payment received. And it's much easier to have a unified interface like that if they are all in one table.

wilsonianb: Maybe once we have invoice withdrawals set up, the wallet could just track and display those withdrawals instead of looking up invoices in Rafiki for transaction history.

packages/backend/src/invoice/routes.ts
```diff
@@ -0,0 +1,148 @@
...
+) {
+  const location = `${deps.config.publicHost}/invoices/${invoice.id}`
+  return {
+    id: location,
+    account: `${deps.config.publicHost}/pay/${invoice.paymentPointerId}`,
+    amount: invoice.amountToReceive?.toString(),
+    assetCode: invoice.account.asset.code,
+    assetScale: invoice.account.asset.scale,
+    description: invoice.description,
+    received: received.toString()
```
wilsonianb, line 138:
```suggestion
    expiresAt: invoice.expiresAt?.toISOString(),
    received: received.toString()
```

wilsonianb (changes requested):

wilsonianb (approved):

packages/backend/migrations/20211022210203_create_invoices_table.js
matdehaast: Also `amountToReceive` would be null for fixedSend as you wouldn't necessarily know as the sender how much the invoice should receive.

packages/backend/migrations/20210422194130_create_assets_table.js
```diff
@@ -13,6 +13,8 @@ exports.up = function (knex) {
     table.uuid('settlementBalanceId').notNullable()
     // TigerBeetle account id tracking reserved outgoing payments balance
     table.uuid('outgoingPaymentsBalanceId').notNullable()
+    // TigerBeetle account id tracking cumulative invoice receive limit
+    table.uuid('receiveLimitBalanceId').notNullable()
```
matdehaast, line 17: @sentientwaffle can you explain the need for the extra balance here?

sentientwaffle: Invoices have a new TigerBeetle account that track their progress towards the Invoice's receive limit. That way, the limit is enforced by TigerBeetle. https://github.com/interledger/rafiki/blob/825737fa98f3e9ed072319eac7640f6727b5846f/packages/backend/src/account/service.ts#L306-L314 This account on the asset counters the new one on the invoice/Account.

When the invoice is created, money is transferred from the invoice receive limit account to the asset's corresponding account. When payments to the invoice arrive, money is transferred the other direction. That way, Prepares which would put too much money in the invoice's receive limit account fail.

PR #193, opened 2021-12-01
feat(backend): add API key authentication

### LINKED ISSUE #149: Implement API key authentication
Allow an API key to be associated with an account which can be used to call that account's APIs. API keys could be added to an account by the admin

sabinebertram:

- adds admin API key to environment
- adds table to store bcrypted API keys to postgres db
- adds method to create API keys for user accounts
- adds functionality to exchange API key for session token
- stores session keys in reddis db
- implements graphql directives that check that valid admin API key or valid session key is provided

**! not yet included:** adding directives to graphql schema

## Context

- closes #149

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


wilsonianb:
Here's a merge to resolve the conflicts:
https://github.com/wilsonianb/rafiki/commit/2860b159245e5d4746e2dc25c92851e413e3f5cd

packages/backend/src/apiKey/service.ts
```diff
@@ -0,0 +1,114 @@
...
+    deleteAll: (options, trx) => deleteAllApiKeys(deps, options, trx)
+  }
+}
+
+async function createApiKey(
+  deps: ServiceDependencies,
+  { accountId }: ApiKeyOptions,
+  trx?: Transaction
+): Promise<NewApiKey> {
+  const keyTrx = trx || (await ApiKey.startTransaction(deps.knex))
```
wilsonianb, line 60: Either the commit and rollback should only happen `if (!trx)` (see [invoice creation](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/invoice/service.ts#L82-L89)) or `createApiKey` could be transaction free (`ApiKey.query(deps.knex).insertAndFetch`).
I'd vote for the latter.

packages/backend/migrations/20211022170432_create_apikey_table.js
```diff
@@ -0,0 +1,15 @@
+exports.up = function (knex) {
```
wilsonianb, line 1: file could be renamed to `20211022170432_create_api_keys_table.js`

packages/backend/src/apiKey/service.ts
```diff
@@ -0,0 +1,114 @@
...
+  trx?: Transaction
+): Promise<void> {
+  const keyTrx = trx || (await ApiKey.startTransaction(deps.knex))
+  try {
+    await ApiKey.query(keyTrx).delete().where('accountId', accountId)
+    await keyTrx.commit()
+  } catch (err) {
+    await keyTrx.rollback()
+    throw err
+  }
```
wilsonianb, line 113:
```suggestion
  { accountId }: ApiKeyOptions
): Promise<void> {
  await ApiKey.query(deps.knex).delete().where('accountId', accountId)
```

packages/backend/src/apiKey/service.ts
```diff
@@ -0,0 +1,114 @@
...
+
+async function redeemSessionKey(
+  deps: ServiceDependencies,
+  { accountId, key }: SessionKeyOptions
+): Promise<SessionKey | ApiKeyError> {
+  const keys = await ApiKey.query()
+    .select('hashedKey')
+    .where('accountId', accountId)
+  if (keys && keys.length) {
+    const match = await bcrypt.compare(key, keys[0].hashedKey)
```
wilsonianb, line 93:
```suggestion
  for (const { hashedKey } of keys) {
    const match = await bcrypt.compare(key, hashedKey)
```

packages/backend/src/apiKey/service.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+  )
+
+  describe('Create / Get Api Key', (): void => {
+    test('An api key can be created/fetched for a certain account', async (): Promise<void> => {
+      const apiKeyOptions = { accountId: account.id }
+      const apiKey = await apiKeyService.create(apiKeyOptions)
+      expect(apiKey.key).toBeDefined()
+      expect(apiKey.hashedKey).toBeDefined()
+      expect(apiKey.accountId).toBeDefined()
```
wilsonianb, line 75:
```suggestion
      expect(apiKey.accountId).toEqual(account.id)
```

packages/backend/src/apiKey/service.ts
```diff
@@ -0,0 +1,114 @@
...
+  })
+  const deps: ServiceDependencies = {
+    logger: log,
+    knex,
+    sessionService: sessionKeyService
+  }
+  return {
+    create: (options, trx) => createApiKey(deps, options, trx),
+    get: (options) => getApiKeys(deps, options),
+    redeem: (options) => redeemSessionKey(deps, options),
```
wilsonianb, line 50:
```suggestion
    redeem: (options) => redeemApiKey(deps, options),
```

packages/backend/src/apiKey/service.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      expect(fetchedKeys[0].hashedKey).toEqual(apiKey.hashedKey)
+      expect(fetchedKeys[0].accountId).toEqual(apiKey.accountId)
+      expect(fetchedKeys[0].createdAt).toEqual(apiKey.createdAt)
+      expect(fetchedKeys[0].updatedAt).toEqual(apiKey.updatedAt)
+      expect(fetchedKeys[0]).not.toHaveProperty('key')
+    })
+  })
+
+  describe('Redeem Session Key', (): void => {
+    test('A session key can be redeemed for a valid api key', async (): Promise<void> => {
```
wilsonianb, line 92:
```suggestion
  describe('Redeem Api Key for Session Key', (): void => {
    test('A valid api key can be redeemed for a session key', async (): Promise<void> => {
```

wilsonianb (changes requested):

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,83 @@
...
+    create: () => createSession(deps),
+    revoke: (options: SessionOptions) => revokeSession(deps, options),
+    refresh: (options: SessionOptions) => refreshSession(deps, options),
+    get: (options: SessionOptions) => getSession(deps, options)
+  }
+}
+
+async function createSession(deps: ServiceDependencies): Promise<SessionKey> {
+  const sessionKey = uuid()
+  const expiresAt = Date.now() + 30 * 60 * 1000 // 30 minutes
```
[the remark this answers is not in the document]
sabinebertram: Configurable per Rafiki deploy or configurable per account?

packages/backend/src/apiKey/service.ts
sabinebertram: I just looked up the meaning of "redeem" and it looks like I had it backwards. Thanks!

packages/backend/src/apiKey/service.ts
sabinebertram: I chatted to @sharafian about that at a point and he said I should make it consistent with what we do in other services. But I forgot to go back and check if all services to that. I checked now, they don't. And I agree, it is not really necessary to pass `trx` because I don't see a scenario where we need to create API keys atomically.

packages/backend/src/session/service.ts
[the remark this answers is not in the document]
wilsonianb: I was thinking per Rafiki deploy, but I'm not sure either way (hard coded or per Rafiki or per account).

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+async function refreshSession(
+  deps: ServiceDependencies,
+  { key }: SessionOptions
+): Promise<SessionKey | SessionError> {
+  const sessionOrError = await getSession(deps, { key })
+  if (isSessionError(sessionOrError)) {
+    return sessionOrError
+  } else {
+    if (sessionOrError.expiresAt > new Date(Date.now())) {
```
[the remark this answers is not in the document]
sabinebertram: I guess that means we don't need `SessionError` at all because there either is a running session or there isn't.

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,83 @@
...
+    revoke: (options: SessionOptions) => revokeSession(deps, options),
+    refresh: (options: SessionOptions) => refreshSession(deps, options),
+    get: (options: SessionOptions) => getSession(deps, options)
+  }
+}
+
+async function createSession(deps: ServiceDependencies): Promise<SessionKey> {
+  const sessionKey = uuid()
+  const expiresAt = Date.now() + 30 * 60 * 1000 // 30 minutes
+  await deps.redis.set(sessionKey, JSON.stringify({ expiresAt }), 'EX', 30 * 60)
```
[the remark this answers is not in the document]
sabinebertram: I don't know if I understand what you mean.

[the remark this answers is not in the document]
wilsonianb: https://stackoverflow.com/questions/25557250/redis-store-key-without-a-value
```suggestion
  await deps.redis.set(sessionKey, 0, 'EX', 30 * 60)
```
Then `getSession` can just call
```
await deps.redis.ttl(sessionKey)
```
and `refreshSession` might be able to just call
```
await deps.redis.expire(sessionKey, 30*60)
```

packages/backend/src/graphql/resolvers/session.test.ts
```diff
@@ -0,0 +1,166 @@
...
+              if (query.data) {
+                return query.data.revokeSession
+              } else {
+                throw new Error('Data was empty')
+              }
+            }
+          )
+
+        expect(response.success).toBe(true)
+        expect(response.code).toEqual('200')
```
wilsonianb, line 162: You could call 
```
sessionService.get({
  key: sessionOrError.key
})
```
to confirm it was revoked

packages/backend/src/graphql/schema.graphql
```diff
@@ -77,6 +77,25 @@ type Mutation {
...
+
+  "Delete all API Keys"
+  deleteAllApiKeys(
+    input: DeleteAllApiKeysInput!
+  ): DeleteAllApiKeysMutationResponse
+
+  "Redeem Session Key"
+  redeemSessionKey(
+    input: RedeemSessionKeyInput!
+  ): RedeemSessionKeyMutationResponse
```
wilsonianb, line 92:
```suggestion
  "Redeem API Key"
  redeemApiKey(
    input: RedeemApiKeyInput!
  ): RedeemApiKeyMutationResponse
```

wilsonianb (changes requested):

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,86 @@
...
+import { uuid } from '../connector/core'
+import IORedis from 'ioredis'
+import { isSessionError, SessionError } from './errors'
+import { Session, SessionKey } from './util'
+
+export interface SessionService {
+  create(): Promise<SessionKey>
+  revoke(session: SessionOptions): void
+  refresh(session: SessionOptions): Promise<SessionKey | SessionError>
+  get(session: SessionOptions): Promise<Session | SessionError>
```
wilsonianb, line 11: What do you think about removing `SessionOptions` in favor of
```suggestion
  revoke(key: string): void
  refresh(key: string): Promise<SessionKey | SessionError>
  get(key: string): Promise<Session | SessionError>
```

packages/backend/src/session/service.ts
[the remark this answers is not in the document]
sabinebertram: Thank you for clarifying! That is smart.

packages/backend/src/session/service.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        key: session.key
+      })
+      expect(retrievedSession).toEqual({ expiresAt: session.expiresAt })
+    })
+
+    test('Cannot fetch non-existing session', async (): Promise<void> => {
+      const sessionOrError = SessionService.get({ key: '123' })
+      expect(sessionOrError).resolves.toEqual(SessionError.UnknownSession)
+    })
+  })
```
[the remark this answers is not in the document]
wilsonianb: :thinking:  I guess jest mock timers wouldn't work with redis.
You _could_ use [ioredis-mock](https://www.npmjs.com/package/ioredis-mock), but I gave up trying to use it with the receipt verifier.
Instead, you could lookup the session's ttl in redis to confirm it's set to expire.

wilsonianb (changes requested):

packages/backend/src/session/service.test.ts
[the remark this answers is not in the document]
sabinebertram: >  Instead, you could lookup the session's ttl in redis to confirm it's set to expire.

That is probably the easiest. Thanks!

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,78 @@
...
+    expiresAt: new Date(Date.now() + sessionExpiry)
+  }
+}
+
+async function revokeSession(
+  deps: ServiceDependencies,
+  { key }: SessionOptions
+) {
+  deps.redis.del(key)
+}
```
wilsonianb, line 55:
```suggestion
async function revokeSession(
  deps: ServiceDependencies,
  { key }: SessionOptions
): Promise<void> {
  await deps.redis.del(key)
}
```

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,78 @@
...
+
+async function refreshSession(
+  deps: ServiceDependencies,
+  { key }: SessionOptions
+): Promise<Session | undefined> {
+  await deps.redis.expire(key, 30 * 60)
```
wilsonianb, line 61: This could check the response to know if the session doesn't exist.
Also what if here and `createSession` calculated `expiresAt` instead of calling `pttl`?
(It looks like we could use [`pexpiretime`](https://redis.io/commands/expiretime) when redis 7 is out.)

packages/backend/src/session/service.test.ts
```diff
@@ -0,0 +1,97 @@
...
+          retrievedSession.expiresAt.getTime() - session.expiresAt.getTime()
+        ).toBeGreaterThanOrEqual(-1)
+      } else {
+        fail()
+      }
+    })
+
+    test('Cannot fetch non-existing session', async (): Promise<void> => {
+      const sessionOrError = sessionService.get({ key: '123' })
+      expect(sessionOrError).resolves.toBeUndefined()
```
wilsonianb, line 60: There's a few more stray `OrError`s below
```suggestion
      const session = sessionService.get({ key: '123' })
      expect(session).resolves.toBeUndefined()
```

packages/backend/src/session/service.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (!refreshedSession) {
+        fail()
+      } else {
+        expect(session.key).toEqual(refreshedSession.key)
+        expect(
+          session.expiresAt.getTime() - refreshedSession.expiresAt.getTime()
+        ).toBeLessThanOrEqual(1)
+        expect(
+          session.expiresAt.getTime() - refreshedSession.expiresAt.getTime()
+        ).toBeGreaterThanOrEqual(-1)
```
wilsonianb, line 88: Since the returned `expiresAt` is calculated using `Date`, this test could use [`setSystemTime`](https://jestjs.io/docs/jest-object#jestsetsystemtimenow-number--date) between `create` and `refresh` to check that the expiration has been extended.

packages/backend/src/session/service.ts
sabinebertram: Don't all services have `options`? I wanted to make it consistent.

packages/backend/src/session/service.ts
sabinebertram: Using `expireAt` is probably cleaner and makes testing easier, too. Thanks!

packages/backend/src/session/service.ts
wilsonianb: Yeah, a lot do especially for `create`.
I was just thinking that since there's just one option here, to deal with it as the string.

packages/backend/src/session/service.ts
sabinebertram: Or I just store the unix timestamp as value in the meantime.

wilsonianb: Yeah, maybe `create` and `refresh` should just calculate the timestamp and pass it to:
```
deps.redis.set(key, sessionExpiry, 'PXEX', sessionExpiry)
```

and `get` can do  
```
deps.redis.get(key)
```

Then if/when there's redis 7, we can go back to an empty value.
`refresh` can call:
```
deps.redis.pexpire(key, sessionExpiry)
```
and `get` can use:
```
deps.redis.pexpiretime(key)

File: packages/backend/src/session/service.test.ts [hunk 19]
```diff
@@ -0,0 +1,95 @@
...
+      if (retrievedSession) {
+        expect(retrievedSession).toEqual(session)
+      } else {
+        fail()
+      }
+    })
+
+    test('Cannot fetch non-existing session', async (): Promise<void> => {
+      const session = sessionService.get({ key: '123' })
+      console.log(session)
```
wilsonianb, line 54:
```suggestion
```

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,88 @@
...
+    refresh: (options: SessionOptions) => refreshSession(deps, options),
+    get: (options: SessionOptions) => getSession(deps, options)
+  }
+}
+
+async function createSession(deps: ServiceDependencies): Promise<Session> {
+  const key = uuid()
+  const expiry = Date.now() + 30 * 60 * 1000
+  await deps.redis.set(key, expiry)
+  await deps.redis.pexpireat(key, expiry)
```
wilsonianb, line 44:
```suggestion
  await deps.redis.set(key, expiry, 'PXAT', expiry)
```

packages/backend/src/session/service.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+async function refreshSession(
+  deps: ServiceDependencies,
+  { key }: SessionOptions
+): Promise<Session | undefined> {
+  const session = await deps.redis.get(key)
+  if (session) {
+    const expiry = Date.now() + 30 * 60 * 1000
+    await deps.redis.pexpireat(key, expiry)
```
wilsonianb, line 65:
```suggestion
    await deps.redis.set(key, expiry, 'PXAT', expiry)
```

packages/backend/src/session/service.ts
sabinebertram: I didn't know that existed. Thanks!

packages/backend/src/session/service.ts
sabinebertram: I don't have a strong opinion on that. I'm happy to change it if you think it makes more sense.

packages/backend/src/session/service.ts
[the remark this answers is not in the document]
sabinebertram: I can just make it configurable per deploy for now and then we can come back to that later.

wilsonianb (approved):

packages/backend/src/graphql/directives/isAdmin.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        schema: schemaWithDirectives,
+        context: ({ req }): { admin: boolean } => {
+          return {
+            admin: req.headers['x-api-key'] == 'qwertyuiop1234567890'
+          }
+        }
+      })
+      await server.listen(3010)
+      const httpLink = createHttpLink({
+        uri: 'http://localhost:3010/graphql',
```
wilsonianb, line 60: I got a random `Error: listen EADDRINUSE: address already in use :::3010` one time while running the tests.
Can this be changed to a different port than the auth directirve test's server?

wilsonianb (changes requested):

wilsonianb (approved):
:+1:

sharafian (approved):

Issue #160, opened 2021-10-18
Add outgoing payment liquidity mutation resolvers


wilsonianb:
Add and remove outgoing payment liquidity

wilsonianb:
This would be obviated by https://github.com/interledger/rafiki/pull/197

PR #197, opened 2021-12-03
chore(backend): remove payment balance accounts

### LINKED ISSUE #196: Remove outgoing payment accounts
### Discussed in https://github.com/interledger/rafiki/discussions/190

<div type='discussions-op-text'>

<sup>Originally posted by **wilsonianb** November 30, 2021</sup>
- related to https://github.com/interledger/rafiki/issues/156

For :point_up: accounts now only exist in Rafiki for:
- peers
- asset liquidity
- invoices
- outgoing payments

I'm wondering if there needs to be accounts for outgoing payments.
Instead of creating a tigerbeetle account and adding liquidity to it, we could send directly from the settlement account. The flow would be:

1. createPayment
2. handleQuoting
3. Rafiki notifies wallet of quote
4. Wallet gets user approval (if necessary) and reserves the quoted amount in the sender's wallet account
4. sendPayment -> Rafiki sends payment from settlement account
5. Rafiki notifies wallet of payment outcome

Not having a payment account avoids having an outstanding balance in Rafiki resulting from either a failed send or a successful fixed delivery payment.
The payment outcome notification allows the wallet to return the leftover amount to the sender or claim as fees.</div>

wilsonianb:

- ~Send outgoing payments from and receive invoice payments to settlement accounts.~
- ~Outgoing payments and invoices have accounts for tracking the amounts
sent and received, but they do not represent (withdraw-able) balances.~
- Remove outgoing payments and invoices accounts used for tracking the amounts
sent and received.
- Add ~sendOutgoingPayment~ `fundOutgoingPayment` mutation resolver. Wallet operators will be
notified when quote is complete (TODO) and call `fundOutgoingPayment`
after payment is approved (if necessary) and funds are reserved in
the sender's wallet account.
- Make incoming and outgoing accounts distinct types
- ~Rename `Funding` payment state as `Ready`~
- Replace web monetization service with Open Payments account spsp fallback tigerbeetle account. (All invoices are now Open Payments invoices.)
- Do not update account balances when sending unfulfillable rate probe packets.

## Context

- ~#196~
- part of #160 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/connector/core/middleware/balance.ts
```diff
@@ -52,9 +52,9 @@ export function createBalanceMiddleware(): ILPMiddleware {
           throw new InsufficientLiquidityError(trxOrError)
         case TransferError.ReceiveLimitExceeded: {
           const receivedAmount = destinationAmountOrError.toString()
-          assert.ok(accounts.outgoing.id)
-          const receiveLimit = await services.accounts.getReceiveLimit(
-            accounts.outgoing.id
+          assert.ok(accounts.outgoing.receivedAccountId)
+          const receiveLimit = await services.accounts.getBalance(
+            accounts.outgoing.receivedAccountId
```
sentientwaffle, line 57: The naming of `receivedAccountId` confuses me. It isn't the amount that has been received, it is the maximum amount that _can_ be received, right?

wilsonianb: It's confusing because it depends:
https://github.com/interledger/rafiki/pull/197/files#diff-e5a9751319994ef97d1413f0c3bc6587f756163958ad21f877a5fe7d84a08661R81-R89

I considered having separate `receivedAccountId` and `receiveLimitAccountId`, but ~`transferFunds`~`sendAndReceive` would treat them the same.

packages/backend/src/graphql/schema.graphql
```diff
@@ -302,6 +304,7 @@ type PaymentQuote {
   minExchangeRate: Float!
   lowExchangeRateEstimate: Float!
   highExchangeRateEstimate: Float!
+  amountSent: UInt64!
```
sentientwaffle, line 307: Why does this parameter need to be exposed on the API?

If it does need to be exposed, it should have a more descriptive name to ensure that the caller doesn't assume that it should be subtracted from (e.g.) the `maxSourceAmount`.

wilsonianb: It doesn't need to be exposed since it's in the `OutgoingPaymentOutcome`

sentientwaffle (dismissed):

wilsonianb:
@matdehaast Here's the possible webhooks for reference https://github.com/interledger/rafiki/pull/199

packages/backend/src/open_payments/invoice/routes.ts
```diff
@@ -50,12 +50,15 @@ async function getInvoice(
...
-  if (amountReceived === undefined) {
+  const balance = await deps.accountingService.getBalance(invoice.id)
+  if (balance === undefined) {
     deps.logger.error({ invoice: invoice.id }, 'balance not found')
     return ctx.throw(500)
   }
 
+  const amountReceived = invoice.amountToReceive
+    ? invoice.amountToReceive + POSITIVE_SLIPPAGE - balance
+    : balance
```
matdehaast, line 61: I'm wondering if we should have a function in the invoice service rather to account for this, if somewhere else wants to quote the invoice balance, they will need to know this has to happen

matdehaast (commented):
This is good work @wilsonianb.

I do still worry that we are losing some accounting principals and accurate tracking of money movement by going in this direction. But I can see that it does simplify things greatly.

My assumption is that this is all to account for the need to enforce limits on certain payment accounts? 

I wonder if there is a better abstraction where we can have another primitive within the account service that is called a limit account or something along those lines and we could bind that to an account. I'm trying to draw up some T accounts for this PR to see if we actually lose anything meaningful.

packages/backend/src/accounting/accounts.ts
```diff
@@ -31,7 +31,9 @@ export async function createAccounts(
       flags:
         account.type === AccountType.Debit
           ? AccountFlags.credits_must_not_exceed_debits
-          : AccountFlags.debits_must_not_exceed_credits,
+          : account.type === AccountType.Credit
+          ? AccountFlags.debits_must_not_exceed_credits
+          : 0,
```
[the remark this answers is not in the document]
wilsonianb: :+1:
A reason I didn't add and expose one is that currently the only non-`Debit`/`Credit` account is being created within the accounting service (in `createAssetAccounts`).

packages/backend/src/open_payments/invoice/service.ts
```diff
@@ -72,13 +78,44 @@ async function createInvoice(
...
+    //
+    // Otherwise, the invoice account balance represents the total amount
+    // received by the invoice.
     await deps.accountingService.createAccount({
       id: invoice.id,
       asset: invoice.account.asset,
-      type: AccountType.Credit,
-      receiveLimit: amountToReceive
+      type: amountToReceive ? AccountType.Debit : AccountType.Credit
     })
```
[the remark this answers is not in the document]
wilsonianb: Yes.
I'd like to separate web monetization from invoices and make the invoice model follow the open payments spec which would make `amountToReceive` a required field.

packages/backend/src/open_payments/invoice/routes.ts
wilsonianb: I started to add a function yesterday and then I added `getTotalReceived` to the `accountingService`. https://github.com/interledger/rafiki/pull/199/commits/2ce777887690604405c2dea93b5f5354b6aa6708
I'll put that in this pr.

packages/backend/src/accounting/service.ts
```diff
@@ -85,10 +70,12 @@ interface AssetAccountOptions {
...
     account: AssetAccount
   }
-  withBalance?: never
 }
 
-export type AccountOptions = BasicAccountOptions | AssetAccountOptions
+export type AccountOptions = (BasicAccountOptions | AssetAccountOptions) & {
+  receivedAccountId?: string
+  sentAccountId?: string
+}
```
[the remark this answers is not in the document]
wilsonianb: It looks like this comment is on an outdated commit. Here's the final form:
https://github.com/interledger/rafiki/pull/197/files#diff-72a5530c2c89f105f75e071479a382e8f8408459d37968fb93b546cca4783b31R75-R91
There's currently not a situation in which an account would have both a received and sent account id.
With this previous iteration, it could have been consolidated to one account whose type could be checked, but I like how it fit into separating the connector's incoming and outgoing account types.
https://github.com/interledger/rafiki/pull/197/files#diff-08011ba8622ef9b1a38af778014f9471eb8ce942adc8609074f019765f7338f6R26-R43

wilsonianb:
>My assumption is that this is all to account for the need to enforce limits on certain payment accounts?

We were already effectively enforcing limits on payments.
This change accomplishes limit enforcement in a simpler way (one tigerbeetle account per invoice and outgoing payment), while giving Rafiki fewer limits to enforce.

For outgoing payments, there's no longer a liquidity balance to enforce. The amount sent is restricted by `@interledger/pay` with its quote.

Invoices do still enforce the receive limit, but there's similarly no enforcement of an invoice's withdrawable liquidity. It has an amount received that can only go up, which the wallet can reflect in the receiver's wallet account however it wants.

wilsonianb:
This won't work as is for an intra-rafiki same asset outgoing payment to invoice. It'll get `TransferError.SameAccounts` for the settlement account sending to itself.
I'm working on a change to fix.

packages/backend/src/open_payments/invoice/service.ts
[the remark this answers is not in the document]
matdehaast: @wilsonianb ah I see that is a required field. I can't for the life of me remember why. I wonder how fixed send amounts are handled if amount is required...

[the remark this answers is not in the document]
wilsonianb: https://github.com/interledger/open-payments/issues/39

wilsonianb:
:point_up: that fixes the same account issue.
The pull request should be good to review.
@matdehaast @sentientwaffle

packages/backend/src/accounting/utils.ts
```diff
@@ -1,10 +1,9 @@
-import { AssetAccount, Balance } from './service'
+import { AssetAccount } from './service'
 
-const ASSET_ACCOUNTS_RESERVED = 32
+const ASSET_ACCOUNTS_RESERVED = 8
```
sentientwaffle, line 3: What is this constant for?

packages/backend/src/graphql/resolvers/outgoing_payment.ts
```diff
@@ -119,6 +118,28 @@ export const requoteOutgoingPayment: MutationResolvers<ApolloContext>['requoteOu
     }))
 }
 
+export const sendOutgoingPayment: MutationResolvers<ApolloContext>['sendOutgoingPayment'] = async (
+  parent,
+  args,
+  ctx
+): ResolversTypes['OutgoingPaymentResponse'] => {
```
sentientwaffle, line 125: If the payment state is not `Ready` should that be a client (400) error?

packages/backend/src/accounting/utils.ts
wilsonianb: It's used in generating deterministic asset account ids:
https://github.com/interledger/rafiki/blob/main/packages/backend/src/accounting/utils.ts#L28-L30
We're now down two types of asset accounts (liquidity & settlement).

I was actually just considering replacing it with:
```
const MAX_ASSETS = 2**16
```
based on Tigerbeetle's two byte unit field. And then `getAssetAccountId` could do:
```
BigInt(type * MAX_ASSETS + unit)

Comment (sentientwaffle) [reply packages/backend/src/accounting/utils.ts:3]: Couldn't miscellaneous account uuids get generated into that range?

Comment (wilsonianb) [reply packages/backend/src/accounting/utils.ts:3]: Yeah, which is why I was still favoring `ASSET_ACCOUNTS_RESERVED` over `MAX_ASSETS` since the range is limited by the number of assets.

However, I _think_ we would safely fail to create an asset if it's id is already used.
https://github.com/interledger/rafiki/blob/main/packages/backend/src/asset/service.ts#L57-L68
We'd end up with a unit sequence gap, and retrying would attempt to create the asset with the next unit value.

Comment (wilsonianb) [reply packages/backend/src/accounting/utils.ts:3]: I also had the thought that `validateId` in `accountingService.createAccount` could disallow ids in the possible asset id range.

File: packages/backend/src/graphql/resolvers/outgoing_payment.ts (hunk 8 above)
Comment (wilsonianb) [reply packages/backend/src/graphql/resolvers/outgoing_payment.ts:125]: Yeah, and maybe the same for `requoteOutgoingPayment` and `cancelOutgoingPayment`.

Should we define a `WrongPaymentState` error that the resolvers can catch and return `400`?

Comment (sentientwaffle) [reply packages/backend/src/graphql/resolvers/outgoing_payment.ts:125]: Yeah. The alternative would be to check the state in the resolver itself, but I think that validation logic is better off in the service (where it already is).

**Review by sentientwaffle — DISMISSED** [review]

**Comment by wilsonianb** [comment]
After talking with Matt, I'm in agreement that sending from and receiving to the settlement account prevents the wallet operator from getting an accurate snapshot of liabilities.
We also determined that we can still get by with a single tigerbeetle account for each payment:
- Invoices can be deactivated when they've received their full amount (a la https://github.com/interledger/rafiki/pull/199/commits/9b68ea501ce0a8045584fd34889623e9b2144fd9). The limit doesn't need to be enforced by Tigerbeetle.
- Outgoing payment total sent amount can be tracked with Tigerbeetle transfer codes. (since total debits won't work due to withdrawals)

I'll convert the payment accounts back to normal in this pr and do liquidity deposits/withdrawals and webhooks in separate pr(s).

File: packages/backend/src/open_payments/invoice/service.ts [hunk 9]
```diff
@@ -94,6 +98,25 @@ async function createInvoice(
...
+    invoiceId
+  )
+  if (!amountReceived) {
+    return
+  }
+  const invoice = await Invoice.query(deps.knex).findById(invoiceId)
+  if (invoice.amount <= amountReceived) {
+    await invoice.$query(deps.knex).patch({
+      active: false
+    })
```
sentientwaffle, line 115: Would it cause problems if this query + patch are in separate transactions?

packages/backend/src/outgoing_payment/service.ts
```diff
@@ -154,32 +154,79 @@ async function createOutgoingPayment(
...
+
+async function fundPayment(
+  deps: ServiceDependencies,
+  { id, amount, transferId }: FundOutgoingPaymentOptions
+): Promise<OutgoingPayment | OutgoingPaymentError> {
+  return deps.knex.transaction(async (trx) => {
+    const payment = await OutgoingPayment.query(trx)
+      .findById(id)
+      .forUpdate()
+      .withGraphFetched('account.asset')
```
sentientwaffle, line 183: Can you double-check that the `forUpdate` doesn't also get carried over the the `account.asset` being fetched? (i.e. that the payment row is locked but not the account/asset). I know we do this elsewhere too but it didn't occur to me until now...

wilsonianb: https://github.com/Vincit/objection.js/issues/1608
I'm guessing it would for `withGraphJoined`

packages/backend/src/open_payments/invoice/service.ts
wilsonianb: I don't think so, but this could be put in a single
```
    await Invoice.query(deps.knex).patch({
      active: false
    }).where('id', invoiceId).andWhere('amount', '<=', amountReceived)
```

wilsonianb: I ran into objection/knex type errors trying to compare with a bigint in `andWhere` so I've put the query and patch in a transaction https://github.com/interledger/rafiki/pull/197/commits/88ba6d1f2c0cb97ca9cdbaaac747fa8b548798dd

sentientwaffle: Did you try passing the `toString` of the bigint?

wilsonianb: Nope :upside_down_face: 
https://github.com/interledger/rafiki/pull/197/commits/6478831124e22700d05251d02dc131a7b3b3618b

sentientwaffle (approved):

PR #214, opened 2022-01-12
fix(backend): manage payment liquidity on webhook response

### LINKED ISSUE #210: Fix webhook concurrency
Outgoing payment webhook events are currently sent while the outgoing payment's row is locked.
The wallet's webhook event handler is expected to call a corresponding liquidity mutation resolver (such as `fundOutgoingPayment`) which needs to be able to modify the payment row.

This can be resolved by removing payment liquidity mutation resolvers and managing the liquidity when receiving a successful webhook response from the wallet.
(Webhook sending should also happen in standalone payment lifecycle state handlers instead of at the end of the previous state in order to support webhook retries.)

The same can be done for invoices.

Resolvers to remove:
- `fundOutgoingPayment`
- `cancelOutgoingPayment`
- `requoteOutgoingPayment`
- `createOutgoingPaymentWithdrawal`
- `createInvoiceWithdrawal`

wilsonianb:

- Remove payment and invoice liquidity management mutation resolvers in favor of managing liquidity on webhook responses.
- Retry sending webhook events (using same webhook id).
- Withdraw payment liquidity on requote. (Payment always has empty account balance when quoting)

## Context

- fixes #210 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


docs/transaction-api.md
```diff
@@ -156,8 +174,8 @@ The intent must include `invoiceUrl` xor (`paymentPointer` and `amountToSend`).
 
 ### `PaymentState`
 
-- `QUOTING`: Initial state. In this state, an empty payment account is generated, and the payment is automatically resolved & quoted. On success, transition to `FUNDING` or `SENDING` if already funded. On failure, transition to `Cancelled`.
-- `FUNDING`: Awaiting the wallet to add payment liquidity. If `intent.autoApprove` is not set, the wallet gets user approval before reserving money from the user's wallet account. On success, transition to `Sending`. Otherwise, transitions to `Cancelled` when the quote expires.
+- `QUOTING`: Initial state. In this state, an empty payment account is generated, and the payment is automatically resolved & quoted. On success, transition to `FUNDING`. On failure, transition to `Cancelled`.
+- `FUNDING`: Awaiting the wallet to add payment liquidity. If `intent.autoApprove` is not set, the wallet gets user approval before reserving money from the user's wallet account. On success, transition to `Sending`. Otherwise, transitions to `Cancelled` if cancelled by the user or when the quote expires.
```
sentientwaffle, line 178: Use consistent capitalization of the state names, e.g.

```suggestion
- `FUNDING`: Awaiting the wallet to add payment liquidity. If `intent.autoApprove` is not set, the wallet gets user approval before reserving money from the user's wallet account. On success, transition to `SENDING`. Otherwise, transitions to `CANCELLED` if cancelled by the user or when the quote expires.
```

packages/backend/src/open_payments/invoice/model.ts
```diff
@@ -25,4 +26,14 @@ export class Invoice extends BaseModel {
...
   public expiresAt!: Date
   public readonly amount!: bigint
+  // The "| null" is necessary so that `$beforeUpdate` can modify a patch to remove the webhookId. If `$beforeUpdate` set `webhookId = undefined`, the patch would ignore the modification.
+  public webhookId?: string | null
+  public webhookAttempts!: number
+
+  $beforeUpdate(opts: ModelOptions, queryContext: QueryContext): void {
+    super.$beforeUpdate(opts, queryContext)
+    if (opts.old && opts.old['active'] === true && this.active === false) {
+      this.webhookId = uuid()
```
sentientwaffle, line 36: Why reassign the `webhookId` instead of just assigning it once (or using the invoice id)?

packages/backend/src/open_payments/invoice/service.ts
```diff
@@ -4,10 +4,13 @@ import { BaseService } from '../../shared/baseService'
 import { Pagination } from '../../shared/pagination'
 import { EventType, WebhookService } from '../../webhook/service'
 import assert from 'assert'
-import { Transaction } from 'knex'
+import { QueryBuilder, Transaction } from 'knex'
 import { ForeignKeyViolationError, TransactionOrKnex } from 'objection'
+import { v4 as uuid } from 'uuid'
 
 export const POSITIVE_SLIPPAGE = BigInt(1)
+// First retry waits 10 seconds, second retry waits 20 (more) seconds, etc.
```
sentientwaffle, line 12: And note that 30 follows 20 (i.e. the increase on retry is linear, not geometric).

packages/backend/src/open_payments/invoice/service.ts
```diff
@@ -110,62 +113,146 @@ async function handleInvoicePayment(
...
       .limit(1)
       // Ensure the invoices cannot be processed concurrently by multiple workers.
       .forUpdate()
       // If an invoice is locked, don't wait — just come back for it later.
       .skipLocked()
-      .where('active', true)
-      .andWhere('expiresAt', '<', now)
+      .where((builder: QueryBuilder) => {
+        builder.where('active', true).andWhere('expiresAt', '<', delayedNow)
+      })
```
sentientwaffle, line 142: The index on `expiresAt` should be updated to include `active`. And include a `TODO` that once knex/objection support partial indexes we should switch to indexing it on just `active:true`.

packages/backend/src/open_payments/invoice/service.ts
```diff
@@ -110,62 +113,146 @@ async function handleInvoicePayment(
...
+      .orWhere((builder: QueryBuilder) => {
+        builder.whereNotNull('webhookId').andWhere((builder: QueryBuilder) => {
+          builder
+            .where('webhookAttempts', 0)
+            // Back off between retries.
+            .orWhereRaw(
+              '"updatedAt" + LEAST("webhookAttempts", 6) * ? * interval \'1 seconds\' < ?',
+              [RETRY_BACKOFF_SECONDS, now]
+            )
+        })
```
sentientwaffle, line 152: How can we ensure that this query executes efficiently?

I know that this is inspired by `OutgoingPayment`'s job queue, but in that case it relies on the index of the payment `state` and the fact that at any given time, most payments are either completed or cancelled.

Can we rely on `active` being `true` here, or not? If so then indexing on that would get a similar benefit as `state`. If not then we might have to store a `nextAttempt` timestamp, and index on that (rather than using `updatedAt` & `attempts`).

docs/transaction-api.md
wilsonianb: It looks like I've still missed a few, but now I'm wondering if either the state should be made all caps in webhook events or mixed case in graphql.

packages/backend/src/open_payments/invoice/service.ts
```diff
@@ -200,34 +185,34 @@ async function handleDeactivated(
...
-  if (error) throw error
+    deps.logger.trace(
+      { amountReceived },
+      'withdrawing deactivated invoice balance'
+    )
+    const error = await deps.accountingService.createWithdrawal({
+      id: invoice.id,
+      accountId: invoice.id,
+      amount: amountReceived,
+      timeout: BigInt(deps.webhookService.timeout) * BigInt(1e9) // ms -> ns
```
sentientwaffle, line 210: If `deps.webhookService.timeout` is milliseconds, doesn't `* 1e9` convert to picoseconds?

docs/transaction-api.md
wilsonianb: I went with all caps based on 
http://spec.graphql.org/October2021/#sec-Enum-Value
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

sentientwaffle (approved):

PR #239, opened 2022-02-18
feat(backend): add web monetization webhook event

### LINKED ISSUE #220: Web monetization liquidity withdrawal
- Allow balance lookup
- Configure time and/or received amount based withdrawal webhook events.

wilsonianb:

- Add `account.web_monetization` webhook event for discrete (non-cumulative) amounts to be withdrawn
- Add optional asset `withdrawalThreshold` to support automatic web monetization balance withdrawals
- Add optional `withdrawalThrottleDelay` to configuration for scheduling web monetization balance withdrawals upon receiving any amount
- Add asset query and mutation resolvers
- Add `triggerAccountEvents` mutation resolver

## Context

- fixes #220 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/app.ts
```diff
@@ -253,6 +256,24 @@ export class App {
     this.koa.use(this.publicRouter.middleware())
   }
 
+  private async processAccount(): Promise<void> {
```
sentientwaffle, line 259: There are enough of these jobs that it might be worth abstracting the job processing a bit so that these `processFoo` functions are less repetitive. (Doesn't need to be in this PR though).

packages/backend/src/open_payments/account/service.ts
```diff
@@ -69,3 +71,82 @@ async function getAccount(
...
+        totalEventsAmount: account.totalEventsAmount
+      },
+      'no amount to withdrawal'
+    )
+    return
+  }
+
+  deps.logger.trace({ amount }, 'creating webhook withdrawal event')
+
+  await AccountEvent.query(deps.knex).insertAndFetch({
```
sentientwaffle, line 139: Just insert, don't need to fetch.

packages/backend/src/open_payments/account/service.ts
```diff
@@ -69,3 +71,82 @@ async function getAccount(
...
+
+    await account.$query(deps.knex).patch({
+      processAt: null
+    })
+
+    return account.id
+  })
+}
+
+async function createWithdrawalEvent(
```
sentientwaffle, line 112: Add a comment noting that for this to be safe, the provided `account` *must* be fetched with `deps.knex`'s transaction.

packages/backend/src/open_payments/account/model.ts
```diff
@@ -25,4 +27,64 @@ export class Account
...
+          totalReceived - this.asset.withdrawalThreshold
+        ])
+        .withGraphFetched('asset')
+      if (account) {
+        return account
+      }
+    }
+    if (withdrawalCron && !this.processAt) {
+      await this.$query().patch({
+        processAt: parser.parseExpression(withdrawalCron).next().toDate()
```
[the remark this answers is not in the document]
wilsonianb: I was picturing the wallet being able to configure regularly scheduled withdrawals like every day at 5pm EST.
Maybe `withdrawalThrottleDelay` could be added in addition to cron (would they need to be mutually exclusive?)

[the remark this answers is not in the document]
sentientwaffle: I think we should try to minimize the handling logic for this in Rafiki, since it is so wallet-specific. Regularly scheduled withdrawals are really better suited to a pull/query model (i.e. the wallet could query all accounts with money to withdraw), rather than webhooks/events.

What about a hybrid? i.e. use the `processAt: new Date(Date.now() + config.withdrawalThrottleDelay`, but allow the wallet operator to disable starting any `processNextAccount` workers when they configure Rafiki. Expose an endpoint that, when called, queries accounts with `now < processAt` and triggers `n` of them (for some integer parameter `n`).

That gives a bit more control to the wallet operator. It's less elegant than a pull model would be, but it reuses more logic from the push model.

packages/backend/src/config/app.ts
```diff
@@ -73,7 +73,10 @@ export const Config = {
   webhookSecret: process.env.WEBHOOK_SECRET, // optional
   webhookTimeout: envInt('WEBHOOK_TIMEOUT', 2000), // milliseconds
 
-  webMonetizationWithdrawalCron: process.env.WEB_MONETIZATION_WITHDRAWAL_CRON, // optional
+  withdrawalThrottleDelay:
```
sentientwaffle, line 76: Why not use `envInt`?

sentientwaffle (approved):

packages/backend/src/config/app.ts
wilsonianb: I wanted `withdrawalThrottleDelay` to be optional without a default (`number | undefined`) and didn't want to change `envInt`'s return type

PR #242, opened 2022-02-24
chore(backend): abstract pagination support and tests

### LINKED ISSUE #240: Consolidate pagination implementation and tests
There is an increasing number of services with a [pagination method](https://github.com/interledger/rafiki/blob/8efb70b4dd02afec86f8283648840f18bbef9222/packages/backend/src/peer/service.ts#L273).
The implementations and tests are largely redundant and could be consolidated.

wilsonianb:

- add [custom query builder](https://vincit.github.io/objection.js/recipes/custom-query-builder.html#extending-the-query-builder-in-typescript) to `BaseModel` to support services' `getPage` functionality
- add common `BaseModel` and query resolver pagination tests

## Context

- fixes #240 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/shared/baseModel.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+      const modelsBackwards = await getPage(paginationBackwards)
+      expect(modelsForwards).toHaveLength(10)
+      expect(modelsBackwards).toHaveLength(10)
+      expect(modelsForwards).toEqual(modelsBackwards)
+    })
+  })
+}
+
+test.todo('test suite must contain at least one test')
```
sentientwaffle, line 80:
```suggestion
```

sentientwaffle (approved):
:tada:

packages/backend/src/shared/baseModel.test.ts
wilsonianb: Tests fail without that.
The file _could_ be excluded from tests either via `jest.config.js` or changing the `.test.ts` extension

sentientwaffle: Ah, nevermind, I didn't think of that.

PR #250, opened 2022-03-04
feat(ci): add workflow for pushing rafiki-backend docker images to GHCR

### LINKED ISSUE #233: Release Docker Images

omertoast:

- adds an actions workflow that pushes the rafiki backend docker image to github container registery. 

## Context

- fixes #233 
- this workflow only works when pushed to main.
- the image gets tagged by commit sha.

## Note
Repo owner must create a personal access token and add it to the repo's actions secrets named by `GHCR_PAT` in order to this workflow work. Without a token, the workflow fails to login GHCR which is required for pushing the image. You can read the instructions from [here](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry) to create a personal access token with the needed scopes.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,29 @@
+on:
+  push:
+    branches:
+    - main
+
+name: Release Docker Image
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: yarn install --immutable --immutable-cache
+
+      - name: Build Docker Image
+        run: yarn docker build backend -t ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}
+
+      - name: Login GHCR
+        # You must create a personal access token on GitHub and add it to the 
+        # repo's actions secrets named by 'GHCR_PAT' in order to use this step.
+        run: echo ${{ secrets.GHCR_PAT }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Push Docker Image to GHCR
+        run: docker push ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}


omertoast:
Also, we can use a shortened version of sha while tagging the image for simplicity purposes. I couldn't decide how much short is short enough to keep the uniqueness of sha, so the length of it is up to your recommendations.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: yarn install --immutable --immutable-cache
+
+      - name: Build Docker Image
+        run: yarn docker build backend -t ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}
```
wilsonianb, line 21:
```suggestion
        run: yarn docker build backend -t ghcr.io/${{ github.repository_owner }}/rafiki-backend:${{ github.sha }}
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
+on:
+  push:
+    branches:
+    - main
```
wilsonianb, line 4:
```suggestion
    branches:
    - main
    tags:
    - '*'
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Build Docker Image
+        run: yarn docker build backend -t ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}
+
+      - name: Login GHCR
+        # You must create a personal access token on GitHub and add it to the 
+        # repo's actions secrets named by 'GHCR_PAT' in order to use this step.
+        run: echo ${{ secrets.GHCR_PAT }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Push Docker Image to GHCR
+        run: docker push ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}
```
wilsonianb, line 29: Tagging with the git sha is fine, but I'd like to also have `latest` and/or `main`.
If we do want to also build and push images for git tags like I suggested above, I'd like to tag the image with tag version as well.
I'm trying to see if the [docker/metadata-action](https://github.com/docker/metadata-action) could be used for tagging even if we aren't using docker/build-push-action

.github/workflows/docker.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Install dependencies
+        run: yarn install --immutable --immutable-cache
+
+      - name: Build Docker Image
+        run: yarn docker build backend -t ghcr.io/${{ github.actor }}/rafiki-backend:${{ github.sha }}
+
+      - name: Login GHCR
+        # You must create a personal access token on GitHub and add it to the 
+        # repo's actions secrets named by 'GHCR_PAT' in order to use this step.
+        run: echo ${{ secrets.GHCR_PAT }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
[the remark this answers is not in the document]
omertoast: The reason I didn't want to use `GITHUB_TOKEN` as the secret name is that there could be multiple PATs with different scopes for different purposes. I can commit this suggestion if it's not a problem.

.github/workflows/docker.yml
omertoast: I added a conditional step to the workflow for tagging the image by tag name if the workflow is triggered by a tag push event.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: yarn install --immutable --immutable-cache
+
+      - name: Build Docker Image
+        run: yarn docker build backend -t ghcr.io/${{ github.repository_owner }}/rafiki-backend:${{ github.sha }} -t ghcr.io/${{ github.repository_owner }}/rafiki-backend:latest
```
wilsonianb, line 23: If we start building/pushing for other branches, the `latest` tag should only be used for `main`.
But I think this is good for now, unless we want to not use `latest` on git tags?

wilsonianb (approved):
:+1: 
I tested this out on my fork and it works as expected.
Thanks!

PR #280, opened 2022-03-16
fix(ci): support autolabelling pull requests from forks

### LINKED ISSUE #277: Labeler doesn't support pull request from fork
https://github.com/interledger/rafiki/blob/main/.github/labeler.yml

https://github.com/interledger/rafiki/pull/250

```
Error: HttpError: Resource not accessible by integration
Error: Resource not accessible by integration
```

omertoast:

- Changed `pull_request` to `pull_request_target` for labeling pull requests coming from forks.

## Context

- fixes #277.
- Current labeler workflow fails when a pull request comes from a fork as you can see from [this PR](https://github.com/interledger/rafiki/pull/250).
- `pull_request_target` workflow trigger additionally gives PR author write permissions compared to `pull_request` trigger. It provides the workflow to tag pull requests coming from forks in this scenario, which is not possible with the `pull_request` trigger.
- Github states that misuse of `pull_request_target` may lead to insecure situations, check [this article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for information. Unless the workflow contains malicious code apart from the labeler action - which is a whole nother security issue -, it doesn't pose a vulnerability.
- [Labeler action](https://github.com/marketplace/actions/labeler) already uses `pull_request_target` trigger at the examples in readme, so the solution in this PR would be good to go.
- I tested this workflow on my fork and [it worked out](https://github.com/omertoast/rafiki/pull/3) as expected.
- You can check [this really long thread](https://github.com/actions/labeler/issues/12) for additional information.

## Checklist

- [x] Related issues linked using `fixes #number `
- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -15,9 +15,12 @@ on:
 
 jobs:
   label:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/labeler@v3
+      - uses: actions/labeler@v4
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           sync-labels: true
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -7,7 +7,7 @@
 
 name: "Pull Request Labeler"
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - reopened


wilsonianb (commented):
Thanks for the detailed context.

Could you add the `permissions` from the latest example workflow (and update to `actions/labeler@v4`)?
https://github.com/actions/labeler#create-workflow

wilsonianb (approved):
:+1: 

I tested it here: https://github.com/wilsonianb/rafiki/pull/85

wilsonianb:
Also, the fact that the repo's `Fork pull request workflows from outside collaborators` appears to be set to `Require approval for first-time contributors` should provide additional protection:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

PR #290, opened 2022-03-23
feat(backend): update get account details to OPv2

### LINKED ISSUE #260: Get account details
The current get account details needs to be updated to the [new OPv2 spec ](https://docs.openpayments.guide/reference/get-public-account)

- [ ] Add the authServer as an env config value to Rafiki
- [ ] Add the ability to specify displayName when creating a paymentPointer on Rafiki
- [ ] Ensure the above properties are returned in the call to [getAccount](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/account/routes.ts)

omertoast:

- adds optional `displayName` value to open payments account model.
- adds `authServer` as an env config value to Rafiki.
- adds `publicName`, `authServer` values to the `getAccount` response.

## Context

- fixes #260 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/config/app.ts
```diff
@@ -61,6 +61,8 @@ export const Config = {
   accountWorkers: envInt('ACCOUNT_WORKERS', 1),
   accountWorkerIdle: envInt('ACCOUNT_WORKER_IDLE', 200), // milliseconds
 
+  authServer: envString('AUTH_SERVER', 'http://127.0.0.1:3006'),
```
wilsonianb, line 64: I think we're going to end up with two urls for the auth server:
- one for grant requests (this)
- one for token introspection

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.1

packages/backend/migrations/20211012172409_create_accounts_table.js
```diff
@@ -4,6 +4,8 @@ exports.up = function (knex) {
     table.uuid('assetId').notNullable()
     table.foreign('assetId').references('assets.id')
 
+    table.string('displayName').nullable()
```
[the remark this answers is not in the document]
omertoast: There is no specific reason other than it's called `displayName` instead of `publicName` at the issue but I can surely change it.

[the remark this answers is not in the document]
sabineschaller: I just realized that we haven't put the account endpoints into the spec on openpayments.guide yet. However, we do have them on hackmd. Please use the property names that are defined [there](https://hackmd.io/YWnLmu9ASiiHIwTPzmZKNg?view#121-Get-Account-Details) and I'll make sure to add them to openpayments.guide.

packages/backend/src/open_payments/account/service.ts
```diff
@@ -53,6 +54,7 @@ async function createAccount(
   return await Account.transaction(deps.knex, async (trx) => {
     const account = await Account.query(trx)
       .insertAndFetch({
+        publicName: options.publicName,
```
sabineschaller, line 57: Maybe add a test to check that creation works when a `publicName` is provided.

packages/backend/src/open_payments/account/routes.ts
```diff
@@ -36,8 +36,9 @@ export async function getAccount(
...
   ctx.body = {
     id: `${config.publicHost}/accounts/${encodeURIComponent(accountId)}`,
-    accountServicer: config.publicHost,
+    publicName: account.publicName,
     assetCode: account.asset.code,
-    assetScale: account.asset.scale
+    assetScale: account.asset.scale,
+    authServer: config.authServerGrantUrl
   }
 }
```
sabineschaller, line 44: Could you update the comment to point to [here](https://docs.openpayments.guide/reference/get-public-account). That's the place in the docs that I still need to update but that will include that endpoint.

sabineschaller: It doesn't show here. I mean the comment above the method.

packages/backend/src/open_payments/account/service.ts
omertoast: Isn't this test enough to check the creation? https://github.com/omertoast/rafiki/blob/9ee418a32a22debe98ac4528071e069a65d0c0d6/packages/backend/src/open_payments/account/routes.test.ts#L100-L121

sabineschaller: The test you linked is for the route. The only test for the create method on the service is [this one](https://github.com/omertoast/rafiki/blob/9ee418a32a22debe98ac4528071e069a65d0c0d6/packages/backend/src/open_payments/account/service.test.ts#L62-L69), which doesn't include the public name. I'm suggesting to add a second test that does include a public name to make sure that writing to the DB works.

omertoast (commented):
How did I manage to break a different test? 🤨

omertoast:
> How did I manage to break a different test? 🤨

Re-run the workflow and it worked out this time somehow.

wilsonianb (approved):

PR #301, opened 2022-04-04
fix(backend): fix AppServices typos

wilsonianb:

- fix `AppServices` typos
- Add IncomingPayment mapper to exclude `receivedAmount` from resolver type.
- Use default [`resolverTypeWrapperSignature`](https://www.graphql-code-generator.com/plugins/typescript-resolvers) of `Promise<T> | T` since `IncomingPaymentEdge`'s node became `ResolversTypes['IncomingPayment']`

## Context

https://github.com/interledger/rafiki/pull/288#issuecomment-1086371488

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/graphql/models/incoming_payment.ts
+++ b/packages/backend/src/graphql/models/incoming_payment.ts
@@ -1,0 +1,15 @@
+import {
+  IncomingPaymentState,
+  Amount
+} from '../../open_payments/payment/incoming/model'
+
+export interface IncomingPaymentModel {
+  id: string
+  description?: string
+  createdAt: string
+  expiresAt: string
+  incomingAmount?: Amount
+  receivedAmount: Omit<Amount, 'amount'>
+  externalRef?: string
+  state: IncomingPaymentState
+}
--- a/packages/backend/src/open_payments/payment/incoming/service.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.ts
@@ -256,9 +256,12 @@ async function getAccountIncomingPaymentsPage(
 ): Promise<IncomingPayment[]> {
   assert.ok(deps.knex, 'Knex undefined')
 
-  return await IncomingPayment.query(deps.knex).getPage(pagination).where({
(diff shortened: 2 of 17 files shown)


packages/backend/src/graphql/models/incoming_payment.ts
```diff
@@ -0,0 +1,15 @@
...
+  Amount
+} from '../../open_payments/payment/incoming/model'
+
+export interface IncomingPaymentModel {
+  id: string
+  description?: string
+  createdAt: string
+  expiresAt: string
+  incomingAmount?: Amount
+  receivedAmount: Omit<Amount, 'amount'>
```
sabineschaller, line 12: Why only omit `amount`?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -256,9 +256,12 @@ async function getAccountIncomingPaymentsPage(
...
 
-  return await IncomingPayment.query(deps.knex).getPage(pagination).where({
-    accountId: accountId
-  })
+  return await IncomingPayment.query(deps.knex)
+    .getPage(pagination)
+    .where({
+      accountId: accountId
+    })
+    .withGraphFetched('asset')
```
sabineschaller, line 264: Why does that not also need to be `.withGraphFetched('[account.asset, asset]')`?

sabineschaller (commented):
Thanks for fixing this!

packages/backend/src/open_payments/payment/incoming/service.ts
wilsonianb: I know I'd previously said otherwise, but I think incoming payment queries really only need `.withGraphFetched('asset')`. The payment's asset should just be checked against the account asset on creation.
`IncomingPayment`'s `account` could be made optional or just removed:
https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/payment/incoming/model.ts#L51

PR #316, opened 2022-04-19
chore(backend):  redis tls config

DonChangfoot:
Connect to redis using a tls connection.

Additional env variables have been added to enable tls and configure the paths to the required tls certificates and key.
https://github.com/luin/ioredis\#tls-options

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/README.md
+++ b/packages/backend/README.md
@@ -21,3 +21,17 @@ In order to build the docker container run the following command.
 ```shell
 yarn docker build backend -t rafiki-backend
 ```
+
+## Configuration
+
+### Redis connection
+
+The connection can be configured by specifying the following environment variables.
+The config is passed to `ioredis` - see https://github.com/luin/ioredis#tls-options.
+
+| Variable                 | Description                                        | Default                  |
+| ------------------------ | -------------------------------------------------- | ------------------------ |
+| REDIS_URL                | Redis connection string.                           | "redis://127.0.0.1:6379" |
+| REDIS_TLS_CA_FILE_PATH   | Path to CA certificate - overrides well-known CAs. | ""                       |
+| REDIS_TLS_KEY_FILE_PATH  | Path to private key for client authentication.     | ""                       |
+| REDIS_TLS_CERT_FILE_PATH | Path to certificate for client authentication.     | ""                       |
--- a/packages/backend/src/config/app.ts
+++ b/packages/backend/src/config/app.ts
@@ -1,4 +1,6 @@
 import * as crypto from 'crypto'
+import { readFileSync } from 'fs'
+import { ConnectionOptions } from 'tls'
 
 function envString(name: string, value: string): string {
   const envValue = process.env[name]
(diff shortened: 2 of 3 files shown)


wilsonianb:
:confused:  I can't get a successful GitHub Actions workflow run, but I'm not seeing how these changes would cause an issue
https://github.com/wilsonianb/rafiki/runs/6081144469?check_suite_focus=true

DonChangfoot:
@wilsonianb There's test setup that relies on the `REDIS_URL` environment variable. Seems IORedis supports sending in the url and options object. So I've reverted back to using `REDIS_URL`.

DonChangfoot:
@wilsonianb I also had an issue with the tests failing locally due to timeouts. I'm guessing it was due to it running out of memory as the tests take up quite a bit of it. I killed some processes to free up some memory and the tests passed. Not sure what the best solution here is? Using `runInBand` or specifying the memory limits or using a larger machine?

packages/backend/src/config/app.ts
```diff
@@ -37,6 +38,21 @@ export const Config = {
...
+          ca: readFileSync(
+            envString('REDIS_TLS_CA_FILE_PATH', '/certs/ca.crt')
+          ),
+          key: readFileSync(
+            envString('REDIS_TLS_KEY_FILE_PATH', '/certs/tls.key')
+          ),
+          cert: readFileSync(
+            envString('REDIS_TLS_CERT_FILE_PATH', '/certs/tls.crt')
+          )
+        }
```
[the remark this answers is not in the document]
DonChangfoot: Good point. The `tls.key` and `tls.crt` are required for client authentication. `ca.crt` is optional and there if you want to use self-signed certs. I'll also then toggle `rejectUnauthorized` if the `ca.crt` is specified.

packages/backend/src/config/app.ts
```diff
@@ -92,3 +99,27 @@ export const Config = {
...
+  const options: ConnectionOptions = {}
+
+  // self-signed certs.
+  if (caFile !== '') {
+    options.ca = readFileSync(caFile)
+    options.rejectUnauthorized = false
+  }
+
+  // Both are required for client authentication.
+  if (keyFile !== '' || certFile !== '') {
```
wilsonianb, line 119:
```suggestion
  if (keyFile !== '' && certFile !== '') {
```
`readFileSync` throws if a path is empty

Or the `readFileSync`s could be moved to their own `if` blocks.

packages/backend/src/config/app.ts
```diff
@@ -92,3 +99,27 @@ export const Config = {
...
+    options.key = readFileSync(keyFile)
+    options.cert = readFileSync(certFile)
+  }
+
+  return options
```
wilsonianb, line 124:
```suggestion
  return Object.keys(options).length ? options : undefined
```
If only `keyFile` or `certFile` are specified, `options` would be empty.
If this check ends up being needed (depending on how only one of `key`/`cert` is handled) the check at the top of the function could be removed.

packages/backend/README.md
```diff
@@ -21,3 +21,16 @@ In order to build the docker container run the following command.
...
+
+## Configuration
+
+### Redis connection
+
+The connection can be configured by specifying the following environment variables.
+The config is passed to `ioredis` - see https://github.com/luin/ioredis#tls-options.
+| Variable | Description | Default |
+|-----------------------------|------------------|
+| REDIS_URL | "redis://127.0.0.1:6379" |
```
wilsonianb, line 33: `REDIS_URL` (and the row above) are missing a `Description` column

DonChangfoot: thanks, good catch

wilsonianb (approved):

Issue #296, opened 2022-03-30
Allow `accountId` to be a string, not just uuid


sabineschaller:
Currently, `accountId`s in incoming and outgoing payments are uuids. For user-friendliness, we would also want to allow for human-readable id's, such that a payment pointer can be e.g. https://openpayments.guide/alice and not only https://openpayments.guide/016da9d5-c9a4-4c80-a354-86b915a04ff8.

DonChangfoot:
@wilsonianb mentioned maintaining a blocklist so that canonincal urls like `/graphql` are not confused with account names (#322 ).

wilsonianb:
> @wilsonianb mentioned maintaining a blocklist so that canonincal urls like `/graphql` are not confused with account names (#322 ).

- That should be less of an issue once https://github.com/interledger/rafiki/issues/441 is resolved

wilsonianb:
- Duplicate of https://github.com/interledger/rafiki/issues/416

PR #322, opened 2022-04-21
fix(backend): make graphql path reachable

DonChangfoot:

Add the graphql middleware first so that `/graphql` is not routed to the `accounting` endpoints.

## Context

Previously, visiting `/graphql` would return an `Invalid account id` error. This was being routed to the accounting endpoints instead of graphql. 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -140,8 +140,8 @@ export class App {
         }
       }
     )
-    await this._setupRoutes()
     this._setupGraphql()
+    await this._setupRoutes()
 
     // Workers are in the way during tests.
     if (this.config.env !== 'test') {


wilsonianb (approved):
:thinking: we're not supporting vanity payment pointers ([yet](https://github.com/interledger/rafiki/issues/296)), but we'll need to include `graphql` on a blocklist

DonChangfoot:
> thinking we're not supporting vanity payment pointers ([yet](https://github.com/interledger/rafiki/issues/296)), but we'll need to include `graphql` on a blocklist

@wilsonianb I've added a comment on that issue just to keep note of it.

PR #328, opened 2022-04-25
feat(auth): key validation service

### LINKED ISSUE #270: Support Client Keys
The AS must:
* Get public key from `openpayments.network/keys/{keyname}
* Verify client details provided in request body
* Decrypt provided signature

njlie:

- Adds a service that will be used to validate a client & public key with the registry that the key's id refers to. 
  - This will handle validation during the grant request & continuation.

## Context

Closes #270.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+async function verifySig(
+  deps: ServiceDependencies,
+  sig: string,
+  jwk: JWKWithRequired,
+  challenge: string
+): Promise<boolean> {
+  const publicKey = (await importJWK(jwk)) as crypto.KeyLike
```
wilsonianb, line 75: If `alg` isn't specified for [`importJWK`](https://github.com/panva/jose/blob/HEAD/docs/functions/key_import.importJWK.md#parameters) and

>Default is the "alg" property on the JWK.

should `JWKWithRequired` make [`alg`](https://github.com/panva/jose/blob/main/src/types.d.ts#L99) required?

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+    jwk.use === 'sig' ||
+    !jwk.key_ops ||
+    (jwk.key_ops.includes('sign') && jwk.key_ops.includes('verify')) ||
+    jwk.alg === 'EdDSA' ||
+    jwk.crv === 'Ed25519'
+  ) {
+    return true
+  }
+
+  return false
```
wilsonianb, line 95: Are `true`/`false` swapped?

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+): Promise<boolean> {
+  const publicKey = (await importJWK(jwk)) as crypto.KeyLike
+  const data = Buffer.from(challenge)
+  return crypto.verify(null, data, publicKey, Buffer.from(sig))
+}
+
+function validateRequiredJwkProperties(jwk: JWKWithRequired): boolean {
+  if (
+    !jwk.kid ||
+    !jwk.x ||
```
wilsonianb, line 83: Wouldn't these be enforced by `JWKWithRequired`?
Should `kid` and `x` instead be checked wherever `clientInfo`'s `key` gets defined?

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+  const { logger, config } = deps
+  const { jwk } = clientInfo.key
+  const { keyRegistries } = config
+
+  // No key registries in list means no validation
+  if (keyRegistries.length === 0) return true
+  if (!validateRequiredJwkProperties(jwk)) return false
+
+  const kidUrl = new URL(jwk.kid)
+  if (keyRegistries.length > 0 && !keyRegistries.includes(kidUrl.origin))
```
wilsonianb, line 111:
```suggestion
  if (!keyRegistries.includes(kidUrl.origin))
```

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    displayInfo.name === registryClientInfo.name &&
+    displayInfo.uri === registryClientInfo.uri
+  )
+}
+
+function verifyJwk(jwk: JWKWithRequired, keys: RegistryKey): boolean {
+  // TODO: update this to reflect eventual shape of response from registry
+  return !!(
+    keys.revoked &&
```
wilsonianb, line 146:
```suggestion
    !keys.revoked &&
```

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,153 @@
...
+  config: IAppConfig
+}
+
+export interface ClientService {
+  verifySig(
+    sig: string,
+    jwk: JWKWithRequired,
+    challenge: string
+  ): Promise<boolean>
+  validateClientWithRegistry(clientInfo: ClientInfo): Promise<boolean>
```
wilsonianb, line 45: You could add some tests that use `nock` for one or more registries.

packages/auth/src/clients/service.ts
njlie: No, though it might have seemed that way because the conditional that would return `true` was erroneously checking if `kty !== 'OKP'`, which I'll fix.

https://github.com/interledger/rafiki/pull/328/files#diff-ebe1c87507722350de717578d84973dc69709f084f1195e4221dfaf97bcdd9daR84

packages/auth/src/tests/app.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const knex = await container.use('knex')
+
+  return {
+    app,
+    port: app.getPort(),
+    knex,
+    connectionUrl: config.databaseUrl,
+    shutdown: async () => {
+      nock.cleanAll()
```
wilsonianb, line 33: I don't think this is necessary.
The backend has it because of nock instances that use [`persist()`](https://www.npmjs.com/package/nock#persist)

packages/auth/migrations/20220502133210_create_test_table.js
```diff
@@ -0,0 +1,9 @@
+exports.up = function (knex) {
+  return knex.schema.createTable('dummyTables', function (table) {
```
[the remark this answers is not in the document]
njlie: I have it so that the build step doesn't break because it doesn't find any migration files. But we could merge in https://github.com/interledger/rafiki/pull/332 first and then rebase this branch off main to avoid having to do this.

packages/auth/src/clients/service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+  )
+
+  describe('Registry Validation', (): void => {
+    describe('Client Properties', (): void => {
+      test('Can validate client properties with registry', async (): Promise<void> => {
+        const expDate = new Date()
+        expDate.setTime(expDate.getTime() + 1000 * 60 * 60)
+
+        const nbfDate = new Date()
+        nbfDate.setTime(nbfDate.getTime() - 1000 * 60 * 60)
```
wilsonianb, line 53: These could be defined once in an outer scope or a `beforeAll`/`beforeEach`.

packages/auth/src/clients/service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+          key: {
+            proof: 'httpsig',
+            jwk: {
+              ...TEST_CLIENT_KEY,
+              kid: KEY_REGISTRY_ORIGIN + '/keys/correct'
+            }
+          }
+        })
+
+        expect(validClient).toEqual(true)
```
wilsonianb, line 78: You could also call [`isDone`](https://github.com/nock/nock#isdone) on the nock scopes.

packages/auth/src/clients/service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+            jwk: TEST_CLIENT_KEY
+          }
+        })
+
+        expect(validClient).toEqual(false)
+      })
+    })
+
+    test('Cannot validate client with key that has invalid properties', async (): Promise<void> => {
+      const futureDate = new Date()
```
wilsonianb, line 146:
```suggestion
      const expDate = new Date()
```

packages/auth/src/tests/app.ts
```diff
@@ -0,0 +1,37 @@
...
+  connectionUrl: string
+  shutdown: () => Promise<void>
+}
+
+export const createTestApp = async (
+  container: IocContract<AppServices>
+): Promise<TestContainer> => {
+  const config = await container.use('config')
+  config.port = 0
+
```
wilsonianb, line 21: You could suppress test logging with:
https://github.com/interledger/rafiki/blob/main/packages/backend/src/tests/app.ts#L40-L49

packages/auth/src/clients/service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+            exp: Math.round(futureDate.getTime() / 1000),
+            nbf: Math.round(nbfDate.getTime() / 1000),
+            revoked: false
+          }
+        })
+
+      nock(KEY_REGISTRY_ORIGIN).get('/wrong').reply(200)
+
+      // Validate kid
+      const validClientKid = await clientService.validateClientWithRegistry({
```
wilsonianb, line 165: Is this intended to test (which would require using a different origin)?
https://github.com/interledger/rafiki/pull/328/files#diff-ebe1c87507722350de717578d84973dc69709f084f1195e4221dfaf97bcdd9daR111
It returns false at
https://github.com/interledger/rafiki/pull/328/files#diff-ebe1c87507722350de717578d84973dc69709f084f1195e4221dfaf97bcdd9daR127

packages/auth/src/clients/service.test.ts
```diff
@@ -0,0 +1,332 @@
...
+            ...TEST_CLIENT_KEY,
+            key_ops: ['wrapKey']
+          }
+        }
+      })
+
+      expect(validClientKeyOps).toEqual(false)
+
+      // Validate "x"
+      const validClientX = await clientService.validateClientWithRegistry({
```
wilsonianb, line 207: You could move this (and maybe the `kid` check?) to its own test, and remove any nock from this test, since I think validation fails before the request is sent.

packages/auth/src/clients/service.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+function verifyJwk(jwk: JWKWithRequired, keys: RegistryKey): boolean {
+  // TODO: update this to reflect eventual shape of response from registry
+  return !!(
+    !keys.revoked &&
+    keys.exp &&
+    new Date() < new Date(keys.exp * 1000) &&
+    keys.nbf &&
+    new Date() > new Date(keys.nbf * 1000) &&
```
wilsonianb, line 154:
```suggestion
    new Date() >= new Date(keys.nbf * 1000) &&
```

packages/auth/src/clients/service.test.ts
njlie: It's supposed to test when a `kid` is a url but doesn't point to an endpoint with a key or is otherwise invalid.

packages/auth/src/tests/app.ts
wilsonianb: bump

packages/auth/src/tests/app.ts
wilsonianb: bump

wilsonianb (approved):

packages/auth/src/tests/app.ts
njlie: I think the log suppression is just below this comment?

wilsonianb: Oops, I'm apparently overly dependent on GitHub marking review comments as `Outdated`

wilsonianb (approved):

PR #342, opened 2022-05-17
feat(backend): list endpoints

### LINKED ISSUE #289: List endpoints for incoming and outgoing payments

sabineschaller:

- list endpoints for incoming payments, outgoing payments, and quotes
- abstract some functionality to avoid repetition in list endpoints

## Context

- fixes #289 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/model.ts
+++ b/packages/backend/src/open_payments/payment/incoming/model.ts
@@ -1,5 +1,5 @@
-import { Model } from 'objection'
-import { Amount } from '../../amount'
+import { Model, Pojo } from 'objection'
+import { Amount, AmountJSON } from '../../amount'
 import { Asset } from '../../../asset/model'
 import { LiquidityAccount, OnCreditOptions } from '../../../accounting/service'
 import { ConnectorAccount } from '../../../connector/core/rafiki'
@@ -165,4 +165,41 @@ export class IncomingPayment
 
     return data
   }
+
+  $formatJson(json: Pojo): Pojo {
+    json = super.$formatJson(json)
+    return {
+      id: json.id,
+      accountId: json.accountId,
+      incomingAmount: this.incomingAmount
+        ? {
+            ...json.incomingAmount,
+            value: json.incomingAmount.value.toString()
+          }
+        : null,
+      receivedAmount: {
+        ...json.receivedAmount,
+        value: json.receivedAmount.value.toString()
+      },
+      state: json.state.toLowerCase(),
+      description: json.description,
+      externalRef: json.externalRef,
+      createdAt: json.createdAt,
+      updatedAt: json.updatedAt,
+      expiresAt: json.expiresAt.toISOString()
+    }
+  }
+}
+
+export type IncomingPaymentJSON = {
+  id: string
+  accountId: string
+  incomingAmount: AmountJSON | undefined
+  receivedAmount: AmountJSON
+  state: string
+  description: string | undefined
+  externalRef: string | undefined
+  createdAt: string
+  updatedAt: string
+  expiresAt: string
 }
--- a/packages/backend/src/open_payments/payment/incoming/routes.test.ts
+++ b/packages/backend/src/open_payments/payment/incoming/routes.test.ts
@@ -441,4 +441,51 @@ describe('Incoming Payment Routes', (): void => {
       })
     })
   })
+
+  describe('list', (): void => {
+    test('returns 200 without query params', async (): Promise<void> => {
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        { accountId: account.id }
+      )
+      await expect(incomingPaymentRoutes.list(ctx)).resolves.toBeUndefined()
+      expect(ctx.status).toBe(200)
+      expect(ctx.response.get('Content-Type')).toBe(
+        'application/json; charset=utf-8'
+      )
+    })
+
+    test('returns 200 with query params', async (): Promise<void> => {
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' },
+          query: { first: 5, cursor: incomingPayment.id }
+        },
+        { accountId: account.id }
+      )
+      await expect(incomingPaymentRoutes.list(ctx)).resolves.toBeUndefined()
+      expect(ctx.status).toBe(200)
+      expect(ctx.response.get('Content-Type')).toBe(
+        'application/json; charset=utf-8'
+      )
+    })
+
+    test('returns 500 if TB account not found', async (): Promise<void> => {
+      jest
+        .spyOn(accountingService, 'getAccountsTotalReceived')
+        .mockResolvedValueOnce([undefined])
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        { accountId: account.id }
+      )
+      await expect(incomingPaymentRoutes.list(ctx)).rejects.toMatchObject({
+        status: 500,
+        message: `Error trying to list incoming payments`
+      })
+    })
+  })
 })
--- a/packages/backend/src/shared/baseModel.ts
+++ b/packages/backend/src/shared/baseModel.ts
@@ -21,6 +21,8 @@ export interface PageInfo {
   endCursor?: string
   hasNextPage?: boolean
   hasPreviousPage?: boolean
+  first?: number
+  last?: number
 }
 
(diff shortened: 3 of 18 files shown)


wilsonianb (commented):
It'd be great if the shared `getPageInfo` could be used by all `getPageInfo` query resolvers (including assets and peers) as well as list endpoints (either shared `list` function or individual routes `list` functions)

packages/backend/src/shared/pagination.ts
```diff
@@ -0,0 +1,165 @@
...
+        ...item.toJSON(),
+        accountId: `${host}/${accountId}`,
+        id:
+          item['receivedAmount'] !== undefined
+            ? `${host}/${accountId}/incoming-payments/${item.id}`
+            : item['sentAmount'] !== undefined
+            ? `${host}/${accountId}/outgoing-payments/${item.id}`
+            : `${host}/${accountId}/quotes/${item.id}`
+      }
+    })
```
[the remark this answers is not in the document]
sabineschaller: I can just create a shared method for all of them, what do you think?

packages/backend/src/shared/pagination.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export async function getPageInfo(
+  service:
+    | IncomingPaymentService
+    | OutgoingPaymentService
+    | QuoteService
+    | AssetService
+    | PeerService,
+  page: (IncomingPayment | OutgoingPayment | Quote | Asset | Peer)[],
+  accountId?: string
```
wilsonianb, line 56:
```suggestion
export async function getPageInfo<T extends BaseModel>(
  getPage: (pagination: Pagination) => Promise<T[]>,
  page: T[],
```

`getAccountPage` can be passed to `getPage` as:
```
const pageInfo = await getPageInfo((pagination) => quoteService.getAccountPage(accountId, pagination), page)
```

sabineschaller: :exploding_head: I knew my solution wasn't great and I'm blown away by the elegance of this. Thank you!

packages/backend/src/shared/pagination.ts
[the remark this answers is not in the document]
sabineschaller: I went for not having a shared `list` function. I think it's cleaner.

sabineschaller:
Tests fail because `ilp-pay` expects non-required fields to be [`undefined` and not `null`](https://github.com/interledgerjs/interledgerjs/blob/37f2205d91c27643113b4b41405f8925be6e5734/packages/pay/src/open-payments.ts#L404-L406), but `undefined` is not valid JSON.

packages/backend/src/shared/pagination.ts
```diff
@@ -0,0 +1,80 @@
...
+  const pageInfo = await getPageInfo((pagination) => getPage(pagination), page)
+  if (pagination?.last) {
+    pageInfo.last = page.length
+  } else {
+    pageInfo.first = page.length
+  }
+  return pageInfo
+}
+
+export async function getPageInfo<T extends BaseModel>(
```
wilsonianb, line 43: What if [`getPage`](https://github.com/interledger/rafiki/blob/557fe12851bb7d34911dcba13d8724cee0403f83/packages/backend/src/shared/baseModel.ts#L41) included `pageInfo` in the query response (similar to [`page`](https://vincit.github.io/objection.js/api/query-builder/other-methods.html#page))?

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -165,4 +165,41 @@ export class IncomingPayment
...
+    json = super.$formatJson(json)
+    return {
+      id: json.id,
+      accountId: json.accountId,
+      incomingAmount: this.incomingAmount
+        ? {
+            ...json.incomingAmount,
+            value: json.incomingAmount.value.toString()
+          }
+        : null,
```
wilsonianb, line 179: As mentioned on the [Rafiki call](https://github.com/interledger/rafiki/discussions/346), we should decide between either excluding optional fields from Open Payments responses when null, or marking them as [`nullable`](https://swagger.io/docs/specification/data-models/data-types/#null) in the OpenAPI spec. I'm leaning towards the former.

packages/backend/src/shared/baseModel.ts
```diff
@@ -16,6 +16,15 @@ export interface Pagination {
   last?: number // Backward pagination: limit.
 }
 
+export interface PageInfo {
+  startCursor?: string
+  endCursor?: string
+  hasNextPage: boolean
+  hasPreviousPage: boolean
+  first?: number
+  last?: number
```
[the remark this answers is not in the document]
sabineschaller: I don't have a strong opinion. When I researched what was usually included in these page info blocks, they sometimes were.

packages/backend/src/open_payments/payment/incoming/model.ts
sabineschaller: Me too. I'll exclude them.

packages/backend/src/shared/pagination.ts
sabineschaller: Since `pageInfo` is always part of these list request, we could move it. However, I'd like to do that in a subsequent PR to get the list endpoints in. We can always refactor later.

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -165,4 +165,41 @@ export class IncomingPayment
...
+}
+
+export type IncomingPaymentJSON = {
+  id: string
+  accountId: string
+  incomingAmount: AmountJSON | null
+  receivedAmount: AmountJSON
+  state: string
+  description: string | null
+  externalRef: string | null
```
wilsonianb, line 201:
```suggestion
  incomingAmount?: AmountJSON
  receivedAmount: AmountJSON
  state: string
  description?: string
  externalRef?: string
```
Ditto for outgoing payments and quotes

packages/backend/src/open_payments/payment/incoming/routes.test.ts
```diff
@@ -441,4 +509,91 @@ describe('Incoming Payment Routes', (): void => {
       })
     })
   })
+
+  describe('list', (): void => {
+    let items: IncomingPayment[]
+    let result: Omit<IncomingPaymentJSON, 'incomingAmount' | 'externalRef'>[]
```
wilsonianb, line 515:
```suggestion
    let result: IncomingPaymentJSON[]
```
Ditto elsewhere

packages/backend/src/open_payments/payment/incoming/model.ts
sabineschaller: I wouldn't want to call the type `IncomingPaymentJSON` in that case since `undefined` is not valid JSON. What do you think?

wilsonianb: This makes me wonder if we could get away with not defining `IncomingPaymentJSON` (just specify `$formatJson`)
https://vincit.github.io/objection.js/api/model/overview.html#model-data-lifecycle
>Note: Most libraries like [express](http://expressjs.com/en/index.html)[ (opens new window)](http://expressjs.com/en/index.html) and [koa](https://koajs.com/)[ (opens new window)](https://koajs.com/) automatically call the [toJSON](https://vincit.github.io/objection.js/api/model/instance-methods.html#tojson) method when you pass the model instance to methods like response.json(model). You rarely need to call [toJSON()](https://vincit.github.io/objection.js/api/model/instance-methods.html#tojson) or [$toJson()](https://vincit.github.io/objection.js/api/model/instance-methods.html#tojson) explicitly.

wilsonianb: I guess that's tricky because `$formatJson` would need `config.publicHost`

sabineschaller: I guess we could find a hacky way to get it in there but I don't know if we want to do that.

wilsonianb: What if we removed the `null`s and renamed as `IncomingPaymentResponse`?

sabineschaller: The thing is that `toJSON()` automatically transforms `undefined` into `null`. That is why I think it makes sense to keep the `null`s and the `IncomingPaymentJSON` name and removing the `null`s from the response the way I do it.

wilsonianb (approved):

PR #339, opened 2022-05-13
chore(backend): validate Open Payments requests via OpenAPI

wilsonianb:
## Changes proposed in this pull request

- Set up Open Payments routes and validate requests using OpenAPI spec.
- Move invalid expiresAt check to incomingPaymentService.

## Context

- fixes #285

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [ ] Documentation added
- [x] Make sure that all checks pass

PR #343, opened 2022-05-19
feat(auth): initiate grant request endpoint

### LINKED ISSUE #268: Handle Initial Request Flow
* Create model for session information, to be used to identify client on subsequent grant request continuations
* Initialize session when grant request is initialized.

njlie:

- Refactor `limit` to `access`
  - A grant can have many "accesses", and each access has a single limit attached to it, so it seemed to make more sense to make them the same entity.
- Add `/auth` endpoint to initiate a grant request flow.
- Add `types` files for each service/model.

## Context

Closes #268.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,134 @@
...
+  grantRequest: GrantRequest
+): grantRequest is GrantRequest {
+  if (typeof grantRequest.access !== 'object') return false
+  for (const access of grantRequest.access) {
+    if (!isAccessRequest(access)) return false
+  }
+
+  return (
+    grantRequest.interact?.start !== undefined &&
+    grantRequest.interact?.finish !== undefined
```
wilsonianb, line 83: Why isn't `finish` optional?

packages/auth/src/grant/model.ts
```diff
@@ -55,17 +42,20 @@ export class Grant extends BaseModel {
   }
 
   public state!: GrantState
-  public type!: AccessType
-  public actions!: Action[]
   public startMethod!: StartMethod[]
+  public locations?: string[]
+  public identifier?: string
+  public interval?: string
```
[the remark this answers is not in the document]
njlie: Nope, this was a part of the refactoring I did that I missed. Thanks for catching.

packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+interface ServiceDependencies extends BaseService {
+  accessService: AccessService
+  config: IAppConfig
+  knex: TransactionOrKnex
+}
+
+// TODO: maybe update docs.openpayments.guide with location
+export interface GrantRequest {
+  access: AccessRequest[]
```
wilsonianb, line 25:
```suggestion
  access_token: {
    access: AccessRequest[]
  }
```

packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const {
+    access,
+    interact: {
+      start,
+      finish: { method: finishMethod, uri: finishUri, nonce: clientNonce }
+    }
+  } = grantRequest
+
+  const grant = await Grant.query(trx || knex).insert({
```
wilsonianb, line 99: This should create a transaction if `trx` is undefined, so that the grant insert can be rolled back if any of the access inserts fail.
See: https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/payment/incoming/service.ts#L120-L145

packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,131 @@
...
+  return {
+    interact: {
+      redirect: config.interactDomain + `/interact/${grant.interactId}`,
+      finish: grant.interactNonce
+    },
+    continue: {
+      access_token: {
+        value: grant.continueToken
+      },
+      uri: config.domain + `/auth/continue/${grant.continueId}`
```
wilsonianb, line 128: `wait` is `RECOMMENDED`
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.1

packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,131 @@
...
+    interactRef: v4(),
+    interactNonce: crypto.randomBytes(8).toString('hex').toUpperCase(), // TODO: factor out nonce generation
+    continueId: v4(),
+    continueToken: crypto.randomBytes(8).toString('hex').toUpperCase()
+  })
+
+  // Associate provided accesses with grant
+  await Promise.all(
+    access.map((accessRequest) => {
+      accessService.createAccess(grant.id, accessRequest, trx)
```
wilsonianb, line 115: What if `accessService.createAccess` accepted an array of accesses _or_ `initiateGrant` performed a [graph insert](https://vincit.github.io/objection.js/guide/query-examples.html#graph-inserts)?

sabineschaller (commented):
Not an entire review. Just things that I noticed when working on token introspection.

packages/auth/src/access/service.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      type: AccessType.IncomingPayment,
+      actions: [Action.Create, Action.Read, Action.List],
+      limits: incomingPaymentLimit
+    }
+
+    const access = await accessService.createAccess(grant.id, [
+      incomingPaymentAccess
+    ])
+
+    expect(access[0].grantId).toEqual(grant.id)
```
wilsonianb, line 83: You could check `access.length`

packages/auth/src/access/service.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    const grant = await Grant.query(trx).insertAndFetch(BASE_GRANT)
+
+    const access = await accessService.createAccess(grant.id, [
+      outgoingPaymentAccess
+    ])
+
+    expect(access[0].grantId).toEqual(grant.id)
+    expect(access[0].type).toEqual(AccessType.OutgoingPayment)
+    expect(access[0].limits).toEqual(outgoingPaymentLimit)
+  })
```
wilsonianb, line 122: You could test how `accessService.createAccess` an empty array of `accessRequests`

packages/auth/src/access/service.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      nock.restore()
+      await appContainer.shutdown()
+    }
+  )
+
+  const BASE_GRANT = {
+    state: GrantState.Pending,
+    startMethod: [StartMethod.Redirect],
+    continueToken: crypto.randomBytes(8).toString('hex').toUpperCase(),
+    continueId: v4(),
```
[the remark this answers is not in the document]
njlie: This is used to create a `Grant` since one needs to exist in order to have a valid foreign key for the `grantId` in `Access` when it is subsequently created.

packages/auth/src/grant/model.ts
```diff
@@ -44,28 +31,28 @@ export class Grant extends BaseModel {
         to: 'accessTokens.grantId'
       }
     },
-    grants: {
+    accesses: {
```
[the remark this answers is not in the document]
wilsonianb: I could go either way

wilsonianb (approved):

packages/auth/src/app.ts
```diff
@@ -112,10 +115,38 @@ export class App {
...
+        ctx.assert(ctx.accepts('application/json'), 406, 'must accept json')
+        ctx.assert(
+          ctx.get('Content-Type') === 'application/json',
+          400,
+          'must send json body'
+        )
+        const { body } = ctx.request
+        const grantService = await this.container.use('grantService')
+        const clientService = await this.container.use('clientService')
+        if (!grantService.validateGrantRequest(body)) {
```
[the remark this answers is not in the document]
wilsonianb: https://github.com/interledger/rafiki/pull/339/commits/4556d21258dc2c8c61d3b0f735b6a50d2f430b14

packages/auth/src/app.ts
```diff
@@ -112,10 +115,38 @@ export class App {
...
+        if (!isValidClient) {
+          ctx.status = 400
+          ctx.body = { error: 'invalid_client' }
+          return
+        }
+
+        const res = await grantService.initiateGrant(body)
+        ctx.status = 200
+        ctx.body = res
+      }
```
sabineschaller, line 148: I'd probably want to move this into a `routes.ts` file, probably under `grant`.

packages/auth/src/app.ts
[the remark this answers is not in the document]
njlie: I'm going to capture this in https://github.com/interledger/rafiki/issues/351 for now.

packages/auth/src/app.ts
```diff
@@ -111,11 +114,9 @@ export class App {
       ctx.status = 200
     })
 
+    const grantRoutes = await this.container.use('grantRoutes')
     // TODO: GNAP endpoints
-    this.publicRouter.post('/auth', (ctx: AppContext): void => {
-      // TODO: generate interaction session
-      ctx.status = 200
-    })
+    this.publicRouter.post('/auth', grantRoutes.auth.post)
```
wilsonianb, line 119: I know the OpenAPI spec currently has `/auth` on the base server url
https://github.com/interledger/open-payments/blob/master/auth-server-open-api-spec.yaml#L13
but I don't think the `/auth` path is required by GNAP.
What do you think about changing this to
```suggestion
    this.publicRouter.post('/', grantRoutes.create)
```
(`create` to distinguish from the `POST /continue` route method)

packages/auth/src/grant/service.ts
```diff
@@ -0,0 +1,144 @@
...
+
+import { BaseService } from '../shared/baseService'
+import { Grant, GrantState, StartMethod, FinishMethod } from './model'
+import { AccessRequest, isAccessRequest } from '../access/types'
+import { ClientInfo } from '../client/service'
+import { IAppConfig } from '../config/app'
+import { AccessService } from '../access/service'
+
+export interface GrantService {
+  validateGrantRequest(grantRequest: GrantRequest): boolean
```
wilsonianb, line 13: What do you think about moving this to the `GrantRoutes` (until it's removed for #351)?

wilsonianb (commented):
There's confllcts from https://github.com/interledger/rafiki/pull/339 that need to resolved

packages/auth/src/access/types.ts
```diff
@@ -0,0 +1,178 @@
...
+    isIncomingPaymentAccessRequest(accessRequest as IncomingPaymentRequest) ||
+    isOutgoingPaymentAccessRequest(accessRequest as OutgoingPaymentRequest) ||
+    isAccountAccessRequest(accessRequest as AccountRequest) ||
+    isQuoteAccessRequest(accessRequest as QuoteRequest)
+  )
+}
+
+// value should hold bigint, serialized as string for requests & storage as jsonb field in postgres
+export interface PaymentAmount {
+  value: string
```
sabineschaller, line 114: Why not BigInt?

packages/auth/src/config/app.ts
```diff
@@ -29,5 +29,8 @@ export const Config = {
         ),
   keyRegistries: envStringArray('KEY_REGISTRIES', [
     'https://openpayments.network'
-  ])
+  ]),
+  interactDomain: envString('INTERACT_URL', 'http://localhost:3004'),
+  domain: envString('DOMAIN', 'http://localhost:3006'), // TODO: replace this with whatever frontend port ends up being
```
sabineschaller, line 34: Just a general question: What are these two?

packages/auth/src/grant/service.test.ts
```diff
@@ -0,0 +1,327 @@
...
+      }
+
+      const isValid = grantService.validateGrantRequest(
+        incomingPaymentGrantRequest as GrantRequest
+      )
+      expect(isValid).toEqual(false)
+    })
+  })
+
+  test('Can create a grant', async (): Promise<void> => {
```
sabineschaller, line 268: Should this one be moved to a `describe('create', ...)` block?

packages/auth/src/app.ts
njlie: That works for me. I take it removing the `/auth/` path applies to all the other routes?

packages/auth/src/config/app.ts
njlie: The `interactDomain` is the domain at which interaction takes place - essentially the domain of the RS. It's dawning on me that something like `resourceServerDomain` would be a better name.

`domain` is the domain of the AS. it's used to create things like the `continue` url when the AS needs to respond with a URL pointing to itself. Also realizing now that `authServerDomain` would be better.

packages/auth/src/access/types.ts
njlie: Because the shape of the `limit` object can change depending on the `access.type` it is for, I stored as a JSON blob in postgres which doesn't play nicely with BigInt unless it is serialized somehow.

I figure any BigInt value being sent to the AS in a request or response would have to be serialized into a string anyway. The AS shouldn't be doing any arithmetic with them either, so I don't think there will be a need for the AS to de-serialize them either.

packages/auth/src/grant/service.test.ts
```diff
@@ -87,31 +87,13 @@ describe('Grant Service', (): void => {
       assetScale: 9
     },
     expiresAt: new Date().toISOString(),
-    description: 'this is a test',
+    description: 'this is a test you fuck',
```
wilsonianb, line 90: :eyes:

packages/auth/src/grant/routes.ts
```diff
@@ -1,17 +1,17 @@
...
 interface ServiceDependencies extends BaseService {
   grantService: GrantService
   clientService: ClientService
 }
 
 export interface GrantRoutes {
-  auth: {
-    post(ctx: AppContext): Promise<void>
-  }
+  validateGrantRequest(grantRequest: GrantRequest): boolean
```
wilsonianb, line 13: This might not need to be exposed by `GrantRoutes`.
It could instead be tested within the `create` tests (checking that it returns `400`)

wilsonianb (approved):

packages/auth/src/grant/service.test.ts
njlie: Forgive my exasperation with getting the tests working 😓

wilsonianb (approved):

Issue #274, opened 2022-03-11
Signing/Request Library


matdehaast:
* Validate requests in accordance with https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-7.3.1

PR #348, opened 2022-05-25
feat(auth): token introspection

### LINKED ISSUE #267: Add Introspection Ability
Implement `POST /auth/introspect` endpoint

sabineschaller:

- adds token introspection endpoint

## Context

-fixes #267 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/accessToken/service.ts
```diff
@@ -0,0 +1,72 @@
...
+  const token = await AccessToken.query(deps.knex).findOne({ value })
+  if (!token) return
+  const now = new Date(Date.now())
+  const expiresAt = token.expiresIn
+    ? token.createdAt.getTime() + token.expiresIn
+    : Infinity
+  if (expiresAt < now.getTime()) {
+    return { active: false }
+  } else {
+    const grant = await Grant.query(deps.knex)
```
njlie, line 53: We should check to see if the grant state is revoked or not

packages/auth/src/grant/model.ts
```diff
@@ -39,8 +40,8 @@ export class Grant extends BaseModel {
         to: 'accesses.grantId'
       }
     }
-  }
-
+  })
+  public access!: Access[]
```
njlie, line 44: Is this field intentional? I don't see a migration for it in the grants table, but I believe the `Access` table already has a foreign key in the form of the grant id as well.

sabineschaller: It is because otherwise I can't get the access property [here](https://github.com/interledger/rafiki/blob/8b4e7859b94991a5a187e30c47c5474cefdfb84d/packages/auth/src/accessToken/routes.ts#L55). Like you said, we do have the `grantId` as foreign key in the access table. Due to the `relationsMapping`, we can easily add `access` as a property to the grant without having to store it in the table.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -0,0 +1,210 @@
...
+        })
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        {}
+      )
+      ctx.request.body = {
+        access_token: token.value,
+        proof: 'httpsig',
```
njlie, line 162: Just as a general question, are we using the GET request to `openpayments.network` as our method of determining client key possession instead of the `httpsig` method?

njlie (approved):

njlie:
@sabineschaller Are you planning to add `httpsig` validation to this endpoint? I don't think it's necessarily in scope since I think it might be captured in https://github.com/interledger/rafiki/issues/274 but if you plan to add it I don't want to accidentally overlap the implementation in another branch.

dclipp (approved):

sabineschaller:
I'll leave signature validation for whoever works on #274.

PR #360, opened 2022-06-09
feat(auth): access token revocation

### LINKED ISSUE #336: Revoke access token
Implement `DEL /auth/token/{id}` endpoint

see:
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6.2
https://docs.openpayments.guide/reference/delete-token

dclipp:

- Implement access token revocation feature

## Context

fixes #336 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/accessToken/routes.test.ts
+++ b/packages/auth/src/accessToken/routes.test.ts
@@ -212,6 +212,7 @@ describe('Access Token Routes', (): void => {
   describe('Revocation', (): void => {
     let grant: Grant
     let token: AccessToken
+    // overriding the `no-explicit-any' rule to avoid having to mock every Objection method for each tokenQuery mock
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let tokenQuery: <M extends { QueryBuilderType: any }>(
       trxOrKnex?: Objection.TransactionOrKnex | undefined
@@ -249,22 +250,13 @@ describe('Access Token Routes', (): void => {
     )
 
     test('Returns status 404 if token does not exist', async (): Promise<void> => {
+      id = v4()
       const ctx = createContext(
         {
-          headers: { Accept: 'application/json' },
-          params: { id }
+          headers: { Accept: 'application/json' }
         },
-        {}
+        { id }
       )
-
-      tokenQuery = () => {
-        return {
-          findById: () => {
-            return undefined
-          }
-        }
-      }
-      AccessToken.query = jest.fn(tokenQuery)
 
       await expect(accessTokenRoutes.revoke(ctx)).rejects.toMatchObject({
         status: 404,
@@ -275,10 +267,9 @@ describe('Access Token Routes', (): void => {
     test('Does not modify token if it has already expired', async (): Promise<void> => {
       const ctx = createContext(
         {
-          headers: { Accept: 'application/json' },
-          params: { id }
+          headers: { Accept: 'application/json' }
         },
-        {}
+        { id }
       )
 
       tokenQuery = () => {
@@ -309,10 +300,9 @@ describe('Access Token Routes', (): void => {
     test('Modifies token if it has not expired', async (): Promise<void> => {
       const ctx = createContext(
         {
-          headers: { Accept: 'application/json' },
-          params: { id }
+          headers: { Accept: 'application/json' }
         },
-        {}
+        { id }
       )
 
       tokenQuery = () => {
--- a/packages/auth/src/accessToken/routes.test.ts
+++ b/packages/auth/src/accessToken/routes.test.ts
@@ -220,6 +220,7 @@ describe('Access Token Routes', (): void => {
     let actualTokenQuery: <M extends { QueryBuilderType: any }>(
       trxOrKnex?: Objection.TransactionOrKnex | undefined
     ) => Objection.QueryBuilderType<M>
+    let id: string
 
     beforeAll(
       async (): Promise<void> => {
@@ -237,6 +238,7 @@ describe('Access Token Routes', (): void => {
           ...BASE_TOKEN
         })
         tokenQuery = AccessToken.query
+        id = v4()
       }
     )
 
@@ -249,19 +251,15 @@ describe('Access Token Routes', (): void => {
     test('Returns status 404 if token does not exist', async (): Promise<void> => {
       const ctx = createContext(
         {
-          headers: { Accept: 'application/json' }
+          headers: { Accept: 'application/json' },
+          params: { id }
         },
         {}
       )
-      ctx.request.body = {
-        access_token: v4(),
-        proof: 'httpsig',
-        resource_server: 'test'
-      }
 
       tokenQuery = () => {
         return {
-          findOne: () => {
+          findById: () => {
             return undefined
           }
         }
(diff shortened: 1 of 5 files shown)


packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
     return { active: true, ...grant, key: { proof: 'httpsig', jwk: keys[0] } }
   }
 }
+
+async function revoke(
+  deps: ServiceDependencies,
+  id: string
+): Promise<RevocationResult> {
```
sabineschaller, line 82: Why not return a boolean instead of the object?

packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
...
+        .where('id', token.id)
+    }
+
+    return {
+      foundToken: true
+    }
+  } else {
+    return {
+      foundToken: false
+    }
```
sabineschaller, line 99: I wonder if that should return an Error instead of the object or just `false`.

packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
     return { active: true, ...grant, key: { proof: 'httpsig', jwk: keys[0] } }
   }
 }
+
+async function revoke(
+  deps: ServiceDependencies,
+  id: string
+): Promise<RevocationResult> {
+  const token = await AccessToken.query(deps.knex).findOne({ id })
```
sabineschaller, line 83: You could use `findById`

packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
     return { active: true, ...grant, key: { proof: 'httpsig', jwk: keys[0] } }
   }
 }
+
+async function revoke(
```
sabineschaller, line 79: Could you add tests for that in `service.test.ts`?

packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
...
+  id: string
+): Promise<RevocationResult> {
+  const token = await AccessToken.query(deps.knex).findOne({ id })
+  if (token) {
+    if (!isTokenExpired(token)) {
+      AccessToken.query(deps.knex)
+        .update({
+          expiresIn: 1
+        })
+        .where('id', token.id)
```
sabineschaller, line 90:
```suggestion
      token.$query(deps.knex).patch({ expiresIn: 1 })
```

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+        {
+          headers: { Accept: 'application/json' }
+        },
+        {}
+      )
+      ctx.request.body = {
+        access_token: v4(),
+        proof: 'httpsig',
+        resource_server: 'test'
+      }
```
sabineschaller, line 260: The revocation request does not have a body.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+        AccessToken.query = actualTokenQuery
+      }
+    )
+
+    test('Returns status 404 if token does not exist', async (): Promise<void> => {
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        {}
```
sabineschaller, line 254: The path parameter `id` is missing here.

packages/auth/src/accessToken/service.ts
```diff
@@ -65,3 +75,27 @@ async function introspect(
...
   }
 }
+
+async function revoke(
+  deps: ServiceDependencies,
+  id: string
+): Promise<RevocationResult> {
+  const token = await AccessToken.query(deps.knex).findOne({ id })
+  if (token) {
+    if (!isTokenExpired(token)) {
```
sabineschaller, line 85: I think I'd add an else statement where an Error is returned.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+      )
+      ctx.request.body = {
+        access_token: v4(),
+        proof: 'httpsig',
+        resource_server: 'test'
+      }
```
sabineschaller, line 288: Same as above.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+              expiresIn: -1
+            }
+          },
+          update: mockUpdate
+        }
+      }
+      AccessToken.query = jest.fn(tokenQuery)
+
+      await accessTokenRoutes.revoke(ctx)
+      expect(mockUpdate).not.toHaveBeenCalled()
```
sabineschaller, line 305: This is functionality I'd test in `service.test.ts`. From looking at the docs, I guess if the token is expired, it should also respond with a 404. This is what I'd test here.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+              expiresIn: 10000
+            }
+          },
+          update: mockUpdate
+        }
+      }
+      AccessToken.query = jest.fn(tokenQuery)
+
+      await accessTokenRoutes.revoke(ctx)
+      expect(mockUpdate).toHaveBeenCalled()
```
sabineschaller, line 340: Same as above. Test that the response is a 204.

packages/auth/src/accessToken/routes.test.ts
dclipp: Updated (assuming you meant status `204` instead of `404`)

packages/auth/src/accessToken/service.ts
dclipp: added

packages/auth/src/accessToken/service.ts
dclipp: changed to return `Error | undefined`

packages/auth/src/accessToken/service.ts
dclipp: changed to return `Error | undefined`

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
   })
+
+  describe('Revocation', (): void => {
+    let grant: Grant
+    let token: AccessToken
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let tokenQuery: <M extends { QueryBuilderType: any }>(
+      trxOrKnex?: Objection.TransactionOrKnex | undefined
+    ) => Objection.QueryBuilderType<M>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
njlie, line 219: Can you comment (in the code) why this and line 215 need to be cast to `any`?

packages/auth/src/accessToken/routes.test.ts
njlie: @dclipp The empty object in the `createContext` call _is_ the params object, so the `id` could be moved there. Does including the `params` in the first argument get it passed into the context?

See https://github.com/interledger/rafiki/blob/236912d7d12ed012d5e7f51496879ccc4f8901d8/packages/auth/src/tests/context.ts#L8 and https://github.com/interledger/rafiki/blob/236912d7d12ed012d5e7f51496879ccc4f8901d8/packages/auth/src/tests/context.ts#L14

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+      )
+
+      tokenQuery = () => {
+        return {
+          findById: () => {
+            return undefined
+          }
+        }
+      }
+      AccessToken.query = jest.fn(tokenQuery)
```
njlie, line 267: It might not be necessary to mock this function if this test's purpose is to verify a particular row doesn't exist.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
+                      where: () => Promise.resolve()
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+      AccessToken.query = jest.fn(tokenQuery)
```
njlie, line 303: It might be better to actually insert a row into the access token table (and the grant & access tables) so that this test covers more.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,136 @@ describe('Access Token Routes', (): void => {
...
```
njlie, line 337: Ditto of https://github.com/interledger/rafiki/pull/360/files#r898562906

packages/auth/src/accessToken/routes.test.ts
dclipp: Good catch, I moved it to the empty object and was able to eliminate the mock for the `404` test

njlie (approved):

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,126 @@ describe('Access Token Routes', (): void => {
...
+
+      await accessTokenRoutes.revoke(ctx)
+      expect(ctx.response.status).toBe(204)
+    })
+
+    test('Modifies token if it has not expired', async (): Promise<void> => {
```
sabineschaller, line 300: Same as above. Tbh, Idk why the token query has to be mocked here since the token should still be valid for a bit.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -207,4 +208,126 @@ describe('Access Token Routes', (): void => {
...
+        { id }
+      )
+
+      await expect(accessTokenRoutes.revoke(ctx)).rejects.toMatchObject({
+        status: 404,
+        message: 'token not found'
+      })
+    })
+
+    test('Does not modify token if it has already expired', async (): Promise<void> => {
```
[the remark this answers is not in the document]
dclipp: good point

sabineschaller (approved):
LGTM

PR #366, opened 2022-06-10
chore(backend): validate outgoing payment grant limits

### LINKED ISSUE #253: Check outgoing payment grant
- add token introspection query to configured AS (https://github.com/interledger/rafiki/pull/313)
- store grant id with payment
  - (lock and) check existing grant payments on payment creation

wilsonianb:
TODO:
- ~abstract to also handle incoming payment limits?~

- validate outgoing payment grant limits

## Context

- fixes #253

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


njlie (approved):

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -108,3 +142,58 @@ export class Grant {
...
+  } else if (
+    (parts[0].length === 1 && parts[0][0] === 'R') ||
+    (parts[0].length > 1 && parts[0][1] === '-')
+  ) {
+    repetitions = Infinity
+  }
+  if (repetitions === undefined || isNaN(repetitions)) return
+
+  const interval = Interval.fromISO(`${parts[1]}/${parts[2]}`)
+  if (!interval.start) return
```
wilsonianb, line 166:
```suggestion
  if (!interval.isValid || !interval.start) return
```

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -108,3 +142,58 @@ export class Grant {
...
+  if (parts[1].length > 1 && parts[1][0] === 'P') {
+    duration = parts[1]
+    forward = false
+  } else if (parts[2].length > 1 && parts[2][0] === 'P') {
+    duration = parts[2]
+    forward = true
+  } else {
+    duration = interval.toDuration().toString()
+    forward = true
+  }
```
wilsonianb, line 180: These checks were already performed inside [Interval.fromISO](https://github.com/moment/luxon/blob/0cbffd9c87eca6ac9a75dce124ce890dd945bc78/src/interval.js#L134)
What if we call `DateTime.fromISO`/`Duration.fromISO`/etc on `parts[1]` and `parts[2]` instead of `Interval.fromISO`, so that we can capture `forward` and the duration (as a `Duration`, and not re-parse the duration for every interval below)?

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -108,3 +142,58 @@ export class Grant {
...
+
+  const intervals = [interval]
```
wilsonianb, line 182: Could `interval` just be reassigned in the loop instead of building an `intervals` array?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+  const estTotalAvailableReceiveAmount =
+    availableReceiveAmount +
+    BigInt(
+      Math.floor(
+        (Number(payment.quote.lowEstimatedExchangeRate.a) /
+          Number(payment.quote.lowEstimatedExchangeRate.b)) *
+          (1 + slippage) *
+          Number(availableSendAmount)
+      )
+    )
```
wilsonianb, line 297: It's not clear to me why we wouldn't just use `availableSendAmount` and `availableReceiveAmount`

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+    paymentAccess,
+    payment,
+    deps.slippage
+  )
+
+  if (
+    estTotalAvailableSendAmount < payment.sendAmount.value ||
+    estTotalAvailableReceiveAmount < payment.receiveAmount.value
+  ) {
+    // Payment amount single-handedly exceeds sendAmount limit(s)
```
wilsonianb, line 382:
```suggestion
    // Payment amount single-handedly exceeds amount limit(s)
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+    return false
+  }
+
+  // TODO: lock to prevent multiple grant payments from being created concurrently?
```
wilsonianb, line 386: bump

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+        grantPayment.id
+      )
+      assert.ok(totalSent !== undefined)
+      if (totalSent === BigInt(0)) {
+        continue
+      }
+      grantPayment.sentAmount = { ...grantPayment.sentAmount, value: totalSent }
+    } else {
+      grantPayment.sentAmount = grantPayment.sendAmount
+    }
```
wilsonianb, line 413: This could probably be moved down into the next `for` loop

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+    const competingUnrelatedAccess = unrelatedAccess.filter((access) =>
+      validateAccess({
+        access,
+        payment: grantPayment,
+        identifier: `${deps.publicHost}/${grantPayment.accountId}`
+      })
+    )
+    if (competingUnrelatedAccess.length > 0) {
+      for (const access of unrelatedAccess) {
+        if (competingUnrelatedAccess.indexOf(access) > -1) {
```
wilsonianb, line 461: Instead of populating and then searching `competingUnrelatedAccess`, could `validateAccess` be called at the start of this `for` loop?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+          if (access.limits?.sendAmount) {
```
wilsonianb, line 462: Maybe we could wait to assign `sentAmount` here (if it's still undefined).
Ditto for `receivedAmount` below

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+            if (access.limits.sendAmount.value > sentAmount) {
+              access.limits.sendAmount.value -= sentAmount
+              sentAmount = BigInt(0)
+              receivedAmount = BigInt(0)
+            } else {
+              receivedAmount -=
+                (access.limits.sendAmount.value *
+                  grantPayment.receiveAmount.value) /
+                grantPayment.sentAmount.value
```
wilsonianb, line 471: Similar to https://github.com/interledger/rafiki/pull/366#discussion_r935884318, why is `sendAmount` limit accounting affecting `receiveAmount` limit accounting?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+              sentAmount -=
+                (access.limits.receiveAmount.value *
+                  grantPayment.sentAmount.value) /
+                grantPayment.receiveAmount.value
+              receivedAmount -= access.limits.receiveAmount.value
+              access.limits.receiveAmount.value = BigInt(0)
+            }
+          }
+        } else {
+          continue
```
wilsonianb, line 490:
```suggestion
```

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,566 @@ describe('OutgoingPaymentService', (): void => {
...
+          })
+          assert.ok(firstPayment)
+          if (failed) {
+            await firstPayment
+              .$query(trx)
+              .patch({ state: OutgoingPaymentState.Failed })
+
+            jest
+              .spyOn(accountingService, 'getTotalSent')
+              .mockResolvedValueOnce(sendAmount ? BigInt(188) : BigInt(188 * 2))
```
wilsonianb, line 621: I tried changing this to:
```suggestion
              .mockResolvedValueOnce(sendAmount ? BigInt(11) : BigInt(188 * 2))
```
but the payment is created successfully and the test fails.

I think this is due to `estTotalAvailableSendAmount` not being updated:
https://github.com/interledger/rafiki/pull/366/files#diff-c7f2b9ad5c5975c96f5355c09c8f95fcf5da3c117b1f073ff2c2ec2ee9d0b338R465
https://github.com/interledger/rafiki/pull/366/files#diff-c7f2b9ad5c5975c96f5355c09c8f95fcf5da3c117b1f073ff2c2ec2ee9d0b338R465

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -1,4 +1,5 @@
 import assert from 'assert'
+import { Interval, Duration, DateTime } from 'luxon'
```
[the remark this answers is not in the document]
sabineschaller: Thank you! Will make the change

packages/backend/src/open_payments/auth/grant.ts
sabineschaller: If don't use `Interval.fromISO`, we can't capture those intervals of the form `repetition/start/end`. But I can move the parsing out of the loop.

packages/backend/src/open_payments/auth/grant.test.ts
```diff
@@ -81,3 +82,59 @@ describe('Grant', (): void => {
     })
   })
 })
+
+describe('Interval', (): void => {
+  it('parses a repeating interval / start + duration', (): void => {
```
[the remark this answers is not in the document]
sabineschaller: I'm not sure I know what you are trying to tell me here.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: I probably don't need to calculate both. I think we should work through this part together next week.

packages/backend/src/open_payments/auth/grant.test.ts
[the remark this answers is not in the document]
wilsonianb: I meant test with `target` (the second `getInterval` argument) as each of those listed times relative to the repeating interval.

packages/backend/src/open_payments/auth/grant.ts
wilsonianb: >If don't use `Interval.fromISO`, we can't capture those intervals of the form `repetition/start/end`.

Not sure I follow.
I think we should be able to replace the call to `Interval.fromISO` by implementing this part of it ourselves (still using luxon to `DateTime.fromISO` and `Duration.fromISO`):
https://github.com/moment/luxon/blob/0cbffd9c87eca6ac9a75dce124ce890dd945bc78/src/interval.js#L137-L167

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,326 @@ async function createOutgoingPayment(
...
+      }
+      if (!access.limits.receiveAmount) {
+        noReceiveAmount = true
+      }
+      if (noSendAmount && noReceiveAmount) {
+        return true
+      }
+      let paymentInterval: Interval | undefined
+      if (access.interval) {
+        paymentInterval = getInterval(access.interval, payment.createdAt)
```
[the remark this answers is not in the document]
sabineschaller: Sounds like a good idea especially when we have very long repetitions.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: Yeah, let's talk through my thinking in person.

packages/backend/src/open_payments/payment/outgoing/service.ts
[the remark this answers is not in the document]
sabineschaller: I decided to pass the interval to `validateAccess` rather than returning it. I had a hard time renaming `validateAccess` to make it clear that it validates access and returns an interval.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: Since I also need it outside of the `if` statement, I don't think I can.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: But then I always loop, even if there is no competing unrelated access. But that may still be faster than doing the filter operation. it gets rid of `competingUnrelatedAccess` so that's a plus.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,566 @@ describe('OutgoingPaymentService', (): void => {
...
+      test.each`
+        sendAmount | description
+        ${true}    | ${'sendAmount'}
+        ${false}   | ${'receiveAmount'}
+      `(
+        'fails if grant limit $description is not enough for payment ',
+        async ({ sendAmount }): Promise<void> => {
+          const amount = {
+            value: BigInt(12),
+            assetCode: 'USD',
```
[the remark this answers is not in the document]
sabineschaller: I changed it. It makes the test better. Thanks!

packages/backend/src/open_payments/payment/outgoing/service.test.ts
sabineschaller: I think that is because 11 doesn't use up the grant and a second payment can still be successfully created.

packages/backend/src/open_payments/payment/outgoing/service.ts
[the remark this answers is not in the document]
wilsonianb: I'm thinking `getInterval` should be called as little as possible since it has a linear search for a particular interval.
We can change `unlimitedAccess` and `unrelatedAccess` to both also be `PaymentAccess[]` to avoid re-calling `getInterval` in the loops below.
We _could_ also check access `limit`s before ever calling `getInterval`, so that `getInterval` potentially never called (as opposed to currently being called for every grant `access`). Maybe we do something like replace the first call to `validatePayment` with `validatePaymentLimits` and change the subsequent calls to `validatePaymentAccess`?

packages/backend/src/open_payments/payment/outgoing/service.test.ts
wilsonianb: You're right. I couldn't keep my amounts straight.
The test passes for `78` but fails for `77`, which is correct with the grant send amount limit of `200` and the payment send amount of `123`. :heavy_check_mark:

packages/backend/src/open_payments/payment/outgoing/service.ts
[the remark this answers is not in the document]
sabineschaller: You are right. I'll change it to use `getInterval` as little as possible.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: Just to make sure I understand correctly, you want to lock querying of the grant payments to prevent another outgoing payment from being created concurrently?

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -108,3 +142,58 @@ export class Grant {
...
   }
 }
+
+// Export for testing
+export function getInterval(
+  repeatingInterval: string,
+  target: Date
+): Interval | undefined {
+  const parts = repeatingInterval.split('/')
+  assert.ok(parts.length === 3)
```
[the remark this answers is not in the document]
sabineschaller: But I don't know how.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,181 @@ async function createOutgoingPayment(
...
+// "payment" is locked by the "deps.knex" transaction.
+async function validateGrant(
+  deps: ServiceDependencies,
+  payment: OutgoingPayment,
+  grant: Grant
+): Promise<boolean> {
+  const grantAccess = grant.getAccess({
+    type: AccessType.OutgoingPayment,
+    action: AccessAction.Create
+  })
```
wilsonianb, line 276: The [auth middleware](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/middleware.ts#L28) already calls `grant.includesAccess`
```suggestion
  const grantAccess = grant.access[0]
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,181 @@ async function createOutgoingPayment(
...
+}: {
+  access: GrantAccess
+  payment: OutgoingPayment
+  identifier: string
+}): PaymentAccess | undefined {
+  let paymentInterval: Interval | undefined
+  if (access.interval)
+    paymentInterval = getInterval(access.interval, payment.createdAt)
+  if (
+    (!access.identifier || access.identifier === identifier) &&
```
wilsonianb, line 201: `identifier` was already checked by auth middleware
```suggestion
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,181 @@ async function createOutgoingPayment(
...
+    (!access.interval || paymentInterval !== undefined) &&
+    (!access.limits || validateAccessLimits(payment, access.limits))
```
wilsonianb, line 203: We could move this ahead of `getInterval` to avoid a call.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +185,181 @@ async function createOutgoingPayment(
...
+  access: PaymentAccess
+  payment: OutgoingPayment
+  identifier: string
+}): boolean {
+  if (
+    (!access.identifier || access.identifier === identifier) &&
+    (!access.paymentInterval ||
+      (access.paymentInterval.start.toMillis() <= payment.createdAt.getTime() &&
+        payment.createdAt.getTime() < access.paymentInterval.end.toMillis())) &&
+    (!access.limits || validateAccessLimits(payment, access.limits))
```
wilsonianb, line 223: :thinking: If grants were immutable, we would only need to check if an existing payment is within the same interval.
Instead, we have the opportunity for an existing payment to no longer match `identifier`, `limits` fields, etc.

packages/backend/src/open_payments/payment/outgoing/service.ts
wilsonianb: Yeah, I'm not sure if it's possible to just do an `insert` lock.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,318 @@ describe('OutgoingPaymentService', (): void => {
...
+              type: AccessType.OutgoingPayment,
+              actions: [AccessAction.Create, AccessAction.Read],
+              identifier: `${Config.publicHost}/${uuid()}`
+            }
+          ]
+        })
+        await expect(
+          outgoingPaymentService.create({ ...options, grant })
+        ).resolves.toEqual(OutgoingPaymentError.InsufficientGrant)
+      })
```
wilsonianb, line 485:
```suggestion
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +178,142 @@ async function createOutgoingPayment(
...
+function validatePaymentLimits({
+  limits,
+  payment
+}: {
+  limits: PaymentLimits
+  payment: OutgoingPayment
+}): boolean {
+  if (
+    (!limits.receiver || payment.receiver === limits?.receiver) &&
+    validateAmountAssets(payment, limits) &&
```
sabineschaller, line 207: I was wondering whether we have to include this check since it was already checked on payment creation of the old payments and this function is only called on old payments under the same grant.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: Maybe we should just allow `limits` changes. Everything else requires a revocation of an old grant and the issuance of a new grant?
@adrianhopebailie @dclipp @njlie

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: That relates to [this comment](https://github.com/interledger/rafiki/pull/366#discussion_r947929309)

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +178,140 @@ async function createOutgoingPayment(
...
+  payment: OutgoingPayment
+}): boolean {
+  if (
+    !limits.paymentInterval ||
+    (limits.paymentInterval.start.toMillis() <= payment.createdAt.getTime() &&
+      payment.createdAt.getTime() < limits.paymentInterval.end.toMillis())
+  ) {
+    return true
+  }
+  return false
```
wilsonianb, line 212:
```suggestion
  return
    !limits.paymentInterval ||
    (limits.paymentInterval.start.toMillis() <= payment.createdAt.getTime() &&
      payment.createdAt.getTime() < limits.paymentInterval.end.toMillis())
```

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: When using `knex.raw`, [we can add a lock](https://stackoverflow.com/questions/52325771/manually-lock-table-in-knex). But I don't know if and how that works with the transaction.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +178,140 @@ async function createOutgoingPayment(
...
+
+// "payment" is locked by the "deps.knex" transaction.
+async function validateGrant(
+  deps: ServiceDependencies,
+  payment: OutgoingPayment,
+  grant: Grant
+): Promise<boolean> {
+  const grantAccess = grant.access[0]
+  if (!grantAccess.limits?.sendAmount && !grantAccess.limits?.receiveAmount) {
+    return true
```
wilsonianb, line 245: This returns `true` before `receiver` is checked.
Maybe it should be `if (!grantAccess.limits)` here and move this below `validateAccessLimits`?

packages/backend/src/open_payments/payment/outgoing/service.ts
wilsonianb: It'd be great if we could limit the lock to prevent inserts with the same `grantId`

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,286 @@ describe('OutgoingPaymentService', (): void => {
...
+        async ({ limits }): Promise<void> => {
+          const grant = new Grant({
+            active: true,
+            grant: uuid(),
+            access: [
+              {
+                type: AccessType.OutgoingPayment,
+                actions: [AccessAction.Create, AccessAction.Read],
+                identifier: `${Config.publicHost}/${accountId}`,
+                interval: `R0/${start.toISOString()}/P1M`,
```
wilsonianb, line 491: `interval`s need to be moved inside `limits`

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,286 @@ describe('OutgoingPaymentService', (): void => {
...
+              interval: `R0/${start.toISOString()}/P1M`,
+              limits: undefined
+            }
+          ]
+        })
+        await expect(
+          outgoingPaymentService.create({ ...options, grant })
+        ).resolves.toBeInstanceOf(OutgoingPayment)
+      })
+      test('succeeds if grant limits do not specify send or receive amount', async (): Promise<void> => {
```
wilsonianb, line 619: These two could be combined into a `test.each` with different `limits` defined.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -428,6 +431,283 @@ describe('OutgoingPaymentService', (): void => {
...
+        async ({ limits }): Promise<void> => {
+          const grant = new Grant({
+            active: true,
+            grant: uuid(),
+            access: [
+              {
+                type: AccessType.OutgoingPayment,
+                actions: [AccessAction.Create, AccessAction.Read],
+                identifier: `${Config.publicHost}/${accountId}`,
+                limits: { ...limits, interval: `R0/${start.toISOString()}/P1M` }
```
wilsonianb, line 491: What if `interval` were defined in the `beforeEach` (instead of `start`)?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -38,9 +40,10 @@ export interface OutgoingPaymentService {
 export interface ServiceDependencies extends BaseService {
   knex: TransactionOrKnex
   accountingService: AccountingService
-  accountService: AccountService
   peerService: PeerService
   makeIlpPlugin: (options: IlpPluginOptions) => IlpPlugin
+  publicHost: string
+  slippage: number
```
wilsonianb, line 46: This doesn't appear to be used
```suggestion
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -159,6 +178,138 @@ async function createOutgoingPayment(
...
+): boolean {
+  if (
+    limits.sendAmount &&
+    (limits.sendAmount.assetCode !== payment.asset.code ||
+      limits.sendAmount.assetScale !== payment.asset.scale)
+  ) {
+    return false
+  }
+  return (
+    (!limits.sendAmount && !limits.receiveAmount) ||
```
wilsonianb, line 224: I think this should be moved into `validateGrant` after the call to `validateAmountAssets` (to return true before querying other grant payments).

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -154,11 +186,144 @@ async function createOutgoingPayment(
       }
     } else if (isOutgoingPaymentError(err)) {
       return err
+    } else if (err instanceof KnexTimeoutError) {
+      return OutgoingPaymentError.GrantLocked
```
wilsonianb, line 190: We're generally returning `OutgoingPaymentError` for caller related issues (invalid options, etc).
I'm thinking this should log and still throw.

wilsonianb (commented):
`main` needs to merged in / conflicts resolved

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -154,11 +186,144 @@ async function createOutgoingPayment(
...
+    (paymentLimits.sendAmount &&
+      paymentLimits.sendAmount.value < payment.sendAmount.value) ||
+    (paymentLimits.receiveAmount &&
+      paymentLimits.receiveAmount.value < payment.receiveAmount.value)
+  ) {
+    // Payment amount single-handedly exceeds amount limit
+    return false
+  }
+
+  const grantPayments = await OutgoingPayment.query(deps.knex)
```
wilsonianb, line 274: :thinking: Technically, we don't need to lock the grant until just before we perform this query.

sabineschaller: That is correct but in the realm of what was said this morning, I'd rather keep it as is. I don't think it makes a big difference.

wilsonianb: :+1: I opened an issue for it:
- https://github.com/interledger/rafiki/issues/485

wilsonianb (commented):
:+1:
Screenshot from 2022-08-23 11-25-10

njlie (approved):

PR #370, opened 2022-06-15
feat(auth): gnap interaction

### LINKED ISSUE #271: Handle user interaction (redirect initially)
On redirect to the interaction URL provided by the AS during the initial grant request:
* Sign an interaction id cookie to be set on the response.
* Redirect the user to the (config-specified in the AS) site where their account is managed (RS).
* Have the user login & approve access to client.
* Have the RS `POST` to AS that session is complete - signed cookie with interaction id should still be provided.
* Redirect user to the redirect URL provided by the client with hash in query params.

njlie:

- Closes #271.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/config/app.ts
```diff
@@ -32,5 +34,6 @@ export const Config = {
   ]),
   resourceServerDomain: envString('INTERACT_URL', 'http://localhost:3004'),
   authServerDomain: envString('DOMAIN', 'http://localhost:3006'), // TODO: replace this with whatever frontend port ends up being
-  waitTime: envInt('WAIT', 5)
+  waitTime: envInt('WAIT', 5),
```
matdehaast, line 37: we should add units to this variable. Difficult to know what it represents when glancing

packages/auth/src/grant/routes.test.ts
```diff
@@ -664,4 +668,273 @@ describe('Grant Routes', (): void => {
...
+              Accept: 'application/json',
+              'Content-Type': 'application/json'
+            }
+          },
+          { interactId: 'unknown_interaction' }
+        )
+
+        await expect(grantRoutes.interaction.get(ctx)).resolves.toBeUndefined()
+        expect(ctx.status).toBe(401)
+        expect(ctx.body).toEqual({ error: 'unknown_request' })
```
matdehaast, line 699: Does the spec say anything about the return codes here?

packages/auth/src/grant/service.ts
```diff
@@ -61,12 +69,43 @@ export async function createGrantService({
...
+    issueGrant: (grantId: string) => issueGrant(deps, grantId)
+  }
+}
+
+async function issueGrant(
+  deps: ServiceDependencies,
+  grantId: string,
+  trx?: Transaction
+): Promise<{ grant: Grant; accessToken: AccessToken }> {
+  // TODO: create access token, update grant state
```
matdehaast, line 89: Is this TODO still required? Seems you are doing that below

matdehaast (commented):
Looks good so far, just some open questions

packages/auth/src/grant/routes.test.ts
njlie: It doesn't seem to specify the return code, just the body: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.6

sabineschaller (commented):
I know that is what the issue says but I'm not so sure that the flow described in the issue is what we want to do. Maybe we should all discussed that on a call again @njlie @matdehaast @adrianhopebailie .

packages/auth/src/grant/routes.ts
```diff
@@ -1,26 +1,31 @@
...
 
 export interface GrantRoutes {
   create(ctx: AppContext): Promise<void>
-  // interaction: {
-  //   get(ctx: AppContext): Promise<void>
-  //   post(ctx: AppContext): Promise<void>
-  // }
+  interaction: {
+    get(ctx: AppContext): Promise<void>
+    post(ctx: AppContext): Promise<void>
```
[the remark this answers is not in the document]
njlie: I think `start` and `finish` would be great. With the way I've structured the grant routes service these endpoints would be structured behind `grantRoutes.interaction.start/finish` so that should efficiently communicate what it is.

packages/auth/src/grant/service.ts
```diff
@@ -120,14 +148,18 @@ async function initiateGrant(
...
-      await invTrx.commit()
+      await invTrx.rollback()
     }
 
     throw err
   }
 }
+
+async function getByInteraction(interactId: string): Promise<Grant> {
+  return Grant.query().where({ interactId }).first()
```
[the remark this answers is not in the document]
njlie: `interactId` is [already unique](https://github.com/interledger/rafiki/blob/a16c71d378074c685f166bd7e89cc26363a5da0f/packages/auth/migrations/20220504161932_create_grants_table.js#L17), so `findOne` would indeed be better, good point.

packages/auth/src/access/types.ts
```diff
@@ -109,6 +109,7 @@ export function isAccessRequest(
   )
 }
 
+// value should hold bigint, serialized as string for requests & storage as jsonb field in postgres
```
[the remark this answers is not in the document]
njlie: `jsonb` is a way that Postgres can store JSON payloads. Rather than storing a stringified form of the JSON like the `json` format does, `jsonb` decomposes it into a binary format so that it may be indexed.
https://www.postgresql.org/docs/current/datatype-json.html

packages/auth/src/grant/routes.ts
```diff
@@ -84,3 +94,83 @@ async function createGrantInitiation(
...
+
+  if (!registryData) {
+    ctx.status = 401
+    ctx.body = {
+      error: 'invalid_client'
+    }
+    return
+  }
+
+  const interactionUrl = new URL(config.resourceServerDomain)
```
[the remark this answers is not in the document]
njlie: @sabineschaller What do you think about shortening it to `accountProviderDomain`?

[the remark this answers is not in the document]
sabineschaller: I googled the term. It's "account servicing payment service provider". What about `ASPSPDomain` or `ASPSPAuthDomain`?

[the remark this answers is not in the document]
sabineschaller: Ok, chatted to Matt and we should call it `identityServerDomain` 💡

packages/auth/src/accessToken/service.ts
```diff
@@ -85,3 +99,31 @@ async function revoke(
...
+async function createAccessToken(
+  deps: ServiceDependencies,
+  grantId: string,
+  opts?: AccessTokenOpts
+): Promise<AccessToken> {
+  const invTrx = opts?.trx || (await AccessToken.startTransaction(deps.knex))
+  try {
+    const accessToken = await AccessToken.query(invTrx).insert({
+      value: crypto.randomBytes(8).toString('hex').toUpperCase(), // TODO: factor out nonce generation
+      managementId: v4(),
```
sabineschaller, line 112: Is that TODO something you still want to include in this PR?

packages/auth/src/grant/routes.ts
```diff
@@ -84,3 +94,83 @@ async function createGrantInitiation(
...
+
+  const clientRedirectUri = new URL(grant.finishUri)
+  const { clientNonce, interactNonce, interactRef } = grant
+  const interactUrl = config.identityServerDomain + `/interact/${interactId}`
+
+  // https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-4.2.3
+  const data = `${clientNonce}\n${interactNonce}\n${interactRef}\n${interactUrl}`
+
+  // TODO: support other hash_methods if provided
+  const hash = crypto.createHash('sha3-512').update(data).digest('base64')
```
sabineschaller, line 172: Same question here.

njlie: Probably not, I think our spec doesn't offer ways to configure this in a grant request.

packages/auth/src/accessToken/service.ts
```diff
@@ -85,3 +99,31 @@ async function revoke(
     return new Error('token not found')
   }
 }
+
+async function createAccessToken(
+  deps: ServiceDependencies,
+  grantId: string,
+  opts?: AccessTokenOpts
+): Promise<AccessToken> {
+  const invTrx = opts?.trx || (await AccessToken.startTransaction(deps.knex))
```
sabineschaller, line 108: Why is that called `invTrx`? 
I actually don't think we have to do the create + commit approach here because there is nothing else going on at the same time. We only create the access token and don't call on another service that could cause problems in which case we don't want to commit the transaction. 
So you could just use `AccessToken.query(deps.knex).insertAndFetch`.

packages/auth/src/grant/routes.ts
```diff
@@ -84,3 +94,82 @@ async function createGrantInitiation(
...
+  const { clientNonce, interactNonce, interactRef } = grant
+  const interactUrl = config.identityServerDomain + `/interact/${interactId}`
+
+  // https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-4.2.3
+  const data = `${clientNonce}\n${interactNonce}\n${interactRef}\n${interactUrl}`
+
+  const hash = crypto.createHash('sha3-512').update(data).digest('base64')
+  clientRedirectUri.searchParams.set('hash', hash)
+  clientRedirectUri.searchParams.set('interact_ref', interactRef)
+  ctx.redirect(clientRedirectUri.toString())
```
sabineschaller, line 174: I guess I'm a bit confused about the user redirect in general. 
`start` will redirect the user to the identity server. There, the user logs in and (for now) accepts the grant. The identity server does a POST to the `finish` endpoint to confirm that the user accepted the grant. Wouldn't the identity server have to redirect the user to the `clientRedirectURI`?

packages/auth/src/grant/service.ts
```diff
@@ -112,22 +140,26 @@ async function initiateGrant(
 
     return {
       interact: {
-        redirect: config.resourceServerDomain + `/interact/${grant.interactId}`,
+        redirect: config.identityServerDomain + `/interact/${grant.interactId}`,
```
sabineschaller, line 143: I wonder whether we should move the build of the request response to the route and just return `grant` here.

packages/auth/src/grant/routes.ts
njlie: Yes, it's the [RO's job to redirect the user](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-4.2.1). I think unless the RO made this POST request in a `<form>` element, it would end up being the one performing the redirect in the `Location` header of this response.

sabineschaller: Got it, thanks!

packages/auth/src/app.ts
```diff
@@ -187,6 +204,27 @@ export class App {
...
     }
 
+    // Interaction
+    this.publicRouter.get(
+      '/interact/:interactId',
+      grantRoutes.interaction.start
+    )
+
+    this.publicRouter.post(
+      '/interact/:interactId/login',
```
sabineschaller, line 214: Why do you want to add the `/login`?

sabineschaller (commented):
You should remove `config` from the `grantService` singleton in index.ts.

packages/auth/src/grant/model.ts
```diff
@@ -41,7 +41,6 @@ export class Grant extends BaseModel {
       }
     }
   })
-  public access!: Access[]
```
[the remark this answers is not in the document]
njlie: Oh, okay. I thought we had forgotten to remove it because it wasn't in the migration. I'll keep it.

packages/auth/src/grant/routes.ts
```diff
@@ -80,7 +90,98 @@ async function createGrantInitiation(
...
+  if (!grant) {
+    ctx.status = 401
+    ctx.body = {
+      error: 'unknown_request'
+    }
+
+    return
+  }
+
+  ctx.session.interactId = grant.interactId
```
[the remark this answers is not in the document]
njlie: I believe session cookies would be written to the AS only, so they should remain first-party.

packages/auth/src/grant/service.test.ts
```diff
@@ -106,47 +103,27 @@ describe('Grant Service', (): void => {
         }
       }
 
-      const grantResponse = await grantService.initiateGrant(grantRequest)
+      const grant = await grantService.initiateGrant(grantRequest)
 
-      expect(grantResponse).toEqual(
+      expect(grant).toEqual(
         expect.objectContaining({
```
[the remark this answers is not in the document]
njlie: None, my initial assumption was that `.toMatch` would expect an _exact_ match but looking at the documentation, it only expects a subset, so I'll use `.toMatchObject`

packages/auth/src/app.ts
sabineschaller: I think we can just leave it. We have the `GET` to this url, which is for `start`, and the `POST`, which is for `finish`. At least that is how I speced it out [here](https://github.com/interledger/open-payments/pull/141).

sabineschaller (approved):
I think that looks good now

PR #372, opened 2022-06-17
feat(auth): grant continue request

### LINKED ISSUE #269: Handle continuation Request

njlie:

- Closes #269 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


njlie:
Will change the base to `main` once https://github.com/interledger/rafiki/pull/370 is merged in.

packages/auth/src/app.ts
```diff
@@ -215,6 +216,8 @@ export class App {
       grantRoutes.interaction.finish
     )
 
+    this.publicRouter.post('/auth/continue/:id', grantRoutes.post)
```
sabineschaller, line 219: I didn't realize that you reintroduced these "individual" routes (for the lack of a better term) in your previous PR. When you look at lines 167-205, I actually included functionality that only creates routes for those speced out. You should add the one for the `continue` request there (and maybe remove the ones for introspect, rotate, and revoke). Now that we've also speced out the interaction endpoints, we could include them in the functionality of lines 167-205 as well. But that can be in another PR.

packages/auth/src/grant/service.ts
```diff
@@ -60,23 +68,44 @@ export async function createGrantService({
...
 async function issueGrant(
   deps: ServiceDependencies,
-  grantId: string
+  grantId: string,
+  trx?: Transaction
 ): Promise<Grant> {
-  return Grant.query(deps.knex).patchAndFetchById(grantId, {
-    state: GrantState.Granted
-  })
+  const invTrx = trx || (await Grant.startTransaction())
```
sabineschaller, line 92: nit pick: let's not call it `invTrx`. I think that still stands for "invoice transaction".

Also, what is the concurrent db transaction that we would pass here? I think what you had before was sufficient.

packages/auth/src/grant/service.ts
```diff
@@ -132,3 +161,11 @@ async function initiateGrant(
 async function getByInteraction(interactId: string): Promise<Grant> {
   return Grant.query().findOne({ interactId })
 }
+
+async function getByContinue(
+  continueId: string,
+  continueToken: string,
+  interactRef: string
+): Promise<Grant> {
+  return Grant.query().where({ continueId, continueToken, interactRef }).first()
```
sabineschaller, line 170: Why do we need all three of them?

packages/auth/src/grant/service.ts
```diff
@@ -7,15 +7,22 @@ import { Grant, GrantState, StartMethod, FinishMethod } from './model'
...
+  getByContinue(
+    continueId: string,
+    continueToken: string,
+    interactRef: string
+  ): Promise<Grant>
 }
 
 interface ServiceDependencies extends BaseService {
   accessService: AccessService
+  config: IAppConfig
```
sabineschaller, line 25: What is that needed for?

packages/auth/src/grant/service.ts
```diff
@@ -60,23 +68,44 @@ export async function createGrantService({
...
     initiateGrant: (grantRequest: GrantRequest, trx?: Transaction) =>
       initiateGrant(deps, grantRequest, trx),
     getByInteraction: (interactId: string) => getByInteraction(interactId),
-    issueGrant: (grantId: string) => issueGrant(deps, grantId)
+    issueGrant: (grantId: string) => issueGrant(deps, grantId),
+    getByContinue: (
+      continueId: string,
+      continueToken: string,
+      interactRef: string
+    ) => getByContinue(continueId, continueToken, interactRef)
```
sabineschaller, line 83: I think we should add some basic tests for these two.

packages/auth/src/index.ts
```diff
@@ -99,6 +99,7 @@ export function initIocContainer(
       return createGrantService({
         logger: await deps.use('logger'),
         accessService: await deps.use('accessService'),
+        config: await deps.use('config'),
```
sabineschaller, line 102: As above, not sure that is needed.

packages/auth/src/grant/routes.ts
```diff
@@ -42,7 +48,8 @@ export function createGrantRoutes({
     interaction: {
       start: (ctx: AppContext) => startInteraction(deps, ctx),
       finish: (ctx: AppContext) => finishInteraction(deps, ctx)
-    }
+    },
+    post: (ctx: AppContext) => continueGrant(deps, ctx)
```
sabineschaller, line 52: Can we call that `continue`? That way, it can easily be included in the mapping functionality in app.ts (the one from spec to code I commented on in app.ts).

packages/auth/src/grant/routes.test.ts
```diff
@@ -987,4 +935,197 @@ describe('Grant Routes', (): void => {
...
+        continueToken: crypto.randomBytes(8).toString('hex').toUpperCase(),
+        continueId: v4(),
+        finishMethod: FinishMethod.Redirect,
+        finishUri: 'https://example.com',
+        clientNonce: crypto.randomBytes(8).toString('hex').toUpperCase(),
+        clientKeyId: KEY_REGISTRY_ORIGIN + TEST_KID_PATH,
+        interactId: v4(),
+        interactRef: v4(),
+        interactNonce: crypto.randomBytes(8).toString('hex').toUpperCase()
+      })
```
sabineschaller, line 953: Is that different from the one you create in `beforeEach`?

packages/auth/src/grant/routes.test.ts
```diff
@@ -987,4 +935,197 @@ describe('Grant Routes', (): void => {
...
+
+      expect(ctx.status).toBe(200)
+      expect(ctx.body).toEqual({
+        access_token: {
+          value: accessToken.value,
+          manage: Config.authServerDomain + `/token/${accessToken.managementId}`
+        }
+      })
+    })
```
sabineschaller, line 990: Can you add the `expect` that checks that in complies with the spec?

packages/auth/src/app.ts
njlie: It actually is already in the openapi routes, and is duplicated in the changed line - I just didn't notice during the rebase so I just need to remove this line.

packages/auth/src/grant/service.ts
njlie: Hmm...good point.

packages/auth/src/grant/routes.test.ts
njlie: The difference is that `GrantState` is "Granted" rather than "Pending", will update this to be more clear.

njlie:
OpenAPI validation will fail in the auth tests until https://github.com/interledger/open-payments/pull/143 is merged in.

matdehaast (approved):

Issue #321, opened 2022-04-20
Fix Docker workflow


wilsonianb:
[Docker workflow](https://github.com/interledger/rafiki/blob/main/.github/workflows/docker.yml) is failing after merging https://github.com/interledger/rafiki/pull/315 with:
```
UsageError: Couldn't find knex@npm:0.21.19 [aa597] in the currently installed PnP map - running an install might help
```

wilsonianb:
I tried updating [@yarnpkg/plugin-docker-build](https://github.com/Dcard/yarn-plugins/tree/master/packages/docker-build) to incorporate https://github.com/Dcard/yarn-plugins/commit/89a71671514b49606b20db52a71a5f3786e92fca but that didn't help.

njlie:
Something about `knex` being added to the `auth` workspace causes link between it and `backend` to be displaced somehow. The build only fails if `knex` is included in `auth` at the same version as `backend`. Removing it from `auth` will cause the `backend` build process to succeed again. Not sure why that is.

rluckom-coil:
I spent yesterday looking into this.

We use [yarn-plugin-docker-build](https://github.com/Dcard/yarn-plugins/tree/master/packages/docker-build) to do our docker builds. The way that it (is supposed to) work(s) is:

1. [resolve all dependencies](https://github.com/Dcard/yarn-plugins/blob/master/packages/docker-build/src/commands/build.ts#L104)
2. [fetch all dependencies](https://github.com/Dcard/yarn-plugins/blob/master/packages/docker-build/src/commands/build.ts#L108)
3. [Make a temporary directory for doing the docker build](https://github.com/Dcard/yarn-plugins/blob/master/packages/docker-build/src/commands/build.ts#L111)
[...]

PR #374, opened 2022-06-22
fix: docker build

DonChangfoot:
This is just a workaround for the failing docker build. Ideally, the dependency issue can be resolved and this can be reverted.

The updated version of the yarn `docker-build` allows adding dev dependencies. So we now do the build inside the docker container. This does mean that the docker image will contain dev dependencies.

- the `tsconfig.json` in each package no longer extends the base one
- the backend `.npmignore` no longer ignores `src` folder

## Context

#321 

The `docker-build` plugin will copy the files across and then generate a `lockfile`. After that it runs the `pack` command. Previously, each package would `yarn build` in the `prepack` step. For some reason, the `backend` package would fail at this step due to a missing `knex` dependency. However, it is present in the `.yarn/cache` inside the container. So it seems the pnp mapping is incorrect.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


rluckom-coil:
Probably ok to close this since we fixed the build by switching the linker.

DonChangfoot:
closed by https://github.com/interledger/rafiki/pull/377

PR #375, opened 2022-06-23
feat(auth): access token rotation

### LINKED ISSUE #335: Rotate access token
Implement `POST /auth/token/{id}` endpoint

see:
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6.1
https://docs.openpayments.guide/reference/post-token

dclipp:

- Implements access token rotation

## Context

- fixes #335 

## Checklist

- [X] Tests added/updated
- [ ] Documentation added


packages/auth/src/accessToken/service.ts
```diff
@@ -78,10 +92,44 @@ async function revoke(
 ): Promise<Error | undefined> {
   const token = await AccessToken.query(deps.knex).findById(id)
   if (token) {
-    if (!isTokenExpired(token)) {
-      await token.$query(deps.knex).patch({ expiresIn: 1 })
+    if (!token.revoked && !isTokenExpired(token)) {
+      await token.$query(deps.knex).patch({ revoked: true })
```
sabineschaller, line 96: Is there a reason we are keeping the old tokens and are setting the `revoked` property? Why not cleaning them up, together with the expired tokens?

packages/auth/migrations/20220621180000_create_access_token_revoked_column.js
```diff
@@ -0,0 +1,11 @@
...
+  return knex.schema.alterTable('accessTokens', function (table) {
+    table.boolean('revoked').nullable()
+  })
+}
+
+exports.down = function (knex) {
+  return knex.schema.alterTable('accessTokens', function (table) {
+    table.dropColumn('revoked')
+  })
+}
```
sabineschaller, line 11: I think it's ok to just add the column to the access token table.

packages/auth/src/accessToken/service.ts
dclipp: According to the spec:
`An AS MUST NOT honor a rotation request for an access token that has been revoked or otherwise disabled`

So I think we have to use the `revoked` property in order to distinguish between tokens that have expired and tokens that were explicitly revoked

packages/auth/migrations/20220621180000_create_access_token_revoked_column.js
dclipp: Does it automatically create the column if I just add the field to the model class?

packages/auth/src/accessToken/service.ts
sabineschaller: But if the token doesn't exist anymore because it expired or was revoked, it can't honor the rotation request because it can't find the original token. Maybe I'm misunderstanding, though. 🤔

dclipp: Sorry, I should've also included this from the spec:

`Note that in many cases, the access token will have expired for
   regular use.  To facilitate token rotation, the AS SHOULD honor the
   rotation request of the expired access token since it is likely that
   the client instance is attempting to refresh the expired token.`

So it seems like we have to keep the token around to be able to determine whether it has been revoked or just expired

sabineschaller: Makes sense. We have an issue for that now. #385

dclipp: @sabineschaller After rereading the spec it seems like we do need to have a `revoked` field rather than just deleting or setting tokens to be expired:

`If the access token has already expired or has been revoked through other means, the AS SHOULD honor the revocation request to the token management URI as valid, since the end result is still the token not being usable.`

If we delete the token (before whatever time we use in #385 at least) then we can't honor the _revocation_ request.
But if we keep the token and just make it expired, then we can't know whether or not to honor the _rotation_ request.

Let me know what you think

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -266,5 +266,94 @@ describe('Access Token Routes', (): void => {
...
+      id = v4()
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        { id }
+      )
+
+      await expect(accessTokenRoutes.rotate(ctx)).rejects.toMatchObject({
+        status: 400,
```
njlie, line 316: Might this make more sense as a 404?

packages/auth/src/accessToken/routes.ts
```diff
@@ -86,3 +88,22 @@ async function revokeToken(
     ctx.status = 204
   }
 }
+
+async function rotateToken(
+  deps: ServiceDependencies,
+  ctx: AppContext
+): Promise<void> {
+  //TODO: verify accessToken with httpsig method
+  const result = await deps.accessTokenService.rotate(ctx.params['id'])
```
njlie, line 97: I used [managementId](https://github.com/interledger/rafiki/blob/422f084017e3f24ddbde7c819b95a2f287fcd768/packages/auth/src/accessToken/model.ts#L23) so that management URLs wouldn't expose internal database ids. You can update the [migration file](https://github.com/interledger/rafiki/blob/dc-token-rotation/packages/auth/migrations/20220504163005_create_access_token_table.js) and add the `.unique()` modifier to it so you can call `.findOne()` with `managementId`.

packages/auth/src/app.ts
```diff
@@ -137,10 +137,7 @@ export class App {
     // Token management
     this.publicRouter.post('/auth/introspect', accessTokenRoutes.introspect)
 
-    this.publicRouter.post('/auth/token/:id', (ctx: AppContext): void => {
-      // TODO: tokenService.rotate
-      ctx.status = 200
-    })
+    this.publicRouter.post('/auth/token/:id', accessTokenRoutes.rotate)
```
njlie, line 140: To make the usage of this endpoint described in https://github.com/interledger/rafiki/pull/375/files#r911538130 not confusing (as I regretfully have written it), maybe we can change the parameter from `id` to `managementId`.

packages/auth/src/accessToken/routes.ts
dclipp: I updated it to use `managementId`, but should `managementId` be unique if we can have expired tokens for the same access token?

njlie: From the GNAP spec [section 3.2.1](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.2.1) (emphasis mine):

> manage (string):  The management URI for this access token.  If
      provided, the client instance MAY manage its access token as
      described in [Section 6](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6).  This management URI is a function of the
      AS and is separate from the RS the client instance is requesting
      access to.  **This URI MUST NOT include the access token value and
      SHOULD be different for each access token issued in a request.**
      OPTIONAL.

So we should be generating a new `managementId` each time a token is rotated. It might make sense to just delete the old token and return a newly created one (with corresponding `managementId`) but it could work either way.

packages/auth/src/accessToken/service.ts
sabineschaller: What if on revocation, the API didn't return a 404 for tokens that weren't found but always a 204 (given that auth is correct, i.e. no 401)? That way, the end result is still the same for the user. Token found - delete token - 204. Token not found - do nothing - 204. The spec doesn't mention a 404. We may just have put it in there without thinking about it.

packages/auth/src/accessToken/routes.ts
```diff
@@ -79,10 +81,27 @@ async function revokeToken(
...
+  const result = await deps.accessTokenService.rotate(
+    ctx.params['managementId']
+  )
+  if (result.success) {
+    ctx.status = 200
+    ctx.body = {
+      access_token: {
+        access: [accessToBody(result.access)]
+      }
+    }
```
sabineschaller, line 102: Isn't that missing `value`, `manage`, and `expires_in`? I realize that the spec don't say that `manage` and `expires_in` are required but I think they should. I'll include that in my spec updates.

packages/auth/src/accessToken/service.ts
```diff
@@ -72,16 +86,39 @@ async function introspect(
...
+  managementId: string
+): Promise<Rotation> {
+  let access: Access | undefined
+  let error: Error | undefined
+
+  const token = await AccessToken.query(deps.knex).findOne({ managementId })
+  if (token) {
+    await token.$query(deps.knex).patch({
+      value: uuid()
+    })
```
sabineschaller, line 107: Note that @njlie pointed out that the `managementId` SHOULD be updated, too.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -285,4 +283,67 @@ describe('Access Token Routes', (): void => {
...
+    test('Can rotate token', async (): Promise<void> => {
+      const ctx = createContext(
+        {
+          headers: { Accept: 'application/json' }
+        },
+        { managementId }
+      )
+
+      await accessTokenRoutes.rotate(ctx)
+      expect(ctx.response.status).toBe(200)
```
sabineschaller, line 333: Can you also test for the body, the conent-type, and that it is satisfies the API spec (` expect(ctx.response).toSatisfyApiSpec()`)?

sabineschaller (commented):
I think you also need to update your PR description to not mention the `revoked` column anymore.

packages/auth/src/accessToken/service.ts
sabineschaller: After reading the spec again, I think we should just delete the old token and create a new one.

packages/auth/src/accessToken/routes.ts
sabineschaller: PR to update the spec: https://github.com/interledger/open-payments/pull/140

packages/auth/src/accessToken/service.ts
```diff
@@ -72,16 +89,41 @@ async function introspect(
...
+async function rotate(
+  deps: ServiceDependencies,
+  managementId: string
+): Promise<Rotation> {
+  let token = await AccessToken.query(deps.knex).findOne({ managementId })
+  if (token) {
+    await token.$query(deps.knex).delete()
+    const newManagementId = uuid()
+    token = await AccessToken.query(deps.knex).insertAndFetch({
+      value: uuid(),
```
njlie, line 108: I'm currently planning to generate access token values [like thus](https://github.com/interledger/rafiki/blob/8cdbc80efc4e823fa7137a6514d50edea39b8d6a/packages/auth/src/accessToken/service.ts#L109), it differentiates them from identifiers that should be used in URLs and is a bit closer to the preconception of a token (though that may just be me).

packages/auth/src/accessToken/service.ts
```diff
@@ -72,16 +90,41 @@ async function introspect(
...
+  let token = await AccessToken.query(deps.knex).findOne({ managementId })
+  if (token) {
+    await token.$query(deps.knex).delete()
+    const newManagementId = uuid()
+    token = await AccessToken.query(deps.knex).insertAndFetch({
+      value: crypto.randomBytes(8).toString('hex').toUpperCase(),
+      grantId: token.grantId,
+      expiresIn: token.expiresIn,
+      managementId: newManagementId
+    })
```
sabineschaller, line 113: I'd call that `newToken` to reduce confusion. Also, I think that should call a create function in the future. But I'm aware of the fact that we don't have that yet.

packages/auth/src/accessToken/service.ts
```diff
@@ -72,16 +90,41 @@ async function introspect(
...
+}
+
+async function rotate(
+  deps: ServiceDependencies,
+  managementId: string
+): Promise<Rotation> {
+  let token = await AccessToken.query(deps.knex).findOne({ managementId })
+  if (token) {
+    await token.$query(deps.knex).delete()
+    const newManagementId = uuid()
```
sabineschaller, line 107: Is there a reason you define that separately? Why not just call the `uuid` function in line 112?

packages/auth/src/accessToken/service.ts
```diff
@@ -72,16 +90,41 @@ async function introspect(
...
+    token = await AccessToken.query(deps.knex).insertAndFetch({
+      value: crypto.randomBytes(8).toString('hex').toUpperCase(),
+      grantId: token.grantId,
+      expiresIn: token.expiresIn,
+      managementId: newManagementId
+    })
+    const access = await Access.query(deps.knex).findOne({
+      grantId: token.grantId
+    })
```
sabineschaller, line 116: There could be multiple accesses attached to that grant / token. So please use `where`.

packages/auth/src/accessToken/routes.ts
```diff
@@ -79,10 +81,30 @@ async function revokeToken(
...
+): Promise<void> {
+  //TODO: verify accessToken with httpsig method
+  const result = await deps.accessTokenService.rotate(
+    ctx.params['managementId']
+  )
+  if (result.success) {
+    ctx.status = 200
+    ctx.body = {
+      access_token: {
+        access: [accessToBody(result.access)],
```
sabineschaller, line 100: I think you need to use a map function here, don't you? `result.access` is an array of accesses.

packages/auth/src/accessToken/service.ts
dclipp: Originally I was returning the value but since I don't have to now I can just call the function

packages/auth/src/accessToken/service.ts
dclipp: Replaced `newManagementId` with a direct function call

njlie (dismissed):
LGTM, just needs merge conflicts resolved

njlie (approved):

PR #377, opened 2022-06-27
fix: use node-modules linker, not pnp

rluckom-coil:

This PR switches from using the `pnp` linker to using the `node-modules` linker, because unlike the pnp linker, the node-modules linker allows `yarn docker build backend` to complete successfully (originally, the plan was to move the docker image build out of yarn. That turned out to require switching back to node modules, since yarn doesn't have any command that packages the dependencies along with the source code. Yarn 1 [apparently did this](https://classic.yarnpkg.com/en/docs/cli/pack); the docs for yarn 2 [remove mention of it](https://yarnpkg.com/cli/pack). So switching back to node_modules was necessary in order to get sufficient access to the packages to put them in a docker image manually. However, after switching back to the `node-modules` linker, the existing process started working again, so I didn't have to create a separate docker build script). I'd like to do some further testing to ensure that the image built using this linker will work.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


matdehaast:
We also moved to the `node-modules` linker internally. So this is probably a good change

matdehaast:
@rluckom-coil May be worth adding an issue to optimize the github actions now to cache the `node-modules`.

matdehaast (dismissed):

packages/openapi/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "openapi",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "files": [
+    "dist/**/*"
+  ],
```
rluckom-coil, line 7: This was strange. After I had set up everything else in this PR, I did `yarn docker build backend`. It completed successfully. I ran the container and it threw an error that the require of `./middleware` from `packages/openapi/index.js` could not be found. This seemed wrong; I looked in the `dist` folder in the repo, it was there. I removed the dist folder and did a fresh yarn build. All good. Then I did a `yarn pack`, and sure enough, the package included `dist/index.js` but no other files in the `dist` directory. According to [this conversation](https://github.com/yarnpkg/yarn/issues/2498), specifying a `files` block in the package.json can help yarn pack do the right thing, so I added this and now it works.

Note that the old `pack` would also include things in the root of the `packages/openapi` directory--stuff like the package json and other configs. Those do not appear to me to be needed, but if they are it should be possible to include them here and have them added to the package.

rluckom-coil (commented):
I believe that `yarn docker build backend` now successfully creates a working `backend` container image. When I try to run it locally, I see:

```
raphael@Raphaels-MacBook-Pro rafiki % docker run --privileged rafiki-backend 
{"level":30,"time":1656441573391,"pid":1,"hostname":"6c9eaf71ece9","msg":"creating knex"}
{"level":50,"time":1656441573418,"pid":1,"hostname":"6c9eaf71ece9","error":"connect ECONNREFUSED 127.0.0.1:5432","msg":"error migrating database"}
{"level":30,"time":1656441573420,"pid":1,"hostname":"6c9eaf71ece9","msg":"creating graphile worker utils"}
{"level":50,"time":1656441573424,"pid":1,"hostname":"6c9eaf71ece9","msg":"Error: connect ECONNREFUSED 127.0.0.1:5432\n    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1159:16)"}
```
which looks about right, given that I haven't supplied its dependencies.

I modified the caching strategy to cache the `node_modules` directories in a cache keyed by the hash of the yarn.lock file. This will hopefully make builds a little faster by making the cached packages available under a common key, but I'm honestly not seeing much change.

matdehaast (approved):

PR #387, opened 2022-06-29
feat(auth): httpsig request validation

### LINKED ISSUE #274: Signing/Request Library

njlie:

- Adds `httpsig` validation to requests where the client must present a public key.

## Context

Closes #274.

Bumped `jest` to 28 to fix: https://github.com/facebook/jest/issues/10221#issuecomment-1090157337, which was affecting one of the tests using `fake-timers`.
https://github.com/facebook/jest/releases/tag/v28.0.0
* `[jest-config]` [BREAKING] Rename config option name to id (https://github.com/facebook/jest/pull/11981)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/accessToken/routes.ts
```diff
@@ -32,19 +34,73 @@ async function introspectToken(
...
+      return
     } else {
-      return ctx.throw(404, 'token not found')
+      ctx.status = 401
+      ctx.body = {
+        error: 'invalid_client'
+      }
+      return
     }
+  }
```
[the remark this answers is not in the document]
njlie: Indeed, this might be better as a middleware we could place in front of the relevant routes. Will look into changing that.

packages/auth/src/accessToken/routes.ts
```diff
@@ -35,19 +37,28 @@ async function introspectToken(
...
-      body['access_token']
-    )
-    if (introspectionResult) {
-      ctx.body = introspectionToBody(introspectionResult)
-    } else {
-      return ctx.throw(404, 'token not found')
+  if (!body['access_token']) {
+    ctx.status = 400
+    ctx.body = {
+      error: 'invalid_request',
```
wilsonianb, line 44: Are these [error responses](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.6) only necessary for the client (not the RO)?

packages/auth/src/client/service.ts
```diff
@@ -166,14 +297,77 @@ function verifyClientDisplay(
...
+  // TODO: replace with HttpMethod types instead of string literals
+  if (path.includes('/introspect') && method === 'POST') {
+    keyName = 'value'
+    value = body['access_token']
+  } else if (path.includes('/token') && method === 'DELETE') {
+    keyName = 'managementId'
+    value = ctx.params['managementId']
+  } else {
+    // Is not a route that requires httpsig validation, somehow
+    next()
```
wilsonianb, line 349: Can we ditch `useHttpSigMiddleware` and use this middleware for all routes if it's making determinations of what requires httpsig validation?

packages/auth/src/client/service.ts
```diff
@@ -74,10 +97,66 @@ export async function createClientService({
...
+
+  // https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-09#section-2.3
+  let signatureBase = ''
+  for (const component of cleanMessageComponents) {
+    if (component === '@method') {
+      signatureBase += `"@method": ${ctx.request.method}\n`
+    } else if (component === '@target-uri') {
+      signatureBase += `"@target-uri": ${ctx.request.url}\n`
+    } else {
+      signatureBase += `"${component}": ${ctx.headers[component]}\n`
```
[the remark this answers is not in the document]
njlie: Yeah that makes sense, will add that/

packages/auth/src/accessToken/routes.ts
njlie: It's only for the client. Though looking at the spec, it does seem like these are only necessary during the grant authorization flow and not necessarily for token management endpoints.

wilsonianb: Yeah, I was thinking that since the RO would be sending token introspection requests that these error responses don't _have_ to be used.

packages/auth/src/client/service.ts
```diff
@@ -166,14 +302,77 @@ function verifyClientDisplay(
...
+    value = ''
+  const { path, method } = ctx
+  // TODO: replace with HttpMethod types instead of string literals
+  if (path.includes('/introspect') && method === 'POST') {
+    keyName = 'value'
+    value = body['access_token']
+  } else if (path.includes('/token') && method === 'DELETE') {
+    keyName = 'managementId'
+    value = ctx.params['managementId']
+  } else {
```
wilsonianb, line 352: Should continuation requests be included here?
https://github.com/interledger/rafiki/pull/387#discussion_r937075684

packages/auth/package.json
```diff
@@ -44,6 +44,6 @@
     "jest-openapi": "^0.14.2",
     "nock": "^13.2.4",
     "openapi-types": "^12.0.0",
-    "typescript": "^4.2.4"
+    "typescript": "^4.3.0"
```
wilsonianb, line 47: I'm still getting the warning and so is GitHub Actions
https://github.com/interledger/rafiki/pull/387#discussion_r937048994
https://github.com/interledger/rafiki/runs/7682195245?check_suite_focus=true

I wonder if typescript should be updated in the root `package.json`

packages/auth/src/client/service.ts
njlie: Oh yeah, they should. Thanks for pointing that out.

packages/auth/package.json
njlie: I'll try it and see what happens

njlie: Bumping typescript breaks a lot of stuff

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -113,40 +111,67 @@ describe('Access Token Routes', (): void => {
...
+        access_token: v4(),
+        proof: 'httpsig',
+        resource_server: 'test'
+      }
+      const { signature, sigInput, contentDigest } = await generateSigHeaders(
+        url,
+        method,
+        requestBody
+      )
       const ctx = createContext(
```
wilsonianb, line 124: Maybe you could add a wrapper for `createContext` that takes `url`, `method`, `requestBody`, etc as parameters and does `generateSigHeaders` + `createContext` instead of having to do this full setup in all these tests.

packages/auth/src/accessToken/routes.ts
sabineschaller: Isn't that covered by checking against the spec?

packages/auth/migrations/20220504163005_create_access_token_table.js
```diff
@@ -2,7 +2,7 @@ exports.up = function (knex) {
   return knex.schema.createTable('accessTokens', function (table) {
     table.uuid('id').notNullable().primary()
     table.string('value').notNullable().unique()
-    table.string('managementId').notNullable()
+    table.string('managementId').notNullable().unique()
```
sabineschaller, line 5: Why do we introduce that? The `id` is already unique, so why adding another unique id?

packages/auth/src/app.ts
```diff
@@ -219,13 +221,6 @@ export class App {
     // Token management
     this.publicRouter.post('/auth/introspect', accessTokenRoutes.introspect)
 
-    this.publicRouter.post('/auth/token/:id', (ctx: AppContext): void => {
-      // TODO: tokenService.rotate
-      ctx.status = 200
-    })
-
-    this.publicRouter.del('/auth/token/:id', accessTokenRoutes.revoke)
```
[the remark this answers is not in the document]
sabineschaller: I guess because we have this section further up that also includes checking against the openapi spec (L168 - L1206). In the end, all of those individual routes should be moved into the block that also checks the openapi spec.

packages/auth/src/client/service.ts
```diff
@@ -88,7 +150,44 @@ async function verifySig(
...
+): Promise<VerifySigFromBoundKeyResult> {
+  const registryData = await getRegistryDataByKid(deps, grant.clientKeyId)
+  if (!registryData)
+    return {
+      success: false,
+      error: 'invalid_client',
+      status: 401
+    }
+  const { keys } = registryData
+  const clientKey = keys[0]
```
sabineschaller, line 171: Why is that an array? Should there be just one key per key id?

packages/auth/src/app.ts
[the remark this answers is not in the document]
njlie: It is redundant to [this block of code now](https://github.com/interledger/rafiki/blob/16deb5bc3605e83594ac7673c8052a7b2a453a93/packages/auth/src/app.ts#L193-L242). I originally stubbed it out when scaffolding the auth server but it didn't get removed once OpenApi was added.

packages/auth/jest.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
   testRegex: `(packages/${packageName}/.*/__tests__/.*|\\.(test|spec))\\.tsx?$`,
   moduleDirectories: [`node_modules`, `packages/${packageName}/node_modules`],
   modulePaths: [`<rootDir>/packages/${packageName}/src/`],
-  name: packageName,
+  id: packageName,
```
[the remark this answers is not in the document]
njlie: Yes, it's part of the breaking jest change. https://github.com/facebook/jest/pull/11981

packages/auth/migrations/20220504163005_create_access_token_table.js
njlie: I figured it would be the right practice to expose an external id instead of an internal database id. I should update it to a `.uuid()` instead of a `.string()` though since in practice it's always a uuid/v4.

packages/auth/src/client/service.ts
njlie: Looking back, I think this was the result of a misinterpretation of Adrian's original doc. Which is probably pretty moot now that the keys aren't being provided by an external registry anymore.

njlie: After discussing with Adrian on a new `kid` response, we settled on the shape described [here](https://github.com/interledger/rafiki/issues/424). I'll move that work to outside this PR so that it doesn't grow any larger than it already has, but the end result will be that we won't need to handle an array anymore.

packages/auth/src/accessToken/routes.test.ts
njlie: I think we can actually just remove it since the test doesn't run the middleware.

packages/auth/src/accessToken/service.test.ts
```diff
@@ -199,9 +201,6 @@ describe('Access Token Service', (): void => {
       token = await AccessToken.query(trx).findById(token.id)
       expect(token).toBeUndefined()
     })
-    test('Cannot revoke nonexistent token', async (): Promise<void> => {
-      expect(accessTokenService.revoke('uuid')).rejects.toBeInstanceOf(Error)
-    })
```
wilsonianb, line 204: Why was this removed?

packages/auth/src/client/service.test.ts
```diff
@@ -43,6 +61,400 @@ describe('Client Service', (): void => {
...
+            'Signature-Input': sigInputHeader,
+            Authorization: 'GNAP test-access-token'
+          }
+        },
+        {}
+      )
+
+      ctx.request.body = { foo: 'bar' }
+      ctx.method = 'GET'
+      ctx.request.url = '/test'
```
wilsonianb, line 96: Maybe add a function that takes `sigInputHeader` and optional `authorization` header that handles all this setup for all these tests.
Or combine all these into a single `test.each`

packages/auth/src/grant/routes.test.ts
```diff
@@ -199,14 +204,23 @@ describe('Grant Routes', (): void => {
...
+            Signature: signature,
+            'Signature-Input': sigInput
           },
-          url: '/',
-          method: 'POST'
+          url,
+          method
         },
         {}
       )
```
wilsonianb, line 226: https://github.com/interledger/rafiki/pull/387#discussion_r946061677

packages/auth/src/accessToken/service.test.ts
njlie: The revoke function originally called `findById` instead of `findOne` using like the `managementId` like it should be. When it was using `findById` it would error if it couldn't find it, but it doesn't if `findOne` fails to find it. I decided to not re-implement this behavior because of the following:

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6.2
```
If the access token has already expired or has been revoked through other means, the AS SHOULD honor the revocation request to the token management URI as valid, since the end result is still the token not being usable.

File: packages/auth/src/client/service.test.ts [hunk 15]
```diff
@@ -43,6 +60,335 @@ describe('Client Service', (): void => {
...
+      const challenge = 'test-challenge'
+      const privateJwk = (await importJWK(TEST_PRIVATE_KEY)) as crypto.KeyLike
+      const signature = crypto.sign(null, Buffer.from(challenge), privateJwk)
+      const verified = await clientService.verifySig(
+        signature.toString('base64'),
+        TEST_PUBLIC_KEY,
+        challenge
+      )
+
+      expect(verified).toBe(true)
```
wilsonianb, line 74:
```suggestion
      await expect(clientService.verifySig(
        signature.toString('base64'),
        TEST_PUBLIC_KEY,
        challenge
      )).resolves.toBe(true)
```

packages/auth/migrations/20220504163005_create_access_token_table.js
wilsonianb: >I should update it to a .uuid() instead of a .string() though since in practice it's always a uuid/v4.

bump

packages/auth/src/client/service.ts
```diff
@@ -166,14 +276,126 @@ function verifyClientDisplay(
...
+  next: () => Promise<any>
+): Promise<void> {
+  const sig = ctx.headers['signature']
+  const sigInput = ctx.headers['signature-input']
+
+  if (
+    !sig ||
+    !sigInput ||
+    typeof sig !== 'string' ||
+    typeof sigInput !== 'string'
```
wilsonianb, line 304: I'm wondering if the openapi schema can enforce this?

njlie: I tried looking to see if it can validate request headers but I wasn't able to find a way to describe that in the spec from the OpenAPI docs.

packages/auth/src/client/service.ts
njlie: Added now in: https://github.com/interledger/rafiki/blob/nl-httpsig/packages/auth/src/client/service.ts#L357-L370

wilsonianb:
I just came across this TODO to use `httpsig` in the backends token introspection request to the AS.
https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/service.ts#L47-L68
Should that be added in this PR or should we open an issue to do it separately?

wilsonianb:
> I just came across this TODO to use `httpsig` in the backend's token introspection request to the AS. https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/service.ts#L47-L68 Should that be added in this PR or should we open an issue to do it separately?

https://github.com/interledger/rafiki/pull/442 adds `auth` as a dependency of `backend` which could be useful for allowing to backend to use `httpsig` in requests.

packages/auth/src/tests/signature.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export async function generateSigHeaders(
+  url: string,
+  method: string,
+  body?: unknown
+): Promise<{ sigInput: string; signature: string; contentDigest?: string }> {
+  const privateKey = {
+    ...TEST_CLIENT_KEY.jwk,
+    d: 'v6gr9N9Nf3AUyuTgU5pk7gyNULQnzNJCBNMPp5OkiqA'
```
sabineschaller, line 28: Why is this hard-coded?

packages/auth/package.json
```diff
@@ -37,9 +37,12 @@
     "@types/koa__cors": "^3.1.1",
     "@types/koa__router": "^8.0.11",
     "@types/uuid": "^8.3.4",
+    "@types/nock": "^11.1.0",
+    "@types/pino": "^7.0.5",
+    "@types/sinonjs__fake-timers": "^8.1.2",
```
matdehaast, line 42: @njlie do we need these definitions with the newer jest?

packages/auth/package.json
matdehaast: That doesn't make any sense because `^4.2.4` actually installs `4.7.2` currently as the caret means any non breaking changes.

packages/auth/src/accessToken/routes.ts
```diff
@@ -87,7 +87,7 @@ async function rotateToken(
       access_token: {
         access: result.access.map((a) => accessToBody(a)),
         value: result.value,
-        manage: result.managementId,
+        manage: deps.config.authServerDomain + `/token${result.managementId}`,
```
matdehaast, line 90: The managementId is a uuid so surely you need a `/` after the token here?

packages/auth/src/client/service.ts
```diff
@@ -166,14 +276,126 @@ function verifyClientDisplay(
...
+    const grant = await Grant.query().findOne(
+      'interactId',
+      ctx.params['interactId']
+    )
+    if (!grant) {
+      ctx.status = 401
+      ctx.body = {
+        error: 'invalid_interaction',
+        message: 'invalid grant'
+      }
```
matdehaast, line 367: @njlie we should not be doing any direct model querying outside of services. These should be access methods on the service and then be using the service method for them. It makes maintaining it far easier.

packages/auth/src/client/service.ts
```diff
@@ -166,14 +276,126 @@ function verifyClientDisplay(
...
+    (!keys.nbf || new Date() >= new Date(keys.nbf * 1000))
   )
 }
+
+function verifyJwk(jwk: JWKWithRequired, keys: RegistryKey): boolean {
+  // TODO: update this to reflect eventual shape of response from registry
+  return !!(!keys.revoked && isJwkViable(keys) && keys.x === jwk.x)
+}
+
+async function tokenHttpsigMiddleware(
```
matdehaast, line 291: Musing here.... This middleware seems to do many thing that are path dependent.. Does it make sense for it to be a singular middleware?

packages/auth/src/client/service.ts
njlie: Is this method disqualified from direct model querying because it's designated as a middleware, or because it crosses the boundary between models and their corresponding services (i.e. it's a `client` service calling into a `Grant` model)?

matdehaast: The later, the models should really be reserved for usage within the service or tests when really required.

packages/auth/package.json
njlie: I think we can remove these.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -83,14 +72,18 @@ describe('Access Token Routes', (): void => {
 
   const BASE_TOKEN = {
     value: crypto.randomBytes(8).toString('hex').toUpperCase(),
-    managementId: 'https://example.com/manage/12345',
+    managementId: v4(),
```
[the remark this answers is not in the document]
njlie: This is supposed to represent a row in the `accessTokens` table, I initially (i.e. the early stages of committing my GNAP notes to code) thought to put an entire URL in as a field on the access token but quickly realized that did not make sense. Now it just holds a path parameter that is inserted into the `manage` URL [returned by grant continuation](https://docs.openpayments.guide/reference/post-continue).

packages/auth/src/tests/signature.ts
njlie: Originally was for bootstrapping the tests faster. Will change.

packages/auth/src/accessToken/routes.ts
njlie: Yep that's a typo, thanks for catching that

matdehaast (approved):

PR #388, opened 2022-07-09
feat(backend): periodically clean up expired rows

### LINKED ISSUE #385: Clean up expired tokens after reasonable time
We need to figure out how long expired tokens should be kept until they can be deleted from the DB.

dclipp:

- add logic to `auth` to periodically delete database rows that we consider "expired"
- add rule for cleaning up `accessTokens` table

## Context
 fixes #385 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -287,7 +287,7 @@ export class App {
               [
                 rule.absoluteStartTimeColumnName,
                 rule.expirationOffsetColumnName,
-                `${rule.minDaysLapsedBeforeDeletion} days`
+                `${rule.defaultExpirationOffsetDays} days`
               ]
             )
             .del()
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -2,7 +2,7 @@ import { Server } from 'http'
 import { EventEmitter } from 'events'
 
 import { IocContract } from '@adonisjs/fold'
-import Knex, { QueryInterface } from 'knex'
+import Knex from 'knex'
 import Koa, { DefaultState, DefaultContext } from 'koa'
 import bodyParser from 'koa-bodyparser'
 import session from 'koa-session'
@@ -34,19 +34,15 @@ export interface DatabaseCleanupRule {
    */
   absoluteStartTimeColumnName: string
   /**
-   * the column which will be used to either set or offset the computed age
-   * if not provided, rows are considered expired when the difference between the current time
-   * and the time specified in `absoluteStartTimeColumnName` is greater than or equal to `minLapseTimeMillis`
+   * the name of the column containing the time offset, in seconds, since
+   * `absoluteStartTimeColumnName`, which specifies when the row expires
    */
-  lapseTime?: {
-    columnName: string
-    absolute?: boolean
-  }
+  expirationOffsetColumnName: string
   /**
-   * the minimum number of milliseconds since expiration before rows of this table will be
-   * considered safe to delete during clean up
+   * the minimum number of days since expiration before rows of
+   * this table will be considered safe to delete during clean up
    */
-  minLapseTimeMillis: number
+  minDaysLapsedBeforeDeletion: number
 }
 
 type ContextType<T> = T extends (
@@ -101,11 +97,9 @@ export class App {
     this.databaseCleanupRules = {
       accessTokens: {
         absoluteStartTimeColumnName: 'createdAt',
-        lapseTime: {
-          columnName: 'expiresIn',
-          absolute: false
-        },
-        minLapseTimeMillis: this.config.accessTokenCleanupMinAge
+        expirationOffsetColumnName: 'expiresIn',
+        minDaysLapsedBeforeDeletion: this.config
+          .accessTokenMinDaysBeforeDeletion
       }
     }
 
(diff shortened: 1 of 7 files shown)


packages/backend/src/app.ts
```diff
@@ -462,4 +500,52 @@ export class App {
...
+    let absoluteLapseTime = row.column[rule.absoluteStartTimeColumnName]
+
+    if (rule.lapseTime) {
+      const timeParameter =
+        row.column[rule.lapseTime.columnName] || rule.minLapseTimeMillis
+      if (rule.lapseTime.absolute) {
+        absoluteLapseTime = timeParameter
+      } else {
+        // offset the working base time by a time span stored in another column
+        absoluteLapseTime += timeParameter
```
sabineschaller, line 519: Interesting that this works given that one is a timestamp and one is an integer.

packages/backend/src/app.ts
```diff
@@ -91,6 +91,28 @@ export type ListContext = Context<
   AppRequest<'accountId', never, PageQueryParams>
 >
 
+export interface DatabaseCleanupRule {
```
sabineschaller, line 94: I think you added this to the wrong package. This is `backend` and not `auth`.

njlie: ☝️

packages/backend/src/app.ts
```diff
@@ -160,6 +185,16 @@ export class App {
     this.koa.context.closeEmitter = await this.container.use('closeEmitter')
     this.messageProducer = await this.container.use('messageProducer')
     this.publicRouter = new Router()
+    this.databaseCleanupRules = {
+      accessTokens: {
+        absoluteStartTimeColumnName: 'updatedAt',
```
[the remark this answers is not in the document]
dclipp: changed to `createdAt`

packages/backend/src/app.ts
dclipp: added `.valueOf()` just to be safe

packages/auth/src/config/app.ts
```diff
@@ -43,5 +43,7 @@ export const Config = {
   authServerDomain: envString('AUTH_SERVER_DOMAIN', 'http://localhost:3006'), // TODO: replace this with whatever frontend port ends up being
   waitTimeSeconds: envInt('WAIT_SECONDS', 5),
   cookieKey: envString('COOKIE_KEY', crypto.randomBytes(32).toString('hex')),
-  accessTokenExpirySeconds: envInt('ACCESS_TOKEN_EXPIRY_SECONDS', 10 * 60) // Default 10 minutes
+  accessTokenExpirySeconds: envInt('ACCESS_TOKEN_EXPIRY_SECONDS', 10 * 60), // Default 10 minutes
+  databaseCleanupWorkers: envInt('DATABASE_CLEANUP_WORKERS', 1),
+  accessTokenMinDaysBeforeDeletion: 30
```
wilsonianb, line 48: Should this be configurable?

packages/auth/src/app.ts
```diff
@@ -227,4 +262,36 @@ export class App {
 
     this.koa.use(this.publicRouter.middleware())
   }
+
+  private async processDatabaseCleanup(): Promise<void> {
+    const knex = await this.container.use('knex')
+
+    const tableNames = Object.keys(this.databaseCleanupRules)
+    for await (const tableName of tableNames) {
```
wilsonianb, line 270:
```suggestion
    for (const tableName of tableNames) {
```

packages/auth/src/app.ts
```diff
@@ -227,4 +262,36 @@ export class App {
...
+           * NOTE: do not remove seemingly pointless interpolations such as '${'??'}'
+           * because they are necessary for preventing SQL injection attacks
+           */
+          await knex(tableName)
+            .whereRaw(
+              `?? + COALESCE(make_interval(0, 0, 0, 0, 0, 0, ??), interval '${'??'}') < NOW()`,
+              [
+                rule.absoluteStartTimeColumnName,
+                rule.expirationOffsetColumnName,
+                `${rule.minDaysLapsedBeforeDeletion} days`
```
wilsonianb, line 284:
```suggestion
                `${rule.defaultExpirationOffsetDays} days`
```
`minDaysLapsedBeforeDeletion` makes me think deletion would happen after max(`expirationOffsetColumnName`, `minDaysLapsedBeforeDeletion`)

packages/auth/src/config/app.ts
dclipp: I thought 10 minutes was reasonable but it's not actually part of the spec. I can just hardcode it if you think that's cleaner, though.

wilsonianb: I'm not sure this default is actually necessary.
We currently only have `databaseCleanupRules` for `accessTokens`, for which `expiresIn` is [nullable](https://github.com/interledger/rafiki/blob/7e32e45bf9259837c95abb34fd110ac45712f2f3/packages/auth/migrations/20220504163005_create_access_token_table.js#L6) but we appear to always [define it](https://github.com/interledger/rafiki/blob/7e32e45bf9259837c95abb34fd110ac45712f2f3/packages/auth/src/accessToken/service.ts#L121).

I vote we make `expiresIn` non-nullable and remove `accessTokenMinDaysBeforeDeletion`

njlie: I second the above ☝️

dclipp: I could remove `accessTokenMinDaysBeforeDeletion` and make `expiresIn` non-nullable, but it seems like the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6.1) recommends against removing expired tokens immediately:

`Note that in many cases, the access token will have expired for
   regular use.  To facilitate token rotation, the AS SHOULD honor the
   rotation request of the expired access token since it is likely that
   the client instance is attempting to refresh the expired token.`

That makes me think that we'd still need to define a constant somewhere to prevent the auth service from cleaning up tokens before the client has a chance to request renewal of an expired token.

Maybe it doesn't need to be configurable, though. Let me know what you think

wilsonianb: Got it. Good point.
How about a configurable `accessTokenDeletionDays` that `processDatabaseCleanup` uses?
An access token would be deleted if (`createdAt` + `expiresIn` +`accessTokenDeletionDays`) < `now`.

Should the old access token always be deleted (expired or otherwise) by token rotation?

dclipp: Yeah, right now the old token is deleted during rotation (see #335 / #375)

packages/auth/src/config/app.ts
```diff
@@ -43,5 +43,7 @@ export const Config = {
   authServerDomain: envString('AUTH_SERVER_DOMAIN', 'http://localhost:3006'), // TODO: replace this with whatever frontend port ends up being
   waitTimeSeconds: envInt('WAIT_SECONDS', 5),
   cookieKey: envString('COOKIE_KEY', crypto.randomBytes(32).toString('hex')),
-  accessTokenExpirySeconds: envInt('ACCESS_TOKEN_EXPIRY_SECONDS', 10 * 60) // Default 10 minutes
+  accessTokenExpirySeconds: envInt('ACCESS_TOKEN_EXPIRY_SECONDS', 10 * 60), // Default 10 minutes
+  databaseCleanupWorkers: envInt('DATABASE_CLEANUP_WORKERS', 1),
+  accessTokenDeletionDays: 30
```
wilsonianb, line 48:
```suggestion
  accessTokenDeletionDays: envInt('ACCESS_TOKEN_DELETION_DAYS', 30)
```

njlie (approved):

PR #403, opened 2022-08-02
Rl tigerbeetle helm chart

rluckom-coil:

Adds a chart for deploying tigerbeetle to kubernetes via helm. 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


rluckom-coil (commented):
This version of the chart can deploy from scratch and come up healthy.

The port-forwarding via nginx is still brittle in a couple of ways. First, I think nginx caches the IP of the pods after resolving them initially. The [docs](https://nginx.org/en/docs/stream/ngx_stream_upstream_module.html#resolver) state that nginx uses the TTL of the response from the DNS server, which [should be 5s by default](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#coredns-configmap-options). The nginx settings to configure a ttl override are marked as "available as part of the commercial subscription".

Second, since nginx is configured with backends for all of the other replicas, if a replica goes down and stays down, no amount of waiting (for DNS resolution or anything else) will allow any of the other pods to restart, because any time they try nginx will fail with a name resolution error regarding the unavailable replica.

rluckom-coil (commented):
I think this is ready for review. I decided not to include the storageclass in this chart. The storageclass is a tiny resource that is tightly bound to the cloud provider one uses for k8s, so including it in this chart would tend to tie this chart to GCP for no great benefit. It can be set as a parameter so it is trivial to bring one's own storage clas. That is how the cockroachdb chart works and it seems like the right decision.

infrastructure/helm/tigerbeetle/templates/configmap.nginx.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-nginx-proxy
+data:
+  # property-like keys; each key maps to a simple value
+  nginx-prestart.sh: |
+    #!/bin/sh
+    sleep 600
```
rluckom-coil, line 9: This sleep is long. It is designed to be long enough to cover the entire period of rolling all of the replicas, because otherwise hostname resolution can fail after an nginx is up, or while it is coming up, causing a kind of round-robin failure that can bring down the whole cluster.

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
```diff
@@ -0,0 +1,229 @@
...
+              FMT_REPLICA=$(echo $REPLICA | awk '{ printf("%03d\n", $1) }')
+              ls /var/lib/tigerbeetle
+              DATA_FILE=/var/lib/tigerbeetle/cluster_${FMT_CLUSTER}_replica_${FMT_REPLICA}.tigerbeetle
+              if [ ! -f "$DATA_FILE" ]; then 
+                /opt/beta-beetle/tigerbeetle init --cluster=$CLUSTER --replica=$REPLICA --directory=/var/lib/tigerbeetle; 
+              fi
+          volumeMounts:
+            - name: data
+              mountPath: /var/lib/tigerbeetle
+    {{- if or .Values.statefulset.nodeAffinity .Values.statefulset.podAffinity .Values.statefulset.podAntiAffinity }}
```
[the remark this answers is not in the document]
rluckom-coil: might be better to switch this back actually to maintain cloud-provider-agnosticity.

infrastructure/helm/tigerbeetle/templates/configmap.nginx.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-nginx-proxy
+data:
+  # property-like keys; each key maps to a simple value
+  nginx-prestart.sh: |
```
rluckom-coil, line 7: Dynamically build the nginx config because the DNS server address is only known at runtime. and it is required for the `resolve` directive for the backends.

rluckom-coil:
Now that this uses the `resolver` directive, nginx _will_ be able to run even if some of the replicas are not assigned hostnames, making this a more robust solution than I previously thought.

infrastructure/helm/tigerbeetle/templates/configmap.nginx.yaml
rluckom-coil: No longer needs a sleep; resolver directive instead.

infrastructure/helm/tigerbeetle/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: '1.16.0'
```
matdehaast, line 24: nit: does this reflect tb current version? I'm not even sure they have one yet. But we should try keep it consistent

infrastructure/helm/tigerbeetle/templates/configmap.nginx.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-nginx-proxy
```
matdehaast, line 4: Missing namespace and labels

infrastructure/helm/tigerbeetle/templates/service.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "tigerbeetle.fullname" . }}
+  labels: {{- include "tigerbeetle.labels" . | nindent 4 }}
```
matdehaast, line 5: Missing namespace

infrastructure/helm/tigerbeetle/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "tigerbeetle.serviceAccountName" . }}
+  labels:
```
matdehaast, line 6: namespace is missing

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
```diff
@@ -0,0 +1,224 @@
...
+              FMT_CLUSTER=$(echo $CLUSTER | awk '{ printf("%010d\n", $1) }')
+              FMT_REPLICA=$(echo $REPLICA | awk '{ printf("%03d\n", $1) }')
+              ls /var/lib/tigerbeetle
+              DATA_FILE=/var/lib/tigerbeetle/cluster_${FMT_CLUSTER}_replica_${FMT_REPLICA}.tigerbeetle
+              if [ ! -f "$DATA_FILE" ]; then 
+                /opt/beta-beetle/tigerbeetle init --cluster=$CLUSTER --replica=$REPLICA --directory=/var/lib/tigerbeetle; 
+              fi
+          env:
+            - name: CLUSTER
+              value: "0"
```
matdehaast, line 52: This should be configurable from the values file

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
```diff
@@ -0,0 +1,224 @@
...
+                  ADDRESSES="$ADDRESSES,127.0.0.1:300$n"
+                else
+                  ADDRESSES="$ADDRESSES,0.0.0.0:4242"
+                fi
+              done
+              ADDRESSES=${ADDRESSES#?}
+              /opt/beta-beetle/tigerbeetle start --cluster=$CLUSTER --replica=$REPLICA --directory=/var/lib/tigerbeetle --addresses=$ADDRESSES
+          env:
+            - name: CLUSTER
+              value: "0"
```
matdehaast, line 148: configurable from values file

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
```diff
@@ -0,0 +1,224 @@
...
+          image: "{{ .Values.statefulset.image.repository }}:{{ .Values.statefulset.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.statefulset.image.pullPolicy }}
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -ex; 
+              REPLICA=${HOSTNAME##*-}
+              ADDRESSES=""
+              for n in $(seq 0 5)
```
matdehaast, line 135: This loop should be based on the number of replicas that are configured from the values file

matdehaast (commented):
@rluckom-coil just given it a high-level pass through so far. Will try deploy and give more detailed feedback

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
rluckom-coil: thanks, fixed

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
rluckom-coil: fixed

infrastructure/helm/tigerbeetle/templates/statefulset.yaml
rluckom-coil: fixed

infrastructure/helm/tigerbeetle/templates/serviceaccount.yaml
rluckom-coil: fixed

infrastructure/helm/tigerbeetle/templates/service.yaml
rluckom-coil: fixed

infrastructure/helm/tigerbeetle/templates/configmap.nginx.yaml
rluckom-coil: thanks fixed.

infrastructure/helm/tigerbeetle/Chart.yaml
rluckom-coil: Thanks, changed to 0.0.0 until I find an official version

infrastructure/helm/tigerbeetle/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
```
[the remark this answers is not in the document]
DonChangfoot: @rluckom-coil do you think it's worth adding a role and role binding? I don't think tigerbeetle needs to access any of the k8s resources yet.

infrastructure/helm/tigerbeetle/values.yaml
```diff
@@ -0,0 +1,169 @@
...
+    # capabilities:
+    #   drop:
+    #   - ALL
+    # readOnlyRootFilesystem: true
+    # runAsNonRoot: true
+    # fsGroup: 2000
+    # runAsUser: 1000
+
+  # List of additional command-line arguments you want to pass to the
+  # `cockroach start` command.
```
DonChangfoot, line 54: nit: remove cockroach reference

infrastructure/helm/tigerbeetle/values.yaml
```diff
@@ -0,0 +1,169 @@
...
+  args: []
+    # - --disable-cluster-name-verification
+
+  # List of extra environment variables to pass into container
+  env: []
+    # - name: COCKROACH_ENGINE_MAX_SYNC_DURATION
+    #   value: "24h"
```
DonChangfoot, line 61: nit: remove cockroach reference

infrastructure/helm/tigerbeetle/templates/serviceaccount.yaml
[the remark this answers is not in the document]
rluckom-coil: It's a good question. Overall I think it might be best not to. I had initially added a storage class to this PR to set up the PVCs in the way that seemed best. But then I remembered that in our cluster we pretty much have a single "database-y" storage class that it's nice to be able to apply to whatever DBs we end up deploying. And I think it might be that way for a role for this serviceaccount as well--in the cases where it is needed (and I agree with you that I'm not sure what those are yet) I think they might turn out to be environment-specific enough that people deploying this chart would rather supply them themselves.

For another perspective, the cockroach chart's [role gives permission to view and create secrets](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/templates/role.yaml) that cockroach uses for client certificates. I think that once tigerbeetle has an auth mechanism (or secure communication between nodes) something like that might become appropriate.

But if you'd rather have them added now for expansion later I can definitely do that.

infrastructure/helm/tigerbeetle/values.yaml
rluckom-coil: done

infrastructure/helm/tigerbeetle/values.yaml
rluckom-coil: done

matdehaast (dismissed):

infrastructure/helm/tigerbeetle/templates/serviceaccount.yaml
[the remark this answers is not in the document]
DonChangfoot: Agreed, let's get this in and take a look at that later.

DonChangfoot (approved):

PR #407, opened 2022-08-06
feat(auth): grant request update api

### LINKED ISSUE #406: Grant patch-continue API
Implement `PATCH /auth/continue/{id}` endpoint

see:
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-5.3
https://docs.openpayments.guide/reference/patch-continue

dclipp:

- Implements grant request update API via HTTP `PATCH`

## Context

fixes #406 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/grant/routes.ts
+++ b/packages/auth/src/grant/routes.ts
@@ -304,13 +304,12 @@ async function continueGrant(
   }
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// eslint-disable @typescript-eslint/no-explicit-any
 function isAccessTokenUpdateParameter(
   param: any
 ): param is {
   access: AccessRequest[]
 } {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return (
     param &&
     param.access &&
@@ -319,6 +318,7 @@ function isAccessTokenUpdateParameter(
     param.access.every((a: any) => isAccessRequest(a))
   )
 }
+// eslint-enable
 
 async function updateGrant(
   deps: ServiceDependencies,
--- a/packages/auth/src/grant/routes.ts
+++ b/packages/auth/src/grant/routes.ts
@@ -1,8 +1,9 @@
 import * as crypto from 'crypto'
 import { URL } from 'url'
+import { TransactionOrKnex } from 'objection'
 import { AppContext } from '../app'
 import { GrantService, GrantRequest } from './service'
-import { GrantState } from './model'
+import { Grant, GrantState } from './model'
 import { Access } from '../access/model'
 import { ClientService } from '../client/service'
 import { BaseService } from '../shared/baseService'
@@ -11,6 +12,7 @@ import { IAppConfig } from '../config/app'
 import { AccessTokenService } from '../accessToken/service'
 import { AccessService } from '../access/service'
 import { accessToBody } from '../shared/utils'
+import { AccessToken } from '../accessToken/model'
 
 interface ServiceDependencies extends BaseService {
   grantService: GrantService
@@ -18,6 +20,7 @@ interface ServiceDependencies extends BaseService {
   accessTokenService: AccessTokenService
   accessService: AccessService
   config: IAppConfig
+  knex: TransactionOrKnex
 }
 
 export interface GrantRoutes {
@@ -37,7 +40,8 @@ export function createGrantRoutes({
   accessTokenService,
   accessService,
   logger,
-  config
+  config,
+  knex
 }: ServiceDependencies): GrantRoutes {
   const log = logger.child({
     service: 'GrantRoutes'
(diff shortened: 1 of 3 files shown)


packages/auth/src/grant/routes.ts
```diff
@@ -253,3 +255,76 @@ async function continueGrant(
...
+    const { config, accessTokenService, grantService, accessService } = deps
+    const grant = await grantService.getByContinue(
+      continueId,
+      continueToken,
+      interactRef
+    )
+
+    if (grant) {
+      await grantService.update(grant.id, accessToken.access)
+      const updatedAccessToken = await accessTokenService.create(grant.id)
```
njlie, line 309: I think we should only be returning an `access_token` if one was already issued, rather than creating one since modifying the grant request shouldn't be a way to authorize it

packages/auth/src/grant/routes.ts
```diff
@@ -253,3 +255,76 @@ async function continueGrant(
...
+      const updatedAccess = await accessService.getByGrant(grant.id)
+
+      // TODO: add "continue" to response if additional grant request steps are added
+      ctx.body = {
+        access_token: {
```
njlie, line 314: We should add the `continue` field here if the level of access has changed, since we probably want the RO to re-authorize the new scope of access. [It looks like the continue token gets updated as well, between the example response to the initial POST and the example PATCH response.](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-5.3)

packages/auth/src/grant/routes.ts
njlie: `access_token` doesn't seem required, and it seems like it should be issued if the grant was modified in a way where the previous access already overlaps with the new one (like if the amount of access was reduced in the modification). Otherwise my personal thought is that if _more_ access was added in the grant update, we should expect the RO to authorize it down the line and return the grant to `pending`.

```
When the AS receives a valid modification request, the AS MUST place
   the grant request into the _processing_ state and re-evaluate the
   authorization in the new context created by the update request.

The client instance MAY include the interact field as described in
   [Section 2.5](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-2.5).  Inclusion of this field indicates that the client
   instance is capable of driving interaction with the end user, and
   this field replaces any values from a previous request.  The AS MAY
   respond to any of the interaction responses as described in
   [Section 3.3](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.3), just like it would to a new request.
```

packages/auth/src/grant/routes.ts
```diff
@@ -253,3 +255,76 @@ async function continueGrant(
...
+  if (
+    !continueId ||
+    !continueToken ||
+    !interactRef ||
+    !isAccessTokenUpdateParameter(accessToken)
+  ) {
+    ctx.status = 401
+    ctx.body = {
+      error: 'invalid_request'
+    }
```
wilsonianb, line 298: Should any of this be specified in the [OpenAPI spec](https://github.com/interledger/open-payments/blob/main/auth-server-open-api-spec.yaml#L200) and so it can be validated using the OpenAPI validation middleware?

packages/auth/src/grant/service.ts
```diff
@@ -156,3 +160,23 @@ async function getByContinue(
...
+  updatedAccessRequests: AccessRequest[]
+): Promise<void> {
+  const { accessService, knex } = deps
+  const trx = await Access.startTransaction(knex)
+  try {
+    const existingAccess = await accessService.getByGrant(grantId)
+    await Access.query(trx)
+      .findByIds(existingAccess.map((a) => a.id))
+      .del()
+    await accessService.createAccess(grantId, updatedAccessRequests, trx)
```
wilsonianb, line 176: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-5.3
```
Fields that aren't included in the request are considered
   unchanged from the original request.
```
I'm not sure that's happening here, but I'm also unsure how a client is supposed to clear optional fields.

dclipp: Here we remove all `Access` objects in `existingAccess` and after this call we "add" the updated Accesses. Since Accesses without any change would be `undefined`, they wouldn't be mapped into `existingAccess` and so the current Access would be retained.

I'll move all of the logic into this call to make it clearer

packages/auth/src/grant/routes.ts
dclipp: Makes sense. I copied the `// TODO: add "continue" to response if additional grant request steps are added` part from another method. Does that still apply here in addition to "if the level of access has changed" or should continue be added _only_ if the level of access has changed?

njlie: I think we should add continue either if there has been an access level change, or if the grant hasn't been issued yet. So it would be excluded if there's no access change and the grant has already been issued. In that second scenario you could include the `access_token`, I think.

njlie:
Closing since it is now no longer in spec for our implementation.

PR #428, opened 2022-08-16
add arm64 docker build

rluckom-coil:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


.github/workflows/docker.yml
```diff
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-    - main
+    - '*'
```
rluckom-coil, line 4: This now runs on all PRs because the GHCR login and docker push steps are conditionally done only on main. This allows the docker build to gate merges, which is appropriate.

packages/backend/Dockerfile
```diff
@@ -4,10 +4,12 @@ WORKDIR /workspace
 
 RUN apt update
 RUN apt install -y curl xz-utils
+COPY package.json yarn.lock tsconfig.build.json .yarnrc.yml ./
```
rluckom-coil, line 7: These changes remove reliance on the [yarn-docker-build](https://github.com/Dcard/yarn-plugins/tree/master/packages/docker-build/src) plugin, which made it impossible to build with buildx for different architectures.

.github/workflows/docker.yml
```diff
@@ -11,30 +11,32 @@ jobs:
   build-and-push:
     runs-on: ubuntu-latest
     steps:
```
rluckom-coil, line 13: The new version of this is cribbed from the [tigerbeetle build step](https://github.com/coilhq/tigerbeetle/blob/main/.github/workflows/docker.yml#L24), which already builds for multiple architectures.

DonChangfoot (approved):

PR #432, opened 2022-08-17
fix(backend): create tigerbeetle account for incoming payment fulfill

### LINKED ISSUE #181: clean up accounts when empty incoming payment is deleted
Per https://github.com/interledger/rafiki/issues/123, the `Invoice` is deleted when it expires without receiving any money. Right now it doesn't clean up the accounts (either the `Account` model or in TigerBeetle).

None of the following approaches prevent someone from just creating spam invoices continuously.

Ideas
- Lazily create the `Account` when a payment arrives.
  - Due to how ILP middlewares work, it would have to be created on Prepare, which doesn't entirely solve the issue.
  - The `paymentTag` needs to be the invoice id instead of the account id.
- Create the `Account` eagerly, but only create the TigerBeetle accounts when money arrives (i.e. on Fulfill).
- Delete the TigerBeetle accounts when the Invoice is deleted.
  - TigerBeetle doesn't support account deletion.

barnarddt:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/connector/core/test/middleware/account-middleware.test.ts
```diff
@@ -97,6 +97,35 @@ describe('Account Middleware', () => {
     expect(ctx.accounts.outgoing).toEqual(outgoingAccount)
   })
 
+  test('create TB account for PENDING incoming payment', async () => {
```
wilsonianb, line 100: You could make this a [`test.each`](url) to also test that the middleware handles `createLiquidityAccount` throwing. (See: [mockRejectedValueOnce](https://jestjs.io/docs/mock-function-api#mockfnmockrejectedvalueoncevalue))

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -536,6 +494,7 @@ describe('Incoming Payment Service', (): void => {
         })
         assert.ok(!isIncomingPaymentError(incomingPaymentOrError))
         incomingPayment = incomingPaymentOrError
+        await accountingService.createLiquidityAccount(incomingPayment)
```
wilsonianb, line 497: This `beforeEach` could be updated to just call [`createIncomingPayment`](https://github.com/interledger/rafiki/blob/main/packages/backend/src/tests/incomingPayment.ts#L9)

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -323,17 +309,12 @@ async function addReceivedAmount(
 ): Promise<IncomingPayment> {
   const received =
     value || (await deps.accountingService.getTotalReceived(payment.id))
```
wilsonianb, line 311:
```suggestion
    value !== undefined ? value : (await deps.accountingService.getTotalReceived(payment.id))
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -210,15 +226,27 @@ async function getAccountPage(
...
-        assetScale: payment.asset.scale
+      if (payment.state == OutgoingPaymentState.Funding) {
+        payment.sentAmount = {
+          value: BigInt(0),
+          assetCode: payment.asset.code,
+          assetScale: payment.asset.scale
+        }
+      } else {
+        payment.sentAmount = {
+          value: BigInt(amounts[i]),
```
wilsonianb, line 237:
```suggestion
      assert.ok(amounts[i] !== undefined || payment.state === OutgoingPaymentState.Funding)
      payment.sentAmount = {
        value: amounts[i] ? BigInt(amounts[i]) : BigInt(0),
        assetCode: payment.asset.code,
        assetScale: payment.asset.scale
      }
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -230,18 +258,26 @@ async function addSentAmount(
...
     }
   } else {
-    deps.logger.error({ outgoingPayment: payment.id }, 'account not found')
+    deps.logger.error(
+      { outgoingPayment: payment.id, state: payment.state, sent: sent },
+      'account not found for addSentAmount'
+    )
     throw new Error(
-      `Underlying TB account not found, payment id: ${payment.id}`
+      `Underlying TB account not found, outgoing payment id: ${payment.id}, ${payment.state}, ${value}, ${sent}, ${fundingZeroOrUndefined}`
```
wilsonianb, line 280:
```suggestion
      `Underlying TB account not found, outgoing payment id: ${payment.id}`
```

packages/backend/src/tests/outgoingPayment.ts
```diff
@@ -36,5 +36,9 @@ export async function createOutgoingPayment(
     quoteId: quote.id
   })
   assert.ok(!isOutgoingPaymentError(outgoingPaymentOrError))
+
+  const accountingService = await deps.use('accountingService')
+  await accountingService.createLiquidityAccount(outgoingPaymentOrError)
+
```
wilsonianb, line 42: Unless there's tests tinkering with the balance, I think this isn't needed
```suggestion

  const accountingService = await deps.use('accountingService')
  await accountingService.createLiquidityAccount(outgoingPaymentOrError)

```

wilsonianb (commented):
`main` needs to be merged into this branch

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -179,6 +176,25 @@ async function fundPayment(
...
+      // Don't complain if liquidity account already exists.
+      if (
+        err instanceof CreateAccountError &&
+        err.code === CreateAccountErrorCode.exists
+      ) {
+        // Do nothing.
+      } else {
+        throw err
+      }
+    }
```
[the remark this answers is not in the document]
sentientwaffle: Actually I think you should always treat `exists` as another name for `ok`. (i.e. leave this code as-is). Consider this series of events:

1. We hit this code, and start `createLiquidityAccount`.
2. TigerBeetle creates the account.
3. Before we receive the reply, Rafiki crashes for unrelated reasons.
4. We recover, and try to create the account again.
5. TigerBeetle says it already `exists`.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -179,6 +175,13 @@ async function fundPayment(
       return FundingError.WrongState
     }
     if (amount !== payment.sendAmount.value) return FundingError.InvalidAmount
+
+    // Create the outgoing payment liquidity account before trying to transfer funds to it.
+    await deps.accountingService.createLiquidityAccount({
+      id: id,
+      asset: payment.asset
+    })
```
sentientwaffle, line 183: See https://github.com/interledger/rafiki/pull/432/files#r951469744; I think this should still check & ignore `exists` errors.

packages/backend/src/tests/outgoingPayment.ts
barnarddt: Some test go Straight from `Funding` state to `Completed` or `Failed` without calling `fund`. So the Tigerbeetle account is never created via the normal flow. This is to accommodate those tests.

packages/backend/src/connector/core/middleware/account.ts
```diff
@@ -34,6 +36,27 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
...
+              if (
+                err instanceof CreateAccountError &&
+                err.code === CreateAccountErrorCode.exists
+              ) {
+                // Do nothing.
+              } else {
+                throw err
+              }
+            }
+          }
```
sentientwaffle, line 58: I'm not positive, but I think there is a race condition here. Assuming I'm reading things correctly, `incomingPayment.state` transitions from `Pending` to `Processing` _after_ the transfer[s] of money are committed. That means that this sequence is possible:

1. `getAccountByDestinationAddress` finds a pending `incomingPayment` and creates its liquidity account.
2. Commit `createTransfer` begins to commit, crediting the liquidity account.
3. (For unrelated reasons) Rafiki crashes before `destinationAccount.onCredit` can update the `state` to `Processing`.
4. Rafiki restarts and tries to create the account again — but it gets an `exists_with_different_{debits,credits}_{pending,posted}` error.

So I think those variants should be ignored here, like it ignores `exists`.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -344,6 +341,25 @@ async function fundPayment(
...
+    } catch (err) {
+      // Don't complain if liquidity account already exists.
+      if (
+        err instanceof CreateAccountError &&
+        err.code === CreateAccountErrorCode.exists
+      ) {
+        // Do nothing.
+      } else {
+        throw err
+      }
```
sentientwaffle, line 360: Similar to my other comment: Ignore `exists_with_different_{debits,credits}_{pending,posted}` here. If this SQL transaction aborts part-way through and restarts, the TB account may already exist (from the previous `accountingService.createLiquidityAccount`) with money (from the previous `accountingService.createDeposit`) without the `state` being updated to `Sending`.

packages/backend/src/connector/core/middleware/account.ts
wilsonianb: Do you think `createLiquidityAccount` should catch errors and roll up all `exists*` errors as a single returnable error, a la `createTransfers`?
https://github.com/interledger/rafiki/blob/main/packages/backend/src/accounting/transfers.ts#L80

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -322,18 +311,15 @@ async function addReceivedAmount(
...
-      assetScale: payment.asset.scale
-    }
-  } else {
-    deps.logger.error({ incomingPayment: payment.id }, 'account not found')
-    throw new Error(
-      `Underlying TB account not found, payment id: ${payment.id}`
-    )
+    value !== undefined
+      ? value
+      : await deps.accountingService.getTotalReceived(payment.id)
```
wilsonianb, line 316:
```suggestion
    value ?? await deps.accountingService.getTotalReceived(payment.id)
```

packages/backend/src/connector/core/middleware/account.ts
barnarddt: Went with `areAllAccountExistsErrors([err.code])` that is used by other parts of the code from when tigerbeetle was updated

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -80,25 +80,7 @@ describe('Incoming Payment Service', (): void => {
...
-        externalRef: '#123'
-      })
-      assert.ok(!isIncomingPaymentError(incomingPayment))
-      await expect(
-        accountingService.getBalance(incomingPayment.id)
-      ).resolves.toEqual(BigInt(0))
-    })
-
-    test('Cannot create incoming payment for nonexistent payment pointer', async (): Promise<void> => {
+    test('Cannot create incoming payment for nonexistent account', async (): Promise<void> => {
```
wilsonianb, line 83:
```suggestion
    test('Cannot create incoming payment for nonexistent payment pointer', async (): Promise<void> => {
```

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -1129,7 +1106,7 @@ describe('OutgoingPaymentService', (): void => {
     })
   })
 
-  describe('getPaymentPointerPage', (): void => {
+  describe('getAccountPage', (): void => {
```
wilsonianb, line 1109:
```suggestion
  describe('getPaymentPointerPage', (): void => {
```

wilsonianb (approved):

wilsonianb (approved):

PR #442, opened 2022-08-18
feat(auth): distributed key verification

cairin:

refactor(auth): remove the registry from the client service
```- Removes the idea of the central registry from the client service.
- Refactors display info url -> uri to match spec.
- Adds clientId to the jwk.
- Makes use of typeguards to verify jwk tokens.
```

feat(backend): create client keys services
```
Allows managing clients:
- Creating a client
- Adding keys to a client

Allows fetching a client key by kid.

Imports the auth package, to allow use of the JWK typing in the backend.
```

feat(backend): routing for client keys

feat(backend): add graphql admin endpoints for adding client and keys

## Context
Removes the need for the key directory, and does 
## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/client/service.ts
```diff
@@ -91,56 +89,47 @@ async function verifySig(
...
+    clientInfo.display.name !== clientDetails.name ||
+    clientInfo.display.uri !== clientDetails.uri ||
+    jwk.clientId !== clientDetails.id
+  )
+    return false
 
-  const { keys, ...registryClientInfo } = await getRegistryDataByKid(
-    deps,
-    jwk.kid
+  const currentKey = keys[0]
```
njlie, line 109: The `kid` should refer to a specific key now, so it doesn't need to account for an array of them here.

njlie:
Looks good so far, I just realize when reading this that I was unclear about the fact that the response from the key registry also has the new shape, which is essentially an inversion of what you've got for [ClientDetails](https://github.com/interledger/rafiki/blob/e8656fc2a2ff0b42cbc6aa85a1ad22a923f876a9/packages/auth/src/client/service.ts#L36) already.

```
{
  ...JWK stuff,
  client: {
    ...client stuff
  }
}

**Comment by cairin** [comment]
@njlie I've pushed the changes 👍  This simplifies the other side in the backend quite nicely. Just need to refactor..

File: packages/auth/src/client/service.ts [hunk 2]
```diff
@@ -1,14 +1,26 @@
 import * as crypto from 'crypto'
 import Axios from 'axios'
 import { importJWK, JWK } from 'jose'
-import { URL } from 'url'
 
 import { BaseService } from '../shared/baseService'
 import { IAppConfig } from '../config/app'
 
+export interface JWKWithRequired extends JWK {
+  // clientId is the custom field representing a client in the backend
```
wilsonianb, line 9:
```suggestion
  // client.id is the custom field representing a client in the backend
```

packages/auth/src/client/service.ts
```diff
@@ -91,56 +88,39 @@ async function verifySig(
...
-  if (!keys || !keys[0].kid || !keys[0].x) {
     return false
-  }
 
-  return !!(
-    verifyClientDisplay(clientInfo.display, registryClientInfo) &&
-    verifyJwk(jwk, keys[0])
-  )
+  if (key.exp && new Date() >= new Date(key.exp * 1000)) return false
+  if (key.nbf && new Date() < new Date(key.nbf * 1000)) return false
```
wilsonianb, line 114: The previous validation check (`verifyJwk`) required `key.exp` and `key.nbf`. Is this more correct?

packages/backend/src/clientKeys/service.ts
```diff
@@ -0,0 +1,38 @@
...
+    knex
+  }
+  return {
+    getKeyById: (keyId) => getKeyById(deps, keyId)
+  }
+}
+
+async function getKeyById(
+  deps: ServiceDependencies,
+  // In the form https://somedomain/keys/kid
```
wilsonianb, line 32:
```suggestion
  // In the form https://somedomain/keys/{keyId}
```

packages/backend/src/clients/service.ts
```diff
@@ -0,0 +1,78 @@
...
+    addKeyToClient: (options) => addKeyToClient(deps, options)
+  }
+}
+
+async function getClient(
+  deps: ServiceDependencies,
+  id: string
+): Promise<Client | undefined> {
+  const client = await Client.query(deps.knex).findById(id)
+  client.keys = await getClientKeys(deps, client.id)
```
wilsonianb, line 52: Does something like this work?
https://vincit.github.io/objection.js/api/query-builder/eager-methods.html#withgraphfetched
```suggestion
  const client = await Client.query(deps.knex).findById(id).withGraphFetched('keys')
```

cairin: It does. But I was trying to `orderBy` which was getting quite complicated to do with the relation mappings. And this is only used by the admin api so I thought it probably wasn't necessary.

packages/auth/src/client/service.ts
cairin: Neither `key.exp` or `key.nbf` are required to be defined (according to the type definition). So I shouldn't need to check that they both are. If they are required I can put them in the typeguard and change the typings.

All the tests passed, so I assumed it was working as @njlie intended  🤷

packages/backend/src/app.ts
```diff
@@ -363,6 +372,19 @@ export class App {
...
               )
+            } else if (path === '/keys/{keyId}' && method === HttpMethod.GET) {
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              this.publicRouter.get(
+                toRouterPath(path),
+                createValidatorMiddleware<ClientKeysContext>(openApi, {
+                  path,
+                  method
+                }),
```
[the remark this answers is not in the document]
cairin: So can I just rip it out for now?

packages/auth/src/client/service.ts
njlie: The type definition is correct, so this is more accurate to what I intended.

packages/backend/src/app.ts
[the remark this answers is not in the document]
wilsonianb: Remove `createValidatorMiddleware` and add this to the router outside of these `for` loops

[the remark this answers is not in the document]
cairin: @wilsonianb I have refactored this, please check it again 👍

packages/backend/src/graphql/schema.graphql
```diff
@@ -479,6 +496,21 @@ input CreateAccountInput {
   publicName: String
 }
 
+type Client implements Model {
```
njlie, line 499: One of the required fields in auth's `ClientDetails` type that isn't present here is `email`.

packages/backend/src/clientKeys/model.ts
```diff
@@ -0,0 +1,15 @@
+import { BaseModel } from '../shared/baseModel'
+
+import { JWKWithRequired } from 'auth'
+
+export class ClientKeys extends BaseModel {
+  public static get tableName(): string {
+    return 'clientKeys'
+  }
+
+  // The id should be the same as the kid.
```
njlie, line 10: Is this still true as of the discussion from this morning? i.e. isn't the `id` the internal database id and the `kid` a fully-qualified URL that contains the `id`?

packages/backend/src/clientKeys/routes.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function createClientKeysRoutes(
+  deps: ServiceDependencies
+): ClientKeysRoutes {
+  return {
+    get: (ctx: ClientKeysContext) => getKeyById(deps, ctx)
+  }
+}
+
+// TODO spec
```
njlie, line 20: Can this be cleaned up or elaborated upon?

packages/backend/src/graphql/schema.graphql
cairin: And `image`. Will update it 👍

packages/backend/src/clientKeys/model.ts
cairin: Yes, will update the comment.

packages/backend/src/graphql/schema.graphql
cairin: This has been fixed.

wilsonianb (approved):

njlie (approved):

PR #455, opened 2022-08-19
feat(auth): match dereferenced name with spec name

dclipp:

- Modifies the implementation of `continueGrant` to look for a parameter named `id` instead of `continueId`

## Context

- During debugging it was noticed that the parameter value of the `continueGrant` method was not being found
- At runtime, the property name is `id` (which matches the spec) but the function was expecting it to be named `continueId`

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/accessToken/routes.test.ts
+++ b/packages/auth/src/accessToken/routes.test.ts
@@ -306,7 +306,7 @@ describe('Access Token Routes', (): void => {
         {
           headers: { Accept: 'application/json' }
         },
-        { managementId }
+        { id: managementId }
       )
 
       await expect(accessTokenRoutes.rotate(ctx)).rejects.toMatchObject({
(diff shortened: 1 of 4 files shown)


njlie:
@dclipp Can you also include all the other methods on https://github.com/interledger/open-payments/blob/8cb2c997160a87d9ccf9307c25b1d29cb55bc0b7/auth-server-open-api-spec.yaml#L306, I think those are also destructuring by `managementId` when it should be looking for `id`.

sabineschaller:
Three of the auth tests are failing at the moment.

dclipp:
> @dclipp Can you also include all the other methods on https://github.com/interledger/open-payments/blob/8cb2c997160a87d9ccf9307c25b1d29cb55bc0b7/auth-server-open-api-spec.yaml#L306, I think those are also destructuring by `managementId` when it should be looking for `id`.

Fixed (as needed) and verified:

`grant` routes:
- [X] continueGrant
- [X] startInteraction
- [X] finishInteraction
- [X] denyInteraction

`access token` routes:
- [X] rotateToken
- [X] revokeToken
- [X] introspectToken

dclipp:
> Three of the auth tests are failing at the moment.

Pushed fixes with latest commit

sabineschaller (approved):

PR #480, opened 2022-08-23
Gnap token

rluckom-coil:
First try of passing gnap token through backend to auth without modifying interledgerjs.

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,10 @@ pnpm run localenv
 pnpm run localenv:seed:auth
 ```
 
-The local environment consists of a primary Rafiki instance and a peer Rafiki instance, each with
-its own docker compose files ([primary](infrastructure/local/docker-compose.yml), [peer](infrastructure/local/peer-docker-compose.yml)).
+The local environment consists of a primary Rafiki instance and a secondary Rafiki instance, each with
+its own docker compose files ([primary](infrastructure/local/docker-compose.yml), [secondary](infrastructure/local/peer-docker-compose.yml)).
 The primary `docker-compose.yml` includes the main Rafiki services `backend`, `auth`, and `rates`, as well
 as the required data stores tigerbeetle, redis, and postgres, so it can be run on its own.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -93,10 +93,10 @@ pnpm run localenv
 pnpm run localenv:seed:auth
```
rluckom-coil, line 93: This can be reverted, it will be part of a different PR

packages/backend/src/open_payments/setup_payment/service.ts
```diff
@@ -0,0 +1,122 @@
...
+    })
+  }
+
+  const queryIncomingPayment = async (
+    url: string,
+    token: string
+  ): Promise<IncomingPaymentJSON> => {
+    if (!createHttpUrl(url)) {
+      console.log('destinationPayment query failed: URL not HTTP/HTTPS.')
+      throw new Error(PaymentError.QueryFailed)
```
rluckom-coil, line 84: There `new Error()` bits semed necessary to make the test recognize that it was actualy throwing. Not consistent with elsewhere in the code.

packages/backend/src/open_payments/setup_payment/service.ts
```diff
@@ -0,0 +1,122 @@
...
+  const setupPayment = async (
+    incomingPaymentUrl: string,
+    gnapToken: string,
+    plugin: IlpPlugin
+  ): Promise<Pay.ResolvedPayment> => {
+    const incomingPayment = await queryIncomingPayment(
+      incomingPaymentUrl,
+      gnapToken
+    )
+    return await Pay.setupPayment({
```
rluckom-coil, line 66: @DonChangfoot suggested that we probably don't need the actual `setupPayment` call here because it only gets details that we already have from the `queryIncomingPayment` call.

README.md
rluckom-coil, line 93: This can be removed, it's in a different PR

PR #502, opened 2022-08-23
feat: add seed file framework

rluckom-coil:

Start setup of seed file / process for rafiki demos. This includes a fetcher for doing graphql and later non-graphql requests, an an example that works against the backend.
- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/mock-account-provider/src/parse_config.ts
```diff
@@ -0,0 +1,34 @@
+import { parse } from 'yaml'
```
rluckom-coil, line 1: This can be used by the Remix app to load the config. Doesn't pull in node-fetch or anything unnecessary.

packages/mock-account-provider/src/requesters.ts
```diff
@@ -0,0 +1,163 @@
...
+  }
+  return (await graphqlQuery({
+    resource: backendUrl,
+    method: 'post',
+    query: createPeerQuery,
+    variables: createPeerInput
+  })) as unknown as [PeerResponseElement]
+}
+
+export interface PeerResponseElement {
```
rluckom-coil, line 115: It is pretty tedious to copy the actual responses I'm getting in here as types, then try to coerce the responses into those types, then use the responses. Placating the type system here is probably 60% of the work. Is there an easier way or is this it?

adrianhopebailie (approved):
I think this works fine for now. Let's merge it in. Someone could optimise it in future to skip the build step and use ts-node to run the TS directly.

packages/mock-account-provider/package.json
```diff
@@ -10,12 +11,20 @@
     "@remix-run/node": "^1.6.8",
     "@remix-run/react": "^1.6.8",
     "@remix-run/serve": "^1.6.8",
+    "@types/node": "^18.7.12",
+    "@types/node-fetch": "^2.6.2",
+    "@types/uuid": "^8.3.4",
+    "lodash": "^4.17.21",
```
cairin, line 17: Would prefer it if we didn't us lodash.

packages/mock-account-provider/src/requesters.ts
cairin: You should be able to generate response types from the schema itself, and then set those types as the return type of this function. You won't need to coerce anything then too. I think code is already being generated by the backend, so you could just use that.

> I was trying to keep the graphql requester generic rather than repeating the fetch calls everywhere. is that a bad idea?

Yeah that seems reasonable. We could use something like the apollo client if this becomes unweildy.

PR #514, opened 2022-08-23
feat(auth): add client id to introspection response

### LINKED ISSUE #505: Auth: Add client id to introspection response
The client id is an opaque identifier that corresponds to multiple keys.

Example: Coil is the client but the web app and the phone app have different keys. When Coil wants to read all payments it has created, it will call the list endpoint with a token that corresponds to a grant which includes the `list` action. The introspection response from the AS must include the `client_id` that can then be used to filter the payments before returning.

sabineschaller:

- adds client id to token introspection response
- bumps the version of the auth spec to validate against

## Context

- fixes #505 

## Checklist

- [x] Tests added/updated
- [x] Documentation added

### DIFF
--- a/packages/auth/src/accessToken/service.ts
+++ b/packages/auth/src/accessToken/service.ts
@@ -25,6 +25,7 @@ interface ServiceDependencies extends BaseService {
 export interface Introspection extends Partial<Grant> {
   active: boolean
   key?: KeyInfo
+  clientId?: string
 }
 
 interface AccessTokenOpts {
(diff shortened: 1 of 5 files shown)


packages/auth/src/accessToken/service.ts
```diff
@@ -93,7 +94,16 @@ async function introspect(
       return { active: false }
     }
     const key = await deps.clientService.getKeyByKid(grant.clientKeyId)
-    return { active: true, ...grant, key: { proof: 'httpsig', jwk: key } }
+    const clientId = crypto
+      .createHash('sha256')
+      .update(key.client.id)
```
njlie, line 99: Should the `id` inside of `key.client` be removed from the response, or is obscuring the client's id not the intent?

sabineschaller: Actually, I think `jwk` shouldn't include the entire `client` object. The spec doesn't have it. I'm surprise the validation didn't pick it up.

njlie (approved):

PR #531, opened 2022-08-24
feat: cleanup add liquidity

rluckom-coil:

Add liquidity cleanup and response type

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/mock-account-provider/app/entry.server.tsx
+++ b/packages/mock-account-provider/app/entry.server.tsx
@@ -10,9 +10,15 @@ declare global {
 }
 
 if (!global.seeded) {
-  runSeed().then(() => {
-    global.seeded = true
-  })
+  runSeed()
+    .then(() => {
+      global.seeded = true
+    })
+    .catch((e) => {
+      console.log(
+        `seeding failed with ${e}. If seeding has already completed this can probably be ignored`
+      )
+    })
 }
 
 const ABORT_DELAY = 5000
--- a/packages/mock-account-provider/app/entry.server.tsx
+++ b/packages/mock-account-provider/app/entry.server.tsx
@@ -6,13 +6,13 @@ import { renderToPipeableStream } from 'react-dom/server'
 import { runSeed } from '../src/run_seed'
 
 declare global {
-  var seeded: boolean
+  let __seeded: boolean | undefined
 }
 
-if (!global.seeded) {
+if (!global.__seeded) {
   runSeed()
     .then(() => {
-      global.seeded = true
+      global.__seeded = true
     })
     .catch((e) => {
       console.log(
--- a/packages/mock-account-provider/app/entry.server.tsx
+++ b/packages/mock-account-provider/app/entry.server.tsx
@@ -3,6 +3,17 @@ import type { EntryContext } from '@remix-run/node'
 import { Response } from '@remix-run/node'
 import { RemixServer } from '@remix-run/react'
 import { renderToPipeableStream } from 'react-dom/server'
+import { runSeed } from '../src/run_seed'
+
+declare global {
+  var seeded: boolean
+}
+
+if (!global.seeded) {
+  runSeed().then(() => {
+    global.seeded = true
+  })
+}
 
 const ABORT_DELAY = 5000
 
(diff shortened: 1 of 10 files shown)


packages/mock-account-provider/src/apolloClient.ts
```diff
@@ -0,0 +1,59 @@
...
+  InMemoryCache
+} from '@apollo/client'
+import { setContext } from '@apollo/client/link/context'
+import { CONFIG } from './parse_config'
+import { onError } from '@apollo/client/link/error'
+import fetch from 'cross-fetch'
+
+const httpLink = createHttpLink({
+  uri: CONFIG.self.graphqlUrl,
+  fetch
```
rluckom-coil, line 15: When I was trying this before i was using a stale built file (stale `run_seed.js`). When I switched to using this one it complained that the Link here needed a supplied `fetch`. This can probably be removed once within remix

packages/mock-account-provider/src/requesters.ts
```diff
@@ -104,39 +93,26 @@ export async function createPeer(
...
-        }
-        staticIlpAddress: string
+  return apolloClient
+    .mutate({
+      mutation: createPeerMutation,
+      variables: createPeerInput
+    })
+    .then(({ data }): CreatePeerMutationResponse => {
+      console.log(data)
+      if (!data.createPeer.success) {
```
rluckom-coil, line 103: This was the one other place where I seemed to have to make a change--the `data` here has a `createPeer` object--I believe the "function name" in the gql--that constitutes another layer between the `data` and the `peer`. I don't think the structure of the response is inferred by the type system since it depends on the contents of the gql query. But the type system isn't fighting me as much any more, so still a big improvement.

cairin: Yeah that looks right. I can look into getting better type inference for the response 👍

packages/mock-account-provider/src/requesters.ts
```diff
@@ -30,13 +35,6 @@ function testGraphqlElementSuccess(el: GraphqlResponseElement) {
...
 }
 
-function isGraphqlArray(a: GraphqlResponseElement | GraphqlResponseArray) {
-  if (_.isArray(a)) {
-    return a as GraphqlResponseArray
-  }
-  return a as GraphqlResponseElement
-}
-
 export async function graphqlQuery(options: GraphqlQueryConfig) {
```
cairin, line 38: Do we still use this or can it be removed now?

packages/mock-account-provider/src/run_seed.ts
```diff
@@ -1,21 +1,45 @@
...
+        peer.initialLiquidity.toString(),
+        transferUid
+      )
+      return [peerResponse, liquidity]
+    })
+  )
+  console.log(JSON.stringify(peerResponses, null, 2))
+  const accountResponses = await Promise.all(
+    _.map(config.accounts, async (account: Account) => {
+      const createAccountResponse = await createAccount(
```
cairin, line 32: You probably don't need the await here with the `Promise.all`. You should just be able to return `createAccount(...)`

cairin (dismissed):

packages/mock-account-provider/src/requesters.ts
rluckom-coil: removed

packages/mock-account-provider/src/run_seed.ts
rluckom-coil: done

cairin (approved):

packages/mock-account-provider/app/entry.server.tsx
```diff
@@ -3,6 +3,23 @@ import type { EntryContext } from '@remix-run/node'
...
+  runSeed()
+    .then(() => {
+      global.seeded = true
+    })
+    .catch((e) => {
+      console.log(
+        `seeding failed with ${e}. If seeding has already completed this can probably be ignored`
+      )
+    })
+}
```
rluckom-coil, line 22: This is different from what was in the example https://github.com/remix-run/blues-stack/blob/main/app/db.server.ts .

as far as I can see that _always_ re-runs when in production mode. I don't think we want that, so I went with a simpler test that only runs if the global is undefined. If we get the kind of reboot that erases the global, then we will log an error on restarts but not block.

cairin (approved):

PR #535, opened 2022-08-24
feat(backend): add clientId to OP resources

### LINKED ISSUE #485: chore(backend): only lock grant when querying existing outgoing payments
https://github.com/interledger/rafiki/pull/366#discussion_r952334980

### LINKED ISSUE #506: Backend: Add client id to resource models
The client id is an opaque identifier. It is used to filter resources when sending a `list` request.

Example: Coil is the client but the web app and the phone app have different keys. When Coil wants to read all payments it has created, it will call the list endpoint with a token that corresponds to a grant which includes the `list` action. The introspection response from the AS must include the `client_id` that can then be used to filter the payments before returning.

sabineschaller:

- reference grant and client in incoming and outgoing payments
- allow to all GET incoming and outgoing payments  (read-all / list-all access) or only those belonging to a client (read / list access)

## Context

- fixes #506 
- fixes #485

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/migrations/20220228162503_create_incoming_payments_table.js
```diff
@@ -11,6 +11,7 @@ exports.up = function (knex) {
     table.string('state').notNullable()
     table.string('externalRef').nullable()
     table.uuid('connectionId').notNullable()
+    table.string('clientId').notNullable()
```
wilsonianb, line 14: What if we store `grantId` on all Open Payments sub-resources (not just outgoing payments) and store the `clientId` on the grant?

packages/backend/src/graphql/schema.graphql
```diff
@@ -491,6 +491,7 @@ type QuoteResponse {
 input CreateOutgoingPaymentInput {
   paymentPointerId: String!
   quoteId: String!
+  grant: String!
```
wilsonianb, line 494: Could `grantId` be made optional for incoming/outgoing payments instead of requiring grants in the admin api?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -247,14 +260,21 @@ async function handleDeactivated(
...
+  page = await IncomingPayment.query(deps.knex)
     .getPage(pagination)
     .where({
       paymentPointerId
     })
-    .withGraphFetched('asset')
+    .withGraphFetched('[asset, grant]')
+  if (clientId) {
+    page = page.filter(
+      (payment: IncomingPayment) => payment.grant.clientId === clientId
```
wilsonianb, line 270: This might need to happen within `getPage` in order to adhere to the `pagination` parameters

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -63,9 +64,15 @@ async function getIncomingPayment(
   ctx: ReadContext
 ): Promise<void> {
   let incomingPayment: IncomingPayment | undefined
+  const clientId = ctx.grant
+    ? ctx.grant?.access[0].actions.includes(AccessAction.ReadAll)
```
wilsonianb, line 68: :thinking: Should the OpenAPI spec disallow actions containing both `Read` and `ReadAll` (and `List` and `ListAll`)?

packages/backend/src/graphql/schema.graphql
sabineschaller: If there is no grant, than there is no `clientId`. Hence, we can't filter by `clientId`.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -33,6 +34,13 @@ export function createAuthMiddleware({
       ) {
         ctx.throw(403, 'Insufficient Grant')
       }
+      await Grant.query()
+        .insert({
+          id: grant.grant,
+          clientId: grant.clientId
+        })
+        .onConflict('id')
+        .ignore()
```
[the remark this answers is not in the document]
sabineschaller: Very good questions.

> Should it only happen for the create incoming and outgoing payment routes? 
I think that is worth exploring.

>Should it happen inside a knex transaction that's passed to the route/service via ctx so that we don't actually insert if the route operation fails?

I don't see the risk of a grant being stored but the payment not being created (only tiny storage).

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -78,11 +80,20 @@ export async function createIncomingPaymentService(
...
+  let incomingPayment: IncomingPayment
+  if (!clientId) {
+    incomingPayment = await IncomingPayment.query(deps.knex)
+      .findOne(key, value)
+      .withGraphFetched('asset')
+  } else {
+    incomingPayment = await IncomingPayment.query(deps.knex)
+      .findOne(`incomingPayments.${key}`, value)
+      .withGraphJoined('[asset, grant]')
+      .where('grant.clientId', clientId)
```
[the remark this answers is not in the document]
sabineschaller: I thought about that, too, and thought that the routes don't return it because it's not included in the `toJSON` and it's also not included in the` OutgoingPayment` type on the admin API.

packages/backend/src/open_payments/payment/incoming/service.ts
sabineschaller: Good point. Didn't think about that. It was incredible hard to change that to `withGraphJoined` so I gave up, but I'll give it another go.

packages/backend/src/open_payments/auth/grantModel.ts
```diff
@@ -0,0 +1,11 @@
+import { Model } from 'objection'
+import { DbErrors } from 'objection-db-errors'
+
+export class Grant extends DbErrors(Model) {
```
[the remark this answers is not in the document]
sabineschaller: And then what happens to `grant.ts`?

packages/backend/src/open_payments/payment/incoming/routes.ts
sabineschaller: I think it makes sense to have one or the other since the `All` one includes the other.

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -97,6 +104,7 @@ async function createIncomingPayment(
 
   const incomingPaymentOrError = await deps.incomingPaymentService.create({
     paymentPointerId: ctx.paymentPointer.id,
+    grantId: ctx.grant.grant,
```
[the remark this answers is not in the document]
sabineschaller: Like I said above, if grant is optional, how do we filter on `clientId`?

packages/backend/src/graphql/schema.graphql
wilsonianb: I'm imagining sub-resources created via the admin API not having a `grantId`/`clientId`.
Get and list via the admin API would always be the equivalent of read-all / list-all. We don't check auth / grant.
That should prevent clients doing read / read-all / list / list-all queries via Open Payments. Sub-resources without `grantId`/`clientId` are query-able for read-all / list-all.

packages/backend/src/open_payments/auth/grantModel.ts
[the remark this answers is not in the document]
wilsonianb: Maybe `grant.ts` and `service.ts` become `grant/service.ts`?
`GrantService` could include:
- `getByToken`
- `create`
- `lock`

Not sure how much sense that makes, especially if we'd be calling `getByToken` then `create`...

packages/backend/src/open_payments/payment/incoming/routes.ts
sabineschaller: https://github.com/interledger/open-payments/issues/178

packages/backend/src/open_payments/auth/grantModel.ts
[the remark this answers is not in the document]
sabineschaller: Should we have
`open_payments/grant` that only has `model.ts` and we add a service to create and lock a grant, and 
`open_payment/auth`, that has everything else?

And maybe `grant` is not the right name for the db table and the model name. Maybe we should call it `grantDetails` or `grantOwners`, ... (I'm not great at naming things).

packages/backend/src/open_payments/auth/grant.ts
```diff
@@ -66,11 +69,13 @@ export class Grant {
...
     this.access = options.access || []
+    this.clientId = options.clientId
   }
 
   public readonly active: boolean
   public readonly grant: string
   public readonly access: GrantAccess[]
+  public readonly clientId: string
 
   public includesAccess({
```
wilsonianb, line 80: [`createAuthMiddleware`](https://github.com/interledger/rafiki/blob/68115bb912ed75bb8d72bb542641df9f5dcff1d7/packages/backend/src/open_payments/auth/middleware.ts#L4-L30) is only ever being called with `AccessAction.Read`/`AccessAction.List` and never `AccessAction.ReadAll`/`AccessAction.ListAll`.

Maybe `includesAccess` could be updated to return true for an `action` of `AccessAction.Read`/`AccessAction.List` (specified by the auth middleware) if the grant access includes `AccessAction.ReadAll`/`AccessAction.ListAll`

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -421,7 +500,8 @@ describe('Incoming Payment Service', (): void => {
       getPage: (pagination: Pagination) =>
         incomingPaymentService.getPaymentPointerPage(
           paymentPointerId,
-          pagination
+          pagination,
+          client ? grant.clientId : undefined
```
wilsonianb, line 504: In order to test that when `clientId` is specified, incoming payments with a different or no `clientId` is returned, you could create a few other incoming payments in `beforeEach`.
We probably should have already been doing that to test that incoming payments for different payment pointers are also excluded.
Maybe the extra incoming payments should include same payment pointer, different `clientId` and vice versa.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -1106,19 +1125,32 @@ describe('OutgoingPaymentService', (): void => {
     })
   })
 
-  describe('getPaymentPointerPage', (): void => {
+  describe.each`
```
wilsonianb, line 1128: I `expect(this).makesTheTestsTakeLonger()`

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -59,61 +60,64 @@ export async function createOutgoingPaymentService(
...
+      .where('grant.clientId', clientId)
+  }
   if (outgoingPayment) return await addSentAmount(deps, outgoingPayment)
   else return
 }
 
 export interface CreateOutgoingPaymentOptions {
   paymentPointerId: string
   quoteId: string
+  grant?: Grant | string
```
wilsonianb, line 96: Can this still just be
```suggestion
  grant?: Grant
```

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -272,6 +279,16 @@ async function validateGrant(
     return false
   }
 
+  //lock grant
+  //TODO: update to use objection once it supports forNoKeyUpdate
+  await deps
+    .knex<GrantModel>('grants')
```
wilsonianb, line 285: I'm assuming this is using the transaction that is `deps.knex`?

packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -227,6 +233,7 @@ describe('QuoteService', (): void => {
         beforeEach(async (): Promise<void> => {
           incomingPayment = await createIncomingPayment(deps, {
             paymentPointerId: receivingPaymentPointer.id,
+            grantId: grant.id,
```
wilsonianb, line 236: I don't think this file needs to be updated now that `grantId` is optional

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -97,6 +108,7 @@ async function createIncomingPayment(
 
   const incomingPaymentOrError = await deps.incomingPaymentService.create({
     paymentPointerId: ctx.paymentPointer.id,
+    grantId: ctx.grant ? ctx.grant.grant : undefined,
```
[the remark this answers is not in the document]
sabineschaller: Should we push that can down the road a bit? While I think it should be required (that's why I had it as required in the first place), I see the benefit of not having it required dev mode purposes.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
sabineschaller: 😂 what?

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: That is how I understood it, yes.

packages/backend/src/open_payments/quote/service.test.ts
sabineschaller: I thought so, too, but then I realized that when querying the incoming payment, we proxy the request to the endpoint (to make it https) and add the `testAccessToken` (see [here](https://github.com/interledger/rafiki/blob/68115bb912ed75bb8d72bb542641df9f5dcff1d7/packages/backend/src/tests/app.ts#L86)-L98). Hence, there is a `grant` and a `clientId` associated with that incoming payment. Since the grant only has access type `read`, it will filter on that `clientId`. If we don't add the `grantId` in the test, the query fails.

packages/backend/src/open_payments/payment/outgoing/service.ts
sabineschaller: If we change that, should incoming payments then also not have `grantId` but `grant`?

packages/backend/src/open_payments/payment/incoming/service.test.ts
sabineschaller: Done for grants. Issue for payment pointers [here](https://github.com/interledger/rafiki/issues/602).

packages/backend/src/open_payments/auth/grant.ts
sabineschaller: Does that depend on 
- https://github.com/interledger/open-payments/issues/178

packages/backend/src/open_payments/payment/incoming/routes.ts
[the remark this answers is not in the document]
wilsonianb: :+1: I also had the thought about dev mode.
I'm currently not setting a grant to `ctx` in https://github.com/interledger/rafiki/pull/595/commits/e7f71d18b74b450df638d72180ee83748f848c56

packages/backend/src/open_payments/auth/grant.ts
wilsonianb: I don't think so.
The problem is that a get or list request will currently be rejected by the auth middleware if the grant has `read-all` or `list-all` (and no `read` or `list`)

packages/backend/src/open_payments/payment/outgoing/service.ts
wilsonianb: I think it's ok for `CreateOutgoingPaymentOptions` to take the entire `grant` and `CreateIncomingPaymentOptions` to have `grantId` because they have differnt needs. Outgoing payments needs to validate grant access limits.

packages/backend/src/open_payments/quote/service.test.ts
sabineschaller: Yes we could do that, also. I just added the change to make sure that 'read-all` includes `read` (same for list).

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -156,6 +158,7 @@ describe('Auth Middleware', (): void => {
     async ({ limitAccount }): Promise<void> => {
       const grant = new Grant({
         active: true,
+        clientId: uuid(),
```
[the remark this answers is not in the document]
sabineschaller: > what if the clientId differs from that of the stored grant?

Are you asking what happens when the introspection returns a grant with a grant id that we have already seen and hence stored but the `clientId` is different? While this shouldn't happen if the AS is functioning correctly, I guess we should make sure we catch that.

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -25,22 +25,24 @@ export const EXPIRY = parse('P90D') // 90 days in future
...
+  grantId?: string
   description?: string
   expiresAt?: Date
   incomingAmount?: Amount
   externalRef?: string
 }
 
 export interface IncomingPaymentService {
-  get(id: string): Promise<IncomingPayment | undefined>
+  get(id: string, clientId?: string): Promise<IncomingPayment | undefined>
```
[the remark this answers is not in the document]
wilsonianb: Matching the `clientId` (and `paymentPointerId`)
What if `incomingPaymentService.get` still just took the `id`, and `IncomingPaymentRoutes.get` would check the `clientId` (and `paymentPointerId`) of the returned incoming payment.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -33,6 +36,17 @@ export function createAuthMiddleware({
       ) {
         ctx.throw(403, 'Insufficient Grant')
       }
+      const grantRef = await grantReferenceService.get(grant.grant)
+      if (grantRef) {
+        if (grantRef.clientId !== grant.clientId) {
+          ctx.throw(409, 'Unknown Client ID')
+        }
+      } else {
+        await grantReferenceService.create({
```
wilsonianb, line 45: I think the `get` and `create` should happen within a db transaction

packages/backend/src/open_payments/grantReference/service.ts
```diff
@@ -0,0 +1,51 @@
+import { TransactionOrKnex } from 'objection'
+import { BaseService } from '../../shared/baseService'
+import { GrantReference } from './model'
+
+export interface GrantReferenceService {
+  get(grantId: string): Promise<GrantReference>
+  create(options: CreateGrantReferenceOptions): Promise<GrantReference>
+  lock(trx: TransactionOrKnex, grantId: string): Promise<void>
```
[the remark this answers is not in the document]
sabineschaller: Yes, that is still a todo. Even though it is already tested via the middleware and the outgoing payment service.

packages/backend/src/shared/routes.test.ts
```diff
@@ -1,20 +1,50 @@
...
 interface BaseResponse {
   id: string
 }
 
+interface setupOptions {
+  reqOpts: httpMocks.RequestOptions
+  params?: Record<string, string>
+  paymentPointer: PaymentPointer
+  grant?: Grant
+}
```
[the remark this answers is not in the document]
sabineschaller: I think we should keep it as is. I don't see the value in adding another interface.

packages/backend/src/open_payments/auth/middleware.ts
sabineschaller: You are saying within the same transaction?

wilsonianb: Yes, otherwise `create` will throw if the grant was created (after `get`) via another another request

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -40,6 +43,17 @@ export function createAuthMiddleware({
       ) {
         ctx.throw(403, 'Insufficient Grant')
       }
+      const grantRef = await grantReferenceService.get(grant.grant)
+      if (grantRef) {
+        if (grantRef.clientId !== grant.clientId) {
+          ctx.throw(409, 'Unknown Client ID')
```
wilsonianb, line 49: Should this instead log and throw `500`?
It seems like more an internal conflict with the AS than the client

packages/backend/src/open_payments/grantReference/service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { BaseService } from '../../shared/baseService'
+import { GrantReference } from './model'
+
+export interface GrantReferenceService {
+  get(grantId: string, trx?: Transaction): Promise<GrantReference>
+  create(
+    options: CreateGrantReferenceOptions,
+    trx?: Transaction
+  ): Promise<GrantReference>
+  lock(grantId: string, trx?: TransactionOrKnex): Promise<void>
```
wilsonianb, line 11: I think `lock` should still require a `trx`

packages/backend/src/open_payments/grantReference/service.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  describe('Create and Get GrantReference', (): void => {
+    test.each`
+      transaction
+      ${false}
+      ${true}
+    `(
+      'a grant reference can be created and fetched - transaction: $transaction',
+      async ({ transaction }): Promise<void> => {
+        const trx = transaction ? knexTrx : undefined
```
wilsonianb, line 47: `knexTrx` is undefined

packages/backend/src/open_payments/grantReference/service.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        const grantRef = await grantReferenceService.create(
+          {
+            id,
+            clientId: uuid()
+          },
+          trx
+        )
+        const retrievedRef = await grantReferenceService.get(id, trx)
+        expect(grantRef).toEqual(retrievedRef)
+      }
```
wilsonianb, line 58:
```suggestion
        if (transaction) {
          await trx.rollback()
          await expect(await grantReferenceService.get(id)).resolves.toBeUndefined()
        }
      }
```

packages/backend/src/open_payments/grantReference/service.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      const grantRef = await grantReferenceService.create({
+        id: uuid(),
+        clientId: uuid()
+      })
+      const lock = async (): Promise<void> => {
+        return await GrantReference.transaction(knex, async (trx) => {
+          await grantReferenceService.lock(grantRef.id, trx)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          await new Promise((f: any) => setTimeout(f, 6000))
+          await grantReferenceService.get(grantRef.id, trx)
```
wilsonianb, line 76: You could also just lock it and then do
```
await expect(grantReferenceService.get(grantRef.id, trx).skipLocked()).resolves.toBeUndefined()
```

packages/backend/src/open_payments/auth/middleware.ts
sabineschaller: I was on the fence between the two.

packages/backend/src/open_payments/grantReference/service.test.ts
sabineschaller: That is probably the faster version. 👍

sabineschaller: So I can't do `skipedLocked` on the `get` method, instead I put 
```
await GrantReference.transaction(async (trx) => {
  await grantReferenceService.lock(grantRef.id, trx)
   await expect(
    GrantReference.query(trx).findById(grantRef.id).skipLocked()
  ).resolves.toBeUndefined()
})
```
However, that resulted in `Rejected to value: [Error: .skipLocked() can only be used after a call to .forShare() or .forUpdate()!]`.

wilsonianb: You could add one of those to query or just leave the test as is.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -40,8 +46,27 @@ export function createAuthMiddleware({
...
+            {
+              id: grant.grant,
+              clientId: grant.clientId
+            },
+            trx
+          )
+        }
+        ctx.grant = grant
+        await next()
+      })
```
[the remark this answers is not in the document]
wilsonianb: In which case `GrantReferenceService`' `deps` wouldn't need `knex` :exploding_head:

packages/backend/src/open_payments/grantReference/service.test.ts
sabineschaller: That also didn't work. It queries it successfully. It probably has to do with the fact that we can't do `forNoKeyUpdate()` using objection.  I'll just leave it as is.

packages/backend/src/open_payments/grantReference/service.ts
```diff
@@ -0,0 +1,44 @@
+import { Transaction, TransactionOrKnex } from 'objection'
+import { GrantReference } from './model'
+
+export interface GrantReferenceService {
+  get(grantId: string, trx?: Transaction): Promise<GrantReference>
+  create(
+    options: CreateGrantReferenceOptions,
+    trx?: Transaction
+  ): Promise<GrantReference>
+  lock(grantId: string, trx: TransactionOrKnex): Promise<void>
```
wilsonianb, line 10:
```suggestion
  lock(grantId: string, trx: Transaction): Promise<void>
```

sabineschaller: This has implications on the `validateGrant` function in the `OutgoingPaymentService`. Right now, one of its input parameters is `deps:ServiceDependencies`, which includes `deps.knex:TransactionOrKnex`. If we don't want to use that, I have to specifically pass the transaction to `validateGrant`. Do you think we should make that change?

wilsonianb: - https://github.com/interledger/rafiki/issues/618

wilsonianb (approved):

PR #544, opened 2022-08-25
feat(backend): use payment pointer url in Open Payments

### LINKED ISSUE #416: Change account to payment pointer
Update Rafiki to not reference accounts but rather be reference provisioning Payment Pointers

wilsonianb:

- Use payment pointer url in Open Payments endpoints
- depends on https://github.com/interledger/open-payments/pull/145

## Context

- fixes #416 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -68,11 +69,20 @@ export class IncomingPayment
         from: 'incomingPayments.assetId',
         to: 'assets.id'
       }
+    },
+    paymentPointer: {
+      relation: Model.HasOneRelation,
+      modelClass: PaymentPointer,
```
matdehaast, line 75: can we change this to use the path based imports like in auth to prevent cyclic imports in the future.

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -68,11 +69,20 @@ export class IncomingPayment
         from: 'incomingPayments.assetId',
         to: 'assets.id'
       }
+    },
+    paymentPointer: {
+      relation: Model.HasOneRelation,
```
matdehaast, line 74: I would think think this should be a belongsTo relation?

packages/backend/src/open_payments/payment/outgoing/routes.ts
```diff
@@ -128,9 +128,9 @@ function outgoingPaymentToBody(
   return Object.fromEntries(
     Object.entries({
       ...outgoingPayment.toJSON(),
-      // paymentPointer: `${deps.config.publicHost}/${outgoingPayment.paymentPointerId}`,
-      accountId: `${deps.config.publicHost}/${outgoingPayment.paymentPointerId}`,
-      id: `${deps.config.publicHost}/${outgoingPayment.paymentPointerId}/outgoing-payments/${outgoingPayment.id}`,
+      id: `${outgoingPayment.paymentPointer.url}/outgoing-payments/${outgoingPayment.id}`,
+      paymentPointer: outgoingPayment.paymentPointer.url,
+      paymentPointerId: undefined,
```
matdehaast, line 133: Should we be returning the this? I don't think so

packages/backend/src/open_payments/payment_pointer/routes.ts
```diff
@@ -24,19 +22,16 @@ export async function getPaymentPointer(
   deps: ServiceDependencies,
   ctx: PaymentPointerContext
 ): Promise<void> {
-  const { accountId: paymentPointerId } = ctx.params
-  const paymentPointer = await deps.paymentPointerService.get(paymentPointerId)
-  if (!paymentPointer) {
+  if (!ctx.paymentPointer) {
     ctx.throw(404)
     return // unreachable, but satisfies typescript
```
matdehaast, line 27: Know its not in the review but I think we can do this here
```
return ctx.throw(404)
```

packages/backend/src/open_payments/payment/outgoing/routes.ts
wilsonianb: as in remove `paymentPointerId: undefined,`?
which should be fine as long as the `toJSON` gets updated to exclude `paymentPointerId`, which should also be fine

packages/backend/src/open_payments/payment/incoming/model.ts
wilsonianb: https://vincit.github.io/objection.js/guide/relations.html#require-loops
>In addition to objection's solutions, you can always organize your code so that such loops are not created.

If we really want it, we can open an issue to change all of the backend model relations in one go.

packages/backend/src/open_payments/payment/outgoing/routes.ts
wilsonianb: I was able to change this for outgoing payments, but `toJSON` is being called for the request to `quoteUrl` which expects `paymentPointerId`:
https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/quote/service.ts#L289

matdehaast (dismissed):

matdehaast (approved):

PR #545, opened 2022-08-25
chore: remove lint-staged commit hook ¯\_(ツ)_/¯

traviscrist:

- Remove lint-staged which prettier and eslint checks from the commit hook
- Adds new `checks` script which runs them manually `pnpm checks`
- Keeps the commit message hook

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


cairin:
Why not just run the formatting and linting rather than just a check? There doesn't seem to be any speed advantages.

traviscrist:
> Why not just run the formatting and linting rather than just a check? There doesn't seem to be any speed advantages.

The script is called checks, I think if we want to do formatting then we should just expand what `format` does and bring in eslint to the format script. @cairin thoughts on adding that?

cairin:
I wouldn't put the linting in the formatting step. 

I asked because I wouldn't want to run this and then have to run formatting again afterwards if it fails. I would just want to format. It's not useful to check formatting on your local machine when you can just format - especially if it takes the same time to format and to check formatting. Same with linting, if eslint can fix some minor things and not error, why not just let it do that?

traviscrist:
> I wouldn't put the linting in the formatting step.
> 
> I asked because I wouldn't want to run this and then have to run formatting again afterwards if it fails. I would just want to format. It's not useful to check formatting on your local machine when you can just format - especially if it takes the same time to format and to check formatting. Same with linting, if eslint can fix some minor things and not error, why not just let it do that?

I think I found a good balance. Take a look at the updated scripts. It makes `format` do all formatting and provides a `checks` which does the lint and formatting checks. Leaves it up to the developer which they want to use.

matdehaast (approved):

cairin (approved):

PR #575, opened 2022-08-29
chore: pin redis to specific version

sublimator:
Pin redis to a spefic version. 

## Context
Rafiki seems to use some new features, so some tests fails if the local machine has an older version of redis.


matdehaast (approved):
Thanks @sublimator 🙏

sublimator:
image

This requires an admin or something to merge?

Issue #351, opened 2022-06-01
Validate GNAP requests with OpenAPI


njlie:
From https://github.com/interledger/rafiki/pull/343#discussion_r880691528:

> 🤔 Maybe this OpenAPI validation middleware (#339) could be a workspace in the monorepo: [https://github.com/interledger/rafiki/tree/bw-openapi/packages/backend/src/openapi](https://github.com/interledger/rafiki/tree/bw-openapi/packages/backend/src/openapi?rgh-link-date=2022-05-24T16%3A04%3A12Z) Then this could validate requests based on [https://github.com/interledger/open-payments/blob/master/auth-server-open-api-spec.yaml](https://github.com/interledger/open-payments/blob/master/auth-server-open-api-spec.yaml?rgh-link-date=2022-05-24T16%3A04%3A12Z)

See:
https://github.com/interledger/rafiki/pull/339/commits/4556d21258dc2c8c61d3b0f735b6a50d2f430b14

sabineschaller:
Reopened to specify the `ctx.body` type in `createGrantInitiation` and elsewhere to make use of the benefits of the OpenAPI validator middleware type guard.
See [comment](https://github.com/interledger/rafiki/pull/588#discussion_r966078660)

PR #588, opened 2022-09-02
feat(auth): allow non-interactive grants for incoming payments

### LINKED ISSUE #415: Support grant issuance without interaction
For some access types, like `incoming-payments`, the AS should support issuing the access tokens for these grants immediately without requiring interaction from the user. `incoming-payments` shouldn't carry any financial risk for the user so interaction isn't necessary.

sabineschaller:

- if the access type is `incoming-payment`, the grant will be issued right away, without interaction

## Context

- fixes #415 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


sabineschaller:
Q: Should we have the same behavior for quote grant requests? @wilsonianb

wilsonianb:
> Q: Should we have the same behavior for quote grant requests? @wilsonianb

I'm not sure.
Creating a quote can indirectly reveal the sender's available balance at the account provider.
OTOH a grant allowing quote creation is likely to also include outgoing payment creation, which would require interaction.

packages/auth/src/accessToken/service.test.ts
```diff
@@ -207,21 +207,21 @@ describe('Access Token Service', (): void => {
       await token.$query(trx).patch({ expiresIn: 1000000 })
       const result = await accessTokenService.revoke(token.managementId)
       expect(result).toBeUndefined()
-      token = await AccessToken.query(trx).findById(token.id)
+      token = (await AccessToken.query(trx).findById(token.id)) as AccessToken
       expect(token).toBeUndefined()
```
wilsonianb, line 211:
```suggestion
      await expect(AccessToken.query(trx).findById(token.id)).resolves.toBeUndefined()
```
Ditto below

packages/auth/src/grant/routes.ts
```diff
@@ -66,7 +71,8 @@ export function createGrantRoutes({
   }
 }
 
-function validateGrantRequest(
+// exported for testing
+export function validateGrantRequest(
```
wilsonianb, line 75: Does this have checks that aren't covered by the OpenAPI validator middleware?
https://github.com/interledger/rafiki/blob/0642a0fa91f892b1398ff10ae07a28c79955b39a/packages/auth/src/app.ts#L239

packages/auth/src/grant/routes.ts
```diff
@@ -102,6 +108,27 @@ async function createGrantInitiation(
     return
   }
 
+  if (
+    body.access_token.access
+      .map((acc) => {
+        return isIncomingPaymentAccessRequest(acc as IncomingPaymentRequest)
+      })
+      .every((el) => el === true) &&
+    !deps.config.incomingPaymentInteraction
```
wilsonianb, line 117: Check the boolean first
```suggestion
    !deps.config.incomingPaymentInteraction &&
    body.access_token.access
      .map((acc) => {
        return isIncomingPaymentAccessRequest(acc as IncomingPaymentRequest)
      })
      .every((el) => el === true)
```

packages/auth/src/grant/service.test.ts
```diff
@@ -125,11 +125,44 @@ describe('Grant Service', (): void => {
...
+        startMethod: expect.arrayContaining([StartMethod.Redirect])
+      })
+
+      const dbAccessGrant = (await Access.query(trx)
+        .where({
+          grantId: grant.id
+        })
+        .first()) as Access
 
       expect(dbAccessGrant.type).toEqual(AccessType.IncomingPayment)
```
wilsonianb, line 167:
```suggestion
      await expect(Access.query(trx)
        .where({
          grantId: grant.id
        })
        .first()
      ).resolves.toMatchObject({
        type: AccessType.IncomingPayment
      })
```
Ditto elsewhere

packages/auth/src/grant/service.test.ts
```diff
@@ -164,7 +197,10 @@ describe('Grant Service', (): void => {
       expect(fetchedGrant?.id).toEqual(grant.id)
     })
     test('Can fetch a grant by its interaction information', async (): Promise<void> => {
-      const fetchedGrant = await grantService.getByInteraction(grant.interactId)
+      const fetchedGrant = await grantService.getByInteraction(
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        grant.interactId!
```
sublimator, line 202:
```typescript
export function notNullOrUndef<T>(
  t: T | null | undefined,
  name = '<unknown>'
): T | never {
  if (t == null) {
    throw new Error(`Expecting not null for ${name}`)
  } else {
    return t
  }
}
```

Found a helper like above is useful to get runtime checks as well. 

Use as:

```typescript 
notNullOrUndef(grant.interactionId, 'grant.interactionId')
```

Pity TypeScript doesn't have some kind of macros

packages/auth/src/accessToken/service.test.ts
```diff
@@ -207,21 +207,21 @@ describe('Access Token Service', (): void => {
       await token.$query(trx).patch({ expiresIn: 1000000 })
       const result = await accessTokenService.revoke(token.managementId)
       expect(result).toBeUndefined()
-      token = await AccessToken.query(trx).findById(token.id)
+      token = (await AccessToken.query(trx).findById(token.id)) as AccessToken
```
sublimator, line 210: Those casts seem unnecessary if token is expected to be undefined.
It also makes it seem like the query/findById api could use enhancement.

packages/auth/src/config/app.ts
```diff
@@ -38,5 +38,7 @@ export const Config = {
   accessTokenExpirySeconds: envInt('ACCESS_TOKEN_EXPIRY_SECONDS', 10 * 60), // Default 10 minutes
   databaseCleanupWorkers: envInt('DATABASE_CLEANUP_WORKERS', 1),
   accessTokenDeletionDays: envInt('ACCESS_TOKEN_DELETION_DAYS', 30),
-  introspectionHttpsig: process.env.INTROSPECTION_HTTPSIG === 'true'
+  introspectionHttpsig: process.env.INTROSPECTION_HTTPSIG === 'true',
```
sublimator, line 41: envBool?

packages/auth/src/grant/routes.ts
sabineschaller: I don't think so but I'd like @njlie to confirm.

packages/auth/src/accessToken/service.test.ts
sabineschaller: You are right. I was sloppy here. I just added the type cast for the compiler to be happy but I should have looked at the entire code block and improved it.

packages/auth/src/config/app.ts
sabineschaller: Sure, I can add that.

packages/auth/src/accessToken/service.test.ts
sublimator: Moaning compilers!

I was just poking around your draft after looking to see if MY pr was merged: 
https://github.com/interledger/rafiki/pull/575

sabineschaller:
> Creating a quote can indirectly reveal the sender's available balance at the account provider.
OTOH a grant allowing quote creation is likely to also include outgoing payment creation, which would require interaction.

Then let's leave it as is for now.

wilsonianb (commented):
- Should `config.authServerSpec` be updated to include https://github.com/interledger/open-payments/pull/168 ?

packages/auth/src/config/app.ts
```diff
@@ -10,6 +10,10 @@ function envInt(name: string, value: number): number {
   return envValue == null ? value : parseInt(envValue)
 }
 
+function envBool(name: string, value: boolean): boolean {
+  const envValue = process.env[name]
+  return envValue == null ? value : envValue === 'true'
```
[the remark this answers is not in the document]
sabineschaller: I didn't know we had that.

packages/auth/src/grant/routes.ts
sabineschaller: I'd hold off on any work there now that Adrian has been working on his auth server implementation.

packages/auth/migrations/20220504161932_create_grants_table.js
```diff
@@ -5,18 +5,18 @@ exports.up = function (knex) {
     table.string('state').notNullable()
     table.specificType('startMethod', 'text[]').notNullable()
 
-    table.string('continueToken').unique()
-    table.string('continueId').unique()
+    table.string('continueToken').notNullable().unique()
+    table.string('continueId').notNullable().unique()
```
njlie, line 9: It shouldn't be necessary to generate continuation information if interaction isn't necessary for a grant. The `POST /` should just return the access token without any continuation information.

packages/auth/src/grant/routes.ts
njlie: > I don't think so but I'd like @njlie to confirm.

Yeah I'm pretty sure this can be excised now, I wrote this before we had brought OpenAPI into the auth service. This basically is a type guard against the `POST /` request which is definitely covered.

packages/auth/src/grant/routes.ts
```diff
@@ -384,12 +411,32 @@ async function continueGrant(
...
+) {
+  return {
     access_token: {
       value: accessToken.value,
-      manage: config.authServerDomain + `/token/${accessToken.managementId}`,
+      manage: domain + `/token/${accessToken.managementId}`,
       access: access.map((a: Access) => accessToBody(a)),
-      expiresIn: accessToken.expiresIn
+      expiresIn: accessToken.expiresIn,
+      continue: {
```
njlie, line 434: We can exclude the `continue` object if there's no interaction required and the AS can just supply an `access_token` instead. Also vice/versa for if interaction _is_ required. It's basically an either/or situation between `continue` and `access_token` with how we're using GNAP.

packages/auth/src/grant/routes.ts
sabineschaller: I removed the validation function and re-opened #351

packages/auth/migrations/20220504161932_create_grants_table.js
sabineschaller: I think we discussed that while you were away. We always want to be able to delete/revoke the grant and for that you need the continue token and id. Hence, it must always be present. We also already changed the spec to make it always required.

packages/auth/src/grant/routes.ts
sabineschaller: See above.

packages/auth/src/grant/service.test.ts
```diff
@@ -149,7 +184,7 @@ describe('Grant Service', (): void => {
       const fetchedGrant = await grantService.getByContinue(
```
wilsonianb, line 184:
```suggestion
      assert.ok(interactRef)
      const fetchedGrant = await grantService.getByContinue(
```
and remove the `as string` cast below

packages/auth/src/grant/service.test.ts
```diff
@@ -164,7 +199,10 @@ describe('Grant Service', (): void => {
       expect(fetchedGrant?.id).toEqual(grant.id)
     })
     test('Can fetch a grant by its interaction information', async (): Promise<void> => {
-      const fetchedGrant = await grantService.getByInteraction(grant.interactId)
+      const fetchedGrant = await grantService.getByInteraction(
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        grant.interactId!
+      )
```
wilsonianb, line 205:
```suggestion
      assert.ok(grant.interactId)
      const fetchedGrant = await grantService.getByInteraction(grant.interactId)
```

packages/auth/src/grant/routes.ts
```diff
@@ -96,14 +87,38 @@ async function createGrantInitiation(
...
+  if (
+    !deps.config.incomingPaymentInteraction &&
+    body.access_token.access
+      .map((acc) => {
+        return isIncomingPaymentAccessRequest(acc as IncomingPaymentRequest)
+      })
+      .every((el) => el === true)
+  ) {
+    const grant = await grantService.create(body)
+    const accessToken = await deps.accessTokenService.create(grant.id)
```
wilsonianb, line 100: Should these happen in a transaction?

packages/auth/src/grant/routes.ts
```diff
@@ -102,6 +108,27 @@ async function createGrantInitiation(
     return
   }
 
+  if (
+    !deps.config.incomingPaymentInteraction &&
+    body.access_token.access
+      .map((acc) => {
+        return isIncomingPaymentAccessRequest(acc as IncomingPaymentRequest)
```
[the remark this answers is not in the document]
sabineschaller: I didn't want to touch this because I thought I would add the change to have the `ctx.body` typed but that's probably going to be a subsequent PR so yes, I'll change that.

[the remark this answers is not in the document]
wilsonianb: `BaseAccessRequest` might work, but I just noticed the other type guards also have the target type as the argument type.
Maybe all they can all be updated as part of #351 instead of this pr.

packages/auth/src/grant/routes.ts
sabineschaller: I don't think so. If grant creation fails, we'll never get to the token creation. And if token creation fails, the grant should still be granted. Or am I misunderstanding this?

wilsonianb: I think the main benefit would be rolling back grant creation if access token creation fails.
>And if token creation fails, the grant should still be granted. 

As is, I think we'd 500 and the request would need to be retried.

packages/auth/src/grant/routes.ts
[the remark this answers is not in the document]
sabineschaller: I think that is a good idea.

packages/auth/src/grant/routes.ts
sabineschaller: When you put it this way, I guess you are right.

sabineschaller: But I get a foreign key error because the transaction to create the grant has not been committed and hence I can't insert the access token with the grant id foreign key. 🤔

wilsonianb: This appears to work: https://github.com/wilsonianb/rafiki/pull/134/commits/05b300dc0c176eeece8cf76d86543d4e7d17e8fe?diff=unified&w=1

packages/auth/src/grant/routes.ts
```diff
@@ -96,17 +96,26 @@ async function createGrantInitiation(
...
+      ctx.status = 200
+      ctx.body = createGrantBody({
+        domain: deps.config.authServerDomain,
+        grant,
+        access,
+        accessToken
+      })
+      return
+    } catch (err) {
+      await trx.rollback()
```
sabineschaller, line 116: Does this actually roll back the transaction if we commit in line 105?

wilsonianb: :thinking: yeah, I'm not sure what'd happen if it threw after the commit and tried to rollback.
Options seem to be:

1. end the try block directly after the commit. (Declare `let access: Access[]` before the try block, or return the access on the grant from `grantService.create` (and declare let grant: Grant))
2. Do this (instead of `commit`/`rollback`): https://github.com/wilsonianb/rafiki/pull/134/commits/726f759baea8eba17f74d5f1de18960a48c4b1ae But then, either `accessService.getByGrant` needs to accept a `trx` or you return `access` with the created `grant`.

packages/auth/src/grant/routes.ts
```diff
@@ -384,12 +411,37 @@ async function continueGrant(
...
-      manage: config.authServerDomain + `/token/${accessToken.managementId}`,
+      manage: domain + `/token/${accessToken.managementId}`,
       access: access.map((a: Access) => accessToBody(a)),
       expiresIn: accessToken.expiresIn
+    },
+    continue: {
+      access_token: {
+        value: grant.continueToken
+      },
+      uri: domain + `continue/${grant.continueId}`
```
wilsonianb, line 444: Should this still include `wait` (`RECOMMENDED`)?
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.1

wilsonianb (approved):

packages/auth/src/grant/routes.ts
sabineschaller: I don't think so because the grant is granted right away. At least that was my interpretation of it.

PR #595, opened 2022-09-08
feat(backend): add Open Payments client service

### LINKED ISSUE #581: Query incoming payments directly
Instead of incoming payment queries happening within ilp-pay's [`setupPayment`](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay#setuppayment).
This will allow the Rafiki backend to include the necessary GNAP access token (#583) in incoming payment requests.

related to:
- https://github.com/interledgerjs/interledgerjs/issues/295
- https://github.com/interledgerjs/interledgerjs/issues/268#issuecomment-1220853279

Existing work:
- #480 
- https://github.com/interledger/rafiki/tree/mdh-resolve-payment-details

### LINKED ISSUE #586: Add a dev mode access token
Add a configurable development-mode-only access token that bypasses the RS doing token introspection at the AS.
If the RS needs to represent all of these privileges as a grant, this may depend on:
- https://github.com/interledger/open-payments/issues/162

### LINKED ISSUE #589: Downgrade axios
I'm getting the following error trying to add a new `get` request with `axios`:
```
    TypeError: Cannot assign to read only property 'isAxiosError' of object 'AxiosError: Request failed with status code 404'

      at Object.toFlatObject (node_modules/.pnpm/axios@0.27.2/node_modules/axios/lib/utils.js:387:23)
      at Function.Object.<anonymous>.AxiosError.from (node_modules/.pnpm/axios@0.27.2/node_modules/axios/lib/core/AxiosError.js:73:9)
      at RedirectableRequest.handleRequestError (node_modules/.pnpm/axios@0.27.2/node_modules/axios/lib/adapters/http.js:356:25)
      at OverriddenClientRequest.eventHandlers.<computed> (node_modules/.pnpm/follow-redirects@1.15.1/node_modules/follow-redirects/index.js:14:24)
      at Socket.<anonymous> (node_modules/.pnpm/propagate@2.0.1/node_modules/propagate/index.js:64:17)
      at node_modules/.pnpm/nock@13.2.9/node_modules/nock/lib/socket.js:100:14
```
I believe it's related to the breaking change to `AxiosError` introduced in [`v0.27.0`](https://github.com/axios/axios/releases/tag/v0.27.0)
I've confirmed that this unreleased change resolves the issue:
https://github.com/axios/axios/commit/bdf493cf8b84eb3e3440e72d5725ba0f138e0451#diff-51cc788f57172f1e4c888aff4d9a70034eacbd462bb0c0a626c50dc9055715caR86

wilsonianb:

- add Open Payments client service
- replace calls to `Pay.setupPayment`
- return correct `ilpStreamConnection` type
- Update `ilp-packet` to pre-release version to match types of other [interledgerjs](https://github.com/interledgerjs/interledgerjs) dependencies.
- downgrade `axios`

## Context

- fixes #581 
- fixes #586
- fixes #589
- related to https://github.com/interledger/open-payments/issues/160

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -134,6 +141,13 @@ export class IncomingPayment
     return `${this.paymentPointer.url}/incoming-payments/${this.id}`
   }
 
+  public get hasConnection(): boolean {
```
wilsonianb, line 136: An alternative to this would be to set `connectionId` to `null` when the incoming payment transitions to `Completed`/`Expired`

packages/backend/src/open_payments/client/service.ts
```diff
@@ -0,0 +1,70 @@
...
+  deps_: Omit<ServiceDependencies, 'validateResponse'>
+): Promise<OpenPaymentsClientService> {
+  const log = deps_.logger.child({
+    service: 'OpenPaymentsClientService'
+  })
+  const validateResponse =
+    deps_.openApi.createResponseValidator<IncomingPaymentJSON>({
+      path: '/incoming-payments/{incomingPaymentId}',
+      method: HttpMethod.GET
+    })
```
sabineschaller, line 31: Why does that have to be overwritten?

packages/backend/src/open_payments/client/service.ts
```diff
@@ -0,0 +1,70 @@
...
+    validateResponse
+  }
+  return {
+    incomingPayment: {
+      get: (url) => getIncomingPayment(deps, url)
+    }
+  }
+}
+
+export async function getIncomingPayment(
```
sabineschaller, line 44: Could we check whether the incoming payment is a resource on the same Rafiki instance and if so, we just use the service to get it instead of the axios call + all the validation?

packages/backend/src/open_payments/payment/incoming/model.ts
sabineschaller: I think that would be cleaner.

packages/backend/src/open_payments/client/service.ts
wilsonianb: `validateResponse` isn't included on `deps_`.
Or are you asking why define a response validator that should already exist in a validator middleware?

packages/backend/src/open_payments/client/service.ts
wilsonianb: I initially tried that but didn't like having two code paths, one dealing with `IncomingPayments` model and the other dealing with `IncomingPaymentJSON`, the former of which duplicated a lot of functionality in middleware and incoming payments `routes.ts`, including the solution for #598

packages/backend/src/open_payments/payment/incoming/model.ts
wilsonianb: :+1: I'll change it.
My only reservation is if we don't want to lose record of the `connectionId`

matdehaast (dismissed):

packages/backend/src/open_payments/payment/incoming/model.ts
sabineschaller: Is there any admin purpose we can think of where it would be beneficial to get the ilp credentials after the payment is `completed` or `expired`?

packages/backend/src/open_payments/client/service.ts
sabineschaller: Yes, the latter.

packages/backend/src/open_payments/payment/incoming/model.ts
wilsonianb: Not returning the connection with a completed/expired payment is an Open Payments OpenAPI thing.
It would technically still be possible to generate ilp credentials for the incoming payment since the `connectionId` use only used for the connection url:
https://github.com/interledger/rafiki/pull/595/files#diff-b9b1a593e29f9fdc18a9826d0da73d506ccca8cca9e777165b4ac86b5a0bfdd5R72

packages/backend/src/open_payments/client/service.ts
wilsonianb: :thinking:
Nothing is being overwritten.
Both this and the auth middleware (only for the get incoming payment route) will each create the same response validator once.
However, the auth middleware does not create response validators when running in production:
https://github.com/interledger/rafiki/blob/main/packages/openapi/src/middleware.ts#L11-L12

packages/backend/src/open_payments/payment/incoming/model.ts
sabineschaller: I know it's an OP thing. I just checked the code again and remembered that it is something completely artificial that's not even used to create the STREAM credentials. In that case, set it to `null`.

packages/backend/src/open_payments/client/service.ts
sabineschaller: Never mind, I misread.

sabineschaller (approved):

PR #603, opened 2022-09-15
feat(backend): support Open Payments connection as receiver

### LINKED ISSUE #596: Support quote/outgoing payment to STREAM connection
A quote's specified [`receiver`](https://docs.openpayments.guide/reference/create-quote) may be either:
>The URL of the incoming payment **or** ILP STREAM connection that is being paid.

The backend's quote creation currently returns an error if a resolved payment destination is not an incoming payment:
https://github.com/interledger/rafiki/blob/668e7cb812221daee209b178217f36fedf7bb4cc/packages/backend/src/open_payments/quote/service.ts#L197-L204

### LINKED ISSUE #609: Support non-https payment pointers in local dev
Currently the Open Payments client cannot query local payment pointer (sub-resources)

wilsonianb:

- Support quote/outgoing payment to Open Payments connection
- Include asset details in connection
- Query receiver via http in development mode

## Context

- fixes #596 
- fixes #609 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/client/service.ts
```diff
@@ -1,50 +1,146 @@
...
 }
 
-export async function getIncomingPayment(
+async function getConnection(
   deps: ServiceDependencies,
   url: string
-): Promise<IncomingPaymentJSON | undefined> {
+): Promise<ConnectionJSON | undefined> {
   const requestHeaders = {
     Authorization: `GNAP ${deps.accessToken}`,
```
wilsonianb, line 145: Is this necessary?

sabineschaller: There is no auth for getting a connection.

packages/backend/src/open_payments/client/service.ts
```diff
@@ -68,3 +164,85 @@ export async function getIncomingPayment(
...
+      throw new Error('unreachable')
+    }
+    return data
+  } catch (_) {
+    return undefined
+  }
+}
+
+const CONNECTION_URL_REGEX = /\/connections\/(.){36}$/
+const INCOMING_PAYMENT_URL_REGEX = /\/incoming-payments\/(.){36}$/
```
sabineschaller, line 198: The id doesn't necessarily have to be a uuid so I'd remove the `(.){36}` constraint. Maybe just check that it is alpha-numerical?

packages/backend/src/open_payments/client/service.ts
```diff
@@ -68,3 +164,85 @@ export async function getIncomingPayment(
...
+      incomingAmount.assetScale !== receivedAmount.assetScale
+    ) {
+      return false
+    }
+    if (incomingAmount.value < receivedAmount.value) {
+      return false
+    }
+    if (incomingAmount.value === receivedAmount.value && !payment.completed) {
+      return false
+    }
```
sabineschaller, line 235: Wouldn't that be a server error because it should always complete the payment once the full amount has been received?

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -255,6 +255,6 @@ export type IncomingPaymentJSON = {
   externalRef?: string
   createdAt: string
   updatedAt: string
-  expiresAt: string
+  expiresAt?: string
```
sabineschaller, line 258: I thought we had that as required since it is either set to the client's value or the server's default.

sabineschaller (commented):
It's late and I had a glass of wine because it's my wedding anniversary. If you need another review, I can look at it first thing tomorrow morning.

packages/backend/src/open_payments/client/service.ts
wilsonianb: It's currently enforced as `uuid` in the OpenAPI spec:
https://github.com/interledger/open-payments/blob/6d128e2298301a3a76d2421581e2ffa097ebf33c/open-api-spec.yaml#L629-L636
and this would make it even more so:
https://github.com/interledger/open-payments/pull/174/files#diff-aaff56ee0c5bb1e404fc4e010a7e38034fbc755542d28199e4cf0acb5ef580c9R625

packages/backend/src/open_payments/payment/incoming/model.ts
wilsonianb: That's correct for Rafiki the Open Payments server.
But Rafiki the client of an arbitrary Open Payments server has to go by the OpenAPI spec in which `expiresAt` isn't required:
https://github.com/interledger/open-payments/blob/6d128e2298301a3a76d2421581e2ffa097ebf33c/open-api-spec.yaml#L915-L939

packages/backend/src/open_payments/client/service.ts
wilsonianb: Similar to https://github.com/interledger/rafiki/pull/603#discussion_r974666108
This check is in the context of Rafiki as the client of an arbitrary Open Payments server.
A correctly implemented server shouldn't ever return an incoming payment where `receivedAmount` matches `incomingAmount` but `completed` is `false` (it might return a 500 / server error instead), but the client has to validate the response.

packages/backend/src/open_payments/client/service.ts
sabineschaller: Interesting 🤔. I'm wondering whether we actually want to enforce that in the spec. It seems to me like more of a business decision.

packages/backend/src/open_payments/client/service.ts
```diff
@@ -1,62 +1,179 @@
...
+    if (
+      incomingPayment.expiresAt &&
+      new Date(incomingPayment.expiresAt).getTime() <= Date.now()
+    ) {
+      return undefined
+    }
+    if (typeof incomingPayment.ilpStreamConnection !== 'object') {
+      return undefined
+    }
```
sabineschaller, line 34: nit pick: I'd move that in front of the previous one.

sabineschaller (dismissed):

sabineschaller (approved):

Issue #600, opened 2022-09-13
Add resource set functionality to AS


njlie:
- [ ] Register resource set to AS (https://github.com/interledger/rafiki/issues/601)
- [ ] Use resource set while performing token introspection (https://github.com/interledger/rafiki/issues/556)

TODO: add links to sub-issues

njlie:
Closing as a consequent of the reasoning in https://github.com/interledger/rafiki/issues/576.

> Closing for now due to AS-RS being in a trusted zone

njlie:
Closing as a consequent of the reasoning in https://github.com/interledger/rafiki/issues/576.

> Closing for now due to AS-RS being in a trusted zone

PR #613, opened 2022-09-20
feat(auth): refactor route mounting to be more easily readable

njlie:

- Clearer enumeration of all the routes that are being mounted to the app

## Context

The old pattern of iterating through a list of routes and mounting each item as a route was not great for readability, so they are individually mounted so the different paths and methods can be seen in the code.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -189,83 +189,117 @@ export class App {
     const signatureService = await this.container.use('signatureService')
 
     const openApi = await this.container.use('openApi')
-    const toRouterPath = (path: string): string =>
-      path.replace(/{/g, ':').replace(/}/g, '')
-    const grantMethodToRoute = {
-      [HttpMethod.POST]: 'continue',
-      [HttpMethod.PATCH]: 'update',
-      [HttpMethod.DELETE]: 'cancel'
-    }
-    const tokenMethodToRoute = {
-      [HttpMethod.POST]: 'rotate',
-      [HttpMethod.DELETE]: 'revoke'
-    }
+    /* Back-channel GNAP Routes */
+    // Grant Initiation
+    this.publicRouter.post(
+      '/',
+      createValidatorMiddleware<ContextType<(ctx: AppContext) => Promise<void>>>(openApi, {
+        path: '/',
+        method: HttpMethod.POST
+      }),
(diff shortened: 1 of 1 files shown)


packages/auth/src/app.ts
```diff
@@ -189,83 +189,137 @@ export class App {
...
+      createValidatorMiddleware<
+        ContextType<(ctx: AppContext) => Promise<void>>
+      >(openApi, {
+        path: '/interact/{id}/{nonce}/finish',
+        method: HttpMethod.GET
+      }),
+      grantRoutes.interaction.finish
+    )
+
+    // Grant lookup
```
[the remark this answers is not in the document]
njlie: We could adopt the same behavior we do with the `/introspect` endpoint and have the authentication be optional. If the option to authenticate is set, the IDP could pre-register in the same manner as an RS to associate its public key with some identifier and use that. That probably falls into the scope of https://github.com/interledger/rafiki/issues/600

packages/auth/src/app.ts
```diff
@@ -189,83 +189,117 @@ export class App {
...
-                ctx.status = 200
-              }
-            )
-          }
-        }
-      }
-    }
+    // Grant Continue
+    this.publicRouter.post(
+      '/continue/id',
```
wilsonianb, line 206:
```suggestion
      '/continue/:id',
```

packages/auth/src/app.ts
```diff
@@ -189,83 +189,117 @@ export class App {
...
+      createValidatorMiddleware(openApi, {
+        path: '/continue/{id}',
+        method: HttpMethod.POST
+      }),
+      signatureService.tokenHttpsigMiddleware,
+      grantRoutes.continue
+    )
```
wilsonianb, line 213: The configured (not latest) spec also has `PATCH` and `DELETE` for `/continue/{id}`.
https://github.com/interledger/open-payments/blob/53236cc92a070ec98a7ab31c6ed0f6fc50a98041/auth-server-open-api-spec.yaml#L200-L299
I'm surprised our old magic `for` loop(s) didn't complain :grimacing:

njlie: Interesting...will update to use the latest spec.

wilsonianb (approved):

PR #616, opened 2022-09-20
chore: add build path for auth and backend to docker compose

sabineschaller:

- Builds auth and backend packages instead of using published images
- Both are still being debugged, so running an old-non functioning image defeats the purpose

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/infrastructure/local/docker-compose.yml
+++ b/infrastructure/local/docker-compose.yml
@@ -1,7 +1,9 @@
 version: '3'
 services:
   auth:
-    image: ghcr.io/interledger/rafiki-auth:latest
+    build:
+      context: ../..
+      dockerfile: ./packages/auth/Dockerfile
     restart: always
     networks:
       rafiki:
@@ -34,7 +36,9 @@ services:
     depends_on:
       - backend
   backend:
-    image: ghcr.io/interledger/rafiki-backend:latest
+    build:
+      context: ../..
+      dockerfile: ./packages/backend/Dockerfile
     restart: always
     privileged: true
     ports:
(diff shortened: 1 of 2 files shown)


infrastructure/local/docker-compose.yml
```diff
@@ -1,7 +1,9 @@
 version: '3'
 services:
   auth:
-    image: ghcr.io/interledger/rafiki-auth:latest
```
wilsonianb, line 4: I vote leaving the `image` to give users the option of pulling or building the images (`pnpm localenv up` / `pnpm localenv up --build`)

infrastructure/local/docker-compose.yml
```diff
@@ -1,7 +1,9 @@
...
+    build:
+      context: ../..
+      dockerfile: ./packages/auth/Dockerfile
```
wilsonianb, line 6: This seems to also work
```suggestion
      context: ../../packages/auth
```

wilsonianb: nvm
That worked when pulling, but context does need to be in the repo root when building

infrastructure/local/docker-compose.yml
sabineschaller: Afaik, the `--build` flag only builds those images that actually have the `- build` property in the docker-compose file. Without the `--build`, it first checks if there are already existing images. And with just `-image` property in the docker-compose, it will never build.

wilsonianb: Correct.
I think the docker compose files should have both `build` and `images`.

Issue #576, opened 2022-08-29
Add httpsig to RS token introspection requests


wilsonianb:
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-7.3

https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/service.ts#L47

- might require: #556 
- required for: #574
- related to: #387

wilsonianb:
Since we're assuming the RS and AS are operated by the same account provider, we can make token introspection http signature validation optional. The AS won't publicly expose that endpoint.

matdehaast:
Closing for now due to AS-RS being in a trusted zone

PR #636, opened 2022-09-22
feat(auth): add resource set service

### LINKED ISSUE #601: Add resource set registration endpoint
* `POST /resource`
* RS should identify itself with a key rather than a string or a `kid`
* Return `instance_id` that may be used to identify an RS instead of a key or by reference during token introspection.

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.4

njlie:

- Adds an endpoint which allows an RS to pre-register information that can be used later for token introspection.

## Context

Closes #601.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/resourceSet/service.ts
```diff
@@ -0,0 +1,59 @@
...
+    keyProof: req.key?.proof,
+    keyJwk: req.key?.jwk
+  })
+
+  if (req.access) {
+    await deps.accessService.createAccessForResourceSet(
+      resourceSet.id,
+      req.access
+    )
+  }
```
wilsonianb, line 56: Should these happen within a knex transaction?

packages/auth/src/resourceSet/service.test.ts
```diff
@@ -0,0 +1,79 @@
...
+  afterEach(async (): Promise<void> => {
+    await truncateTables(knex)
+  })
+
+  afterAll(async (): Promise<void> => {
+    await appContainer.shutdown()
+  })
+
+  describe('create', (): void => {
+    test('Can create a resource set', async (): Promise<void> => {
```
wilsonianb, line 44: You could also test without `access` and/or `key`

packages/auth/src/access/service.ts
```diff
@@ -35,22 +40,33 @@ export async function createAccessService({
...
   accessRequests: AccessRequest[],
+  grantId?: string,
+  resourceId?: string,
   trx?: Transaction
 ): Promise<Access[]> {
-  const accessRequestsWithGrant = accessRequests.map((access) => {
-    return { grantId, ...access }
+  if (!grantId && !resourceId) {
+    throw new Error('Missing required properties')
+  }
```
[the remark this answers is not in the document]
njlie: I assume the `createAccess` signature in this snippet not have the `grantId` parameter that's outside the `AccessOptions` parameter?

packages/auth/migrations/20220920224349_add_resource_id_to_accesses_table.js
```diff
@@ -0,0 +1,21 @@
+exports.up = function (knex) {
+  return knex.schema.alterTable('accesses', function (table) {
```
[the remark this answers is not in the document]
njlie: Was initially wary of messing with the dates on the migration files since `create_accesses_table` would have to come after `create_resource_sets_table`, but moving them around like that doesn't seem to be an issue.

packages/auth/src/access/service.ts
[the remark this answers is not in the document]
wilsonianb: yep, fixed

packages/auth/src/resourceSet/routes.ts
```diff
@@ -0,0 +1,41 @@
...
+
+async function createResourceSet(
+  deps: ServiceDependencies,
+  ctx: AppContext
+): Promise<void> {
+  const { body } = ctx.request
+  const resourceSet = await deps.resourceSetService.create(body)
+
+  ctx.body = {
+    resource_reference: resourceSet.id
```
wilsonianb, line 39: It sounds like we want to return `instance_id`
- https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.4
- #601

wilsonianb: Could this also return `introspection_endpoint` to save the RS from having to configure:
https://github.com/interledger/rafiki/blob/c2a7805500d944928befbe71449e6a7d5a16ec3f/packages/backend/src/config/app.ts#L78-L81
If so, should that only happen if `token_introspection_required` is included in the request?

packages/auth/src/resourceSet/service.ts
```diff
@@ -0,0 +1,67 @@
...
+import { BaseService } from '../shared/baseService'
+import { AccessService } from '../access/service'
+import { ResourceSet } from './model'
+
+interface ServiceDependencies extends BaseService {
+  accessService: AccessService
+}
+
+interface ResourceSetRequest {
+  access?: AccessRequest[]
```
wilsonianb, line 13: `access` is technically `REQUIRED` in the resource set registration request
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.4

packages/auth/src/resourceSet/routes.ts
```diff
@@ -0,0 +1,41 @@
...
+    create: (ctx: AppContext) => createResourceSet(deps, ctx)
+  }
+}
+
+async function createResourceSet(
+  deps: ServiceDependencies,
+  ctx: AppContext
+): Promise<void> {
+  const { body } = ctx.request
+  const resourceSet = await deps.resourceSetService.create(body)
```
wilsonianb, line 36: I'm confused about where the `key` comes from
>The RS MUST identify itself with its own key and sign the request.

but the immediate subsequent example doesn't include the `key` in the request body.
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.4

packages/auth/src/resourceSet/service.ts
```diff
@@ -0,0 +1,67 @@
...
+        trx
+      )
+    }
+
+    await trx.commit()
+    return resourceSet
+  } catch (err) {
+    await trx.rollback()
+    throw err
+  }
```
wilsonianb, line 66:
```suggestion
  return await ResourceSet.transaction(async trx => {
    const resourceSet = await ResourceSet.query(trx).insertAndFetch({
      keyProof: req.key?.proof,
      keyJwk: req.key?.jwk
    })

    if (req.access) {
      await deps.accessService.createAccessForResourceSet(
        resourceSet.id,
        req.access,
        trx
      )
    }

    return resourceSet
  })
```

wilsonianb (commented):
Continuing the discussion from https://github.com/interledger/rafiki/issues/601

I'm not convinced we should be using resource set registration for the RS / key registration.
I'm not sure what it'd mean for an RS to register an `access` with a subset of the Open Payments API.

It seems like we just want to accomplish:
>   The AS MAY require an RS to pre-register its keys or could allow
   calls from arbitrary keys in a trust-on-first-use model.

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.2

If so, I wonder if we can either have a dedicated key registration endpoint (not defined in the GNAP RS/AS spec) or allow the AS to be configured with the RS `resource_server`(`instance_id`) and `key`. (Maybe the RS hosts its own `jwks.json` a la https://github.com/interledger/open-payments/pull/190 ?)

njlie:
Closing as a consequent of the reasoning in https://github.com/interledger/rafiki/issues/576.

> Closing for now due to AS-RS being in a trusted zone

PR #638, opened 2022-09-26
feat(backend): provision private key

### LINKED ISSUE #606: Self-provisioned keys
When a Rafiki instance calls out to other open payment endpoints, it needs to provide a key for getting a grant. This is required to identify itself. Hence, the Rafiki instance needs to provision its own key(s).

sabineschaller:

- load private key or create one
- part of backend `config`

## Context

- fixes #606 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


wilsonianb (dismissed):

packages/backend/src/config/app.ts
```diff
@@ -137,3 +138,21 @@ function parseRedisTlsConfig(
...
+      if (jwk.crv === 'Ed25519') {
+        return key
+      } else {
+        console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+      }
+    } catch {
+      console.log('Private key could not be loaded. Generating new key.')
+    }
+  }
+  const keypair = crypto.generateKeyPairSync('ed25519')
```
wilsonianb, line 156: Actually...
should this be storing the generated key to at least a tmp file?

packages/backend/src/config/app.ts
```diff
@@ -137,3 +138,21 @@ function parseRedisTlsConfig(
 
   return Object.keys(options).length > 0 ? options : undefined
 }
+
+function parseOrProvisionKey(keyPath: string): crypto.KeyObject {
+  if (keyPath !== '') {
+    try {
+      const key = crypto.createPrivateKey(readFileSync(keyPath))
```
wilsonianb, line 145: Should this include support for reading an encrypted key file?

packages/backend/src/config/app.ts
sabineschaller: I was wondering something similar and wanted to make a comment but then had to run. 
I was wondering whether we should store the keys somewhere when Rafiki "goes down" for whatever reason. Is that what you want to use the tmp file for?
Also, where would you like to put the tmp file?

packages/backend/src/config/app.ts
sabineschaller: Yes, probably. I'll include that

packages/backend/src/config/app.ts
wilsonianb: I don't have a location in mind, but maybe whatever we choose could be the default value for `PRIVATE_KEY_PATH`

packages/backend/src/config/app.ts
sabineschaller: Now that I thought more about it, I'm not so sure. We would also have to specify which encryption mechanism to use. And those key files are usually not encrypted. 🤔

wilsonianb: I'm fine leaving it out for now.

sabineschaller: #642

sabineschaller:
Should this PR also add the key to the internal key registry?

packages/backend/src/config/app.ts
```diff
@@ -123,17 +126,38 @@ function parseRedisTlsConfig(
...
+    if (jwk.crv === 'Ed25519') {
+      return key
+    } else {
+      console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+    }
+  } catch {
+    console.log('Private key could not be loaded. Generating new key.')
+  }
+  const keypair = crypto.generateKeyPairSync('ed25519')
+  fs.mkdirSync('../../tmp')
```
wilsonianb, line 157: This fails if the directory already exists

packages/backend/src/config/app.ts
```diff
@@ -109,6 +109,9 @@ export const Config = {
     'AUTH_SERVER_SPEC',
     'https://raw.githubusercontent.com/interledger/open-payments/77462cd0872be8d0fa487a4b233defe2897a7ee4/auth-server-open-api-spec.yaml'
   ),
+  privateKey: parseOrProvisionKey(
+    envString('PRIVATE_KEY_PATH', '../../tmp/private.pem')
```
wilsonianb, line 113:
```suggestion
    envString('PRIVATE_KEY_PATH', './tmp/private.pem')
```
If the `backend` is run from a working directory like `/root` the target directory isn't possible.

packages/backend/src/config/app.ts
```diff
@@ -123,17 +126,38 @@ function parseRedisTlsConfig(
...
+    } else {
+      console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+    }
+  } catch {
+    console.log('Private key could not be loaded. Generating new key.')
+  }
+  const keypair = crypto.generateKeyPairSync('ed25519')
+  fs.mkdirSync('../../tmp')
+  fs.writeFileSync(
+    '../../tmp/private.pem',
```
wilsonianb, line 159: Maybe make a const string used both here and in the `privateKey` assignment above

sabineschaller: Which `privateKey` assignment?

wilsonianb: https://github.com/interledger/rafiki/pull/638/files#diff-6300c4703a2fb930745656aed4f5d68507eb07a20ae1f817a1ca3c9d90901facR112

packages/backend/src/config/app.ts
```diff
@@ -123,17 +126,40 @@ function parseRedisTlsConfig(
...
 
   if (keyFile !== '') {
-    options.key = readFileSync(keyFile)
+    options.key = fs.readFileSync(keyFile)
   }
 
   return Object.keys(options).length > 0 ? options : undefined
 }
+
+function parseOrProvisionKey(keyPath: string): crypto.KeyObject {
```
wilsonianb, line 144: Is it worth adding a `config.test.ts` for this?

packages/backend/src/config/app.ts
wilsonianb: maybe something like
```
const TMP_DIR = './tmp'
const PRIVATE_KEY_FILE = `${TMP_DIR}/private-key.pem`

File: packages/backend/src/config/app.ts [hunk 7]
```diff
@@ -123,17 +126,40 @@ function parseRedisTlsConfig(
...
+      console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+    }
+  } catch {
+    console.log('Private key could not be loaded. Generating new key.')
+  }
+  const keypair = crypto.generateKeyPairSync('ed25519')
+  if (!fs.existsSync('./tmp')) {
+    fs.mkdirSync('./tmp')
+  }
+  fs.writeFileSync(
```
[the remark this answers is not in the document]
sabineschaller: I was wondering that, too. But I don't see a reason to keep the old ones as long as the new one is also associated with this RS. And by associating I mean adding it to the internal key registry. Should this PR also do that?

packages/backend/src/config/app.ts
sabineschaller: Probably yes 👍

packages/backend/src/config/app.ts
sabineschaller: Oh, that far above 😄

packages/backend/src/config/app.ts
```diff
@@ -123,17 +129,41 @@ function parseRedisTlsConfig(
...
+    const jwk = key.export({ format: 'jwk' })
+    if (jwk.crv === 'Ed25519') {
+      return key
+    } else {
+      console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+    }
+  } catch {
+    console.log('Private key could not be loaded. Generating new key.')
+  }
+  const keypair = crypto.generateKeyPairSync('ed25519')
```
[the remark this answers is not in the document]
sabineschaller: Well, we'll need the key itself for the signature, won't we? But I agree, we should also provide the jwk.

packages/backend/src/config/app.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    afterEach(async (): Promise<void> => {
+      fs.rmSync(TMP_DIR, { recursive: true, force: true })
+    })
+
+    test('can provision key', async (): Promise<void> => {
+      expect(fs.existsSync(TMP_DIR)).toBe(false)
+      expect(fs.existsSync(PRIVATE_KEY_FILE)).toBe(false)
+      const { key, jwk } = parseOrProvisionKey(PRIVATE_KEY_FILE)
+      expect(key).toBeInstanceOf(crypto.KeyObject)
+      expect(jwk).toMatchObject({
```
wilsonianb, line 24:
```suggestion
```suggestion
      expect(key).toMatchObject({
        asymmetricKeyType: 'ed25519',
        type: 'private'
      })
      expect(jwk).toEqual(key.export({ format: 'jwk' }))
      expect(jwk).toEqual({
```

packages/backend/src/config/app.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        crv: 'Ed25519',
+        kty: 'OKP',
+        d: expect.any(String),
+        x: expect.any(String)
+      })
+      expect(key.export({ format: 'pem', type: 'pkcs8' })).toEqual(
+        keypair.privateKey.export({ format: 'pem', type: 'pkcs8' })
+      )
```
wilsonianb, line 52: https://nodejs.org/api/crypto.html#keyobjectequalsotherkeyobject
```suggestion
      assert.ok(key.equals(keypair.privateKey))
      expect(jwk).toEqual(key.export({ format: 'jwk' }))
      expect(jwk).toEqual({
        crv: 'Ed25519',
        kty: 'OKP',
        d: expect.any(String),
        x: expect.any(String)
      })
```

packages/backend/src/config/app.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      expect(fs.existsSync(PRIVATE_KEY_FILE)).toBe(false)
+      const { key, jwk } = parseOrProvisionKey(PRIVATE_KEY_FILE)
+      expect(key).toBeInstanceOf(crypto.KeyObject)
+      expect(jwk).toMatchObject({
+        crv: 'Ed25519',
+        kty: 'OKP',
+        d: expect.any(String),
+        x: expect.any(String)
+      })
+      expect(fs.existsSync(PRIVATE_KEY_FILE)).toBe(true)
```
wilsonianb, line 30:
```suggestion
      expect(fs.readFileSync(PRIVATE_KEY_FILE, 'utf8')).toEqual(
        key.export({ format: 'pem', type: 'pkcs8' })
      )
```

packages/backend/src/config/app.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    beforeEach(async (): Promise<void> => {
+      fs.rmSync(TMP_DIR, { recursive: true, force: true })
+    })
+
+    afterEach(async (): Promise<void> => {
+      fs.rmSync(TMP_DIR, { recursive: true, force: true })
+    })
+
+    test('can provision key', async (): Promise<void> => {
+      expect(fs.existsSync(TMP_DIR)).toBe(false)
```
wilsonianb, line 20: Test this with or without preexisting `TMP_DIR`

packages/backend/src/config/app.ts
```diff
@@ -123,17 +129,47 @@ function parseRedisTlsConfig(
...
+  if (!fs.existsSync(TMP_DIR)) {
+    fs.mkdirSync(TMP_DIR)
+  }
+  fs.writeFileSync(
+    PRIVATE_KEY_FILE,
```
wilsonianb, line 168: Now I'm wondering if `parseOrProvisionKey` should only operate with `keyPath` as opposed to `TMP_DIR`/`PRIVATE_KEY_FILE`, to prevent an operator from not realizing that the default `PRIVATE_KEY_PATH` is being used and not their configured one (even though that is being logged).

packages/backend/src/config/app.ts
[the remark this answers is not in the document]
wilsonianb: I'm leaning towards not overwriting.
I don't like the idea of potentially overwriting a file not intended for for Rafiki, however unlikely such a misconfiguration may be.

packages/backend/src/config/app.ts
```diff
@@ -123,17 +129,47 @@ function parseRedisTlsConfig(
...
+// exported for testing
+export function parseOrProvisionKey(keyPath: string): {
+  key: crypto.KeyObject
+  jwk: crypto.JsonWebKey
+} {
+  try {
+    const key = crypto.createPrivateKey(fs.readFileSync(keyPath))
+    const jwk = key.export({ format: 'jwk' })
+    if (jwk.crv === 'Ed25519') {
+      return { key, jwk }
```
wilsonianb, line 156: Given that the `jwk` isn't actually being included in the config and I think`parseOrProvisionKey` doens't actually need to generate the `jwk`, I vote we drop it.
(I'm trying to see how we'd generate an ed25519 private `KeyObject` from just `randomBytes` below)

```suggestion
export function parseOrProvisionKey(keyPath: string): crypto.KeyObject {
  try {
    const key = crypto.createPrivateKey(fs.readFileSync(keyPath))
    if (key.asymmetricKeyType === 'ed25519') {
      return key
```

packages/backend/src/config/app.ts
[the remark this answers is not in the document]
wilsonianb: >And by associating I mean adding it to the internal key registry. Should this PR also do that?

I think we should hold off until we decide if we're going to switch the key registry to using JWKS
- https://github.com/interledger/open-payments/pull/190

packages/backend/src/config/app.test.ts
sabineschaller: `Property 'equals' does not exist on type 'KeyObject'.`

packages/backend/src/config/app.ts
sabineschaller: I'm not sure I understand. `PRIVATE_KEY_FILE` does include the path.

packages/backend/src/config/app.ts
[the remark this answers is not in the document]
sabineschaller: > I'm leaning towards not overwriting.
I don't like the idea of potentially overwriting a file not intended for for Rafiki, however unlikely such a misconfiguration may be.

Should we use unique tmp dirs?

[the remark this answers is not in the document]
sabineschaller: > I think we should hold off until we decide if we're going to switch the key registry to using JWKS

I'll create an issue to also create a payment pointer for the Rafiki instance itself that will then be linked to the key, in whatever way and form we end up doing that (probably JWKS endpoint).

packages/backend/.gitignore
```diff
@@ -3,3 +3,4 @@ build
 /dist
 .env
 src/coverage
+tmp
```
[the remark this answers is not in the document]
wilsonianb: *re-add`

packages/backend/src/config/app.test.ts
wilsonianb: :confused: 
https://nodejs.org/dist/latest-v16.x/docs/api/crypto.html#keyobjectequalsotherkeyobject
Is that a typescript build error?

packages/backend/src/config/app.ts
wilsonianb: As in `parseOrProvisionKey` would only read or write (maybe not overwrite) the file at `keyPath`.
As is, it reads from `keyPath` but writes to `PRIVATE_KEY_FILE`.

packages/backend/src/config/app.test.ts
sabineschaller: yes
```
src/config/app.test.ts:59:21 - error TS2339: Property 'equals' does not exist on type 'KeyObject'.
59       assert.ok(key.equals(keypair.privateKey))
                       ~~~~~~
Found 1 error.
```
I even made sure to use node v16.18.0.

packages/backend/src/config/app.ts
sabineschaller: But what if it only has read access and not write access at `keyPath`?

wilsonianb: I'm imagining an operating running multiple instances of rafiki `backend` in a k8s cluster and configuring `PRIVATE_KEY_FILE` as a path to a secret.
In which case, I think we should throw if we can't use (or write to) the configured file (`keyPath`).
Otherwise you end up with individual instances provisioning their own private keys.

mkurapov (approved):
I was able to test successfully, both with existing and new, generated key. 

As you two are discussing in the comment, similarly, I do see the benefit of maybe having the `parseOrProvisionKey` take `undefined` as a possible argument, and doing the `keyPath` generation inside the method itself. This way if a defined `keyPath` _is_ given, we know its's coming from a user-set config, not a default value, in which case we will throw as @wilsonianb mentions

packages/backend/src/config/app.ts
```diff
@@ -145,17 +143,22 @@ function parseRedisTlsConfig(
...
-    } else {
-      console.log('Private key is not EdDSA-Ed25519 key. Generating new key.')
+export function parseOrProvisionKey(
+  keyFile: string | undefined
+): crypto.KeyObject {
+  if (keyFile) {
+    try {
+      const key = crypto.createPrivateKey(fs.readFileSync(keyFile))
+      const jwk = key.export({ format: 'jwk' })
+      if (jwk.crv === 'Ed25519') {
```
mkurapov, line 153: It would be nice to have `'Ed25519'` coming from a type declaration, but it looks like we are limited to just a string here.

mkurapov (approved):

packages/backend/src/config/app.ts
wilsonianb: Maybe we could eventually do schema validation using:
https://github.com/interledger/open-payments/blob/b125873edb6d33acd40eb304e026e5111ec1dba9/openapi/shared/schemas.yaml#L201-L234

wilsonianb:
Merging/rebasing main _should_ resolve the failing Docker checks

PR #643, opened 2022-09-27
refactor(auth): refactor httpsig into more specialized per-route middlewares

### LINKED ISSUE #641: Break out httpsig middleware into separate middlewares
Because routes are now mounted explicitly rather than iteratively, we can use more specific middlewares for each of the routes. There should be separate middlewares for:
* Validating an httpsig signature using a key provided in the request body
* Validating an httpsig signature using a key that is bound to a grant (via an access token in the request body)
  * via an access token in the request body (for grant initiation)
  * via an interaction id in the request body (for grant continuation)
  * via a management id in the request body (for token management)

njlie:

- Breaks down the middleware for parsing httpsig headers into separate smaller ones for each route, using their particular methods for acquiring a key.

## Context

Fixes #641.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/signature/middleware.test.ts
+++ b/packages/auth/src/signature/middleware.test.ts
@@ -10,7 +10,6 @@ import { Config } from '../config/app'
 import { IocContract } from '@adonisjs/fold'
 import { initIocContainer } from '../'
 import { AppServices } from '../app'
-import { SignatureService } from './service'
 import { ClientKey, JWKWithRequired } from '../client/service'
 import { createContext, createContextWithSigHeaders } from '../tests/context'
 import {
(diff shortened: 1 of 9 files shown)


packages/auth/src/signature/service.ts
```diff
@@ -124,6 +159,7 @@ async function verifySigFromBoundKey(
   return verifySigAndChallenge(deps, sig, sigInput, jwk, ctx)
 }
 
+// TODO: Replace with public httpsig library
 function sigInputToChallenge(sigInput: string, ctx: AppContext): string | null {
```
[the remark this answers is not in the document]
njlie: I think it might be better to break this up into smaller functions like `getSigInputComponents`, `validateSigInputComponents`, and `generateChallengeFromSigInput`. Basically, `getSigInputComponents` is going to have to be called in both `validateSigInputComponents` (which would hold the code you're suggesting be moved) and `generateChallengeFromSigInput`.

packages/auth/src/tests/signature.ts
```diff
@@ -58,29 +58,30 @@ export async function generateSigHeaders(
   privateKey: JWKWithRequired,
   url: string,
   method: string,
-  body?: unknown
+  body?: unknown,
+  authorization?: string
```
wilsonianb, line 62: This argument list could be converted to an object

wilsonianb (dismissed):
bump
https://github.com/interledger/rafiki/pull/643#discussion_r985901638

packages/auth/src/signature/middleware.test.ts
```diff
@@ -375,17 +328,16 @@ describe('Signature Service', (): void => {
...
       ctx.request.body = {
         access_token: token.value,
         proof: 'httpsig',
         resource_server: 'test'
       }
-      await signatureService.tokenHttpsigMiddleware(ctx, next)
-
-      expect(next).toHaveBeenCalled()
+      await tokenHttpsigMiddleware(ctx, next)
       expect(ctx.response.status).toEqual(400)
```
wilsonianb, line 341: This could check the message, and tests could be added for the other possible errors

wilsonianb (approved):

wilsonianb (approved):

PR #645, opened 2022-09-28
feat(backend): validate httpsig on rs requests

### LINKED ISSUE #585: Validate httpsig on RS requests
The backend should use the `key` returned in the token introspection response to validate the httpsig on the request from the client.
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.3

dclipp:

- Add local `http-message-signatures` and `structured-headers` packages
- validate the signature received after and introspection request

## Context

- fixes #585 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/package.json
```diff
@@ -66,6 +66,7 @@
     "graphql": "^16.6.0",
     "graphql-scalars": "^1.18.0",
     "graphql-tools": "^8.3.3",
+    "http-message-signatures": "link:../http-message-signatures",
```
wilsonianb, line 69:
```suggestion
    "http-message-signatures": "workspace:../http-message-signatures",
```

packages/backend/src/open_payments/auth/service.ts
```diff
@@ -72,6 +84,53 @@ async function introspectToken(
       }
     )
 
+    const keyType: KeyType = 'public' //dc_TODO
+    const extractable = false //dc_TODO 
+    const typedRequest = requestLike(request)
+    const signatures = httpis.parseSignatures(typedRequest)
```
wilsonianb, line 90: I think this is verifying the backend's token introspection request to the AS, whereas I think we want to validate `ctx.request` that the backend is currently handling.
So this may need to move to the auth middleware, somewhere after the call to `authService.introspect`
https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/middleware.ts#L49

wilsonianb (commented):
Should `http-message-signatures` and `structured-headers` be added to any GitHub Actions workflows?
https://github.com/interledger/rafiki/blob/main/.github/workflows/lint_test_build.yml

sabineschaller:
Why are we adding the `http-message-signatures` and `structured-headers` packages locally and not adding them as a dependency?

sabineschaller:
I see. Maybe that was proposed when I had sound issues... 
It's just very hard to review that way and we could still add the github repo as the dependency. What do you guys think?

wilsonianb:
If we expect regular updates to those dependencies, I'd vote for having them in the monorepo.
Otherwise, I don't have a strong opinion.

packages/backend/package.json
wilsonianb: I think I led you astray.
It looks like this is the way to add it as a workspace:
```
pnpm --filter backend add -P http-message-signatures --workspace
```
which will also update `pnpm-lock.yaml` (which still has `link`) and result in:
```
    "http-message-signatures": "workspace:^0.1.2",
```
because `http-message-signatures`'s `package.json` has a `version`

dclipp:
Adrian's version also has some changes that aren't in the main `http-message-signatures` repo

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,70 @@ import { AccessType, AccessAction } from './grant'
...
+              usages: parseJwkUsages(clientKeys.jwk)
+            }))
+          }
+        })
+
+        if (!success) {
+          throw new Error('signature is not valid')
+        }
+      }
+    }
```
wilsonianb, line 67: `pnpm format` needs to be run.
However, you may want to add `http-message-signatures` to the prettier and eslint ignore files for now (assuming it stays in the monorepo).

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -37,6 +118,17 @@ export function createAuthMiddleware({
       if (!grant || !grant.active) {
         ctx.throw(401, 'Invalid Token')
       }
+      if (!config.skipSignatureVerification) {
+        try {
+          const clientKeysService = await ctx.container.use('clientKeysService')
+          const clientKeys = await clientKeysService.getKeyByClientId(
+            grant.clientId
+          )
```
wilsonianb, line 126: It looks like the token introspection response will include the client's `key`
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.3
https://github.com/interledger/rafiki/blob/3d30ce755b4218a777ec77f408be5d15dfa0e687/packages/auth/src/accessToken/routes.ts#L63
You'll need to update `authService.introspect` to also return the `key` from the response.
You could add `key` to the `Grant` class, but that'd require changing a lot of files (and conflicting with #647 ) and I don't think we actually want to be passing around the key everywhere.
:thinking: Maybe just add a:
```
interface TokenInfo extends GrantOptions {
  key: KeyInfo // import { KeyInfo } from 'auth'?
}
```
for `authService.introspect` return type.
Then use it to construct the `new Grant` in the auth middleware.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -37,6 +118,17 @@ export function createAuthMiddleware({
       if (!grant || !grant.active) {
         ctx.throw(401, 'Invalid Token')
       }
+      if (!config.skipSignatureVerification) {
```
wilsonianb, line 121: auth middleware tests should be updated to test signature verfication:
https://github.com/interledger/rafiki/blob/3d30ce755b4218a777ec77f408be5d15dfa0e687/packages/backend/src/open_payments/auth/middleware.test.ts#L76
(which would require making `skipSignatureVerification` false just for that test file ?)

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,87 @@ import { AccessType, AccessAction } from './grant'
...
+  } else if (use === undefined) {
+    return []
+  } else {
+    throw new Error('invalid usage')
+  }
+}
+
+async function verifyRequest(
+  request: KoaRequest,
+  clientKeys: ClientKeys
```
wilsonianb, line 50:
```suggestion
  jwk: JWKWithRequired
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,87 @@ import { AccessType, AccessAction } from './grant'
...
+    use === 'sign' ||
+    use === 'unwrapKey' ||
+    use === 'verify' ||
+    use === 'wrapKey'
+  ) {
+    return [use]
+  } else if (use === undefined) {
+    return []
+  } else {
+    throw new Error('invalid usage')
```
wilsonianb, line 44: :confused:  The OpenAPI spec appears to only allow `sig`
https://github.com/interledger/open-payments/blob/4c873dba89164decffbe84905d12f1d4ec045389/auth-server-open-api-spec.yaml#L677-L680

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,87 @@ import { AccessType, AccessAction } from './grant'
...
+    headers: request.headers,
+    url: request.url
+  }
+}
+
+function parseJwkKeyType(jwk: JWKWithRequired): KeyType {
+  if (jwk.kty === 'private' || jwk.kty === 'public' || jwk.kty === 'secret') {
+    return jwk.kty
+  } else {
+    throw new Error('invalid key type')
```
wilsonianb, line 24: The OpenAPI spec only allows `OKP`
https://github.com/interledger/open-payments/blob/s2-split-auth-spec/openapi-specs/shared/schemas.yaml#L188

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,87 @@ import { AccessType, AccessAction } from './grant'
...
+): Promise<void> {
+  const keyType = parseJwkKeyType(clientKeys.jwk)
+  const typedRequest = requestLike(request)
+  const signatures = httpis.parseSignatures(typedRequest)
+  for (const [, { keyid, alg }] of signatures) {
+    if (!keyid) {
+      throw new Error(`The signature input is missing the 'keyid' parameter`)
+    } else if (alg !== 'ed25519') {
+      throw new Error(
+        `The signature parameter 'alg' is using an illegal value '${alg}'. Only 'ed25519' is supported.`
```
wilsonianb, line 60: OpenAPI says `EdDSA`
https://github.com/interledger/open-payments/blob/4c873dba89164decffbe84905d12f1d4ec045389/auth-server-open-api-spec.yaml#L664-L668

packages/backend/src/open_payments/auth/middleware.ts
dclipp: I had to do 
```
class TokenInfo extends Grant {
    public readonly key: KeyInfo
    ...
}
```

because `Grant` is a class

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,105 @@ import { AccessType, AccessAction } from './grant'
...
+    throw new Error('signature is missing')
+  }
+
+  if (!koaRequest.headers['signature-input']) {
+    throw new Error('signature-input is missing')
+  }
+
+  const signatureInputMap = new Map<string, string>()
+  signatureInputMap.set('keyid', kid)
+  signatureInputMap.set('alg', parseAlgorithmName(jwk.alg))
```
wilsonianb, line 53: `signatureInputMap` isn't used anywhere.
Was it intended to be added to `typedRequest`?

re: `alg`
I'm trying to decide if we should either:
- update our `http-message-signatures` to not expect `alg` in the `Signature-Input` header given that:
>   If the signing algorithm is a JOSE signing algorithm from the JSON
   Web Signature and Encryption Algorithms Registry established by
   [RFC7518], the JWS algorithm definition determines the signature and
   hashing algorithms to apply for both signing and verification.  There
   is no use of the explicit "alg" signature parameter when using JOSE
   signing algorithms.

https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-06#section-3.3.5
- not use `http-message-signatures` for now, and just `export`/`import` the AS's [httpsig verification](https://github.com/interledger/rafiki/blob/6cb56141013d38197b13b31e392ebbee402adcd6/packages/auth/src/signature/service.ts)

wilsonianb (commented):
I'm starting to lean towards not using `http-message-signatures` for now, and instead import the existing httpsig verification from the `auth` package:
- the verification functionality added to the fork expects `alg` in the `Signature-Input` header
- the fork has already drifted (`verifyContentDigest` has been removed) from what's been imported here
- the open question of adding it to the Rafiki monorepo: https://github.com/interledger/rafiki/pull/645#issuecomment-1262414368

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,105 @@ import { AccessType, AccessAction } from './grant'
...
+              [new ByteSequence(propValue[0].toString()), new Map()]
+            ] as [string, Item | InnerList]
+          })
+        ])
+      )
+    },
+    url: koaRequest.url
+  }
+
+  const verifier = createVerifier('ecdsa-p256-sha256', key)
```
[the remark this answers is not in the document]
dclipp: Because of this change, I was able to remove the `TokenInfo` changes and just keep `introspect` returning a Grant

packages/backend/src/open_payments/auth/middleware.ts
dclipp: For the purposes of this PR, I just did the first option and removed the check for `alg` in the `Signature-Input` header

packages/backend/src/open_payments/auth/middleware.ts
dclipp: Tests have been added

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -38,6 +41,51 @@ describe('Auth Middleware', (): void => {
...
+    const authService = await deps.use('authService')
+    jest.spyOn(authService, 'introspect').mockImplementation(async () => {
+      return new TokenInfo(grant, {
+        jwk: {
+          alg: (jwk && jwk.alg) || 'ed25519',
+          kty: (jwk && jwk.kty) || 'OKP',
+          crv: (jwk && jwk.crv) || 'Ed25519',
+          x: (jwk && jwk.x) || '051208da-f6b6-4ed0-b49b-8b0043900eee',
+          kid: (jwk && jwk.kid) || 'gnap-key'
+        }
```
wilsonianb, line 84:
```suggestion
        },
        proof: 'httpsig'
```

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -266,4 +316,145 @@ describe('Auth Middleware', (): void => {
...
+    })
+    await grantReferenceService.create({
+      id: grant.grant,
+      clientId: grant.clientId
+    })
+    const scope = mockAuthServer(grant.toJSON())
+    await mockIntrospect(grant)
+    await expect(middleware(ctx, next)).resolves.not.toThrow()
+    expect(next).toHaveBeenCalled()
+    scope.isDone()
```
wilsonianb, line 347:
```suggestion
    scope.done()
```
- https://github.com/interledger/rafiki/issues/656

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -266,4 +316,145 @@ describe('Auth Middleware', (): void => {
...
+          identifier: ctx.paymentPointer.url
+        }
+      ]
+    })
+    await grantReferenceService.create({
+      id: grant.grant,
+      clientId: grant.clientId
+    })
+    const scope = mockAuthServer(grant.toJSON())
+    await mockIntrospect(grant)
```
wilsonianb, line 344: We should use one or the other.
Since there currently aren't any `AuthService` tests, I vote using `mockAuthServer` but updating it to also return the key.

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -266,4 +316,145 @@ describe('Auth Middleware', (): void => {
     expect(ctx.grant).toEqual(grant)
     scope.isDone()
   })
+
+  test('returns 200 with valid http signature', async (): Promise<void> => {
+    Config.skipSignatureVerification = false
```
wilsonianb, line 321: I don't think this will suffice.
I think the auth middleware will use the value from when when we call:
```
    deps = await initIocContainer(Config)
    appContainer = await createTestApp(deps)
```
in the `beforeAll`

dclipp: I removed the `skipSignatureVerification` option entirely since existing tests should still pass

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -2,6 +2,56 @@ import { AccessType, AccessAction } from './grant'
...
+      )
+    } else {
+      const success = await httpis.verify(koaRequest, {
+        format: 'httpbis',
+        verifiers: {
+          [kid]: verifier
+        }
+      })
+
+      if (!success || !verifyContentDigest(koaRequest)) {
```
wilsonianb, line 49: Only call `verifyContentDigest` if the request has a body.

packages/http-message-signatures/src/digest/index.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (!value.startsWith(':') || !value.endsWith(':')) {
+        throw new Error('Error parsing digest value')
+      }
+      const digest = value.substring(1, value.length - 1)
+      const hash = createHash(nodeAlgo(key.trim()))
+        .update((request as any).body || '')
+        .digest('base64')
+      return digest === hash
+    })
+    .every((isValid) => isValid)
```
wilsonianb, line 68: I couldn't understand how this was returning true without a body (`digest` did not equal `hash`).
The `.map` is actually returning promises, and `Promise { false }` evaluates to `true`-ish...

packages/backend/src/clientKeys/service.ts
```diff
@@ -5,6 +5,7 @@ import { BaseService } from '../shared/baseService'
 
 export interface ClientKeysService {
   getKeyById(keyId: string): Promise<ClientKeys>
+  getKeyByClientId(clientId: string): Promise<ClientKeys>
```
wilsonianb, line 8: I don't think this is needed anymore

packages/http-message-signatures/src/digest/index.ts
wilsonianb: This has me feeling more strongly about not using `http-message-signatures` for now (https://github.com/interledger/rafiki/pull/645#pullrequestreview-1130574554).
I vote we use the `auth` package's existing httpsig verification and save content-digest verification for a separate pr:
- https://github.com/interledger/rafiki/issues/655

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -264,6 +311,205 @@ describe('Auth Middleware', (): void => {
     await expect(middleware(ctx, next)).resolves.toBeUndefined()
     expect(next).toHaveBeenCalled()
     expect(ctx.grant).toEqual(grant)
-    scope.isDone()
+    scope.done()
+  })
+
+  test('returns 200 with valid http signature', async (): Promise<void> => {
```
wilsonianb, line 317: Add a test for `'returns 401 for invalid http signature'`

packages/http-message-signatures/src/httpbis/index.ts
```diff
@@ -0,0 +1,338 @@
...
+
+  signatures.forEach(
+    ({ value, components, signatureParams, keyid, alg }, signatureName) => {
+      if (!keyid) {
+        return false
+      }
+
+      const verifier = opts.verifiers[keyid]
+      if (!verifier || verifier.alg !== alg) {
+        return false
```
wilsonianb, line 322: `verify` doesn't appear to work as expected.
We're hitting this in our `'returns 200 with valid http signature'` test, but `verify` is returning true.

wilsonianb (commented):
`http-message-signatures` is striking out for me.
I vote we drop it and use the existing httpsig verification in the `auth` package.

packages/http-message-signatures/src/digest/index.ts
```diff
@@ -54,14 +67,25 @@ export function verifyContentDigest(request: RequestLike) {
...
   return digests
-    .map(async (digestHeader) => {
-      const [key, value] = digestHeader.split('=')
-      if (!value.startsWith(':') || !value.endsWith(':')) {
+    .map((digestHeader) => {
+      /**
+       * the expected format is:
+       *    <optional whitespace><key><optional whitespace>=<optional whitespace><value><optional whitespace>
+       * where
+       *    <optional whitespace> contains only spaces and tabs or no characters at all
```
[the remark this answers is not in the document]
dclipp: Right, just to be robust enough to handle possible input

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -264,6 +311,205 @@ describe('Auth Middleware', (): void => {
...
+    mockKeyInfo.jwk.x = '051208da-f6b6-4ed0-b49b-8b004390FFFF'
```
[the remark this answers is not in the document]
dclipp: `mockKeyInfo` is already setting the `x` field so I just removed this one-off instance

packages/http-message-signatures/src/httpbis/index.ts
dclipp: Isn't it supposed to return true if the signature is valid?

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -264,6 +311,205 @@ describe('Auth Middleware', (): void => {
...
+            identifier: ctx.paymentPointer.url
+          }
+        ]
+      },
+      mockKeyInfo
+    )
+    await grantReferenceService.create({
+      id: grant.grant,
+      clientId: grant.clientId
+    })
```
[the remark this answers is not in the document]
dclipp: This is also being called in the other signature tests

PR #647, opened 2022-09-29
chore(backend): match subresources by payment pointer

### LINKED ISSUE #598: Validate payment pointer in Open Payments sub-resource requests
Check that a queried sub-resource belongs to the payment pointer of the url, under which grant `location` the request was authorized.

The paymentPointer is already being added to [request ctx](https://github.com/interledger/rafiki/blob/6e3c33f840b801600b8ae57e626ff1dbcb0cafc3/packages/backend/src/open_payments/payment_pointer/middleware.ts). The subsequent sub-resource route method should confirm that the retrieved sub-resources shares the same payment pointer id.

Otherwise for example, someone with a grant to `READ` outgoing payments for `https://wallet.example/bob` could access `https://wallet.example/alice/outgoing-payments/bf85a520-245c-4398-ae04-f868ffa26eeb` by querying `https://wallet.example/bob/outgoing-payments/bf85a520-245c-4398-ae04-f868ffa26eeb`

### LINKED ISSUE #602: Add payments with different `paymentPointer` to pagination tests
See [discussion](https://github.com/interledger/rafiki/pull/535#discussion_r969888248)

### LINKED ISSUE #644: Store grant client id on quote creation
To limit access when querying a quote with a grant to `read` (not `read-all`)

related to #506

wilsonianb:

- match subresources by payment pointer on GET requests
- store `grantId` on quote
- distinguish `read`/`read-all` (and `list`/`list-all`) in auth middleware instead of routes
- add abstract `PaymentPointerSubresource` model.
- add shared subresource tests

## Context

- fixes #598 
- fixes #644 
- fixes #602

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


wilsonianb:
What if we moved the following into the `packages/backend/src/open_payments/payment_pointer` directory
-`packages/backend/src/open_payments/payment/incoming`
-`packages/backend/src/open_payments/payment/incoming`
-`packages/backend/src/open_payments/quote`
to enforce the idea of them as payment pointer sub-resources?

(What if we then moved everything out of the `packages/backend/src/open_payments` directory?)

packages/backend/src/app.ts
```diff
@@ -84,6 +84,7 @@ export type ClientKeysContext = Context<AppRequest<'keyId'>>
 export interface PaymentPointerContext extends AppContext {
   paymentPointer: PaymentPointer
   grant?: Grant
+  clientId?: string
```
sabineschaller, line 87: Isn't that also a property of `grant`?

packages/backend/src/open_payments/payment_pointer/model.ts
```diff
@@ -91,3 +91,97 @@ export class PaymentPointerEvent extends WebhookEvent {
...
+  get({ id, paymentPointerId, clientId }: GetOptions) {
+    if (paymentPointerId) {
+      this.where(
+        `${this.modelClass().tableName}.paymentPointerId`,
+        paymentPointerId
+      )
+    }
+    if (clientId) {
+      this.byClientId(clientId)
+    }
```
sabineschaller, line 138: Is this doing two queries if `paymentPointerId` and `clientId` are given?

wilsonianb: No, this is all a single query builder that isn't executed until we `await`
https://vincit.github.io/objection.js/api/query-builder/#class-querybuilder

packages/backend/src/app.ts
wilsonianb: Yes, but `ctx.clientId` is now being defined in the auth middleware if the request has limited access to client-specific resources.
(instead of checking the grant in every relevant route)
Maybe it'd be clearer if we replace `grant` here with `grantId`, but also add a `limits?: AccessLimits` (to be used in outgoing payment creation)?

sabineschaller: > Maybe it'd be clearer if we replace grant here with grantId, but also add a limits?: AccessLimits (to be used in outgoing payment creation)?
Let's do that.

wilsonianb: I started making this change but it seems like it'll be a bit extensive, so I'd prefer addressing it as a separate issue:
- https://github.com/interledger/rafiki/issues/688

sabineschaller (approved):
Let's just get it in and fix potential issues later.

wilsonianb:
Apparently I should have merged `main` and re-run checks (despite no conflicts), because builds on `main` are now failing.
This branch removed `packages/backend/src/shared/routes.test.ts` from which an import has since been added
- https://github.com/interledger/rafiki/pull/658

I'll resolve this as a part of:
- https://github.com/interledger/rafiki/issues/686

Issue #663, opened 2022-10-07
Create Open Payments client library


wilsonianb:
The `backend` has some existing Open Payments client functionality for resolving quote/outgoing payment `receiver`s:
- https://github.com/interledger/rafiki/pull/595

This functionality should exist in its own library that is generally usable for any Open Payments client to perform and validate Open Payments requests and responses.

It can include the following pending functionality:
- https://github.com/interledger/rafiki/issues/606
- https://github.com/interledger/rafiki/issues/583
- #587 

This could potentially also export generated Open Payments types (as well as OpenAPI request/response validators):
- #662

mkurapov:
### Remaining work:

#### Handlers

Incoming payment:
- [ ] create
- [ ] complete
- [ ] list/pagination 
   
Outgoing payment:
- [ ] get
- [ ] create
- [ ] list/pagination 

Quote
- [ ] get
- [ ] create

Grant
- [ ] continue
- [ ] cancel

Token
- [ ] rotation
- [ ] revoking

mkurapov:
Remaining issues:

https://github.com/interledger/rafiki/issues/809
https://github.com/interledger/rafiki/issues/810
https://github.com/interledger/rafiki/issues/811
https://github.com/interledger/rafiki/issues/812
https://github.com/interledger/rafiki/issues/813
https://github.com/interledger/rafiki/issues/814
https://github.com/interledger/rafiki/issues/815
https://github.com/interledger/rafiki/issues/816
https://github.com/interledger/rafiki/issues/817

Currently blocked:
https://github.com/interledger/rafiki/issues/818

mkurapov:
Remaining issues:

- https://github.com/interledger/rafiki/issues/809
- https://github.com/interledger/rafiki/issues/810
- https://github.com/interledger/rafiki/issues/811
- https://github.com/interledger/rafiki/issues/812
- https://github.com/interledger/rafiki/issues/813
- https://github.com/interledger/rafiki/issues/814
- https://github.com/interledger/rafiki/issues/815
- https://github.com/interledger/rafiki/issues/816
- https://github.com/interledger/rafiki/issues/817
- https://github.com/interledger/rafiki/issues/662
- https://github.com/interledger/rafiki/issues/818

mkurapov:
Feature parity complete, closing this issue. Only outstanding related issues to the `open-payments` client are 
- #662 
- #970

Issue #662, opened 2022-10-07
Use generated Open Payments OpenAPI types


wilsonianb:
https://github.com/drwpow/openapi-typescript

Replace existing types in the `auth` and `backend` packages such as:
https://github.com/interledger/rafiki/blob/f17fdb9981d57d6199062a94ef032d47339b5459/packages/backend/src/open_payments/payment/incoming/model.ts#L260-L272

Wherever these types live could also export the corresponding Open Payments spec(s) version(s), replacing the configured versions:
https://github.com/interledger/rafiki/blob/f17fdb9981d57d6199062a94ef032d47339b5459/packages/auth/src/config/app.ts#L30-L33
https://github.com/interledger/rafiki/blob/f17fdb9981d57d6199062a94ef032d47339b5459/packages/backend/src/config/app.ts#L104-L111

PR #669, opened 2022-10-12
feat(open-payments): adding open-payments package

mkurapov:

This is a first PR for roadmap item: https://github.com/interledger/roadmap/issues/20. More discussion can be found there.

This PR deals with creating a new package, `open-payments`. It uses the [openapi-typescript](https://github.com/drwpow/openapi-typescript) package to generate Typescript types from the Open API [resource server spec](https://github.com/interledger/open-payments/blob/main/open-api-spec.yaml), and then exports the types (to be used in the `/backend` package). It also exports a client to be used when making rafiki to rafiki calls via the Open Payments API. 

The usage of the client is as follows:
Screen Shot 2022-10-12 at 3 27 16 PM

Additional changes are to follow, want to keep the PR sizes manageable:
- Open API validation (from `open-api` package)
- Usage of client in the `/backend` package, replacing how the incoming-payments and connections are grabbed in the existing `OpenPaymentsClientService`

## Context

The related issues are #663, and #662. 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/open-payments-client/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    headers['Signature'] = 'TODO'
+    headers['Signature-Input'] = 'TODO'
+  }
+
+  const { data } = await axios.get(url, {
+    headers
+  })
+
+  return data
+}
```
mkurapov, line 39: openAPI response validation can live here. This means leveraging the `open-api` package as well.

packages/open-payments-client/src/index.ts
```diff
@@ -0,0 +1,64 @@
+import { components } from './types'
+import axios, { AxiosInstance } from 'axios'
```
sublimator, line 2: considered universal/isomorphic fetch ? 
node 18 LTS is due very soon
https://dev.to/cloudx/nodejs-18-fetch-api-test-runner-module-and-more-2ckg

mkurapov: Yes, looks to be out Oct 25th: https://github.com/nodejs/release#release-schedule. I have `axios` for now as I was mirroring the existing `OpenPaymentsClientService` https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/client/service.ts#L153-L180. I know we use `axios` in a few places so maybe have a team discussion whether we want to upgrade node and replace `axios` with `fetch` altogether once node 18 is out

packages/open-payments/package.json
```diff
@@ -0,0 +1,24 @@
...
+  ],
+  "scripts": {
+    "build": "pnpm clean && tsc --build tsconfig.json",
+    "clean": "rm -fr dist/",
+    "prepack": "pnpm build",
+    "generate:types": "npx ts-node scripts/generate-types.ts"
+  },
+  "devDependencies": {
+    "@types/node": "^18.7.12",
+    "openapi-typescript": "^4.5.0",
```
mkurapov, line 17: I wrested quite a bit with this, but I wasn't able to use the newest version of this library. Using the programmatic way of generating types via `openapiTS` anything above v5 does not work: `ERR_REQUIRE_ESM`. Versions 5 and above do not support backwards compatibility with CommonJS: given the fact that having `type: module` in `package.json` break using this package as a workspace in the `/backend` project

packages/open-payments/src/config.ts
```diff
@@ -0,0 +1,5 @@
+export default {
+  OPEN_PAYMENTS_OPEN_API_URL:
```
mkurapov, line 2: want to make sure the same URL is used for type generation and (following) open API response validation

packages/open-payments/src/client.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return {
+    incomingPayment: {
+      get: (args: GetArgs) => get<IncomingPayment>(axios, args)
+    },
+    ilpStreamConnection: {
+      get: (args: GetArgs) => get<ILPStreamConnection>(axios, args)
+    }
+  }
+}
```
mkurapov, line 67: Wanted to keep very simple for the first PR, will be adding open-api validation next.

packages/open-payments/src/client.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    test('sets headers properly if accessToken provided', async (): Promise<void> => {
+      nock(baseUrl)
+        .get('/incoming-payment')
+        .reply(200, () => ({
+          validReceiver: 0
+        }))
+
+      await get(axiosInstance, {
+        url: `${baseUrl}/incoming-payment`,
+        accessToken: 'accessToken'
```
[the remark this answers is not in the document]
mkurapov: (this is a test file)

packages/open-payments/src/client.test.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+
+      expect(axiosInstance.get).toHaveBeenCalledWith(
+        `${baseUrl}/incoming-payment`,
+        {
+          headers: {
+            Authorization: 'GNAP accessToken',
```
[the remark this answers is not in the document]
mkurapov: (this is a test file)

mkurapov:
Will add documentation once there is a proper usage of the client in the `/backend` package.

packages/open-payments/src/generated/types.ts
```diff
@@ -0,0 +1,702 @@
+/**
```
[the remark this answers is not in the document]
mkurapov: @wilsonianb what we can do is instead of that check is actually just run the script on a pre-commit hook to generate the types and stage the updated file for committing. 

Instead of doing it on every commit, we can just do that _only_ if the `src/types.ts` file has been changed, for example. Thoughts?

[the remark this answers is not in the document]
wilsonianb: That sounds good to me.
We did recently remove pre-commit hooks, but hopefully it won't be an issue to support this
- https://github.com/interledger/rafiki/pull/545

wilsonianb (approved):

mkurapov:
@wilsonianb the openpayments spec split has caused the type generation to break, it will require a change to that repo to support generating these types. Basically, when referencing the [shared schema](https://github.com/interledger/open-payments/blob/main/openapi/shared/schemas.yaml), the types are generated as such:


Screen Shot 2022-10-21 at 3 11 19 PM

Basically, nothing gets generated for the shared schema:
Screen Shot 2022-10-21 at 3 11 28 PM

This is because the shared schemas do not use `components.schemas` types, but directly use fields at the root level, eg `$ref: ../../shared/schemas.yaml#/access`, instead of `../../shared/schemas.yaml#/components/schemas/access`

wilsonianb (approved):

PR #671, opened 2022-10-13
feat(auth): handle grant choice params correctly

njlie:

- Current openAPI spec treats the "accept/reject" in the `/accept` or `/reject` grant endpoints as a query param, so the handler has been modified to reflect that.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/routes.test.ts
```diff
@@ -445,7 +445,7 @@ describe('Grant Routes', (): void => {
         )
 
         const redirectUrl = new URL(config.identityServerDomain)
-        redirectUrl.searchParams.set('interactRef', grant.interactRef)
+        redirectUrl.searchParams.set('interactId', grant.interactId)
```
wilsonianb, line 448: These were just changed in
- https://github.com/interledger/rafiki/pull/670

packages/auth/src/grant/routes.test.ts
```diff
@@ -637,19 +637,25 @@ describe('Grant Routes', (): void => {
...
           headers: {
             Accept: 'application/json',
             'Content-Type': 'application/json'
           }
         },
-        { id: grant.interactId, nonce: grant.interactNonce }
+        {
+          id: grant.interactId,
+          nonce: grant.interactNonce,
+          choice: GrantChoices.Accept
```
wilsonianb, line 652: Add test for non-`GrantChoices` `choice`

packages/auth/src/grant/routes.ts
```diff
@@ -265,15 +269,19 @@ async function acceptGrant(
     ctx.body = {
       error: 'request_denied'
     }
+    return
```
mkurapov, line 272: Not related to the changes, but is it ok to use status = 400 above on L268 even though it's not listed as a possible returned status code in the Open API spec?

https://github.com/interledger/open-payments/blob/f365dbec4b9dec98b9f622bc49a92aea9ee01568/openapi/AS/IdP/openapi.yaml#L122-L128

mkurapov (approved):

packages/auth/src/grant/routes.ts
njlie: Some of this stuff (to my recollection) was written before the OpenAPI spec was brought in, so it's more like an initial stab. There could be something to adding a response to the spec for unexpected stuff, but later I will go over all these responses and decide if they'll conform to an existing response or a new one should be added to the spec.

PR #676, opened 2022-10-17
fix: only store grant and client IDs on create routes

### LINKED ISSUE #599: Selectively store grant and client id only on create routes
Right now, every new grant id and corresponding client id is stored by the auth middleware (packages/backend/src/open_payments/auth/middleware.ts). It should only happen for the create incoming payment / outgoing payment / quote routes.
https://github.com/interledger/rafiki/pull/535#discussion_r967206941

SarahCoilAccount:

- Only store grant and client IDs on create routes

## Context

- Fixes #599

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -220,6 +220,42 @@ describe('Auth Middleware', (): void => {
...
+    }
+  )
+  test.each`
+    type                          | description
+    ${AccessType.IncomingPayment} | ${'incoming payments'}
+    ${AccessType.OutgoingPayment} | ${'outgoing payments'}
+    ${AccessType.Quote}           | ${'quotes'}
+  `(
+    'Stores grant details for $description on the the create path',
+    async ({ type }): Promise<void> => {
```
wilsonianb, line 241: You could test all the actions with something like:
```
  Object.values(AccessAction).forEach((action) => {
    Object.values(AccessType).forEach((type) => {
      if (action === AccessAction.Create) {
        test(`stores grant details for ${type} on the the ${action} path`, async (): Promise<void> => {
        })
      } else {
        test(`does not store grant details for ${type} on the ${action} path`, async (): Promise<void> => {
        })
      }
```

wilsonianb: Actually, since that :point_up_2: would fail OpenAPI validation, maybe you keep your existing test for all types, but then loop through all actions for just incoming payments.

SarahCoilAccount: @wilsonianb  we can still test them all like this?
```
  const table: { type: AccessType, action: AccessAction, description: string }[] = []
  Object.values(AccessType).forEach((type) => {
    Object.values(AccessAction).forEach((action) => {
      // The complete path is not valid for outgoing payments
      if(!(type === AccessType.OutgoingPayment && action == AccessAction.Complete) && type !== AccessType.Quote) {
        table.push({
          type: type,
          action: action,
          description: action === AccessAction.Create ? 'stores grant details' : 'does not store grant details'
        })
        // Create and read are the only valid paths for quotes
      } else if(type === AccessType.Quote && (action === AccessAction.Create || action === AccessAction.Read)) {
        table.push({
          type: type,
          action: action,
          description: action === AccessAction.Create ? 'stores grant details' : 'does not store grant details'
        })
      }
    })
  })
```
Or does this make the tests fragile to future changes?

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -223,17 +223,95 @@ describe('Auth Middleware', (): void => {
...
+          description:
+            action === AccessAction.Create
+              ? 'stores grant details'
+              : 'does not store grant details'
+        })
+      }
+    })
+  })
+  test.each(table)(
+    '$description for $type on the $action path',
```
wilsonianb, line 273: Somewhat more concise.
You're right to call it fragile (until `open-payments` library potentially exposes the type/action pairings).
`read-all` `quotes` is actually in the latest spec but not the version currently used in `backend`

```suggestion
  const types = {
    [AccessType.IncomingPayment]: Object.values(AccessAction),
    [AccessType.OutgoingPayment]: Object.values(AccessAction).filter(
      (action) => action !== AccessAction.Complete
    ),
    [AccessType.Quote]: [
      AccessAction.Create,
      AccessAction.Read
      // TODO
      // AccessAction.ReadAll
    ]
  }

  Object.values(AccessType).forEach((type) => {
    for (const action of types[type]) {
      const description =
        action === AccessAction.Create
          ? 'stores grant details'
          : 'does not store grant details'

      test(`${description} for ${type} on the ${action} path`, async (): Promise<void> => {
```

wilsonianb (approved):

wilsonianb (approved):

PR #681, opened 2022-10-19
feat(backend): use payment pointers as clients

omertoast:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/resolvers/clientKeys.ts
```diff
@@ -101,16 +44,19 @@ export const addKeyToClient: MutationResolvers<ApolloContext>['addKeyToClient']
     ctx
   ): Promise<ResolversTypes['AddKeyToClientMutationResponse']> => {
     try {
-      const clientService = await ctx.container.use('clientService')
-      const client = await clientService.addKeyToClient({
+      const paymentPointerService: PaymentPointerService =
+        await ctx.container.use('paymentPointerService')
```
mkurapov, line 48:
```suggestion
      const paymentPointerService =
        await ctx.container.use('paymentPointerService')
```
assuming it isn't needed here?

packages/backend/src/graphql/resolvers/payment_pointer.ts
```diff
@@ -98,5 +98,14 @@ export const paymentPointerToGraphql = (
   url: paymentPointer.url,
   asset: assetToGraphql(paymentPointer.asset),
   publicName: paymentPointer.publicName ?? undefined,
+  keys:
+    paymentPointer.keys != null
+      ? paymentPointer.keys.map((key) => ({
+          id: key.id,
+          paymentPointerId: key.id,
```
mkurapov, line 105:
```suggestion
          paymentPointerId: key.paymentPointerId,
```

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -110,7 +118,8 @@ async function getPaymentPointerByUrl(
 ): Promise<PaymentPointer | undefined> {
   const paymentPointer = await PaymentPointer.query(deps.knex)
     .findOne({ url })
-    .withGraphJoined('asset')
+    .withGraphFetched('asset')
+    .withGraphFetched('keys')
```
mkurapov, line 122: do we also need this in the 
`src/open_payments/payment_pointer/service.ts` -> `processNextPaymentPointers` ?

packages/backend/src/graphql/schema.graphql
```diff
@@ -527,19 +517,9 @@ input CreatePaymentPointerInput {
...
-  image: String!
-  email: String!
-  keys: [ClientKeys!]!
-  createdAt: String!
-}
-
 type ClientKeys implements Model {
   id: ID!
-  clientId: String!
+  paymentPointerId: String!
```
mkurapov, line 522:
```suggestion
  paymentPointerId: ID!
```
not exactly sure if that is correct, both seem to be used in this file, so probably fine to leave as string

packages/backend/src/clientKeys/routes.ts
```diff
@@ -37,10 +39,8 @@ export async function getKeyById(
     key: key.jwk,
     client: {
```
mkurapov, line 40: do we want update the response body on this as well? what do we need this route for btw?

mkurapov (commented):
LGTM, I was able to test and validate the results in the DB. just have a few comments/nitpicks

adrianhopebailie (approved):

mkurapov (commented):
Looks good, just the `paymentPointerToGraphql` needs an edit

packages/backend/src/app.ts
```diff
@@ -345,7 +347,8 @@ export class App {
     }
     router.get(
       '/keys/{keyId}',
```
adrianhopebailie, line 349: Shouldn't this route have been removed?

wilsonianb: I'm guessing it will with:
- https://github.com/interledger/rafiki/pull/678

packages/backend/migrations/20220819120358_create_payment_pointer_keys_table.js
```diff
@@ -1,8 +1,8 @@
 exports.up = function (knex) {
-  return knex.schema.createTable('clientKeys', function (table) {
+  return knex.schema.createTable('paymentPointerKeys', function (table) {
```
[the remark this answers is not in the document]
adrianhopebailie: @wilsonianb -my suggestion to @omertoast was to rename this so that we can differentiate between these keys and the keys we cache from incoming clients. i.e. Client keys would be keys we lookup/cache for clients making incoming requests and payment pointer keys are keys we have on record for our own payment pointers

packages/backend/src/clientKeys/routes.ts
omertoast: Key inspection from auth server I guess? I updated the body.

packages/backend/src/graphql/schema.graphql
```diff
@@ -253,16 +253,9 @@ input CreatePaymentPointerWithdrawalInput {
...
-input CreateClientInput {
-  name: String!
-  uri: String!
-  image: String!
-  email: String!
-}
-
-input AddKeyToClientInput {
+input AddKeyToPaymentPointerInput {
   id: String!
```
wilsonianb, line 257: Could the backend auto-generate the id instead?

packages/backend/src/graphql/schema.graphql
```diff
@@ -109,13 +108,14 @@ type Mutation {
...
   withdrawEventLiquidity(eventId: String!): LiquidityMutationResponse
 
-  "Create Client"
-  createClient(input: CreateClientInput!): CreateClientMutationResponse
-
   "Add Key to Client"
-  addKeyToClient(input: AddKeyToClientInput!): AddKeyToClientMutationResponse
+  addKeyToPaymentPointer(
+    input: AddKeyToPaymentPointerInput!
+  ): AddKeyToPaymentPointerMutationResponse
```
wilsonianb, line 114:
```suggestion
  addPaymentPointerKey(
    input: AddPaymentPointerKeyInput!
  ): AddPaymentPointerKeyMutationResponse
```

packages/backend/src/open_payments/payment_pointer/model.ts
```diff
@@ -14,16 +15,26 @@ export class PaymentPointer
...
+      modelClass: PaymentPointerKeys,
+      join: {
+        from: 'paymentPointers.id',
+        to: 'paymentPointerKeys.paymentPointerId'
+      }
     }
-  }
+  })
+
+  public keys!: PaymentPointerKeys[]
```
wilsonianb, line 37: I don't think we should do a join on keys with every payment pointer lookup.
```suggestion
  public keys?: PaymentPointerKeys[]
```

packages/backend/src/paymentPointerKeys/model.ts
```diff
@@ -2,14 +2,14 @@ import { BaseModel } from '../shared/baseModel'
 
 import { JWKWithRequired } from 'auth'
 
-export class ClientKeys extends BaseModel {
+export class PaymentPointerKeys extends BaseModel {
```
wilsonianb, line 5:
```suggestion
export class PaymentPointerKey extends BaseModel {
```

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -21,6 +23,9 @@ export interface PaymentPointerService {
   create(options: CreateOptions): Promise<PaymentPointer | PaymentPointerError>
   get(id: string): Promise<PaymentPointer | undefined>
   getByUrl(url: string): Promise<PaymentPointer | undefined>
+  addKeyToPaymentPointer(
```
wilsonianb, line 26: Can this instead be `paymentPointerKeysService.create`?
Either that or we could remove `PaymentPointerKeysService` and include the following here:
- `addKey`
- `getKeys`
- `revokeKey`

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -93,6 +99,7 @@ async function createPaymentPointer(
       assetId: asset.id
     })
     .withGraphFetched('asset')
+    .withGraphFetched('keys')
```
wilsonianb, line 102: ditto below
A separate graphql resolver can be added to fetch a payment pointer's keys (for when `keys` is included in a `paymentPointers` query)
```suggestion
```

omertoast:
FYI, I merged this PR with the other one. https://github.com/interledger/rafiki/pull/678

packages/backend/src/graphql/schema.graphql
```diff
@@ -527,19 +519,9 @@ input CreatePaymentPointerInput {
   publicName: String
 }
 
-type Client implements Model {
+type PaymentPointerKey implements Model {
```
wilsonianb, line 522: I think `PaymentPointer` ([above](https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/graphql/schema.graphql#L337)) should include the following to allow querying existing keys:
```
  keys(
    "Paginating forwards: the cursor before the the requested page."
    after: String
    "Paginating backwards: the cursor after the the requested page."
    before: String
    "Paginating forwards: The first **n** elements from the page."
    first: Int
    last: Int
  ): PaymentPointerKeyConnection
```

packages/backend/src/paymentPointerKey/routes.ts
```diff
@@ -0,0 +1,36 @@
...
+): Promise<void> {
+  if (!ctx.paymentPointer) {
+    return ctx.throw(404)
+  }
+
+  const keys = await deps.paymentPointerKeyService.getKeysByPaymentPointerId(
+    ctx.paymentPointer.id
+  )
+
+  ctx.body = keys.map((key) => key.jwk)
```
wilsonianb, line 35: https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-set-properties
```suggestion
  ctx.body = {
    keys: keys.map((key) => key.jwk)
  }
```
If we want to include [client interact display info](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-2.3.2) in the `jwks.json` response (instead of the payment pointer query response), it might make sense to move this to `PaymentPointerRoutes` (and remove `PaymentPointerKeyRoutes`)

packages/backend/src/paymentPointerKey/model.ts
```diff
@@ -2,14 +2,14 @@ import { BaseModel } from '../shared/baseModel'
 
 import { JWKWithRequired } from 'auth'
 
-export class ClientKeys extends BaseModel {
+export class PaymentPointerKey extends BaseModel {
```
wilsonianb, line 5: I'm tempted to try to make this work:
```suggestion
export class PaymentPointerKey extends PaymentPointerSubresource {
```
which would let you utilize the existing [`list` query](https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/open_payments/payment_pointer/model.ts#L141).
See:
https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/open_payments/payment/outgoing/service.ts#L357-L362
Note that:
https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/open_payments/payment/outgoing/service.ts#L32-L33
At the moment, `PaymentPointerSubresource` (added in https://github.com/interledger/rafiki/pull/647) requires an `assetId` though...

packages/backend/src/paymentPointerKey/service.ts
```diff
@@ -0,0 +1,100 @@
+import { TransactionOrKnex } from 'objection'
+
+import { PaymentPointerKey } from './model'
+import { BaseService } from '../shared/baseService'
+import { JWKWithRequired } from 'auth'
+
+export interface PaymentPointerKeyService {
+  create(options: CreateOptions): Promise<PaymentPointerKey>
+  getKeyById(keyId: string): Promise<PaymentPointerKey>
```
wilsonianb, line 9:
```suggestion
```

packages/backend/src/paymentPointerKey/service.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+async function getKeysByPaymentPointerId(
+  deps: ServiceDependencies,
+  paymentPointerId: string
+): Promise<PaymentPointerKey[]> {
+  const keys = await PaymentPointerKey.query(deps.knex).where({
+    paymentPointerId
+  })
```
wilsonianb, line 98: This seems like a unique case where we want relay pagination for the admin api, but wouldn't support pagination query parameters for `/jwks.json` requests.
Should the route query the largest possible page?
In which case, should we limit the number of keys for a single payment pointer to the maximum page size (100)?
https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/shared/baseModel.ts#L57

packages/backend/src/paymentPointerKey/routes.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        headers: { Accept: 'application/json' },
+        url: `/`
+      })
+      ctx.paymentPointer = paymentPointer
+
+      await expect(
+        paymentPointerKeyRoutes.getKeysByPaymentPointerId(ctx)
+      ).resolves.toBeUndefined()
+      expect(ctx.body[0]).toEqual(key.jwk)
+      expect(ctx.body).toHaveLength(1)
```
wilsonianb, line 88:
```suggestion
      expect(ctx.response).toSatisfyApiSpec()
      expect(ctx.body).toEqual({
        keys: [key.jwk]
      })
```
Using `jestOpenAPI` (`toSatisfyApiSpec`) probably requires updating [`openPaymentsSpec`](https://github.com/interledger/rafiki/blob/92c89a35aca2cfde7903abbef04fcfffb2b6f369/packages/backend/src/config/app.ts#L107-L109) to include https://github.com/interledger/open-payments/commit/f365dbec4b9dec98b9f622bc49a92aea9ee01568

packages/backend/src/paymentPointerKey/model.ts
omertoast: @adrianhopebailie and I talked about this and if we assume not every `PaymentPointerSubresource` should have an `assetId`, it would require a change to `PaymentPointerSubresource` in order to `PaymentPointerKey` implement it. I think we can create a new PR and make the `assetId` an explicit property to `Quote`, `IncomingPayment`, and `OutgoingPayment` models there.

packages/backend/src/paymentPointerKey/service.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        paymentPointerId: paymentPointer.id,
+        jwk: TEST_KEY
+      }
+
+      const paymentPointerKey = await paymentPointerKeyService.create(options)
+
+      await expect(paymentPointerKey.paymentPointerId).toEqual(
+        options.paymentPointerId
+      )
+      await expect(paymentPointerKey.jwk).toEqual(options.jwk)
```
wilsonianb, line 72:
```suggestion
      await expect(
        paymentPointerKeyService.create(options)
      ).resolves.toMatchObject(options)
```

packages/backend/src/paymentPointerKey/service.ts
```diff
@@ -0,0 +1,88 @@
+import { TransactionOrKnex } from 'objection'
+
+import { PaymentPointerKey } from './model'
+import { BaseService } from '../shared/baseService'
+import { JWKWithRequired } from 'auth'
+
+export interface PaymentPointerKeyService {
+  create(options: CreateOptions): Promise<PaymentPointerKey>
+  revokeKeyById(keyId: string): Promise<string>
```
wilsonianb, line 9: I realize this is from `ClientKeysService`, but I'd prefer
```suggestion
  revoke(id: string): Promise<PaymentPointerKey | undefined>
```
especially now that `id` !== `jwk.keyid`.
It should return `undefined` if the key doesn't exist. (This needs to be tested)
I'm not sure what should happen if the key is already revoked.

packages/backend/src/graphql/schema.graphql
wilsonianb: - https://github.com/interledger/rafiki/issues/702

packages/backend/src/paymentPointerKey/model.ts
wilsonianb: - https://github.com/interledger/rafiki/issues/701

packages/backend/src/paymentPointerKey/model.ts
```diff
@@ -2,14 +2,14 @@ import { BaseModel } from '../shared/baseModel'
...
   public static get tableName(): string {
-    return 'clientKeys'
+    return 'paymentPointerKeys'
   }
 
   // The id should be the same as the id in the kid.
   public id!: string
```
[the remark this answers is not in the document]
omertoast: Actually, I didn't notice I left that comment. I think they should be different too.

packages/backend/src/paymentPointerKey/service.ts
omertoast: Why don't we throw when the key doesn't exist? Isn't it dangerous if the undefined value is not handled? I think we can return revoked `PaymentPointerKey`if the key is already revoked if there is not an exceptional case.

packages/backend/src/paymentPointerKey/service.ts
```diff
@@ -0,0 +1,91 @@
...
+    paymentPointerId: options.paymentPointerId,
+    jwk: options.jwk
+  })
+  return key
+}
+
+async function revoke(
+  deps: ServiceDependencies,
+  id: string
+): Promise<PaymentPointerKey | undefined> {
```
wilsonianb, line 57: I'm good with this as is, but another option would be to return `PaymentPointerKeyError` for unknown or already revoked keys.
See:
https://github.com/interledger/rafiki/blob/603df09615a2f458227baf3e23f99b3a0b103921/packages/backend/src/open_payments/payment/incoming/service.ts#L285-L308

packages/backend/src/paymentPointerKey/service.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      const key = await paymentPointerKeyService.create(keyOption)
+      await expect(
+        paymentPointerKeyService.getKeysByPaymentPointerId(paymentPointer.id)
+      ).resolves.toEqual([key])
+    })
+  })
+
+  describe('Revoke Payment Pointer Keys', (): void => {
+    test('Can revoke a key', async (): Promise<void> => {
```
wilsonianb, line 85: Also test already revoked key.

packages/backend/src/paymentPointerKey/model.ts
```diff
@@ -2,14 +2,13 @@ import { BaseModel } from '../shared/baseModel'
...
-    return 'clientKeys'
+    return 'paymentPointerKeys'
   }
 
-  // The id should be the same as the id in the kid.
   public id!: string
-  public clientId!: string
+  public paymentPointerId!: string
 
   public jwk!: JWKWithRequired
```
wilsonianb, line 13: What do you think about the idea of flattening the jwk?
https://github.com/interledger/rafiki/pull/681#discussion_r1006835706

omertoast: Last time `jwk` had `clientDetails` in it, it was excluded because it made it hard to work with the tools related to jwk. I think it's fine right now without the flattening.

packages/backend/src/paymentPointerKey/service.ts
omertoast: I think this could be implemented when we have a few more error cases.

packages/backend/src/paymentPointerKey/routes.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+export async function getKeysByPaymentPointerId(
+  deps: ServiceDependencies,
+  ctx: PaymentPointerContext
+): Promise<void> {
+  if (!ctx.paymentPointer) {
+    return ctx.throw(404)
+  }
```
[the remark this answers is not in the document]
wilsonianb: It seems like typescript would still enforce that the route middleware is called with a `PaymentPointerContext`, which is the case with test for this check.

wilsonianb (approved):

Issue #682, opened 2022-10-20
new dev environment permissions 


mankins:
I was spinning up a new dev environment to take a look an open issue for hacktober (https://github.com/interledger/rafiki/issues/664), but have run into a postgres permissions issue (I think?). 

After  `pnpm localenv up -d --build`   I can see the docker containers started up.

I then was curling to look at the signature:

`curl -X post http://localhost:3006/auth`

And the connection closes immediately. I see this in the Docker logs:

```
2022-10-20 10:22:29.035 UTC [1] LOG:  database system is ready to accept connections
2022-10-20 10:22:29.223 UTC [67] ERROR:  permission denied for schema public at character 14
2022-10-20 10:22:29.223 UTC [67] STATEMENT:  create table "auth_knex_migrations" ("id" serial primary key, "name" varchar(255), "batch" integer, "migration_time" timestamptz)
```

I tried to add. `GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO peerauth` etc to each, but that didn't alter anything. 

I'm guessing I'm missing some setup step, but not sure what it might be so any help would be appreciated.

mkurapov:
@mankins thanks for opening up the issue! We are having the same problem on another new users' machine, so we will keep this thread updated. Which machine do you have, by the way?

mankins:
@mkurapov it's a mac m1 running 12.6. It's fairly new so I keep running into dev dependencies I haven't installed, so that may be part of the issue. 

I have used knex in another project recently and didn't have this issue though.

Let me know if I can provide more...

mkurapov:
@mankins thank you for the info! 

I ended up finding the root cause of the issue: it's a result of Postgres's `latest` Docker image being updated to postgres version 15 [(commit)](https://github.com/docker-library/postgres/commit/747be5974512adedb4ccacf4a77002d41137de15), which results in postgres being upgraded from version 14 to 15 for the app. Based on some [new changes](https://andreas.scherbaum.la/blog/archives/1120-Changes-to-the-public-schema-in-PostgreSQL-15-and-how-to-handle-upgrades.html) in version 14 related to public schema, it breaks permissions & consequently, migrations. 

Will look into how to best address this. 

CC @wilsonianb

mkurapov:
Yup, I was able to reproduce locally as well earlier. I've been playing around with a few ways of making it work in 15, but we should pin the major version going forward to avoid any unexpected issues, whether that is 15 or 14.

[...]

Issue #664, opened 2022-10-10
Make sig validation not processed in dev mode


matdehaast:
If a user sets the `Config.env = "dev"` we should disable signature validation for incoming requests on the AS and the RS. For the AS we can remove the current mechanism we use to do that and rather just assume it from the mode.

This will make it initially much easier to work with Postman as we still need to add signature support to it

mkurapov:
@matdehaast wanted to add more context here.

I believe the current mechanism actually lives on the RS, since on an Open Payments requests you can just set `authorization: GNAP dev-access-token`, and it'll skip through the actual signature verification: 

https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/middleware.ts#L30

However, what's a bit tricky here though is the fact that you can bypass sig verification, you still need to have `signature` and `signature-input` variables in the header (even if they are blank) as that checking is done inside the OpenAPI validation:

https://github.com/interledger/rafiki/blob/main/packages/backend/src/app.ts#L339

I don't believe there is anything to bypass signature validation on the AS side, however.

mankins:
Ok, my dev problems were fixed, so I was able to run these tests locally and make some changes.

I'd love to some feedback to see if the approach is right and worth continuing to do the same thing for RS or if there's some other better way to do this.

mkurapov:
@mankins This is a good approach, I was just clarifying previously that the existing mechanism (with the `dev-access-token`) lives on the resource server side and not originally on the authorization server side. 

I can take a look at the PR later tonight, and do some own testing too.

wilsonianb:
`dev-access-token` helps the RS skip HTTP signature validation on client-RS requests.
#683 will help the AS skip HTTP signature validation on client-AS requests.

wilsonianb:
resolved by:
- https://github.com/interledger/rafiki/pull/683 / https://github.com/interledger/rafiki/pull/735
- https://github.com/interledger/rafiki/pull/741

Issue #656, opened 2022-10-05
Audit tests using nock `.isDone`


wilsonianb:
Assert that the `.isDone` response is being checked or replace with `.done`

https://github.com/nock/nock#isdone

wilsonianb:
The nock scope in this test is not being called and can be removed
https://github.com/interledger/rafiki/blob/c8d42f60386d1c5a0e387c4e8bd30d6f74846f41/packages/auth/src/signature/middleware.test.ts#L338

PR #683, opened 2022-10-21
feat(auth): Make sig validation not processed in dev mode

mankins:

- Disable signature validation for incoming requests on the AS to make development easier with Postman
- New config setting `bypassSignatureValidation` to bypass signature validation

## Context

Attempts to fix part of #664 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

## Discussion

This is a ~WIP / Draft~ PR that attempts some of the work discussed in #664. I thought a bit about the use case and thought against keying off the `Config.env === 'development'` because it may be common to want signature validation in development but sometimes disable validation. As such I propose `bypassSignatureValidation` to be set from an environmental variable `BYPASS_SIGNATURE_VALIDATION` to disable signature checks.

While there are tests and code, this current draft has not been tested because I haven't been able to get my local dev environment up for various reasons including #682  

I'm submitting this ~Draft PR~ in case it helps others and to start a discussion about the approach taken and to help contribute to #hacktoberfest ... apologies in advance if I missed documentation of contributor guidelines about draft prs.


packages/auth/src/signature/middleware.test.ts
```diff
@@ -29,6 +29,7 @@ import {
   grantContinueHttpsigMiddleware,
   grantInitiationHttpsigMiddleware
 } from './middleware'
+import test from 'node:test'
```
wilsonianb, line 32:
```suggestion
```

mankins:
Bump: Can you run the tests again @wilsonianb ? My local still isn't working.

wilsonianb:
Re-ran the tests.
You can also try opening a pull request within your fork in order to run the checks without waiting for someone to push that button for you.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,10 +401,9 @@ describe('Signature Service', (): void => {
       scope.isDone()
     })
 
-    test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is set', async (): Promise<void> => {
-      deps = await initIocContainer(Config)
+    test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is true with bad signature', async (): Promise<void> => {
+      const defaultByPassSignatureValidation = Config.bypassSignatureValidation
```
mankins, line 405: The `Config` is a global to the tests and rather than doing a reset each time on the const I'm saving the state of the variable changed and then resetting at the end of the test. I can't recall how Jest treats failed tests (does it run all the way to the end of the test?) so this approach may need to be re-thought.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,6 +401,101 @@ describe('Signature Service', (): void => {
...
+      ctx.headers['signature'] = 'wrong-signature'
+
+      await grantInitiationHttpsigMiddleware(ctx, next)
+
+      expect(ctx.response.status).toEqual(200)
+      expect(next).toHaveBeenCalled()
+
+      Config.bypassSignatureValidation = defaultByPassSignatureValidation
+
+      scope.isDone()
```
wilsonianb, line 446:
```suggestion
      scope.done()
```
- https://github.com/interledger/rafiki/issues/656

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,6 +401,101 @@ describe('Signature Service', (): void => {
...
+    })
+
+    test('middleware fails if BYPASS_SIGNATURE_VALIDATION is false with bad signature', async (): Promise<void> => {
```
wilsonianb, line 449: I think you could remove this test, and just add a check that `bypassSignatureValidation` is false in:
https://github.com/interledger/rafiki/blob/c8d42f60386d1c5a0e387c4e8bd30d6f74846f41/packages/auth/src/signature/middleware.test.ts#L372

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,6 +401,101 @@ describe('Signature Service', (): void => {
       scope.isDone()
     })
 
+    test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is true with bad signature', async (): Promise<void> => {
+      const defaultByPassSignatureValidation = Config.bypassSignatureValidation
+      Config.bypassSignatureValidation = true
```
wilsonianb, line 406: This appears to work
```suggestion
      const config = await appContainer.container.use('config')
      config.bypassSignatureValidation = true
```

packages/auth/src/signature/middleware.test.ts
mankins: Agree, I had added it because the `Config` is a global and I wasn't convinced it was resetting between runs. 

Your suggestions below look better and solve this so I can do both and get rid of the extra test.

packages/auth/src/signature/middleware.test.ts
mankins: Even with resetting this on `beforeEach` I wasn't able to reset the Config to the default state at the start of the test, so I've reset the state within the test, although I'm still not convinced this is the right way to do this

wilsonianb: Adding this to the `beforeEach` (I tried the one inside `describe('Signature middleware'`) appears to work
```
      const config = await appContainer.container.use('config')
      config.bypassSignatureValidation = false
```

mankins: It works, @wilsonianb  but that assumes that the default state is `false` (A reasonable assumption, but not always correct, as Config is set from `process.env` or may have been modified?). 

I'm fine with it, but realized that I might not understand how that `container.use` system works at all, nor do I get the relationship between it and the `Config` import at the top.

It's probably best for you to push any commits changing that because of my lack of understanding.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,6 +401,51 @@ describe('Signature Service', (): void => {
...
+
+      ctx.headers['signature'] = 'wrong-signature'
+
+      await grantInitiationHttpsigMiddleware(ctx, next)
+
+      expect(ctx.response.status).toEqual(200)
+      expect(next).toHaveBeenCalled()
+      config.bypassSignatureValidation = defaultByPassSignatureValidation
+
+      scope.done()
```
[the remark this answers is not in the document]
mankins: Could we move the `isDone` -> `done` work to another PR? It looks like the original `scope.isDone` "works" ...and then future work could continue https://github.com/interledger/rafiki/issues/656 ?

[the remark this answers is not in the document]
wilsonianb: I'd be ok with changing this to
```suggestion
      // TODO: https://github.com/interledger/rafiki/issues/656
      // scope.done()
```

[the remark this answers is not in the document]
mankins: Ok I verified this still allows tests to pass locally.

packages/auth/src/signature/middleware.test.ts
wilsonianb: >I might not understand how that `container.use` system works at all

Does anyone really, though? :sweat_smile: 
The imported `Config` is used to construct the container:
https://github.com/interledger/rafiki/blob/c8d42f60386d1c5a0e387c4e8bd30d6f74846f41/packages/auth/src/signature/middleware.test.ts#L44-L45
which is eventually the `ctx.container` in the middleware.
I don't _think_ subsequent changes made to `Config` would be reflected in the container.

You can copy or cherry-pick this: https://github.com/wilsonianb/rafiki/commit/9bf207af6e3895e0391fc8beaa8c819fe21f0433

mankins: I incorporated your changes @wilsonianb in commit 1ee50db5f31ff69cb55cf191926a48c4f7e36b66 

I then realized there was a cleaner way of allowing tests with multiple configs, so made a subsequent commit to allow the general case which backed out many of these changes. It seems to be working for me and is cleaner, let me know what you think.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -32,7 +32,8 @@ import {
 
 describe('Signature Service', (): void => {
   let deps: IocContract<AppServices>
-  let appContainer: TestContainer
+  const appContainers: TestContainer[] = []
```
mankins, line 35: Rather than a single appContainer, this allows for multiple with different configs. Because the container takes a config as input and creates a singleton this is required if configs are materially different per test.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -57,7 +59,9 @@ describe('Signature Service', (): void => {
 
   afterAll(async (): Promise<void> => {
     nock.restore()
-    await appContainer.shutdown()
+    for (let i = 0; i < appContainers.length; i++) {
+      await appContainers[i].shutdown()
+    }
```
mankins, line 64: We just keep track of the containers and then shut them down at the end.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -408,7 +406,13 @@ describe('Signature Service', (): void => {
     })
 
     test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is true with bad signature', async (): Promise<void> => {
-      config.bypassSignatureValidation = true
+      const altDeps = await initIocContainer({
+        ...Config,
+        bypassSignatureValidation: true
```
mankins, line 411: And creating a new container can happen in tests like this so we don't have to modify `Config`

packages/auth/src/signature/middleware.test.ts
```diff
@@ -401,6 +405,55 @@ describe('Signature Service', (): void => {
...
     })
 
+    test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is true with bad signature', async (): Promise<void> => {
+      const altDeps = await initIocContainer({
+        ...Config,
+        bypassSignatureValidation: true
+      })
+
+      const altContainer = await createTestApp(altDeps)
+      appContainers.push(altContainer)
```
wilsonianb, line 415: It might also work to keep `altContainer` **contain**ed to this test, call `shutdown` within the test (maybe in a `try`/`finally`), and remove `appContainers`.

wilsonianb (approved):

packages/auth/src/signature/middleware.test.ts
mankins: That would be nice, but that makes the following test fail with:

```
   Unable to acquire a connection
```

(from knex)

I dug into it a little and the shutdown calls `knex.destroy()` 

https://github.com/interledger/rafiki/blob/main/packages/auth/src/tests/app.ts#L47-L49

https://github.com/interledger/rafiki/blob/3807051d636f0c523d8f885c3d71b91d04cbc4c4/packages/auth/src/index.ts#L144

So the current approach works around this by keeping the test connections open until done and then closing them. The container system could get a refactor but I'm not sure it's worth it.

wilsonianb: Thanks for checking :vulcan_salute:

wilsonianb (approved):

packages/auth/docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - "5432:5432"
     environment:
       POSTGRES_PASSWORD: password
+      BYPASS_SIGNATURE_VALIDATION: false
```
mkurapov, line 13: This could be worth an addition to the `docker-compose`s in the `infrastructure/local` folder as well, to make it obvious this can be flipped on and off in the local environment

packages/auth/src/signature/middleware.test.ts
mkurapov, line 415: This might be good to have in a separate function (something that takes in Partial<Config> of sorts) so that anyone won't forget to add the additional container will be properly added to `appContainers`

mkurapov (commented):
Looks good! I realized this was merged right as I pulled it down locally and started to take a look, but I had a few non-blocking comments

mkurapov:
Maybe we could also add this flag to `validateHttpSigHeaders`, or directly as one of the first checks in `grantContinueHttpsigMiddleware` ? otherwise we will still need to provide the http header `signature` and `signature-input` properly formatted, if I'm not mistaken.

Thoughts @mankins @wilsonianb ?

mkurapov:
@wilsonianb I like that idea, keeps it pretty tidy. I wonder if there is a way to add a flag to `createValidatorMiddleware` to skip parts of validation, instead of needing an additional `dummySigHeaderMiddleware`. Maybe somehow setting default values incase this flag is true? https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-default-setter

wilsonianb:
>Maybe somehow setting default values in case this flag is true? https://github.com/kogosoftwarellc/open-api/tree/master/packages/openapi-default-setter

Interesting idea.
Maybe our `openapi`'s `createRequestValidator` could accept an optional set of default params to incorporate into its already existing default setter
https://github.com/interledger/rafiki/blob/889807585763faaf4fc68d9a5d49140776d12501/packages/openapi/src/index.ts#L78-L92

mankins:
> So... should we revert the commit from this pr? 👍 or 👎

I'm fine either way but wanted to comment:

The change to the mechanism that decides to validate signature or not is a small part of this pr:

https://github.com/interledger/rafiki/pull/683/files#diff-649ee9ffe9608a0c6296afe23fa744654797ec47dd45e41c5b4bd99f4195b536R25-R28

Most of it deals with testing that change.

As such you could keep this PR / tests / behavior, and follow on with a modification for how it works with a better implementation, including some ideas by @mkurapov about making it obvious these things can be toggled.

wilsonianb:
I guess it depends on if we choose to put the check in `app.ts` middleware setup, in which case the added tests wouldn't be relevant because the signature middlewares wouldn't be called at all with `bypassSignatureValidation`.

mkurapov:
I think no need to revert for now especially if it's functional, but I'd like to explore the idea of having this be bypass handled in `app.ts` and default parameters as @wilsonianb mentioned

PR #684, opened 2022-10-21
fix(auth): bring in Open API schemas for IdP and RS

adrianhopebailie:
## Changes proposed in this pull request

Import the Open API schemas from the Open Payments project that are Rafiki specific.
Add script to import shared schemas.

## Context

Some of the Open API specs in the Open Payments repo are not supposed to be part of the Open Payments standard. They are implemented by Rafiki and should be in the Rafiki repo.

Fixes https://github.com/interledger/open-payments/issues/198

## Checklist

- [X] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [X] Documentation added
- [ ] Make sure that all checks pass

Issue #630, opened 2022-09-22
Latest AS OpenAPI spec breaking auth middleware tests


wilsonianb:
Grants in the backend [auth middleware tests](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/auth/middleware.test.ts) fail token introspection OpenAPI response validation when using an [`AUTH_SERVER_SPEC`](https://github.com/interledger/rafiki/blob/main/packages/backend/src/config/app.ts#L108-L111) version that includes this commit:
https://github.com/interledger/open-payments/commit/499d551e36fb42614feeadd4aa3af90bb2a4fcff

wilsonianb:
The change to the spec seems to be valid.
The spec (and its examples) pass the [spectral validation workflow](https://github.com/interledger/open-payments/actions/runs/3069867464/jobs/4958993404).
I also successfully validated one the the grants from the auth middleware test against the spec using https://json-schema.hyperjump.io/

This seems to suggest a shortcoming in the [`openapi-response-validator`](https://www.npmjs.com/package/openapi-response-validator) used by our openapi validation middleware
https://github.com/interledger/rafiki/blob/cd9cd2b2cf9e5dd98ac92ff37b1fc4b0540d0eb3/packages/openapi/src/index.ts#L8-L10

wilsonianb:
`openapi-response-validator` appears to use `Ajv` without OpenAPI 3.1's JSON Schema 2020-12
https://github.com/kogosoftwarellc/open-api/blob/master/packages/openapi-response-validator/index.ts#L1-L6
https://ajv.js.org/json-schema.html#draft-2020-12

This seems to prevent keywords such as `const` and `contains` to not be supported.

An attempt to pnpm patch the dependency was unsuccessful
https://github.com/wilsonianb/rafiki/commit/b59fb79db6ae35be94c4cb72db850a40d8fb85c3

wilsonianb:
resolved by https://github.com/interledger/rafiki/pull/692

PR #692, opened 2022-10-25
feat(auth): use split auth specs in app

njlie:

- Updates the AS to use the split openAPI specs.

## Context

With the updated OpenAPI specs, this branch now suffers from https://github.com/interledger/rafiki/issues/630
Also, should be updated to use https://github.com/interledger/rafiki/pull/684 once it is merged in.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


wilsonianb:
> Also, should be updated to use https://github.com/interledger/rafiki/pull/684 once it is merged in.

I vote this waits for #684

sabineschaller:
@njlie Can you please finish that now that #684 has been merged? I want to see if this solves my `response must be null` issue when requesting a grant.

njlie:
> @njlie Want to open a pull request in [interledger/open-payments](https://github.com/interledger/open-payments) removing/fixing whatever `access` restriction are causing problems (maybe just use the `uniqueItems`/`maxItems` for now), and update this to use the specs on `main`?

@wilsonianb https://github.com/interledger/open-payments/pull/208

packages/auth/package.json
```diff
@@ -11,7 +11,7 @@
     "build": "pnpm build:deps && tsc --build tsconfig.json",
     "clean": "rm -fr dist/",
     "fetch-schemas": "./scripts/get-op-schema.sh",
-    "test": "jest --passWithNoTests --maxWorkers=50%",
+    "test": "pnpm fetch-schemas && jest --passWithNoTests --maxWorkers=50%",
```
wilsonianb, line 14: In the spirit of https://github.com/interledger/rafiki/pull/684#discussion_r1009661953, I think we shouldn't cause a local dev to inadvertently nuke their local schemas.

packages/auth/src/index.ts
```diff
@@ -104,7 +104,16 @@ export function initIocContainer(
 
   container.singleton('openApi', async (deps) => {
     const config = await deps.use('config')
-    return await createOpenAPI(config.authServerSpec)
+    const clientSpec = await createOpenAPI(config.authServerSpec)
```
wilsonianb, line 107: What if `get-op-schema.sh` also fetched the auth server spec, and we just used the local file (remove `config.authServerSpec`) as with the other specs?
That way we ensure all the OpenAPI specs in `auth` are using the same version of the schemas.

packages/auth/scripts/get-op-schema.sh
```diff
@@ -10,6 +10,5 @@ done
 OUTDIR=$( cd -P "$( dirname "$SOURCE" )/../openapi" >/dev/null 2>&1 && pwd )
 
 # TODO: revert to using main once https://github.com/interledger/rafiki/issues/630 is resolved
-# curl -o "$OUTDIR/schemas.yaml" https://raw.githubusercontent.com/interledger/open-payments/main/openapi/schemas.yaml
-curl -o "$OUTDIR/schemas.yaml" https://raw.githubusercontent.com/interledger/open-payments/67429f0c7d658c2aaeb59a330be3ec305413f827/openapi/schemas.yaml
-
+curl -o "$OUTDIR/schemas.yaml" https://raw.githubusercontent.com/interledger/open-payments/main/openapi/schemas.yaml
+curl -o "$OUTDIR/auth-server.yaml" https://raw.githubusercontent.com/interledger/open-payments/main/openapi/auth-server.yaml
```
wilsonianb, line 14: I think we should either pin to (current) head of `main`
or support an env var when running `fetch-schemas`, something like:
```suggestion
curl -o "$OUTDIR/schemas.yaml" https://raw.githubusercontent.com/interledger/open-payments/${OPEN_PAYMENTS_COMMIT:-main}/openapi/schemas.yaml
curl -o "$OUTDIR/client.yaml" https://raw.githubusercontent.com/interledger/open-payments/${OPEN_PAYMENTS_COMMIT:-main}/openapi/auth-server.yaml
```

packages/auth/src/index.ts
```diff
@@ -103,8 +103,7 @@ export function initIocContainer(
   })
 
   container.singleton('openApi', async (deps) => {
-    const config = await deps.use('config')
-    const clientSpec = await createOpenAPI(config.authServerSpec)
+    const clientSpec = await createOpenAPI('./openapi/auth-server.yaml')
```
wilsonianb, line 106: What if we rename the fetched AS spec to match the existing local specs:
```suggestion
    const clientSpec = await createOpenAPI('./openapi/client.yaml')
```

njlie: I feel compelled to use the original name since it's being pulled in from somewhere external. The association with that spec's externality seems stronger that way, and it's holding the actual GNAP implementation so I feel auth-server might be a better name anyways.

wilsonianb (approved):

.github/workflows/lint_test_build.yml
```diff
@@ -46,6 +46,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ./.github/workflows/rafiki/env-setup
       - run: pnpm --filter openapi build
+      - run: pnpm --filter auth fetch-schemas
```
wilsonianb, line 49: Actually, I wonder if `fetch-schemas` also needs to be run in the Dockerfile
https://github.com/interledger/rafiki/blob/main/packages/auth/Dockerfile#L18

wilsonianb: Yeah, local env fails as is with:
```
local-peer-auth-1     | {"level":50,"time":1667568836819,"pid":1,"hostname":"f5cbb380f8b7","msg":"ResolverError: Error opening file \"/workspace/openapi/auth-server.yaml\" \nENOENT: no such file or directory, open '/workspace/openapi/auth-server.yaml'\n    at ReadFileContext.callback (/workspace/node_modules/.pnpm/@apidevtools+json-schema-ref-parser@9.0.9/node_modules/@apidevtools/json-schema-ref-parser/lib/resolvers/file.js:52:20)\n    at FSReqCallback.readFileAfterOpen [as oncomplete] (node:fs:323:13)"}
```
I'm thinking instead of adding `fetch-schemas` to the Dockerfile, we just add it to https://github.com/interledger/rafiki#environment-setup
That way localenv can run with modified specs.

packages/auth/src/index.ts
```diff
@@ -102,9 +102,17 @@ export function initIocContainer(
...
 
-  container.singleton('openApi', async (deps) => {
-    const config = await deps.use('config')
-    return await createOpenAPI(config.authServerSpec)
+  container.singleton('openApi', async () => {
+    const authServerSpec = await createOpenAPI('./openapi/auth-server.yaml')
+    const resourceServerSpec = await createOpenAPI(
+      './openapi/resource-server.yaml'
+    )
+    const idpSpec = await createOpenAPI('./openapi/id-provider.yaml')
```
wilsonianb, line 110: `auth` build script needs to add these to the `dist` folder:
https://github.com/interledger/rafiki/blob/889807585763faaf4fc68d9a5d49140776d12501/packages/auth/package.json#L11
See `backend`'s `copy-files`:
https://github.com/interledger/rafiki/blob/889807585763faaf4fc68d9a5d49140776d12501/packages/backend/package.json#L8-L10
(this is causing local env to fail even after running `fetch-schemas`)

wilsonianb (changes requested):

packages/auth/package.json
```diff
@@ -8,11 +8,12 @@
...
     "knex": "knex",
     "build:deps": "pnpm --filter openapi build",
-    "build": "pnpm build:deps && tsc --build tsconfig.json",
+    "build": "pnpm build:deps && tsc --build tsconfig.json && pnpm copy-files",
     "clean": "rm -fr dist/",
     "fetch-schemas": "./scripts/get-op-schema.sh",
     "test": "jest --passWithNoTests --maxWorkers=50%",
-    "prepack": "pnpm build"
+    "prepack": "pnpm build",
+    "copy-files": "pnpm fetch-schemas && cp -r ./openapi ./dist"
```
wilsonianb, line 16:
```suggestion
    "copy-files": "cp -r ./openapi ./dist"
```
Otherwise `copy-files` & `build` nuke local modified specs.

wilsonianb (approved):
:+1: 
local env works as long after running:
```
pnpm --filter auth fetch-schemas
```

PR #706, opened 2022-10-28
663/mk/use open payments client

mkurapov:

- This PR uses the client from the `open-payments` package to make remote open-payments server calls when trying to find a receiver of a payment, either through an `incoming-payment` or a `connection`. This basically incrorporates the client into the existing flow that `OpenPaymentsClientService` had for fetching the remote open-payments resources. Since `OpenPaymentsClientService` will no longer be used in the way that its name suggests, it's been renamed to `Receiver` service, which is an abstraction of an OpenPayments incoming-payment or connection.
- In addition, the incoming payment and connection models now have a `.toOpenPaymentsType` method, that enables entities to go from the domain model to an OpenPayments type, the type that comes from the open api spec type generation in the `open-payments` package.

## Context
 
Progress toward #663, and roadmap item https://github.com/interledger/roadmap/issues/20

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  if (!connection) {
+    return undefined
+  }
+
+  return incomingPayment.toOpenPaymentsType({ ilpStreamConnection: connection })
+}
+
+// Validate referential integrity, which cannot be represented in OpenAPI
+function isValidIncomingPayment(
```
wilsonianb, line 182: Should this be moved into `open-payments`' `IncomingPaymentRoutes`?

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -237,6 +238,30 @@ export class IncomingPayment
     }
     return payment
   }
+
+  public toOpenPaymentsType({
```
wilsonianb, line 242: I expect this can replace a number of existing `toJSON` calls, especially in `routes.ts` files.

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -0,0 +1,212 @@
...
+    return await deps.openPaymentsClient.ilpStreamConnection.get({
+      url,
+      accessToken: deps.accessToken
+    })
+  } catch (_) {
+    return undefined
+  }
+}
+
+function parseIncomingPaymentUrl(
```
wilsonianb, line 105: What do you think about this living in `open-payments` as something like `client.incomingPayment.parseUrl`?

mkurapov: I think I could be swayed to put it in the client later on, especially that it's the one that knows about routes... for now this helper is ok here IMO

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,217 @@
+import { IocContract } from '@adonisjs/fold'
```
mkurapov, line 1: This test looks quite a lot different that the previous OpenPaymentsClientService one, there are some duplicated code but IMO more straightforward

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,217 @@
...
+          }
+        })
+
+        const clientGetIncomingPaymentSpy = jest
+          .spyOn(openPaymentsClient.incomingPayment, 'get')
+          .mockImplementationOnce(async () =>
+            incomingPayment.toOpenPaymentsType({
+              ilpStreamConnection: connectionService.get(incomingPayment)
+            })
+          )
```
mkurapov, line 157: Instead of using `nock` here, I just stub out the return value of the openPaymentsClient call

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -0,0 +1,117 @@
...
+  IncomingPayment as OpenPaymentsIncomingPayment,
+  ILPStreamConnection as OpenPaymentsConnection
+} from 'open-payments'
+import { ConnectionBase } from '../connection/model'
+import { IlpAddress, isValidIlpAddress } from 'ilp-packet'
+
+interface OpenPaymentsConnectionWithIlpAddress
+  extends Omit<OpenPaymentsConnection, 'ilpAddress'> {
+  ilpAddress: IlpAddress
+}
```
mkurapov, line 16: Unfourtunately need this because the base constructor needs to have `ilpAddress: IlpAddress` instead of a `string`. Using static `fromOpenPaymentsConnection` as a wrapper for the constructor, since we can return `undefined` from it directly

packages/backend/src/open_payments/connection/model.ts
```diff
@@ -0,0 +1,74 @@
+import { StreamCredentials } from '@interledger/stream-receiver'
+import base64url from 'base64url'
+import { IlpAddress } from 'ilp-packet'
+import { ILPStreamConnection } from 'open-payments'
+import { IncomingPayment } from '../payment/incoming/model'
+
+export type ConnectionJSON = {
```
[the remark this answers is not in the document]
mkurapov: It is still used in the connection routes

packages/backend/src/open_payments/connection/service.ts
```diff
@@ -1,68 +1,8 @@
...
-export abstract class ConnectionBase {
-  protected constructor(
-    public readonly ilpAddress: IlpAddress,
-    public readonly sharedSecret: Buffer,
-    public readonly assetCode: string,
-    public readonly assetScale: number
-  ) {}
-}
-
-export class Connection extends ConnectionBase {
```
mkurapov, line 25: Moved to its own `model.ts` file

packages/backend/src/open_payments/receiver/model.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  afterAll(async (): Promise<void> => {
+    await appContainer.shutdown()
+  })
+
+  describe('fromIncomingPayment', () => {
+    test('creates receiver', async () => {
+      const paymentPointer = await createPaymentPointer(deps)
+      const incomingPayment = await createIncomingPayment(deps, {
+        paymentPointerId: paymentPointer.id
+      })
```
mkurapov, line 42: We should be able to just stub out the instances here instead of actually hitting the services/dbs for example, but that is for a future improvement

mkurapov:
I merged off of an earlier branch (which is now merged in), hence the many commits

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        const paymentPointer = await createPaymentPointer(deps, {
+          mockServerPort: Config.openPaymentsPort
+        })
+        const { connectionId } = await createIncomingPayment(deps, {
+          paymentPointerId: paymentPointer.id
+        })
+
+        const localUrl = `${Config.openPaymentsUrl}/${CONNECTION_PATH}/${connectionId}`
+
+        await expect(receiverService.get(localUrl)).resolves.toMatchObject({
```
wilsonianb, line 60:
```suggestion
        await expect(receiverService.get(localUrl)).resolves.toEqual({
```
Ditto throughout

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        const incomingPayment = await createIncomingPayment(deps, {
+          paymentPointerId: paymentPointer.id
+        })
+
+        const remoteUrl = new URL(
+          `${paymentPointer.url}/${CONNECTION_PATH}/${incomingPayment.connectionId}`
+        )
+
+        const clientGetConnectionSpy = jest
+          .spyOn(openPaymentsClient.ilpStreamConnection, 'get')
```
wilsonianb, line 82: Could you spy (and not mock) this for the local tests, and check that it isn't called?

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+          url: remoteUrl.href
+        })
+      })
+
+      test('returns undefined for unknown connection', async (): Promise<void> => {
+        const paymentPointer = await createPaymentPointer(deps)
+
+        await expect(
+          receiverService.get(
+            `${paymentPointer.url}/${CONNECTION_PATH}/${uuid()}`
```
wilsonianb, line 108: Also test this with a "remote" url

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        })
+
+        await expect(
+          receiverService.get(
+            `${paymentPointer.url}/${INCOMING_PAYMENT_PATH}/${uuid()}`
+          )
+        ).resolves.toBeUndefined()
+      })
+
+      test('returns undefined when fetching remote incoming payment throws', async (): Promise<void> => {
```
wilsonianb, line 195: Also test this for connection

packages/backend/src/open_payments/client/service.ts
```diff
@@ -1,313 +0,0 @@
-import { Counter, ResolvedPayment } from '@interledger/pay'
-import { createMockContext } from '@shopify/jest-koa-mocks'
```
wilsonianb, line 2: `@shopify/jest-koa-mocks` can be removed as dependency

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -237,6 +238,32 @@ export class IncomingPayment
...
+      incomingAmount: this.incomingAmount
+        ? {
+            ...this.incomingAmount,
+            value: this.incomingAmount.value.toString()
+          }
+        : undefined,
+      receivedAmount: {
+        ...this.receivedAmount,
+        value: this.receivedAmount.value.toString()
+      },
```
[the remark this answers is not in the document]
mkurapov: I will incorporate `serializeAmount` but will hold off removing `AmountJSON`, as it will balloon this PR a bit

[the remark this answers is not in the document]
wilsonianb: I keep having to remind myself that you've managed to avoid changing `open-payments` at all in this pr.

packages/backend/src/open_payments/receiver/model.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      expect(receiver.incomingAmount).toBeUndefined()
+      expect(receiver.receivedAmount).toStrictEqual({
+        assetCode: incomingPayment.asset.code,
+        assetScale: incomingPayment.asset.scale,
+        value: BigInt(0)
+      })
+      expect(receiver.ilpAddress).toEqual(expect.any(String))
+      expect(receiver.sharedSecret).toEqual(expect.any(Buffer))
+      expect(receiver.expiresAt).toEqual(incomingPayment.expiresAt)
```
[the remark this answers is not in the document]
mkurapov: I had that originally, but because it wasn't a normal `Object` but a `Receiver` class, my assertion was failing. Going to try again, maybe I was just using a wrong jest method

packages/backend/src/open_payments/receiver/service.test.ts
mkurapov: Yup, good call

packages/backend/src/open_payments/receiver/model.test.ts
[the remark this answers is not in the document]
mkurapov: Ended up with:
```ts
      expect(receiver).toEqual({
        assetCode: incomingPayment.asset.code,
        assetScale: incomingPayment.asset.scale,
        incomingAmountValue: undefined,
        receivedAmountValue: BigInt(0),
        ilpAddress: expect.any(String),
        sharedSecret: expect.any(Buffer),
        expiresAt: incomingPayment.expiresAt
      })
```

trick was not using the `getters` like `receiver. asset`

wilsonianb (approved):

PR #714, opened 2022-10-31
chore: quote expiration cannot be set past the receiver's expiration

### LINKED ISSUE #593: Do not set quote expiration past receiver expiration
Use receiver's `expiresAt` if it is sooner than `quoteLifespan`
https://github.com/interledger/rafiki/blob/5e08158f445fca5fd8fd640bf0a66a71fd475993/packages/backend/src/open_payments/quote/service.ts#L339

SarahCoilAccount:

- In cases where the receiver is an incoming payment it wouldn't make sense for the quotation's expiration to be set past the expiration date of the incoming payment.

## Context

- Fixes #593

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/client/service.ts
+++ b/packages/backend/src/open_payments/client/service.ts
@@ -54,7 +54,7 @@ export class Receiver extends ConnectionBase {
     connection: ConnectionJSON,
     private readonly incomingAmountValue?: bigint,
     private readonly receivedAmountValue?: bigint,
-    private readonly expiresAtValue?: Date
+    public readonly expiresAt?: Date
   ) {
     super(
       connection.ilpAddress,
@@ -89,13 +89,6 @@ export class Receiver extends ConnectionBase {
         assetCode: this.assetCode,
         assetScale: this.assetScale
       }
-    }
-    return undefined
-  }
-
-  public get expiresAt(): Date | undefined {
-    if (this.expiresAtValue !== undefined) {
-      return this.expiresAtValue
     }
     return undefined
   }
--- a/packages/backend/src/open_payments/client/service.ts
+++ b/packages/backend/src/open_payments/client/service.ts
@@ -31,10 +31,10 @@ export class Receiver extends ConnectionBase {
     if (typeof incomingPayment.ilpStreamConnection !== 'object') {
       return undefined
     }
-    if (
-      incomingPayment.expiresAt &&
-      new Date(incomingPayment.expiresAt).getTime() <= Date.now()
-    ) {
+    const expiryDate = incomingPayment.expiresAt
+      ? new Date(incomingPayment.expiresAt)
+      : undefined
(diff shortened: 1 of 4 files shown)


SarahCoilAccount:
When doing this work I had issues using the Rafiki GraphQL interface to create a quote. I tracked down the problem in the getIncomingPayment function in client/services.ts. I logged the error being caught there and it seems to still be jest related.

```
ReferenceError: jest is not defined
local-backend-1  |     at createContext (/workspace/node_modules/.pnpm/@shopify+jest-koa-mocks@5.0.1/node_modules/@shopify/jest-koa-mocks/build/cjs/create-mock-context/create-mock-context.js:29:22)
local-backend-1  |     at createReadContext (/workspace/packages/backend/dist/open_payments/client/service.js:121:78)
local-backend-1  |     at getIncomingPayment (/workspace/packages/backend/dist/open_payments/client/service.js:168:25)
local-backend-1  |     at runMicrotasks (<anonymous>)
local-backend-1  |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
local-backend-1  |     at async getReceiver (/workspace/packages/backend/dist/open_payments/client/service.js:209:33)
```

packages/backend/src/open_payments/client/service.ts
```diff
@@ -45,14 +45,16 @@ export class Receiver extends ConnectionBase {
...
+      expiryDate
     )
   }
 
   private constructor(
     connection: ConnectionJSON,
     private readonly incomingAmountValue?: bigint,
-    private readonly receivedAmountValue?: bigint
+    private readonly receivedAmountValue?: bigint,
+    private readonly expiresAtValue?: Date
```
wilsonianb, line 57:
```suggestion
    public readonly expiresAt?: Date
```
I don't think this needs the getter below

wilsonianb: Can you update client service / receiver tests to incorporate `expiresAt`?
https://github.com/interledger/rafiki/blob/2c40d222006e5beef2fe850d0208c27c70bfcc69/packages/backend/src/open_payments/client/service.test.ts

wilsonianb (commented):
I wasn't expecting this to affect the client service, but it makes sense.
There'll likely be conflicts between this and:
- https://github.com/interledger/rafiki/pull/706

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -271,6 +271,11 @@ export async function finalizeQuote(
     'Content-Type': 'application/json'
   }
 
+  const receiver = await deps.clientService.receiver.get(quote.receiver)
```
wilsonianb, line 274: We shouldn't need to make another request to get the receiver here.
We could instead pass `receiver` as an argument to `finalizeQuote`.

wilsonianb (approved):

PR #718, opened 2022-11-01
fix(auth): do not check for content-digest in signature if body is empty

njlie:

- Don't check for `Content-Digest` if the body object is empty.

## Context

Even if a body was not provided in the request, koa will default to an empty object.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/signature/middleware.test.ts
+++ b/packages/auth/src/signature/middleware.test.ts
@@ -70,36 +70,6 @@ describe('Signature Service', (): void => {
       ).resolves.toBe(true)
     })
 
-    // test('can construct a challenge from signature input', (): void => {
-    //   const sigInputHeader =
-    //     'sig1=("@method" "@target-uri" "content-digest" "content-length" "content-type" "authorization");created=1618884473;keyid="gnap-key"'
-    //   const ctx = createContext(
-    //     {
-    //       headers: {
-    //         'Content-Type': 'application/json',
-    //         'Content-Digest': 'sha-256=:test-hash:',
-    //         'Content-Length': '1234',
-    //         'Signature-Input': sigInputHeader,
-    //         Authorization: 'GNAP test-access-token'
-    //       },
-    //       method: 'GET',
-    //       url: '/test'
-    //     },
-    //     {},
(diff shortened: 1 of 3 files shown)


packages/auth/src/signature/middleware.test.ts
```diff
@@ -70,35 +70,89 @@ describe('Signature Service', (): void => {
...
-            'Signature-Input': sigInputHeader,
-            Authorization: 'GNAP test-access-token'
-          },
-          method: 'GET',
-          url: '/test'
-        },
-        {},
-        deps
-      )
+    // test('can construct a challenge from signature input', (): void => {
```
wilsonianb, line 73: remove?

packages/auth/src/signature/middleware.ts
```diff
@@ -80,7 +80,9 @@ function validateSigInputComponents(
   return !(
     !sigInputComponents.includes('@method') ||
     !sigInputComponents.includes('@target-uri') ||
-    (ctx.request.body && !sigInputComponents.includes('content-digest')) ||
+    (ctx.request.body &&
+      Object.keys(ctx.request.body).length > 0 &&
```
wilsonianb, line 84: Afaict, this is is due to `koa-bodyparser`
https://github.com/koajs/bodyparser#usage
> if nothing was parsed, body will be an empty object {}

Would it be worth also enabling [`strict` mode](https://github.com/koajs/bodyparser#options)?
https://github.com/interledger/rafiki/blob/1b03f3e766735ec7c9618bc044838364723ad4f2/packages/auth/src/app.ts#L173

njlie: Looks like `strict` mode is true by default. So that specifically would be what's causing this behavior, I think.

wilsonianb: >Default is `true`

Oops.
I wasn't thinking about avoiding this behavior but making sure our new `Object.keys(ctx.request.body).length` is always dealing with an object.

wilsonianb (approved):

PR #719, opened 2022-11-01
feat(map): display account transactions

### LINKED ISSUE #717: Display account transactions

wilsonianb:

- add `/accounts/{accountId}` page with incoming and outgoing account transactions to mock account provider
- as is, payments are queried via the admin api and not stored by the mock account provider (which makes it difficult to pre-seed transactions...)

## Context

- fixes #717

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/mock-account-provider/app/lib/transactions.server.ts
+++ b/packages/mock-account-provider/app/lib/transactions.server.ts
@@ -1,0 +1,72 @@
+import { mockAccounts } from './accounts.server'
+import { getPaymentPointerPayments } from './requesters'
+
+export interface Amount {
+  value: bigint
+  assetCode: string
+  assetScale: number
+}
+
+export type IncomingPayment = {
+  id: string
+  description: string
+  externalRef: string
+  incomingAmountValue: string
+  receivedAmountValue: string
+  assetCode: string
+  assetScale: number
+  state: string
+  createdAt: string
+}
+
+export type OutgoingPayment = {
+  id: string
+  receiver: string
+  description: string
+  externalRef: string
+  sendAmountValue: string
+  sentAmountValue: string
+  receiveAmount: Amount
+  assetCode: string
+  assetScale: number
+  state: string
+  createdAt: string
+}
+
+export async function getAccountTransactions(
+  accountId: string
+): Promise<Array<IncomingPayment | OutgoingPayment>> {
+  const account = await mockAccounts.get(accountId)
+  const { incomingPayments, outgoingPayments } =
+    await getPaymentPointerPayments(account.paymentPointerID)
+  const transactions = incomingPayments.edges.map(({ node }) => {
+    return {
+      id: node.id,
+      description: node.description,
+      externalRef: node.externalRef,
+      incomingAmountValue: node.incomingAmount.value,
+      receivedAmountValue: node.receivedAmount.value,
+      assetCode: node.receivedAmount.assetCode,
+      assetScale: node.receivedAmount.assetScale,
+      state: node.state,
+      createdAt: node.createdAt
+    }
+  })
+  return transactions.concat(
+    outgoingPayments.edges.map(({ node }) => {
+      return {
+        id: node.id,
+        receiver: node.receiver,
+        description: node.description,
+        externalRef: node.externalRef,
+        sendAmountValue: node.sendAmount.value,
+        sentAmountValue: node.sentAmount.value,
+        receiveAmount: node.receiveAmount,
+        assetCode: node.sendAmount.assetCode,
+        assetScale: node.sendAmount.assetScale,
+        state: node.state,
+        createdAt: node.createdAt
+      }
+    })
+  )
+}
--- a/packages/mock-account-provider/app/lib/transactions.server.ts
+++ b/packages/mock-account-provider/app/lib/transactions.server.ts
@@ -7,30 +7,31 @@ export interface Amount {
   assetScale: number
 }
 
-export type IncomingPayment = {
-  id: string
-  description: string
-  externalRef: string
-  incomingAmountValue: string
-  receivedAmountValue: string
-  assetCode: string
-  assetScale: number
-  state: string
-  createdAt: string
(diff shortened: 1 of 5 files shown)


packages/mock-account-provider/app/routes/accounts.$accountId.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <tr>
+          <th>Date</th>
+          <th>Type</th>
+          <th>Amount</th>
+          <th>Description</th>
+        </tr>
+        {transactions.map((trx) => (
+          <tr key={trx.id}>
+            <td>{trx.createdAt}</td>
+            <td>{trx.sendAmountValue ? 'Outgoing' : 'Incoming'}</td>
```
mkurapov, line 31: this is a bit implicit determination, IMO would be better to programatically set a `type` on each payment or something similar

packages/mock-account-provider/app/routes/accounts.$accountId.tsx
```diff
@@ -0,0 +1,48 @@
...
+            <td>
+              {(
+                Number(trx.sendAmountValue ?? trx.incomingAmountValue) / 100
```
mkurapov, line 34: Do we want to show `receivedAmountValue` at all here?

Just for my own understanding: `incomingAmountValue` is the amount originally the incoming payment was created for, but  `receivedAmountValue` was the actual amount transferred?

mkurapov: Also, as an aside, can I partially fulfill an incoming payment, create a "partial" quote, create an outgoing payment, then try creating a quote again for the same incoming payment?

mkurapov (commented):
Looks good, just a few comments/clarifications

packages/mock-account-provider/app/routes/accounts.$accountId.tsx
wilsonianb: > Do we want to show `receivedAmountValue` at all here?

Yes, this should use it instead of `incomingAmountValue` because :point_down: 
> Just for my own understanding: `incomingAmountValue` is the amount originally the incoming payment was created for, but `receivedAmountValue` was the actual amount transferred?

Yes, and `incomingAmount` is optional, an incoming payment can be created without it.

>Also, as an aside, can I partially fulfill an incoming payment, create a "partial" quote, create an outgoing payment, then try creating a quote again for the same incoming payment?

Yes

packages/mock-account-provider/app/lib/transactions.server.ts
```diff
@@ -0,0 +1,72 @@
+import { mockAccounts } from './accounts.server'
+import { getPaymentPointerPayments } from './requesters'
+
+export interface Amount {
+  value: bigint
+  assetCode: string
+  assetScale: number
+}
```
sabineschaller, line 8: Since I also added that in [my last MAP PR](https://github.com/interledger/rafiki/blob/24f486145000fb908d712ebe43bca10cab01313a/packages/mock-account-provider/app/routes/webhooks.ts#L22-L26), should we share that?

packages/mock-account-provider/app/routes/accounts.$accountId.tsx
sabineschaller: I agree, we should use `sentAmount` and `receivedAmount`.

sabineschaller (commented):
Thank you for adding this!

mkurapov:
More for my understanding, why do we do this for an outgoing payment completion? 
```ts
  const amtSend = parseAmount(payment['sendAmount'])
  const amtSent = parseAmount(payment['sentAmount'])

  const toVoid = amtSend.value - amtSent.value

  await mockAccounts.debit(acc.id, amtSent.value, true)
  if (toVoid > 0) {
    await mockAccounts.voidPendingDebit(acc.id, toVoid)
  }
```

ie where does `amtSend` come into play here? If I was sending money to a user, and an incoming payment on their account was for $10, then I was quoted $15 to send $10, would I have to pay $15 on the outgoing payment for them to receive $10? In other words, when am I actually charged $15?

wilsonianb:
>ie where does amtSend come into play here? If I was sending money to a user, and an incoming payment on their account was for $10, then I was quoted $15 to send $10 (sendAmount = $15, receiveAmount = $10), when would I have to pay $15 on the outgoing payment for them to receive $10? In other words, when am I actually charged $15?

Preface: an outgoing payment's `sendAmount` is always equal to its quote's `sendAmount` (ditto for `receiveAmount`)

An outgoing payment's `sendAmount` isn't necessarily relative to the incoming payment's `incomingAmount` because:
- incoming payment might not have an `incomingAmount`
- the quote could have been for created with a specified `sendAmount` or `receiveAmount` instead of for the incoming payment's `incomingAmount`

If the quote _was_ created without a specified `sendAmount` or `receiveAmount` it _would_ use the incoming payment's `incomingAmount` as the `receiveAmount`.
However the quote's `sendAmount` could end up not equally `incomingAmount` due to
- currency conversion
- the sending account provider increasing the calculated `sendAmount` to include a fee

mkurapov:
Thanks for the explanation @wilsonianb ! On the last part
> However, the quote's sendAmount could end up not equally incomingAmount due to
> - currency conversion
> - the sending account provider increasing the calculated sendAmount to include a fee

I'm not 100% clear is why the difference isn't reflected in the balances for the mock account provider

wilsonianb:
>I'm not 100% clear is why the difference isn't reflected in the balances for the mock account provider

I think the mock account provider currently isn't "claiming the fee".
On a completed outgoing payment, it should debit the full `sendAmt`.

mkurapov (approved):
LGTM, it would be good to have info about the original incomingPayment's amount later on, since it might be misleading initially to create an incoming payment for an `incomingAmount.value = $10` but then seeing $0 on the page until the payment is completed

sabineschaller (approved):

PR #672, opened 2022-10-13
feat(backend): validate httpsig on rs requests using auth implementation

dclipp:
- Export `verifySig` from `auth` package
- use `auth`'s `verifySig` implementation to validate the signature received after and introspection request

## Changes proposed in this pull request

- fixes #585 

## Checklist

- [X] Related issues linked using `fixes #number`
- [X] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass

PR #722, opened 2022-11-01
feat(open-payments): add HTTP message signatures

### LINKED ISSUE #587: Generate httpsig for internal OP calls
Any interactions that occur where Rafiki calls out to any open payments api's needs to actually generate the httpsig signature correctly and supply it in the headers along with the GNAP auth token.

wilsonianb:

- 

## Context

- fixes #587 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/open-payments/src/client/index.ts
+++ b/packages/open-payments/src/client/index.ts
@@ -1,3 +1,4 @@
+import { KeyLike } from 'crypto'
 import { createOpenAPI, OpenAPI } from 'openapi'
 import createLogger, { Logger } from 'pino'
 import config from '../config'
@@ -19,6 +20,8 @@ import { AxiosInstance } from 'axios'
 export interface CreateOpenPaymentClientArgs {
   requestTimeoutMs?: number
   logger?: Logger
+  privateKey: KeyLike
+  keyId: string
 }
 
 export interface ClientDeps {
@@ -34,9 +37,11 @@ export interface OpenPaymentsClient {
 }
 
 export const createClient = async (
-  args?: CreateOpenPaymentClientArgs
+  args: CreateOpenPaymentClientArgs
 ): Promise<OpenPaymentsClient> => {
   const axiosInstance = createAxiosInstance({
+    privateKey: args.privateKey,
+    keyId: args.keyId,
     requestTimeoutMs:
       args?.requestTimeoutMs ?? config.DEFAULT_REQUEST_TIMEOUT_MS
   })
--- a/packages/open-payments/src/client/requests.test.ts
+++ b/packages/open-payments/src/client/requests.test.ts
@@ -1,28 +1,35 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import { createAxiosInstance, get } from './requests'
+import { generateKeyPairSync } from 'crypto'
 import nock from 'nock'
 import { mockOpenApiResponseValidators, silentLogger } from '../test/helpers'
 
 describe('requests', (): void => {
   const logger = silentLogger
+  const privateKey = generateKeyPairSync('ed25519').privateKey
+  const keyId = 'myId'
 
   describe('createAxiosInstance', (): void => {
     test('sets timeout properly', async (): Promise<void> => {
       expect(
-        createAxiosInstance({ requestTimeoutMs: 1000 }).defaults.timeout
+        createAxiosInstance({ requestTimeoutMs: 1000, privateKey, keyId })
+          .defaults.timeout
       ).toBe(1000)
     })
     test('sets Content-Type header properly', async (): Promise<void> => {
       expect(
-        createAxiosInstance({ requestTimeoutMs: 0 }).defaults.headers.common[
-          'Content-Type'
-        ]
+        createAxiosInstance({ requestTimeoutMs: 0, privateKey, keyId }).defaults
+          .headers.common['Content-Type']
       ).toBe('application/json')
     })
   })
 
   describe('get', (): void => {
-    const axiosInstance = createAxiosInstance({ requestTimeoutMs: 0 })
+    const axiosInstance = createAxiosInstance({
+      requestTimeoutMs: 0,
+      privateKey,
+      keyId
+    })
     const baseUrl = 'http://localhost:1000'
     const responseValidators = mockOpenApiResponseValidators()
 
@@ -30,8 +37,37 @@ describe('requests', (): void => {
       jest.spyOn(axiosInstance, 'get')
     })
 
+    afterEach(() => {
+      jest.useRealTimers()
+    })
+
     test('sets headers properly if accessToken provided', async (): Promise<void> => {
-      nock(baseUrl).get('/incoming-payment').reply(200)
+      // https://github.com/nock/nock/issues/2200#issuecomment-1280957462
+      jest
+        .useFakeTimers({
+          doNotFake: [
+            'nextTick',
+            'setImmediate',
+            'clearImmediate',
+            'setInterval',
+            'clearInterval',
+            'setTimeout',
+            'clearTimeout'
+          ]
+        })
+        .setSystemTime(new Date())
+
+      const scope = nock(baseUrl)
+        .matchHeader('Signature', /sig1=:([a-zA-Z0-9+/]){86}==:/)
+        .matchHeader(
+          'Signature-Input',
+          `sig1=("@method" "@target-uri" "authorization");created=${Math.floor(
+            Date.now() / 1000
+          )};keyid="${keyId}";alg="ed25519"`
+        )
+        .get('/incoming-payment')
+        // TODO: verify signature
+        .reply(200)
 
       await get(
         { axiosInstance, logger },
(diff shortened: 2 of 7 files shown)


packages/open-payments/src/client/requests.test.ts
```diff
@@ -1,37 +1,73 @@
...
+      const scope = nock(baseUrl)
+        .matchHeader('Signature', /sig1=:([a-zA-Z0-9+/]){86}==:/)
+        .matchHeader(
+          'Signature-Input',
+          `sig1=("@method" "@target-uri" "authorization");created=${Math.floor(
+            Date.now() / 1000
+          )};keyid="${keyId}";alg="ed25519"`
+        )
+        .get('/incoming-payment')
+        // TODO: verify signature
```
wilsonianb, line 69: Should we move the meat of `auth`'s [signature validation](https://github.com/interledger/rafiki/blob/main/packages/auth/src/signature/middleware.ts) into `open-payments` (after #672 is merged)?

sabineschaller: I'd say yes.

packages/open-payments/src/client/requests.ts
```diff
@@ -65,12 +65,36 @@ export const get = async <T>(
 
 export const createAxiosInstance = (args: {
   requestTimeoutMs: number
+  privateKey: KeyLike
+  keyId: string
```
sabineschaller, line 69: This is the Rafiki instance's private key, right? Just wanted to clarify.

packages/open-payments/src/client/requests.test.ts
sabineschaller: What do you mean by `open-payments`?

packages/open-payments/src/client/signatures.ts
```diff
@@ -0,0 +1,52 @@
...
+    },
+    keyId,
+    signer: createSigner(privateKey),
+    format: 'httpbis'
+  })
+  return {
+    Signature: headers['Signature'] as string,
+    'Signature-Input': headers['Signature-Input'] as string
+  }
+}
```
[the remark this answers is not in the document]
mkurapov: @sabineschaller 
This is directly in the `open-payments` lib, we just have it as a sibling package to 
`auth`/`backend`/etc for now

packages/open-payments/src/client/requests.ts
wilsonianb: correct

packages/open-payments/src/client/requests.test.ts
wilsonianb: https://github.com/interledger/rafiki/tree/main/packages/open-payments

packages/open-payments/src/client/requests.ts
```diff
@@ -65,12 +65,36 @@ export const get = async <T>(
...
+        privateKey: args.privateKey,
+        keyId: args.keyId
+      })
+      config.headers['Signature'] = sigHeaders['Signature']
+      config.headers['Signature-Input'] = sigHeaders['Signature-Input']
+      return config
+    },
+    null,
+    {
+      runWhen: (config) => !!config.headers['Authorization']
```
mkurapov, line 95: Because we don't need `Authorization` header on the AS calls, this would effectively handle both AS and RS with no problem, correct?

packages/open-payments/src/client/requests.ts
```diff
@@ -65,12 +65,36 @@ export const get = async <T>(
...
+  privateKey: KeyLike
+  keyId: string
 }): AxiosInstance => {
   const axiosInstance = axios.create({
     timeout: args.requestTimeoutMs
   })
-
   axiosInstance.defaults.headers.common['Content-Type'] = 'application/json'
 
+  axiosInstance.interceptors.request.use(
```
mkurapov, line 76: Confirming: does this apply to all GET, POST, DELETE requests?

packages/open-payments/src/client/requests.ts
wilsonianb: I think there _are_ some AS calls that require the `Authorization` header
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-7.3.1.1

I was actually thinking that the grant initiation request might be the only one with it, but I'm not sure

packages/open-payments/src/client/requests.ts
wilsonianb: I believe so, but it can be filtered.

packages/open-payments/src/client/requests.ts
mkurapov: Does that mean that the auth server open payment specs are missing the `signature-input` and `signature` header parameters in the open api spec files? If they need to provide an `Authorization` token?

wilsonianb: Yeah, looks like that's the case.
I just opened:
- https://github.com/interledger/open-payments/issues/209

packages/open-payments/src/client/index.ts
```diff
@@ -19,6 +20,8 @@ import { AxiosInstance } from 'axios'
 export interface CreateOpenPaymentClientArgs {
   requestTimeoutMs?: number
   logger?: Logger
+  privateKey: KeyLike
+  keyId: string
```
wilsonianb, line 24: - https://github.com/interledger/rafiki/issues/731

Thinking about :point_up_2:, if there are clients that only want to use `open-payments` to query payment pointers and/or [their keys](https://docs.openpayments.guide/reference/get-payment-pointer-keys) and/or connections, it seems like it should be possible to create a client without requiring `privateKey`/`keyId`.

wilsonianb: Should `open-payments` `createClient` be split into:
- `createAuthenticatedClient`
- `createUnauthenticatedClient`
?
The latter would return the subset of requests that are possible without httpsig.

packages/open-payments/src/client/requests.ts
```diff
@@ -65,12 +65,36 @@ export const get = async <T>(
...
+    async (config) => {
+      const sigHeaders = await createSignatureHeaders({
+        request: {
+          method: config.method.toUpperCase(),
+          url: config.url,
+          // https://github.com/axios/axios/issues/5089#issuecomment-1297761617
+          headers: JSON.parse(JSON.stringify(config.headers)),
```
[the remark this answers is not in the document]
wilsonianb: At some point, I was getting the error described in that axios github issue.
But now I can't reproduce. (Tests pass with `config.headers)
I'll tentatively remove the workaround.

packages/open-payments/src/client/requests.test.ts
wilsonianb: I'm thinking the "receiver"/RS will also be using `open-payments` at least for the types (why I was reluctant to call `open-payments` a "client" library), so why not also provide httpsig verification?

sabineschaller: @wilsonianb wouldn't the RS also use it for sig creation when requesting resources at other RS's?

packages/open-payments/src/client/requests.test.ts
```diff
@@ -1,61 +1,101 @@
...
+      scope.done()
+
       expect(axiosInstance.get).toHaveBeenCalledWith(
         `${baseUrl}/incoming-payment`,
         {
           headers: {
-            Authorization: 'GNAP accessToken',
-            Signature: 'TODO',
-            'Signature-Input': 'TODO'
+            Authorization: 'GNAP accessToken'
```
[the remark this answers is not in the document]
sabineschaller: Can we add some sort of exception for this?

packages/backend/src/config/app.ts
```diff
@@ -112,6 +112,7 @@ export const Config = {
     'AUTH_SERVER_SPEC',
     'https://raw.githubusercontent.com/interledger/open-payments/77462cd0872be8d0fa487a4b233defe2897a7ee4/auth-server-open-api-spec.yaml'
   ),
+  keyId: envString('KEY_ID', 'rafiki'),
```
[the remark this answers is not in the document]
wilsonianb: Can you elaborate?

packages/open-payments/src/client/requests.test.ts
wilsonianb: Correct, the RS will also use ~`createClient`~`createAuthenticatedClient` to send requests.

packages/open-payments/src/client/requests.test.ts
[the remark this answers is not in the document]
wilsonianb: Do you mean for codeql?
We could open an issue to exclude test files in https://github.com/interledger/rafiki/blob/main/.github/workflows/codeql-analysis.yml

mkurapov (approved):

packages/open-payments/src/client/signatures.ts
```diff
@@ -0,0 +1,52 @@
...
+  return signer
+}
+
+export const createSignatureHeaders = async ({
+  request,
+  privateKey,
+  keyId
+}: SignOptions): Promise<SignatureHeaders> => {
+  const components = ['@method', '@target-uri']
+  if (request.headers['Authorization']) {
```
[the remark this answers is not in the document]
wilsonianb: It appears to matter.
The [`sets headers properly if accessToken provided`](https://github.com/interledger/rafiki/pull/722/files#diff-1195d882d785bd3f5b02a6e61335bdcd07f01e527af0c25b03221ab70cfb0fe5R44) test fails if I make this lower case.

mkurapov (approved):

packages/open-payments/src/client/requests.test.ts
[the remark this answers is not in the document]
sabineschaller: Let's do that.

packages/backend/src/config/app.ts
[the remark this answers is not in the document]
sabineschaller: I was just wondering if `rafiki` is a good value for the default. But now that I have thought a bit more about it, I think it is.

sabineschaller (approved):

packages/open-payments/src/client/requests.test.ts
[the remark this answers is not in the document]
wilsonianb: - https://github.com/interledger/rafiki/issues/750

packages/open-payments/src/client/requests.ts
mkurapov: @wilsonianb for my own understanding, wouldn't grant initiation be the only AS request _without_ the `Authorization: GNAP {accessToken}` header?

wilsonianb: yeah, I think that's correct

PR #725, opened 2022-11-03
chore(auth): Remove nock .isDone

### LINKED ISSUE #656: Audit tests using nock `.isDone`

mankins:

- Removes the `scope.isDone()` from tests

## Context

- Fixes #656 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

## Discussion

In auditing `scope.isDone()` for #656 I was assuming we want to convert

```
scope.isDone()
```

to an assertion:

```
expect(scope.isDone()).toBe(true)
```

When I re-ran the tests I found that many were failing. Digging a little further to get more information, I replaced the isDone with:

```
   expect(scope.pendingMocks()).toEqual([])
```

To get a better idea of what was going on. I haven't used `nock` much, but found that it's not likely working in the way imagined / in a parallel testing environment:

https://github.com/nock/nock/issues/30#issuecomment-3826860

Consequently I recommend removing the `scope.isDone()` usage because even as an assertion other tests could influence the "done-ness" and cause this to fail. (In a parallel testing environment you can have multiple nocks for the same url in different tests and the `isDone()` function looks to see if there are any pending requests which appear to be true for some tests, even though the test's scope is finished.)

This PR removes the isDone() usage because it wasn't doing anything before and doesn't work reliably for future use so doesn't actually provide information for tests.

### DIFF
--- a/packages/auth/src/grant/routes.test.ts
+++ b/packages/auth/src/grant/routes.test.ts
@@ -181,7 +181,7 @@ describe('Grant Routes', (): void => {
     })
 
     test('Can initiate a grant request', async (): Promise<void> => {
-      const scope = nock(KEY_REGISTRY_ORIGIN)
+      nock(KEY_REGISTRY_ORIGIN)
         .get(KID_PATH)
         .reply(200, {
           ...TEST_CLIENT_KEY.jwk,
@@ -220,12 +220,10 @@ describe('Grant Routes', (): void => {
           wait: Config.waitTimeSeconds
         }
       })
-
-      scope.isDone()
     })
 
     test('Can get a software-only authorization grant', async (): Promise<void> => {
-      const scope = nock(KEY_REGISTRY_ORIGIN)
+      nock(KEY_REGISTRY_ORIGIN)
         .get(KID_PATH)
         .reply(200, {
           ...TEST_CLIENT_KEY.jwk,
@@ -279,14 +277,12 @@ describe('Grant Routes', (): void => {
           uri: expect.any(String)
         }
       })
-
-      scope.isDone()
     })
     test('Does not create grant if token issuance fails', async (): Promise<void> => {
       jest
         .spyOn(accessTokenService, 'create')
         .mockRejectedValueOnce(new Error())
-      const scope = nock(KEY_REGISTRY_ORIGIN)
+      nock(KEY_REGISTRY_ORIGIN)
         .get(KID_PATH)
         .reply(200, {
           ...TEST_CLIENT_KEY.jwk,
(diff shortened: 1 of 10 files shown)


wilsonianb:
> https://github.com/nock/nock/issues/30#issuecomment-3826860

I had no idea `nock` was rocking back in 2012#683 

`scope.done()` should work as an alternative to `expect(scope.isDone()).toBe(true)`
https://github.com/nock/nock#expectations

>In a parallel testing environment you can have multiple nocks for the same url in different tests

I'd prefer to try making urls unique (where necessary).
I think checking that an expected request was made is important for some (but not necessarily all of these) tests.

Was it just a subset of these that had failing tests?
If you want, you could just limit this pr to tests that could successfully be converted to `scope.done()`.

mankins:
> Was it just a subset of these that had failing tests?

It was a subset, usually the ones that failed so something couldn't continue.

>If you want, you could just limit this pr to tests that could successfully be converted to scope.done().

Ok. `scope.done()` looks to be a bit of sugar around `isDone` (adds the assertion) so I'm not sure if it truly scopes as you'd expect because of how nock implemented `isDone` looking at "pending". I'll give it a check when I have a moment though.

mankins:
Here's the `scope.done` conversions that don't (on my machine) cause an error. I backed out the `backend` (ha) changes.

packages/auth/src/grant/routes.test.ts
```diff
@@ -384,7 +384,7 @@ describe('Grant Routes', (): void => {
         ).resolves.toBeUndefined()
         expect(ctx.status).toBe(401)
         expect(ctx.body).toEqual({ error: 'unknown_request' })
-        scope.isDone()
+        // scope.done() // TODO: fails
```
wilsonianb, line 387:
```suggestion
```
This tests `nock` can be removed

packages/auth/src/grant/routes.test.ts
```diff
@@ -458,7 +458,7 @@ describe('Grant Routes', (): void => {
         expect(redirectSpy).toHaveBeenCalledWith(redirectUrl.toString())
         expect(ctx.session.nonce).toEqual(grant.interactNonce)
 
-        scope.isDone()
+        // scope.done() // TODO: fails
```
wilsonianb, line 461: This test _should_ work once `nock` is removed from the test above
```suggestion
        scope.done()
```

packages/auth/src/signature/middleware.test.ts
```diff
@@ -370,11 +370,11 @@ describe('Signature Service', (): void => {
       expect(ctx.response.body.error).toEqual('invalid_request')
       expect(ctx.response.body.message).toEqual('invalid signature headers')
 
-      scope.isDone()
+      // scope.done() // TODO: fails
```
wilsonianb, line 373:
```suggestion
```
remove `nock` from test

packages/auth/src/signature/middleware.test.ts
```diff
@@ -402,7 +402,7 @@ describe('Signature Service', (): void => {
         grantContinueHttpsigMiddleware(ctx, next)
       ).rejects.toHaveProperty('status', 401)
 
-      scope.isDone()
+      // scope.done() // TODO: fails
```
wilsonianb, line 405: This should work when test above is fixed.
```suggestion
      scope.done()
```

PR #727, opened 2022-11-04
Adding AS types & grant requests support to open-payments client

mkurapov:

- Generate open payment AS open APIs types, export them
- Enable open payments client functionality for interactive and non-interactive grant creation (using AS spec under the hood)

## Context

- Continued work for the open payments client, roadmap item: https://github.com/interledger/roadmap/issues/20. 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/open-payments/src/config.ts
```diff
@@ -1,5 +1,7 @@
 export default {
-  OPEN_PAYMENTS_OPEN_API_URL:
-    'https://raw.githubusercontent.com/interledger/open-payments/7bb2e6a03d7dfe7ecb0553afb6c70741317bb489/openapi/RS/openapi.yaml',
+  OPEN_PAYMENTS_RS_OPEN_API_URL:
+    'https://raw.githubusercontent.com/interledger/open-payments/1e3e118d8b22c5d2942f972e28ebf0f0114d04f5/openapi/resource-server.yaml',
+  OPEN_PAYMENTS_AS_OPEN_API_URL:
+    'https://raw.githubusercontent.com/interledger/open-payments/1e3e118d8b22c5d2942f972e28ebf0f0114d04f5/openapi/auth-server.yaml',
```
mkurapov, line 5: Originally, I was thinking of doing something like
```suggestion
  COMMIT_HASH: '1e3e118d8b22c5d2942f972e28ebf0f0114d04f5',
  OPEN_PAYMENTS_RS_OPEN_API_URL:
    'https://raw.githubusercontent.com/interledger/open-payments/<commitHash>/openapi/resource-server.yaml',
  OPEN_PAYMENTS_AS_OPEN_API_URL:
    'https://raw.githubusercontent.com/interledger/open-payments/<commitHash>/openapi/auth-server.yaml',
```
and then replace the `<commitHash>` whenever `OPEN_PAYMENTS_AS_OPEN_API_URL` was grabbed, but I thought overkill with only two configs

packages/open-payments/src/types.ts
```diff
@@ -1,9 +1,36 @@
-import { components, paths as Paths } from './generated/types'
+import {
```
mkurapov, line 1: decided to keep them in a single file, we don't have _too_ many types

packages/open-payments/tsconfig.json
```diff
@@ -7,5 +7,5 @@
     "declaration": true
   },
   "include": ["src/**/*"],
-  "exclude": ["**/*.test.ts", "src/scripts/*", "src/test/*"]
```
mkurapov, line 10: we don't have a `src/scripts` folder

packages/open-payments/src/client/grant.ts
```diff
@@ -0,0 +1,58 @@
...
+  const createInteractiveGrantValidator =
+    openApi.createResponseValidator<InteractiveGrant>({
+      path: getASPath('/'),
+      method: HttpMethod.POST
+    })
+  const createNonInteractiveGrantValidator =
+    openApi.createResponseValidator<NonInteractiveGrant>({
+      path: getASPath('/'),
+      method: HttpMethod.POST
+    })
```
mkurapov, line 38: unfourtunately, needed both to pass type checking when given as arguments to the `request*Grant` methods

packages/open-payments/src/client/grant.ts
```diff
@@ -0,0 +1,58 @@
...
+
+interface RequestGrantArgs<T> {
+  url: string
+  request: T
+}
+
+export interface GrantRoutes {
+  requestInteractiveGrant(
+    args: RequestGrantArgs<InteractiveGrantRequest>
+  ): Promise<InteractiveGrant>
```
wilsonianb, line 20: My concern with splitting `requestInteractiveGrant`/`requestNonInteractiveGrant` is that I don't _think_ we've standardized what grant requests are interactive. And thus, a client might not be expected to know when to use either.
For example, Rafiki has made `incoming-payment`-only grants non-interactive, but I don't know that that's required:
- https://github.com/interledger/rafiki/pull/588

@sabineschaller or @njlie  or @adrianhopebailie might be able to correct me on this.

A compromise could be to do:
```suggestion
  requestInteractiveGrant(
    args: RequestGrantArgs<InteractiveGrantRequest>
  ): Promise<InteractiveGrant | NonInteractiveGrant>
```

packages/open-payments/src/config.ts
wilsonianb, line 5:
```suggestion
  OPEN_PAYMENTS_RS_OPEN_API_URL:
    'https://raw.githubusercontent.com/interledger/open-payments/62c4b4a9875e3adaa21f89f597e88db43016fe0b/openapi/resource-server.yaml',
  OPEN_PAYMENTS_AS_OPEN_API_URL:
    'https://raw.githubusercontent.com/interledger/open-payments/62c4b4a9875e3adaa21f89f597e88db43016fe0b/openapi/auth-server.yaml',
```

packages/open-payments/src/client/grant.ts
wilsonianb: `requestNonInteractiveGrant` seems fine as is, since the request should just fail if the AS requires RO interaction for that grant.

mkurapov: I separated these, since at some point, I thought the client need to know that they need to pass in `interact` ? 
https://github.com/interledger/open-payments/blob/main/openapi/auth-server.yaml#L93-L94

wilsonianb: Given our mix of grants for different resources needing interact or not, I've been assuming a client might just always assume a grant request requires interaction, but also be able to handle getting back a `NonInteractiveGrant` response.

packages/open-payments/src/client/grant.ts
wilsonianb: If we keep `requestInteractiveGrant`/`requestNonInteractiveGrant` separate, I think we'll need custom validators that include the `openApi.createResponseValidator` as well as a subsequent type guard that the response is the expected `InteractiveGrant` or `NonInteractiveGrant`. (In which case, we might be able to have a single `openApi.createResponseValidator` for both?)
As is, I think the OpenAPI validator would allow a `NonInteractiveGrant` response for `requestInteractiveGrant` and vice versa.

packages/open-payments/src/client/grant.ts
```diff
@@ -0,0 +1,58 @@
...
+  requestNonInteractiveGrant(
+    args: RequestGrantArgs<NonInteractiveGrantRequest>
+  ): Promise<NonInteractiveGrant>
+}
+
+export const createGrantRoutes = (clientDeps: ClientDeps): GrantRoutes => {
```
wilsonianb, line 26: We'll also need the grant continuation route, in order to support the full interactive grant flow.
https://github.com/interledger/open-payments/blob/62c4b4a9875e3adaa21f89f597e88db43016fe0b/openapi/auth-server.yaml#L163-L171

packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -34,11 +34,11 @@ export const createIncomingPaymentRoutes = (
 }
 
 export const getIncomingPayment = async (
-  clientDeps: Pick<ClientDeps, 'axiosInstance' | 'logger'>,
+  deps: Pick<RouteDeps, 'axiosInstance' | 'logger'>,
```
wilsonianb, line 37:
```suggestion
  deps: BaseDeps,
```

packages/open-payments/src/client/grant.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  return grant
+}
+
+export const isInteractiveGrant = (
+  grant: InteractiveGrant | NonInteractiveGrant
+): grant is InteractiveGrant =>
+  !('access_token' in grant) &&
+  'interact' in grant &&
+  !!(grant as InteractiveGrant).interact
```
wilsonianb, line 110: Could this just be
```suggestion
  !!(grant as InteractiveGrant).interact
```
I'm _assuming_ the OpenAPI validator would disallow `access_token` + `interact`.
ditto below

mkurapov:
Going to add grant continuation as a separate PR

packages/open-payments/src/client/grant.ts
wilsonianb: I'm still leaning towards
```suggestion
  requestInteractiveGrant(
    args: RequestGrantArgs<InteractiveGrantRequest>
  ): Promise<InteractiveGrant | NonInteractiveGrant>
```
and clients can utilize `isInteractiveGrant`/`isNonInteractiveGrant`.

I got the impression from Adrian that their policy agent AS might be determining what grant requests require interaction (vs having it Open Payments standardized).

mkurapov: Ah, I see what you mean, it's more of a configurable, per-instance thing instead of standard, makes sense. 

Going to make changes to assume client will always require interaction, but call might return either or.

packages/open-payments/src/client/grant.ts
mkurapov: Yeah I wasn't sure how lenient the validator was with `oneOf` in the spec, but this can be less strict

packages/open-payments/src/client/grant.ts
wilsonianb: I'm ok also keeping:
```
  requestNonInteractiveGrant(
    args: RequestGrantArgs<NonInteractiveGrantRequest>
  ): Promise<NonInteractiveGrant>
```
But maybe requiring the `interact` field to be specified isn't unreasonable? :thinking:

mkurapov: Yeah even in the demo Postman example, there is an interact field present when requesting the incoming payment grant -> which returns a non-interactive one. 
We could run with a single call to make things simple for now

packages/open-payments/src/types.ts
```diff
@@ -1,9 +1,44 @@
...
+  interact: ASComponents['schemas']['interact-response']
+  continue: ASComponents['schemas']['continue']
+}
+export const isInteractiveGrant = (
+  grant: InteractiveGrant | NonInteractiveGrant
+): grant is InteractiveGrant => !!(grant as InteractiveGrant).interact
+
+export const isNonInteractiveGrant = (
+  grant: InteractiveGrant | NonInteractiveGrant
+): grant is NonInteractiveGrant => !!(grant as NonInteractiveGrant).access_token
```
mkurapov, line 44: could live in `src/client/grants.ts` or `src/type-guard.ts` but decided to leave them here

packages/open-payments/src/types.ts
```diff
@@ -1,9 +1,44 @@
...
+  client: ASOperations['post-request']['requestBody']['content']['application/json']['client']
+}
+export type NonInteractiveGrant = {
+  access_token: ASComponents['schemas']['access_token']
+  continue: ASComponents['schemas']['continue']
+}
+export type GrantRequest = {
+  access_token: ASOperations['post-request']['requestBody']['content']['application/json']['access_token']
+  client: ASOperations['post-request']['requestBody']['content']['application/json']['client']
+  interact?: ASOperations['post-request']['requestBody']['content']['application/json']['interact']
```
mkurapov, line 32: @wilsonianb single grant request with optional `interact?` in line with the Open Payments API

wilsonianb: I was thinking we'd require `interact``
https://github.com/interledger/rafiki/pull/727#discussion_r1025369118
https://github.com/interledger/rafiki/pull/727#discussion_r1025384369
but I'm ok as is. The grant request would just throw if interaction is required and `interact` wasn't specified.

packages/open-payments/src/client/grant.ts
```diff
@@ -0,0 +1,33 @@
...
+} from '../types'
+import { post } from './requests'
+
+interface RequestGrantArgs {
+  url: string
+  request: GrantRequest
+}
+
+export interface GrantRoutes {
+  requestGrant(
```
wilsonianb, line 17:
```suggestion
  request(
```
or
```suggestion
  requestAccess(
```
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-2

wilsonianb (approved):

packages/open-payments/src/client/grant.ts
mkurapov: yup, had `request` in my follow-up PR, will do it here

packages/open-payments/src/types.ts
mkurapov: Was going back and forth but less errors and throws are better here IMO. Made it required.

wilsonianb (approved):

wilsonianb:
commit 300! :tada:

PR #728, opened 2022-11-04
chore(docker): use node:18-slim base image

### LINKED ISSUE #673: Docker build workflows timing out
Seemingly on this step (but only for `arm64`?):
https://github.com/interledger/rafiki/blob/0f0ff4e2f40e042a315d8c85889b4f71d37d451a/packages/auth/Dockerfile#L13

See:
- https://github.com/interledger/rafiki/pull/671
- https://github.com/interledger/rafiki/pull/638

wilsonianb:

- Update repo to [Node 18](https://nodejs.org/en/blog/release/v18.12.0/)
  This will require updating node locally and possibly [reinstalling `pnpm`](https://github.com/pnpm/pnpm/issues/4839)
- Removes Docker build action timeouts. This may result in longer Docker builds (up to ~1 hour), but presumably only when dependencies change. It seems like we should be able to improve (caching) performance for:
https://github.com/interledger/rafiki/blob/025991895b0e3f42b5dcdfd51359df393a4d41b8/packages/auth/Dockerfile#L13
- Limits building arm64 Docker images and pushing any Docker images to pushes to `main` & tags, which should help with:
https://github.com/interledger/rafiki/issues/985

## Context

- fixes #673 

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


wilsonianb:
Converting to draft until I can test subsequent Docker build times (using cache from these sloooow builds)

packages/auth/Dockerfile
```diff
@@ -1,12 +1,13 @@
-FROM node:16.17.1-slim as builder
+FROM node:18-slim as builder
```
sabineschaller, line 1: Should we specify a specific version to make use of the cache? I think `18-slim` resolves to the latest version.

sabineschaller (commented):
I'm happy to abandon my PR in favor for yours.

packages/auth/Dockerfile
wilsonianb: `latest` is actually `19` now.
https://hub.docker.com/_/node
Do you mean `node:18.13-slim` or specifying the `sha256`?

sabineschaller: Yeah, I meant the latest version of `18`. I was actually thinking of something like `18.13.0-slim` to have a very specific version.

wilsonianb: I'm ok with that.
Can we get renovatebot to help us update it?

wilsonianb: https://docs.renovatebot.com/modules/manager/dockerfile/#open-feature-requests

sabineschaller: Looks like the guy who wanted to do the pinning feature hasn't gotten to it yet.

sabineschaller (approved):

PR #735, opened 2022-11-08
feat(auth): fix httpsig bypass env

njlie:

- 
Conditionally adds httpsig middleware depending on the setting of the env variable for bypassing httpsig validation.

## Context

This conditional used to be in the code but refactors of `app.ts` have caused it to be removed.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -195,7 +195,9 @@ export class App {
         path: '/',
         method: HttpMethod.POST
       }),
-      grantInitiationHttpsigMiddleware,
+      this.config.bypassSignatureValidation
+        ? (ctx, next) => next()
+        : grantInitiationHttpsigMiddleware,
       grantRoutes.create
     )
 
@@ -206,7 +208,9 @@ export class App {
         path: '/continue/{id}',
         method: HttpMethod.POST
       }),
-      grantContinueHttpsigMiddleware,
+      this.config.bypassSignatureValidation
+        ? (ctx, next) => next()
+        : grantContinueHttpsigMiddleware,
       grantRoutes.continue
     )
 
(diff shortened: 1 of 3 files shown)


packages/auth/src/app.ts
```diff
@@ -195,7 +195,9 @@ export class App {
         path: '/',
         method: HttpMethod.POST
       }),
-      grantInitiationHttpsigMiddleware,
+      this.config.bypassSignatureValidation
```
wilsonianb, line 198: Should the existing bypass be removed?
https://github.com/interledger/rafiki/blob/d808c85c428bfc5bc1193882905ce5fdc8fa5888/packages/auth/src/signature/middleware.ts#L25-L28

wilsonianb (approved):
This should work until:
- https://github.com/interledger/open-payments/issues/209

at which point OpenAPI validation will fail without the signature headers.
See:
- https://github.com/interledger/rafiki/pull/683#issuecomment-1302329483

packages/auth/src/app.ts
njlie: it is pretty redundant now, yeah.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -405,55 +405,6 @@ describe('Signature Service', (): void => {
...
     })
 
-    test('middleware succeeds if BYPASS_SIGNATURE_VALIDATION is true with bad signature', async (): Promise<void> => {
-      const altDeps = await initIocContainer({
-        ...Config,
-        bypassSignatureValidation: true
-      })
-
-      const altContainer = await createTestApp(altDeps)
-      appContainers.push(altContainer)
```
wilsonianb, line 415: You could also revert `appContainers` to `appContainer`
https://github.com/interledger/rafiki/pull/683/files#diff-9c6c5d794c9dfa6afb4129ccc2ffb701bcdecb70308c63c7e4d615e8d903859c

wilsonianb (approved):

Issue #712, opened 2022-10-28
Return 200 on webhook event processing error


wilsonianb:
`mock-account-provider` should respond `200` to webhook event requests on at least some errors (such as insufficient sender account balance on `outgoing_payment.created` events).

As is `backend` will persistently retry sending the event until it receives `200`.
```
local-fynbos-1        | received webhook:  {"id":"60522d2f-fabb-48eb-b6b5-2e0455a41710","type":"outgoing_payment.created","data":{"payment":{"id":"3bdf94e8-8772-45ef-9137-e45d83a96cfe","paymentPointerId":"e3d47f7b-d721-44b5-9ca2-152d52f18263","state":"FUNDING","receiver":"https://peer-backend/accounts/pfry/incoming-payments/8c33aef2-c0f8-4a70-bfa7-f62b2abfdff7","sendAmount":{"value":"2060606060707","assetCode":"USD","assetScale":2},"receiveAmount":{"value":"2000000000000","assetCode":"USD","assetScale":2},"sentAmount":{"value":"0","assetCode":"USD","assetScale":2},"stateAttempts":0,"createdAt":"2022-10-28T22:28:47.124Z","updatedAt":"2022-10-28T22:28:47.173Z","balance":"0","description":"yolo","externalRef":"INV-001","peerId":"00525911-bfe8-4930-ac0b-d96aa0727a2b"}}}
local-fynbos-1        | Error: invalid pending debit amount, insufficient funds
local-fynbos-1        |     at AccountProvider.pendingDebit (/workspace/packages/mock-account-provider/build/index.js:258:13)
local-fynbos-1        |     at handleOutgoingPaymentCreated (/workspace/packages/mock-account-provider/build/index.js:460:29)
local-fynbos-1        |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
local-fynbos-1        | POST /webhooks 500 - - 5.569 ms
local-backend-1       | {"level":40,"time":1666996698737,"pid":1,"hostname":"94ca679fa3cf","service":"WebhookService","event":"60522d2f-fabb-48eb-b6b5-2e0455a41710","attempts":13,"error":"Request failed with status code 500","msg":"webhook request failed"}
[...]

PR #736, opened 2022-11-08
fix(map): return 200 on webhook event processing error

raducristianpopa:

- 

## Context

* fixes https://github.com/interledger/rafiki/issues/712

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


raducristianpopa:
In what other cases `backend` should not retry the event ?

wilsonianb (commented):
>In what other cases backend should not retry the event ?

Maybe this existing `500` if the account doesn't exist:
https://github.com/interledger/rafiki/blob/d808c85c428bfc5bc1193882905ce5fdc8fa5888/packages/mock-account-provider/app/routes/webhooks.ts#L96-L100

packages/mock-account-provider/app/lib/accounts.server.ts
```diff
@@ -185,7 +186,9 @@ export class AccountProvider implements AccountsServer {
     assert.ok(acc)
 
     if (acc.creditsPosted < acc.debitsPosted + acc.debitsPending + amount) {
-      throw new Error('invalid pending debit amount, insufficient funds')
+      throw json('invalid pending debit amount, insufficient funds', {
+        status: 200
```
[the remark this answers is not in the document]
raducristianpopa: Should status `200` be returned when the account doesn't exist for `handleOutgoingPaymentCompletedFailed` and `handleIncomingPaymentCompletedExpired` ?

https://github.com/interledger/rafiki/blob/ef46348e04d2355f9741475d4a303dd47bdf78ce/packages/mock-account-provider/app/routes/webhooks.ts#L66-L70
https://github.com/interledger/rafiki/blob/ef46348e04d2355f9741475d4a303dd47bdf78ce/packages/mock-account-provider/app/routes/webhooks.ts#L151-L155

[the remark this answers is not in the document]
wilsonianb: :+1:

packages/mock-account-provider/app/routes/webhooks.ts
```diff
@@ -96,11 +96,17 @@ export async function handleOutgoingPaymentCreated(wh: WebHook) {
...
   }
 
   const amt = parseAmount(payment['sendAmount'])
-  await mockAccounts.pendingDebit(acc.id, amt.value)
+
+  try {
+    await mockAccounts.pendingDebit(acc.id, amt.value)
+  } catch (e) {
+    const errorInfo = e instanceof Error && e.stack ? e.stack : String(e)
+    throw json(errorInfo, { status: 200 })
```
wilsonianb, line 108: Another option is for us to just reply with `200` after successfully parsing webhook events, instead of throwing `200` on subsequent errors.

See:
https://stripe.com/docs/webhooks#acknowledge-events-immediately

packages/mock-account-provider/app/routes/webhooks.ts
```diff
@@ -96,11 +106,19 @@ export async function handleOutgoingPaymentCreated(wh: WebHook) {
...
+    console.log('No account found for payment pointer')
+    return
   }
 
   const amt = parseAmount(payment['sendAmount'])
-  await mockAccounts.pendingDebit(acc.id, amt.value)
+
+  try {
+    await mockAccounts.pendingDebit(acc.id, amt.value)
+  } catch (e) {
```
wilsonianb, line 117: I was going to suggest that we don't have to catch here, since we're not `await`ing `handleOutgoingPaymentCompletedFailed` (we return `200` even if this throws).
But the mock account provider restarts with the uncaught error:
```
local-fynbos-1        | POST /webhooks 200 - - 1.919 ms
local-fynbos-1        | 
local-fynbos-1        | /workspace/packages/mock-account-provider/build/index.js:338
local-fynbos-1        |       throw new Error("invalid pending debit amount, insufficient funds");
local-fynbos-1        |             ^
local-fynbos-1        | Error: invalid pending debit amount, insufficient funds
local-fynbos-1        |     at AccountProvider.pendingDebit (/workspace/packages/mock-account-provider/build/index.js:338:13)
local-fynbos-1        |     at handleOutgoingPaymentCreated (/workspace/packages/mock-account-provider/build/index.js:617:29)
local-fynbos-1        |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
local-fynbos-1        | undefined
local-fynbos-1        | /workspace/packages/mock-account-provider:
local-fynbos-1        |  ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  mock-account-provider@ start: `remix-serve build`
local-fynbos-1        | Exit status 1
local-fynbos-1 exited with code 1
local-fynbos-1        | 
local-fynbos-1        | > mock-account-provider@ start /workspace/packages/mock-account-provider
local-fynbos-1        | > remix-serve build
local-fynbos-1        | 
local-fynbos-1        | calling run_seed
```

packages/mock-account-provider/app/routes/webhooks.ts
```diff
@@ -34,19 +34,28 @@ export function parseAmount(amount: AmountJSON): Amount {
...
-      return handleOutgoingPaymentCompletedFailed(wh)
+      handleOutgoingPaymentCompletedFailed(wh)
+      break
     case EventType.IncomingPaymentCompleted:
     case EventType.IncomingPaymentExpired:
-      return handleIncomingPaymentCompletedExpired(wh)
+      handleIncomingPaymentCompletedExpired(wh)
+      break
+    default:
+      console.log(`unknown event type: ${wh.type}`)
```
wilsonianb, line 58: this could throw `400`

wilsonianb (dismissed):

packages/mock-account-provider/app/routes/webhooks.ts
raducristianpopa: Doesn't this mean that every action which is going to modify the account's balance should be wrapped in a `try-catch` block, even for `handleOutgoingPaymentCompletedFailed` and `handleIncomingPaymentCompletedExpired` ?

Eg, if calling `mockAccouts.debit` and an error is thrown, this would not restart the MAP too ?

https://github.com/interledger/rafiki/blob/df6311d8a24432c640b876ce8877fbc33efb5903/packages/mock-account-provider/app/routes/webhooks.ts#L87-L90

https://github.com/interledger/rafiki/blob/df6311d8a24432c640b876ce8877fbc33efb5903/packages/mock-account-provider/app/routes/webhooks.ts#L171

wilsonianb: :thinking: yeah
I was going to suggest we wrap the `switch` block in a `try`/`catch` but then we couldn't respond `400` in the `default`...
https://github.com/interledger/rafiki/blob/73f573d7dcd72eadaac921fc5cbece5a59002fa1/packages/mock-account-provider/app/routes/webhooks.ts#L41-L50

wilsonianb: ~but then we couldn't respond 400 in the default~
Actually, I think that would work (I was thinking `throw` instead of just `return json` for the `400`).
I'm leaning towards having the only one of these :point_down: in a `catch` wrapping the `switch` statement
```
    const errorInfo = parseError(e)
    console.log(errorInfo)
```
That way we're catching all errors, including:
https://github.com/interledger/rafiki/blob/73f573d7dcd72eadaac921fc5cbece5a59002fa1/packages/mock-account-provider/app/routes/webhooks.ts#L126

Then any `json` can be removed from the `handle*` functions. (`throw json` can be converted to `throw new Error`)
Maybe the `handle*` functions could even be moved to a new `/lib/webhooks.server.ts`.

wilsonianb (approved):

PR #741, opened 2022-11-10
feat(backend): bypass sig validation

sabineschaller:

- bypass sig validation using same flag as auth server (see #735)

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/config/app.ts
+++ b/packages/backend/src/config/app.ts
@@ -19,7 +19,7 @@ function envFloat(name: string, value: number): number {
 
 function envBool(name: string, value: boolean): boolean {
   const envValue = process.env[name]
-  return envValue == null ? value : Boolean(envValue)
+  return envValue == null ? value : envValue === 'true'
 }
 
 export type IAppConfig = typeof Config
--- a/packages/backend/src/config/app.ts
+++ b/packages/backend/src/config/app.ts
@@ -17,10 +17,10 @@ function envFloat(name: string, value: number): number {
   return envValue == null ? value : +envValue
 }
 
-// function envBool(name: string, value: boolean): boolean {
-//   const envValue = process.env[name]
-//   return envValue == null ? value : Boolean(envValue)
-// }
+function envBool(name: string, value: boolean): boolean {
+  const envValue = process.env[name]
+  return envValue == null ? value : Boolean(envValue)
+}
 
 export type IAppConfig = typeof Config
 
@@ -103,6 +103,7 @@ export const Config = {
 
   signatureSecret: process.env.SIGNATURE_SECRET, // optional
   signatureVersion: envInt('SIGNATURE_VERSION', 1),
+  bypassSignatureValidation: envBool('BYPASS_SIGNATURE_VALIDATION', false),
 
   openPaymentsSpec: envString(
     'OPEN_PAYMENTS_SPEC',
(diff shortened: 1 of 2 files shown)


njlie (approved):

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -27,7 +27,7 @@ export function createAuthMiddleware({
       const token = parts[1]
       if (
         process.env.NODE_ENV !== 'production' &&
-        token === config.devAccessToken
+        (token === config.devAccessToken || config.bypassSignatureValidation)
```
wilsonianb, line 30: I think the check should be moved below to wrap:
https://github.com/interledger/rafiki/blob/e796dedeee7c3aedd43c5c977ce83f420ea8667c/packages/backend/src/open_payments/auth/middleware.ts#L48-L55

wilsonianb (commented):
You could add tests that `bypassSignatureValidation`:
- works with invalid signature
- doesn't work with invalid token

packages/backend/src/config/app.ts
```diff
@@ -17,10 +17,10 @@ function envFloat(name: string, value: number): number {
   return envValue == null ? value : +envValue
 }
 
-// function envBool(name: string, value: boolean): boolean {
-//   const envValue = process.env[name]
-//   return envValue == null ? value : Boolean(envValue)
-// }
+function envBool(name: string, value: boolean): boolean {
+  const envValue = process.env[name]
+  return envValue == null ? value : Boolean(envValue)
```
wilsonianb, line 22:
```suggestion
  return envValue == null ? value : envValue === 'true'
```
https://github.com/interledger/rafiki/pull/588#discussion_r966041927
https://github.com/interledger/rafiki/blob/e796dedeee7c3aedd43c5c977ce83f420ea8667c/packages/auth/src/config/app.ts#L13-L16

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -45,13 +45,15 @@ export function createAuthMiddleware({
       if (!access) {
         ctx.throw(403, 'Insufficient Grant')
       }
-      try {
-        if (!(await verifySigAndChallenge(grant.key.jwk, ctx))) {
-          ctx.throw(401, 'Invalid signature')
+      if (config.bypassSignatureValidation) {
```
wilsonianb, line 48:
```suggestion
      if (!config.bypassSignatureValidation) {
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -45,13 +45,15 @@ export function createAuthMiddleware({
       if (!access) {
         ctx.throw(403, 'Insufficient Grant')
       }
-      try {
-        if (!(await verifySigAndChallenge(grant.key.jwk, ctx))) {
-          ctx.throw(401, 'Invalid signature')
+      if (!config.bypassSignatureValidation) {
```
mkurapov, line 48: Log might be handy when signature validation is being bypassed, could also do an early return to avoid nesting

mkurapov (approved):

packages/backend/src/open_payments/auth/middleware.ts
wilsonianb: >could also do an early return to avoid nesting

I think we do still want the `ctx.grant = grant` below

PR #742, opened 2022-11-10
fix(auth): use correct port for peer-auth instance

njlie:

- Adds a missing `/` to the `continue` field in a grant response.
- Updates the port used by the `peer-auth` container to use 4006.
- Updates the default domain config to use the configured port.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/infrastructure/local/peer-docker-compose.yml
+++ b/infrastructure/local/peer-docker-compose.yml
@@ -9,12 +9,13 @@ services:
     networks:
       rafiki:
     ports:
-      - "4006:3006"
+      - "4006:4006"
     environment:
       NODE_ENV: development
       AUTH_DATABASE_URL: postgresql://peerauth:peerauth@database/peerauth
       INTROSPECTION_HTTPSIG: "false"
       BYPASS_SIGNATURE_VALIDATION: "true"
+      PORT: 4006
   peer-backend:
     image: ghcr.io/interledger/rafiki-backend:latest
     build:
--- a/infrastructure/local/peer-docker-compose.yml
+++ b/infrastructure/local/peer-docker-compose.yml
@@ -9,13 +9,13 @@ services:
     networks:
       rafiki:
     ports:
-      - "4006:4006"
+      - "4006:3006"
     environment:
       NODE_ENV: development
       AUTH_DATABASE_URL: postgresql://peerauth:peerauth@database/peerauth
       INTROSPECTION_HTTPSIG: "false"
       BYPASS_SIGNATURE_VALIDATION: "true"
-      PORT: 4006
+      AUTH_SERVER_DOMAIN: "http://localhost:4006"
   peer-backend:
     image: ghcr.io/interledger/rafiki-backend:latest
     build:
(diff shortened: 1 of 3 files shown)


infrastructure/local/peer-docker-compose.yml
```diff
@@ -9,11 +9,12 @@ services:
     networks:
       rafiki:
     ports:
-      - "4006:3006"
+      - "4006:4006"
```
wilsonianb, line 12: Should these be updated?
https://github.com/interledger/rafiki/blob/d5c620b06d7417b6f71a839723ab94141465bd97/infrastructure/local/peer-docker-compose.yml#L40-L41

wilsonianb: In fact, could this stay as
```suggestion
      - "4006:3006"
```
(and `PORT` not be specified)?

sabineschaller: I updated these locally but then, when I got a grant, the wrong port was displayed.

sabineschaller (dismissed):

infrastructure/local/peer-docker-compose.yml
wilsonianb: I'm assuming you're referring to the `localhost` port in the returned grant's `access_token.manage` and `continue.url` fields.
Those are both populated by `config.authServerDomain`, which is why I think `AUTH_SERVER_GRANT_URL` (and `AUTH_SERVER_INTROSPECTION_URL`) just need to be updated.
https://github.com/interledger/rafiki/blob/7b911b17f5b567a47a2a5e671cfaa83e87e34d91/packages/auth/src/grant/routes.ts#L400-L413
https://github.com/interledger/rafiki/blob/7b911b17f5b567a47a2a5e671cfaa83e87e34d91/packages/auth/src/grant/routes.ts#L113-L114

njlie: > In fact, could this stay as
> 
> (and `PORT` not be specified)?

Probably not, the port that's returned in `continue.uri` and other uris pointing to an AS should reflect the one that's accessible by the front-channel.

wilsonianb (dismissed):
~The configured port could remain unchanged, but this works as is~
~https://github.com/interledger/rafiki/pull/742/files#r1019268676~
https://github.com/interledger/rafiki/pull/742#discussion_r1019510317

infrastructure/local/peer-docker-compose.yml
wilsonianb: Got it. I wasn't thinking with container ports.
https://docs.docker.com/compose/networking/
>It is important to note the distinction between HOST_PORT and CONTAINER_PORT. In the above example, for db, the HOST_PORT is 8001 and the container port is 5432 (postgres default). Networked service-to-service communication uses the CONTAINER_PORT

njlie: @wilsonianb My bad, I see what you're getting at now. I've just configured the `AUTH_SERVER_DOMAIN` but left `PORT` at its default value so the backend can still hit the AS at 3006.

wilsonianb (approved):
ugh, I was also conflating auth's `AUTH_SERVER_DOMAIN` with backend's `AUTH_SERVER_GRANT_URL`
sorry for the confusion

Issue #583, opened 2022-08-30
Request grant at other Open Payments server(s)


wilsonianb:
The backend needs to request a grant to read incoming payments (for all payment pointers, no specific `identifier`) at any other Open Payments server where an incoming payment is the `receiver` of a quote / outgoing payment.

wilsonianb:
Don't query local incoming payments/receiver over http that exist locally, in order to not have manage grants for itself
https://github.com/interledger/rafiki/pull/595#discussion_r969502037

wilsonianb:
- [ ] configure client display info (`name`, `uri`)
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-2.3
- [ ] validate interaction finish hash
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-12.10

PR #758, opened 2022-11-19
feat(backend): serve jwks.json from client payment pointer

wilsonianb:

- Configure the RS's client payment pointer and serve its jwk (derived from the configured private key)

## Context

- progress toward #583 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment_pointer/middleware.ts
+++ b/packages/backend/src/open_payments/payment_pointer/middleware.ts
@@ -5,13 +5,15 @@ export function createPaymentPointerMiddleware() {
     ctx: AppContext,
     next: () => Promise<unknown>
   ): Promise<void> => {
-    const url = `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
+    ctx.paymentPointerUrl = `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
     const config = await ctx.container.use('config')
-    if (url !== config.paymentPointerUrl) {
+    if (ctx.paymentPointerUrl !== config.paymentPointerUrl) {
       const paymentPointerService = await ctx.container.use(
         'paymentPointerService'
       )
-      const paymentPointer = await paymentPointerService.getByUrl(url)
+      const paymentPointer = await paymentPointerService.getByUrl(
+        ctx.paymentPointerUrl
+      )
       if (!paymentPointer) {
         ctx.throw(404)
       }
--- a/packages/backend/src/open_payments/payment_pointer/middleware.ts
+++ b/packages/backend/src/open_payments/payment_pointer/middleware.ts
@@ -5,16 +5,18 @@ export function createPaymentPointerMiddleware() {
     ctx: AppContext,
     next: () => Promise<unknown>
   ): Promise<void> => {
-    const paymentPointerService = await ctx.container.use(
-      'paymentPointerService'
-    )
-    const paymentPointer = await paymentPointerService.getByUrl(
-      `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
-    )
-    if (!paymentPointer) {
-      ctx.throw(404)
+    const url = `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
+    const config = await ctx.container.use('config')
+    if (url !== config.paymentPointerUrl) {
+      const paymentPointerService = await ctx.container.use(
+        'paymentPointerService'
+      )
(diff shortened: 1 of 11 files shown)


packages/open-payments/src/client/index.ts
```diff
@@ -125,7 +126,8 @@ export const createAuthenticatedClient = async (
     grant: createGrantRoutes({
       axiosInstance,
       openApi: authServerOpenApi,
-      logger
+      logger,
+      client: args.client
```
wilsonianb, line 130: I was going to have `createAuthenticatedClient` query `${client}/jwks.json` to validate that the key with the specified `keyid` is served there, but `backend` calls `createAuthenticatedClient` long before its routes are set up.

mkurapov: Yeah I was thinking you'd possibly get some kind of race condition type behaviour - could add some kind of retry mechanism but gets too complicated for what it's worth. 

 Is there a more straightforward name than `client`? Even just `paymentPointerUrl`? or `clientPaymentPointer`?

packages/backend/src/open_payments/payment_pointer/middleware.ts
```diff
@@ -5,16 +5,18 @@ export function createPaymentPointerMiddleware() {
...
-      'paymentPointerService'
-    )
-    const paymentPointer = await paymentPointerService.getByUrl(
-      `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
-    )
-    if (!paymentPointer) {
-      ctx.throw(404)
+    const url = `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
+    const config = await ctx.container.use('config')
+    if (url !== config.paymentPointerUrl) {
```
mkurapov, line 10: In which situations will this equality happen?

wilsonianb: When this is an Open Payments resource request for the RS's own payment pointer, which will now/soon be happening when a separate AS is querying RS's client payment pointer `/jwks.json`

packages/backend/src/paymentPointerKey/routes.ts
```diff
@@ -24,15 +38,22 @@ export async function getKeysByPaymentPointerId(
...
+    )
 
-  ctx.body = {
-    keys: keys.map((key) => key.jwk)
+    ctx.body = {
+      keys: keys.map((key) => key.jwk)
+    }
+  } else if (
+    deps.config.paymentPointerUrl ===
+    `https://${ctx.request.host}/${ctx.params.paymentPointerPath}`
```
mkurapov, line 51: Should this be made into a function we can use across this package? Any function name will also help with the self-commenting of the code IMO

packages/open-payments/src/client/index.ts
```diff
@@ -93,6 +93,7 @@ export interface CreateAuthenticatedClientArgs
   extends CreateUnauthenticatedClientArgs {
```
mkurapov, line 93: Kind of outside the scope of this PR, but I'm a bit confused why TS doesn't complain about
```ts
const axiosInstance = createAxiosInstance({
    privateKey: args.privateKey,
    keyId: args.keyId,
    requestTimeoutMs:
      args?.requestTimeoutMs ?? config.DEFAULT_REQUEST_TIMEOUT_MS
}
```
when privateKey and keyId are optional in `createUnauthenticatedClient`.

mkurapov (dismissed):
Few comments, otherwise LGTM :+1:

packages/backend/src/paymentPointerKey/routes.ts
wilsonianb: I'd like to be able to do this via a `paymentPointerUrl` getter on `AppContext`(?), but I'm struggling to pull it off.

packages/open-payments/src/client/index.ts
wilsonianb: I think this due to
https://github.com/interledger/rafiki/blob/2f17be36bb705e3afbe449d577d380e79e876646/tsconfig.build.json#L28
which was toggled in
- https://github.com/interledger/rafiki/pull/758

@matdehaast @traviscrist Do either of you remember why we disabled strict type checking?

wilsonianb: >If you enable it how many issues does it find?

tbd
- https://github.com/interledger/rafiki/pull/768

Issue #755, opened 2022-11-18
Generate content-digest header


wilsonianb:
https://github.com/interledger/rafiki/blob/a2d48c494350903d71430e442d7e29bee84f6edd/packages/open-payments/src/client/signatures.ts#L35-L39

related to:
- https://github.com/interledger/rafiki/issues/655

`open-payments` grant requests will not currently pass `auth`'s http signature verification because it requires `content-digest` to be included in the `Signature-Input` header components (despite not yet validating the digest itself)
https://github.com/interledger/rafiki/blob/a2d48c494350903d71430e442d7e29bee84f6edd/packages/auth/src/signature/middleware.ts#L112-L114

PR #739, opened 2022-11-09
chore(auth): query client key via payment pointer

wilsonianb:
## Changes proposed in this pull request

- Store both client payment pointer and key id with grant
(we wouldn't have to store the latter with #737)
- Retrieve client key from payment pointer's `/jwks.json`
- Add payment pointer keys request :point_up: to `open-payments`' client

## Context

- fixes #731 
- requires https://github.com/interledger/open-payments/pull/211
- partially addresses https://github.com/interledger/rafiki/issues/656

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [ ] Documentation added
- [x] Make sure that all checks pass

PR #777, opened 2022-11-22
feat(auth): validate content-digest header

### LINKED ISSUE #655: Add content-digest verification to httpsig verification.
The content-digest signature component contains a sha-256 hash of the request body. The AS httpsig verifier should verifiy this hash by hashing the request body and making sure it matches.

njlie:

- Adds Content-Digest generation to `open-payments`
- Adds Content-Digest verification to `auth`

## Context

- Closes #655.
- fixes https://github.com/interledger/rafiki/issues/755

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -21,6 +21,7 @@
     "@koa/router": "^12.0.0",
     "ajv": "^8.11.0",
     "axios": "^0.27.2",
+    "httpbis-digest-headers": "github:interledger/httpbis-digest-headers",
     "jose": "^4.9.0",
     "knex": "^0.95",
     "koa": "^2.13.4",
--- a/packages/auth/src/signature/middleware.test.ts
+++ b/packages/auth/src/signature/middleware.test.ts
@@ -77,7 +77,7 @@ describe('Signature Service', (): void => {
         let expectedChallenge = `"@method": GET\n"@target-uri": http://example.com/test\n"content-type": application/json\n`
         const contentDigest = createContentDigestHeader(
           JSON.stringify(testRequestBody),
-          ['sha-256']
+          ['sha-512']
         )
 
         if (withRequestBody) {
@@ -134,7 +134,7 @@ describe('Signature Service', (): void => {
               'Content-Type': 'application/json',
               'Content-Digest': createContentDigestHeader(
                 JSON.stringify(testRequestBody),
-                ['sha-256']
+                ['sha-512']
               ),
               'Content-Length': '1234',
               'Signature-Input': sigInputHeader,
--- a/packages/auth/src/signature/middleware.test.ts
+++ b/packages/auth/src/signature/middleware.test.ts
@@ -4,6 +4,7 @@ import { faker } from '@faker-js/faker'
 import { importJWK } from 'jose'
 import { v4 } from 'uuid'
 import { Knex } from 'knex'
+import { createContentDigestHeader } from 'httpbis-digest-headers'
 
 import { createTestApp, TestContainer } from '../tests/app'
 import { truncateTables } from '../tests/tableManager'
@@ -58,6 +59,8 @@ describe('Signature Service', (): void => {
       ).resolves.toBe(true)
     })
 
+    const testRequestBody = { foo: 'bar' }
+
     test.each`
       title                                 | withAuthorization | withRequestBody
       ${''}                                 | ${true}           | ${true}
@@ -72,13 +75,16 @@ describe('Signature Service', (): void => {
           'Content-Type': 'application/json'
         }
         let expectedChallenge = `"@method": GET\n"@target-uri": /test\n"content-type": application/json\n`
+        const contentDigest = createContentDigestHeader(
+          JSON.stringify(testRequestBody),
+          ['sha-256']
+        )
 
         if (withRequestBody) {
           sigInputHeader += ' "content-digest" "content-length"'
-          headers['Content-Digest'] = 'sha-256=:test-hash:'
+          headers['Content-Digest'] = contentDigest
           headers['Content-Length'] = '1234'
-          expectedChallenge +=
-            '"content-digest": sha-256=:test-hash:\n"content-length": 1234\n'
+          expectedChallenge += `"content-digest": ${contentDigest}\n"content-length": 1234\n`
         }
 
         if (withAuthorization) {
@@ -104,7 +110,7 @@ describe('Signature Service', (): void => {
           deps
         )
 
-        ctx.request.body = withRequestBody ? { foo: 'bar' } : {}
+        ctx.request.body = withRequestBody ? testRequestBody : {}
 
         const challenge = sigInputToChallenge(sigInputHeader, ctx)
 
@@ -126,7 +132,10 @@ describe('Signature Service', (): void => {
           {
             headers: {
               'Content-Type': 'application/json',
-              'Content-Digest': 'sha-256=:test-hash:',
+              'Content-Digest': createContentDigestHeader(
+                JSON.stringify(testRequestBody),
+                ['sha-256']
+              ),
               'Content-Length': '1234',
               'Signature-Input': sigInputHeader,
               Authorization: 'GNAP test-access-token'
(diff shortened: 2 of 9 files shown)


packages/open-payments/src/client/requests.test.ts
```diff
@@ -175,7 +175,17 @@ describe('requests', (): void => {
         id: 'id'
       }
 
-      nock(baseUrl).post('/grant', body).reply(status, body)
+      const scope = nock(baseUrl)
+        .matchHeader('Signature', /sig1=:([a-zA-Z0-9+/]){86}==:/)
+        .matchHeader(
+          'Signature-Input',
+          `sig1=("@method" "@target-uri" "content-digest" "content-length" "content-type");created=${Math.floor(
+            Date.now() / 1000
```
wilsonianb, line 183: This test also needs
https://github.com/interledger/rafiki/blob/632bdfb15a920062637dc8a273034e58b382205b/packages/open-payments/src/client/requests.test.ts#L45-L58

packages/auth/src/tests/signature.ts
```diff
@@ -78,10 +79,7 @@ export async function generateSigHeaders({
   let challenge = `"@method": ${method}\n"@target-uri": ${url}\n`
   let contentDigest
   if (body) {
-    const hash = crypto.createHash('sha256')
-    hash.update(Buffer.from(JSON.stringify(body)))
-    const bodyDigest = hash.digest()
-    contentDigest = `sha-256:${bodyDigest.toString('base64')}:`
+    contentDigest = createContentDigestHeader(JSON.stringify(body), ['sha-256'])
```
wilsonianb, line 82:
```suggestion
    contentDigest = createContentDigestHeader(JSON.stringify(body), ['sha-512'])
```
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-7.3.1
>   When the proof method is specified in string form, the signing
   algorithm MUST be derived from the key material (such as using the
   JWS algorithm in a JWK formatted key), and the content digest
   algorithm MUST be sha-512.

That said, since we've switched to specifying clients by payment pointer, we now only have `proof` in the introspection response:
- https://github.com/interledger/rafiki/pull/739

I vote we follow the spirit of the spec and do sha512 when not otherwise specified.

packages/auth/package.json
```diff
@@ -21,6 +21,7 @@
     "@koa/router": "^12.0.0",
     "ajv": "^8.11.0",
     "axios": "^0.27.2",
+    "httpbis-digest-headers": "github:interledger/httpbis-digest-headers",
```
mkurapov, line 24: how does this install work?

mkurapov (commented):
few comments

packages/open-payments/src/client/requests.test.ts
```diff
@@ -175,7 +175,17 @@ describe('requests', (): void => {
         id: 'id'
       }
 
-      nock(baseUrl).post('/grant', body).reply(status, body)
+      const scope = nock(baseUrl)
```
wilsonianb, line 178:
```suggestion
      const scope = nock(baseUrl)
        .matchHeader('Content-Digest', /sha-512=:([a-zA-Z0-9+/]){86}==:/)
        .matchHeader('Content-Length', 11)
        .matchHeader('Content-Type', 'application/json')
```

packages/open-payments/src/client/requests.ts
```diff
@@ -128,6 +129,12 @@ export const createAxiosInstance = (args: {
   if (args.privateKey && args.keyId) {
     axiosInstance.interceptors.request.use(
       async (config) => {
+        if (config.data) {
+          config.headers['Content-Digest'] = createContentDigestHeader(
+            JSON.stringify(config.data),
+            ['sha-512']
+          )
```
wilsonianb, line 136: Upon further inspection, `Content-Length` and `Content-Type` headers aren't yet populated in the interceptor
```suggestion
          const data = JSON.stringify(config.data)
          config.headers['Content-Digest'] = createContentDigestHeader(data, [
            'sha-512'
          ])
          config.headers['Content-Length'] = Buffer.from(data, 'utf-8').length
```
You could either also add `Content-Type` here or above as
```
axiosInstance.defaults.headers.post['Content-Type'] = 'application/json'
```
but then I think we'd need to go back to https://github.com/interledger/rafiki/pull/722#discussion_r1016045236

packages/auth/src/signature/middleware.ts
```diff
@@ -110,7 +111,11 @@ function validateSigInputComponents(
     !sigInputComponents.includes('@target-uri') ||
```
wilsonianb, line 111: I'm finally having `open-payments` http signatures verified by `auth`.
`@target-uri` needs to use `ctx.request.href` instead of
https://github.com/interledger/rafiki/blob/632bdfb15a920062637dc8a273034e58b382205b/packages/auth/src/signature/middleware.ts#L137
https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-09#section-2.2.3

wilsonianb (commented):
I'm finally having `open-payments` http signatures verified by `auth`.
`auth` needs to

packages/auth/package.json
wilsonianb: https://pnpm.io/cli/add#install-from-git-repository
I don't think https://github.com/interledger/httpbis-digest-headers has been published to npm yet

packages/auth/src/tests/signature.ts
njlie: Works for me

packages/auth/src/signature/middleware.test.ts
```diff
@@ -71,14 +74,17 @@ describe('Signature Service', (): void => {
         const headers = {
           'Content-Type': 'application/json'
         }
-        let expectedChallenge = `"@method": GET\n"@target-uri": /test\n"content-type": application/json\n`
+        let expectedChallenge = `"@method": GET\n"@target-uri": http://example.com/test\n"content-type": application/json\n`
+        const contentDigest = createContentDigestHeader(
+          JSON.stringify(testRequestBody),
+          ['sha-256']
```
wilsonianb, line 80: Should these also change?
```suggestion
          ['sha-512']
```

packages/auth/src/signature/middleware.ts
```diff
@@ -105,12 +106,20 @@ function validateSigInputComponents(
...
+        ctx.headers['content-digest'] as string
+      ))
+
   return !(
     !sigInputComponents.includes('@method') ||
     !sigInputComponents.includes('@target-uri') ||
-    (ctx.request.body &&
-      Object.keys(ctx.request.body).length > 0 &&
-      !sigInputComponents.includes('content-digest')) ||
+    isValidContentDigest ||
```
wilsonianb, line 122:
```suggestion
    !isValidContentDigest ||
```
and maybe move it ahead of the `@method` check

packages/open-payments/src/client/requests.ts
```diff
@@ -126,27 +127,37 @@ export const createAxiosInstance = (args: {
...
+        keyId: args.keyId
+      })
+      config.headers['Signature'] = sigHeaders['Signature']
+      config.headers['Signature-Input'] = sigHeaders['Signature-Input']
+      return config
+    }
+    axiosInstance.interceptors.request.use(interceptor, null, {
+      runWhen: (config) =>
+        config.method.toLowerCase() === 'post' ||
+        !!config.headers['Authorization']
```
wilsonianb, line 156:
```suggestion
        config.method.toLowerCase() === 'post'
```

packages/open-payments/src/client/requests.ts
```diff
@@ -126,27 +127,37 @@ export const createAxiosInstance = (args: {
...
+    })
+    axiosInstance.interceptors.request.use(interceptor, null, {
+      runWhen: (config) => !!config.headers['Authorization']
+    })
```
wilsonianb, line 160: I think the interceptor would be run twice for `POST` requests with `Authorization`
https://github.com/axios/axios#multiple-interceptors

njlie: Sounds like we should just keep using one interceptor then

wilsonianb (approved):

Issue #737, opened 2022-11-09
Return client payment pointer in introspection response


wilsonianb:
Instead of returning the client jwk and an opaque client id in the token introspection response, return the client payment pointer.
https://github.com/interledger/rafiki/blob/d808c85c428bfc5bc1193882905ce5fdc8fa5888/packages/auth/src/openapi/resource-server.yaml#L37-L41
This not only makes us consistently pass the client (and key) by reference (see: https://github.com/interledger/open-payments/pull/211), the RS will be able to check if a client request fits the grant access before querying the client's `jwks.json`

Implications:
- client (id) no longer opaque to RS
- RS makes request to client payment pointer's `jwks.json` (and/or caches key(s)), instead of AS
- grants are bound to client instead of a specific client key

wilsonianb:
This kind of requires:
- https://github.com/interledger/rafiki/issues/740

because it is currently using the token introspection definition from the old `auth-server-open-api-spec.yaml` instead of https://github.com/interledger/rafiki/blob/main/packages/auth/src/openapi/resource-server.yaml

PR #779, opened 2022-11-23
feat(backend): request grant to query incoming payment receiver 

### LINKED ISSUE #583: Request grant at other Open Payments server(s)

wilsonianb:

- The receiver service now uses a stored grant if one exists or requests a new grant when querying a remote incoming payment.

## Context

- fixes #583 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


wilsonianb:
Using a combination of this branch + https://github.com/interledger/rafiki/pull/777, I was able to run localenv with AS signature verification enabled:
https://github.com/interledger/rafiki/blob/632bdfb15a920062637dc8a273034e58b382205b/infrastructure/local/peer-docker-compose.yml#L17
and successfully create a quote (with `backend` auto-requesting a grant to read the receiver incoming payment)

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -179,3 +188,49 @@ async function getLocalIncomingPayment({
...
+    accessType: AccessType.IncomingPayment,
+    accessActions: [AccessAction.ReadAll]
+  }
+
+  const existingGrant = await deps.grantService.get(grantOptions)
+  if (existingGrant) {
+    return existingGrant
+  }
+
+  const grant = await deps.openPaymentsClient.grant.request({
```
wilsonianb, line 213: Is it worth (get or creating and) locking the auth server's db row here to avoid simultaneously sending redundant grant requests?

mkurapov:
> Using a combination of this branch + #777, I was able to run localenv with AS signature verification enabled:
> 
> https://github.com/interledger/rafiki/blob/632bdfb15a920062637dc8a273034e58b382205b/infrastructure/local/peer-docker-compose.yml#L17
> 
> and successfully create a quote (with `backend` auto-requesting a grant to read the receiver incoming payment)

what sig headers did you use for making it work?

wilsonianb:
Screenshot from 2022-11-23 12-50-21
This query :point_up_2: 
The signature headers are unchanged (`TODO`)
But the signature headers are there for the RS->AS grant request under the hood.

mkurapov:
Is `accessToken` just `dev-access-token` for now? I'm working on updating those seed scripts.

wilsonianb:
Yeah, I changed it from the (Postman) default `example-access-token`.

packages/backend/src/open_payments/grant/model.ts
```diff
@@ -0,0 +1,34 @@
+import { Model, QueryContext } from 'objection'
+
+import { AccessType, AccessAction } from '../auth/grant'
+import { AuthServer } from '../authServer/model'
+import { BaseModel } from '../../shared/baseModel'
+
+export class Grant extends BaseModel {
```
mkurapov, line 7: Anything we can do here (naming wise) to make it more explicit what the difference between Grant and GrantReference is?

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -179,3 +188,49 @@ async function getLocalIncomingPayment({
...
+  const grantOptions = {
+    authServer: paymentPointer.authServer,
+    accessType: AccessType.IncomingPayment,
+    accessActions: [AccessAction.ReadAll]
+  }
+
+  const existingGrant = await deps.grantService.get(grantOptions)
+  if (existingGrant) {
+    return existingGrant
+  }
```
mkurapov, line 211: Should we be storing & checking the `expiresAt` for `Grant`s?

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -179,3 +188,49 @@ async function getLocalIncomingPayment({
...
+  const paymentPointer = await deps.openPaymentsClient.paymentPointer.get({
+    url: paymentPointerUrl
+  })
+  if (!paymentPointer) {
+    return undefined
+  }
+  const grantOptions = {
+    authServer: paymentPointer.authServer,
+    accessType: AccessType.IncomingPayment,
+    accessActions: [AccessAction.ReadAll]
```
mkurapov, line 205: I guess it all depends on the AS, but would just `Read` be "safer" here? Having to create a grant per incoming payment isn't great, but makes it easier on auth servers to restrict access

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    await appContainer.shutdown()
+  })
+
+  describe('Create and Get Grant', (): void => {
+    test.each`
+      newAuthServer
+      ${false}
+      ${true}
+    `(
+      'Grant can be created and fetched (new auth server: $newAuthServer)',
```
mkurapov, line 44: imo might be easier to split into two separate tests to avoid the `if (newAuthServer)` checks

packages/backend/src/open_payments/payment_pointer/model.ts
```diff
@@ -84,6 +85,20 @@ export class PaymentPointer
...
     }
   }
+
+  public toOpenPaymentsType({
+    authServer
+  }: {
+    authServer: string
+  }): OpenPaymentsPaymentPointer {
+    return {
+      id: this.url,
```
mkurapov, line 95: Would returning `url: this.url` be more clear from an Open Payments API perspective?

packages/backend/src/open_payments/receiver/service.ts
mkurapov: Is this a case for when we request the receiver on the outgoing payment lifecycle hook?

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -179,3 +188,49 @@ async function getLocalIncomingPayment({
...
+          {
+            type: grantOptions.accessType,
+            actions: grantOptions.accessActions
+          }
+        ]
+      },
+      interact: {
+        start: ['redirect']
+      }
+    } as GrantRequest
```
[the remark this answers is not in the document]
wilsonianb: Or could these be exported
https://github.com/interledger/rafiki/blob/48cdfa261ce6857810cd1e5a080c411788db5da2/packages/open-payments/src/generated/auth-server-types.ts#L250-L287
as `IncomingPaymentAccess`, etc?
Then `backend`/`auth` could utilize these types instead having a bunch of pairings of arbitrary `AccessType`/`AccessAction`.

packages/backend/src/open_payments/receiver/service.ts
wilsonianb: I don't have a strong opinion either way, and I'm unsure which would make auth server locking more of a necessity:
https://github.com/interledger/rafiki/pull/779#discussion_r1029968121

packages/backend/src/open_payments/payment_pointer/model.ts
wilsonianb: As in, change `id` to `url` in the Open Payments spec?
https://github.com/interledger/open-payments/blob/b363d33038fe789e5388f04f80ddd06a4fa97093/openapi/resource-server.yaml#L809-L813

packages/backend/src/open_payments/receiver/service.ts
wilsonianb: I was thinking if (multiple instances of) the backend were handling two quotes with the same receiver at the same time, should it avoid performing multiple grant requests?

packages/backend/src/open_payments/grant/model.ts
wilsonianb: What about `GrantReference` -> `ResourceGrant` (or (PaymentPointer)SubresourceGrant)?

packages/auth/src/access/types.ts
```diff
@@ -8,15 +7,15 @@ export enum AccessType {
...
   List = 'list',
+  ListAll = 'list-all',
   Complete = 'complete'
 }
 
 interface BaseAccessRequest {
   actions: Action[]
-  locations?: string[]
   identifier?: string
-  interval?: string
```
njlie, line 19: Should `interval` be moved to the `IncomingPaymentRequest` and `OutgoingPaymentRequest` types?

wilsonianb: It should be added to:
https://github.com/interledger/rafiki/blob/48cdfa261ce6857810cd1e5a080c411788db5da2/packages/auth/src/access/types.ts#L120-L124
https://github.com/interledger/open-payments/blob/main/openapi/schemas.yaml#L151-L163
I'll try to include that without scope-creeping too much more. I expect a lot of outstanding cleanup will be addressed with:
- https://github.com/interledger/rafiki/issues/662

packages/backend/src/open_payments/grant/model.ts
wilsonianb: I'm also tempted to backtrack on this: https://github.com/interledger/rafiki/pull/535#discussion_r954891956

We could instead store `clientId` (soon to be `client` payment pointer #737) on subresources.
Rename `GrantReference` as `OutgoingPaymentGrants` (since it's only really needed to lock on limit checking :point_down:) and drop the `clientId` column
https://github.com/interledger/rafiki/blob/a5c36d7212c6dadc600f574186d5d5f08462c718/packages/backend/src/open_payments/payment/outgoing/service.ts#L255-L256

njlie (dismissed):

packages/backend/src/open_payments/grant/model.ts
wilsonianb: :grimacing: 
- https://github.com/interledger/rafiki/pull/800

infrastructure/local/peer-docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       dockerfile: ./packages/auth/Dockerfile
     restart: always
     networks:
-      rafiki:
+      local_rafiki:
```
sabineschaller, line 10: Nit pick: I'm always confused when I see the colon at the end of the network name. Could we not do the following?
```
networks:
  - local_rafiki
```

packages/backend/migrations/20221012013413_create_grants_table.js
```diff
@@ -0,0 +1,23 @@
...
+    table.specificType('accessActions', 'text[]')
+
+    table.timestamp('expiresAt').nullable()
+
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
+    table.timestamp('updatedAt').defaultTo(knex.fn.now())
+
+    table.unique(['authServerId', 'accessType', 'accessActions'])
+  })
+}
```
sabineschaller, line 19: Is it sufficient to not store the entire access array of the grant?

packages/backend/src/open_payments/grant/model.ts
sabineschaller: Renaming is the right move since we only store references to outgoing payment grants. I'll have to look into the other PR to understand why we can drop `clientId`.

packages/backend/src/open_payments/receiver/service.ts
sabineschaller: I think just `Read` wouldn't work because I'd imagine in most cases, this Rafiki instance wasn't the client creating the incoming payment. It just wants to read it to pay it.

packages/backend/src/open_payments/grant/model.ts
wilsonianb: At the moment, we store references for grants for incoming payments, quotes, and outgoing payments because it is how we associate `cilentId` with those resources.

packages/backend/migrations/20221012013413_create_grants_table.js
wilsonianb: It is for now since the RS is only requesting read-all incoming payment grants, and the new `GrantService` only supports creating grants of a single access type.
Do we anticipate the RS requesting multi-type grants that we should consider future-proofing for?

packages/backend/src/open_payments/payment_pointer/model.ts
mkurapov: > As in, change id to url in the Open Payments spec?

Exactly, wasn't a comment of the code, more in general. `id`s for other resources (incoming/outgoing payments) have a different meaning that the `id` for the payment pointer.

wilsonianb: :thinking: Want to elaborate on how the `id` is different for the payment pointer (here or in an [`open-payments`](https://github.com/interledger/open-payments) issue)?
Part of me wants to say payment pointers are probably going to be out of scope for Open Payments soon.
Also:
- https://github.com/interledger/open-payments/pull/223

mkurapov: Well, for example in our case, we call incoming payments by `baseUrl/incoming-payment/{{id}}`, id being a uuid, a sub resource identifier. You don't necessarily have a similar "baseUrl/payment-pointer" resource, where you could GET a payment pointer using the `id` path param. 

Likewise in `backend`, when we use `paymentPointer.id` its usually the model's uuid. I don't think it's worth changing, just might not be always 100% clear, is all.

packages/backend/src/open_payments/receiver/service.ts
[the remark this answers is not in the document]
wilsonianb: I've gone with your original suggestion but opened:
- https://github.com/interledger/rafiki/issues/808

packages/backend/src/open_payments/receiver/service.ts
mkurapov: Pretty edge-casey IMO, not sure if worth adding just yet

mkurapov (approved):

PR #783, opened 2022-11-23
fix(auth): use expires_in in grant initiation response body

### LINKED ISSUE #749: Fix grant initiation response body
Currently, we return `expiresIn` as part of the grant initiation request:
https://github.com/interledger/rafiki/blob/74e6f642501f1f3caf66ec5cd66d13a624b997f5/packages/auth/src/grant/routes.ts#L420

However, open payments API expects snake case `expires_in`:

https://github.com/interledger/open-payments/blob/0523804eeb21d9b8cd2eb65e4119738a0512a095/openapi/auth-server.yaml#L64

We do have an issue open in the open-payments repo for converting to pascal case: https://github.com/interledger/open-payments/issues/195, but I'm more concerned that the Open API request/response validation we have set up allows this to pass. Is it because of the fact that `access_token` is optional in the case where an interactive grant gets generated?

wilsonianb:

- 

## Context

- fixes #749 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/scripts/get-op-schema.sh
+++ b/packages/auth/scripts/get-op-schema.sh
@@ -9,5 +9,5 @@ while [ -L "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
 done
 OUTDIR=$( cd -P "$( dirname "$SOURCE" )/../src/openapi" >/dev/null 2>&1 && pwd )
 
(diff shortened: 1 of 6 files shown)


mkurapov (approved):

mkurapov (changes requested):
I think we need to update the AS token rotation response as well

Issue #703, opened 2022-10-27
Add `grantReferenceService.getOrCreate`


wilsonianb:
https://github.com/interledger/rafiki/pull/672#discussion_r1000773412

replace
https://github.com/interledger/rafiki/blob/603df09615a2f458227baf3e23f99b3a0b103921/packages/backend/src/open_payments/auth/middleware.ts#L48-L66

with a method: `grantReferenceService.getOrCreate`

mkurapov:
@raducristianpopa we'd need to add you to the org to be able to assign you to the issue I believe

AlexLakatos:
Sent @raducristianpopa an invite.

mkurapov:
Closed since `grantReferences` are no longer a thing: https://github.com/interledger/rafiki/pull/800

PR #787, opened 2022-11-24
refactor(backend): add `getOrCreate` method for `grantReferenceService`

raducristianpopa:

- Add `getOrCreate` method for `grantReferenceService`
## Context

- fixes https://github.com/interledger/rafiki/issues/703

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -55,26 +53,18 @@ export function createAuthMiddleware({
...
-      })
+
+      try {
+        await grantReferenceService.getOrCreate(
+          { id: grant.grant, clientId: grant.clientId },
+          action
+        )
+      } catch (e) {
+        const errInfo = e && typeof e === 'object' && e.stack ? e.stack : e
+        logger.debug(errInfo)
```
mkurapov, line 64:
```suggestion
      } catch (error) {
        const errorMessage = error?.message ?? 'Failed to get or create grant reference'
        logger.error({ error }, errorMessage)
```
would prefer something like this, without needing to check the type etc

packages/backend/src/open_payments/grantReference/service.test.ts
```diff
@@ -79,4 +80,60 @@ describe('Grant Reference Service', (): void => {
...
+        clientId: uuid()
+      })
+
+      await expect(
+        async () =>
+          await grantReferenceService.getOrCreate(
+            { id, clientId: uuid() },
+            AccessAction.List
+          )
+      ).rejects.toThrowError('does not match internal reference clientId')
```
mkurapov, line 108: I thought we would have needed a RegEx here, but it seems to work? 
```suggestion
      ).rejects.toThrowError(/does not match internal reference clientId/)
```

mkurapov (commented):
Looks solid, just a few small comments. Would test this by

packages/backend/src/open_payments/grantReference/service.test.ts
raducristianpopa: I was thinking the same, but based on Jest docs, RegEx and strings are equivalent in this case.

mkurapov (approved):
Looks great!
In order to properly test this we need to:
- Create or seed an initial grant in the auth server
- Make a request (e.g. get incoming payment) with the token related to the created grant, and verify that the grant reference was created in backend's database during token introspection
- Make another request (e.g. duplicate get incoming payment) with the token related to the created grant, and this time verify the grant reference was fetched correctly from backend's database.

Easiest way would be to test this with the Create Incoming Payment request in the Postman P2P payment transfer example collection, after my [PR](https://github.com/interledger/rafiki/pull/786) is merged in. Once that is merged, I can walk you through the flow

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -55,26 +53,19 @@ export function createAuthMiddleware({
...
-            trx
-          )
-        }
-      })
+
+      try {
+        await grantReferenceService.getOrCreate(
+          { id: grant.grant, clientId: grant.clientId },
+          action
+        )
```
wilsonianb, line 61:
```suggestion
        if (action === AccessAction.Create) {
          await grantReferenceService.getOrCreate({
            id: grant.grant,
            clientId: grant.clientId
          })
        } else {
          await grantReferenceService.get(grant.grant)
        }
```

raducristianpopa: By moving the action check to the caller, the `clientId` check 
https://github.com/interledger/rafiki/blob/c6eaaa91825bc840d0b6bd4aa10b9b3e4cb1896c/packages/backend/src/open_payments/grantReference/service.ts#L58-L63
will probably need to be moved to the `getGrantRerefence` method
https://github.com/interledger/rafiki/blob/c6eaaa91825bc840d0b6bd4aa10b9b3e4cb1896c/packages/backend/src/open_payments/grantReference/service.ts#L27-L29
and will have to take both arguments (`id` and `clientId`)
```ts
async function getGrantReference(
  options: CreateGrantReferenceOptions,
  trx: Transaction
) {
// ...
}
```
otherwise there will be no validation for the `clientId` when calling `grantReferenceService.get`

wilsonianb: I'm leaning towards not passing `clientId` to `grantReferenceService.get` (even though this is currently the only place calling it).
What if we moved the `clientId` check out here?
```suggestion
        if (action === AccessAction.Create) {
          await grantReferenceService.getOrCreate({
            id: grant.grant,
            clientId: grant.clientId
          })
        } else {
          const grantRef = await grantReferenceService.get(grant.grant)
          if (grantRef && grantRef.clientId !== grant.clientId) {
            throw new Error(
              `clientID ${grant.clientId} for grant ${grant.id} does not match internal reference clientId ${grantRef.clientId}.`
            }
          }
        }
```
:thinking: that looks a lot like how it was before...

Is there an argument for not actually using `getOrCreate`?
It looks like the `AccessAction.Create` check was added after this issue was opened:
- https://github.com/interledger/rafiki/pull/676

wilsonianb:
This will completely remove `grantReferenceService` (if we decide we want to):
- https://github.com/interledger/rafiki/pull/800

raducristianpopa:
Closing this PR since https://github.com/interledger/rafiki/pull/800 got merged in.

Issue #661, opened 2022-10-07
Audit usages of `assert` outside of tests 


mkurapov:
There are a handful of usages of `assert` outside of tests. An example in the mock account provider:
https://github.com/interledger/rafiki/blob/main/packages/mock-account-provider/app/lib/accounts.server.ts#L49-L50

We should convert these to throw applicable `Error`s instead, and reserve `assert` usages for tests only.

jpsousa78:
Hey there! I'm interested in working on this issue.

wilsonianb:
It's yours! :vulcan_salute:

PR #791, opened 2022-11-28
Replace usages of assert outside of tests for error throws

jpsousa78:

- This pull request will change all `assert` usage outside of tests to throw their respectives errors.

## Context
In this PR we are fixing the following issue: https://github.com/interledger/rafiki/issues/661 
It's as much self explanatory as it could be.
The issue focus on replacing asserts ouside of tests, so no tests were updated.
This change doesn't reflect on any new feature or business rule, so no need to update documentation.
- fixes https://github.com/interledger/rafiki/issues/661
## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/tests/incomingPayment.ts
```diff
@@ -1,4 +1,3 @@
-import assert from 'assert'
```
wilsonianb, line 1: :thinking: Are we ok keeping `assert` in these `src/tests/*` files?

wilsonianb (commented):
You'll need to run `pnpm format` to satisfy the checks.

packages/backend/src/tests/incomingPayment.ts
mkurapov: I would argue errors are better here since these functions are just test setup/prerequisites, and not actual test cases where we should be `assert`ing some condition

mkurapov (commented):
Thanks for contributing! Other than `pnpm format`ting , just one comment

wilsonianb (approved):
:tada:

PR #792, opened 2022-11-28
feat(auth): bypassOrCallHttpsigMiddleware

mkurapov:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/signature/middleware.ts
+++ b/packages/auth/src/signature/middleware.ts
@@ -7,6 +7,7 @@ import { JWK } from 'open-payments'
 import { AppContext } from '../app'
 import { Grant } from '../grant/model'
 import { Context } from 'koa'
+import { Logger } from 'pino'
 
 export async function verifySig(
   sig: string,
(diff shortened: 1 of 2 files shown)


packages/auth/src/signature/middleware.ts
```diff
@@ -277,3 +278,24 @@ export async function tokenHttpsigMiddleware(
...
+  bypassSignatureValidation: boolean,
+  middleware: Middleware
+): Middleware {
+  if (bypassSignatureValidation) {
+    return async (ctx: AppContext, next) => {
+      ctx.logger.info('Skipping httpsig validation')
+
+      ctx.clientKeyId = getSigInputKeyId(
+        ctx.headers['signature-input'] as string
+      ) //TODO: remove once workaround is found for clientKeyId definition
```
mkurapov, line 294: Makes it possible to provide proper keyId via:

`Signature-Input: keyid="{{keyId}}"`

wilsonianb: Should this call `clientService.getKey`?
Otherwise, I think this will allow any `keyId`, which would cause token introspection to fail (if it isn't also bypassed)
https://github.com/interledger/rafiki/blob/48cdfa261ce6857810cd1e5a080c411788db5da2/packages/auth/src/accessToken/service.ts#L116-L119

Issue #649, opened 2022-10-03
Add webhook event for a created incoming payment


AlexLakatos:

wilsonianb:
It seems like this would resemble the call to the account provider that happens during quote creation which is distinct from webhook events:
https://github.com/interledger/rafiki/blob/main/packages/backend/src/config/app.ts#L89-L93
https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/quote/service.ts#L262
https://github.com/interledger/rafiki/blob/main/docs/transaction-api.md#quote

I've actually been considering doing this for outgoing payments as well to replace the asynchronous flow of:

1. `outgoing_payment.created` webhook event
2. account provider calls `Mutation.depositEventLiquidity`

https://github.com/interledger/rafiki/blob/main/docs/transaction-api.md#outgoing_paymentcreated
(I'm not it makes sense though, because it'd require being able to withdraw liquidity from an outgoing payment that may have failed to successfully be created)

>The Open Payments Spec needs to be changed in order to be able to handle this new behaviour.

RS <> Account provider interaction isn't defined in Open Payments.

wilsonianb:
>RS <> Account provider interaction isn't defined in Open Payments.

yet:
- https://github.com/interledger/rafiki/issues/626#issue-1381270487

mkurapov:
@wilsonianb 

Thanks for the notes. A few questions (and please correct any of my assumptions):
1. What is the route for the backend <> account provider? Based on your comment
> during quote creation which is distinct from webhook events:
it seems that I would _not_ reuse the `/webhooks` route?
2. In the logic for the account provider, what do I actually need to validate there? I'm guessing no balance needs to be check if the incoming payment is a credit to the user. Do I just need to check for the existence of the account by `paymentPointerId`?
3. In the [webhook](https://github.com/interledger/rafiki/blob/main/docs/transaction-api.md#webhooks) section of the transaction api doc, the `incoming_payment.completed` and `incoming_payment.expired` are describing the same behaviour (credit the receiver). This seems wrong, I feel like there should be no liquidity movement in the expired case?

wilsonianb:
1. 
>it seems that I would _not_ reuse the `/webhooks` route?

[...]

PR #796, opened 2022-11-29
feat(open-payments): add create incoming payment

raducristianpopa:

- Method for creating an incoming payment

## Context

- https://github.com/interledger/rafiki/issues/663#issuecomment-1327622353
## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/open-payments/src/client/incoming-payment.test.ts
+++ b/packages/open-payments/src/client/incoming-payment.test.ts
@@ -1,4 +1,5 @@
 import {
+  createIncomingPayment,
   createIncomingPaymentRoutes,
   getIncomingPayment,
   validateIncomingPayment
@@ -13,6 +14,8 @@ import {
   silentLogger
 } from '../test/helpers'
 import nock from 'nock'
+import { CreateIncomingPaymentArgs } from '../types'
+import { PostArgs } from './requests'
 
 describe('incoming-payment', (): void => {
   let openApi: OpenAPI
@@ -38,6 +41,21 @@ describe('incoming-payment', (): void => {
       expect(openApi.createResponseValidator).toHaveBeenCalledWith({
         path: '/incoming-payments/{id}',
         method: HttpMethod.GET
+      })
+    })
+
+    test('creates createIncomingPaymentOpenApiValidator properly', async (): Promise<void> => {
+      jest.spyOn(openApi, 'createResponseValidator')
+
+      createIncomingPaymentRoutes({
+        axiosInstance,
+        openApi,
+        logger
+      })
+
+      expect(openApi.createResponseValidator).toHaveBeenCalledWith({
+        path: '/incoming-payments',
+        method: HttpMethod.POST
       })
     })
   })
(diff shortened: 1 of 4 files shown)


packages/open-payments/src/client/requests.ts
```diff
@@ -4,11 +4,12 @@ import { ResponseValidator } from 'openapi'
 import { BaseDeps } from '.'
 import { createSignatureHeaders } from './signatures'
 
-interface GetArgs {
+export interface GetArgs {
```
mkurapov, line 7: I think we should keep these defined per resource, some routes have different `GetArgs`

packages/open-payments/src/types.ts
```diff
@@ -44,3 +45,6 @@ export const isInteractiveGrant = (
 export const isNonInteractiveGrant = (
   grant: InteractiveGrant | NonInteractiveGrant
 ): grant is NonInteractiveGrant => !!(grant as NonInteractiveGrant).access_token
+
+export type CreateIncomingPaymentArgs =
+  RSOperations['create-incoming-payment']['requestBody']['content']['application/json']
```
mkurapov, line 50: let's move this up to the other IncomingPayment types

mkurapov (commented):
Small comments

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -94,6 +110,95 @@ describe('incoming-payment', (): void => {
...
+            body: {
+              incomingAmount,
+              expiresAt,
+              description,
+              externalRef
+            }
+          },
+          openApiValidators.successfulValidator
+        )
+
```
wilsonianb, line 151:
```suggestion
        const scope = nock(baseUrl).post('/incoming-payment').reply(200, incomingPayment)

        const result = await createIncomingPayment(
          {
            axiosInstance,
            logger
          },
          {
            url: `${baseUrl}/incoming-payment`,
            body: {
              incomingAmount,
              expiresAt,
              description,
              externalRef
            }
          },
          openApiValidators.successfulValidator
        )

        scope.done()
```
ditto below

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -94,6 +110,95 @@ describe('incoming-payment', (): void => {
...
+    test('throws if the created incoming payment does not pass validation', async (): Promise<void> => {
+      const receivedAmount = {
+        assetCode: 'USD',
+        assetScale: 2,
+        value: '10'
+      }
+
+      const incomingPayment = mockIncomingPayment({
+        receivedAmount
+      })
```
wilsonianb, line 165: I had to do some digging to understand why the incoming payment was invalid.
How about something like this to make it explicit?
```suggestion
      const amount = {
        assetCode: 'USD',
        assetScale: 2,
        value: '10'
      }

      const incomingPayment = mockIncomingPayment({
        incomingAmount: amount,
        receivedAmount: amount,
        completed: false
      })
```

wilsonianb: :thinking: Should we actually do extra validation create incoming payment response?
Disallow non-zero `receivedAmount` / `completed`?

mkurapov: Do you mean non-zero `incomingAmount`?

wilsonianb: non-zero `receivedAmount.value` on a newly created incoming payment

mkurapov: Ah I see now, yeah, I think it would be good to add those additional checks :+1:

packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -1,15 +1,21 @@
...
 interface GetArgs {
   url: string
-  accessToken: string
+  accessToken?: string
+}
+
+interface PostArgs<T> {
+  url: string
+  body: T
 }
```
mkurapov, line 14: I think both `GetArgs` and `PostArgs` should have the `accessToken` and have it be a required field

mkurapov, line 14: I think both `GetArgs` and `PostArgs` should have the `accessToken` and have it be a required field

mkurapov (commented):
Comment + additional validation as @wilsonianb described, and we should be good to do

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -94,6 +111,106 @@ describe('incoming-payment', (): void => {
...
+    `(
+      'returns the incoming payment on success',
+      async ({
+        incomingAmount,
+        expiresAt,
+        description,
+        externalRef
+      }): Promise<void> => {
+        const incomingPayment = mockIncomingPayment({
+          incomingAmount,
```
wilsonianb, line 128: `mockIncomingPayment` always defines `incomingAmount`
https://github.com/interledger/rafiki/blob/465b537fe19a417ce45a79fe90cdefb88f661b6c/packages/open-payments/src/test/helpers.ts#L55-L59
At first I was going to suggest only define `incomingAmount` in `mockIncomingPayment` if it specified in the `overrides`.
Now I'm wondering if it should actually be allowed for a created incoming payment to have an `incomingAmount` if it isn't specified in the request? Or if it _is_ specified, should the created incoming payment be required to have that `incomingAmount` value?
I'm surprised neither of two have been mentioned in:
- https://github.com/interledger/rafiki/issues/649

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -94,6 +111,106 @@ describe('incoming-payment', (): void => {
...
+            logger
+          },
+          {
+            url: `${baseUrl}/incoming-payment`,
+            body: {},
+            accessToken: 'accessToken'
+          },
+          openApiValidators.failedValidator
+        )
+      ).rejects.toThrowError()
```
wilsonianb, line 210:
```suggestion
      const scope = nock(baseUrl)
        .post('/incoming-payment')
        .reply(200, incomingPayment)

            await expect(() =>
        createIncomingPayment(
          {
            axiosInstance,
            logger
          },
          {
            url: `${baseUrl}/incoming-payment`,
            body: {},
            accessToken: 'accessToken'
          },
          openApiValidators.failedValidator
        )
      ).rejects.toThrowError()
     scope.done()
```

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -241,4 +358,59 @@ describe('incoming-payment', (): void => {
...
+      )
+    })
+
+    test('throws if the created incoming payment is completed', async (): Promise<void> => {
+      const incomingPayment = mockIncomingPayment({
+        incomingAmount: {
+          assetCode: 'USD',
+          assetScale: 2,
+          value: '5'
+        },
```
wilsonianb, line 407:
```suggestion
```

packages/open-payments/src/client/incoming-payment.test.ts
wilsonianb: I don't think it's possible in OpenAPI 3.1's JSON Schemsa 2020-12 to validate the presence or value of response's `incomingAmounnt` based on the that of the request.

I think we should at least validate that if `incomingPayment` is specified, it is included in the response (with an equal or lesser value?)

wilsonianb (approved):
I'm ok tackling additional `incomingAmount` validation later once we've actually figured out if/how to do it:
- https://github.com/interledger/rafiki/pull/796#discussion_r1039576791
- https://github.com/interledger/rafiki/issues/649#issuecomment-1339418943

packages/open-payments/src/client/incoming-payment.test.ts
mkurapov: `incomingPayment.value` should be the same in the request and the response, so we could just check for equality.

But in line with what was said previously, we can merge this in and then follow up with the extra validation

mkurapov (approved):

mkurapov:
Merging as @raducristianpopa doesn't have access to do so: 
image

Issue #740, opened 2022-11-10
Update backend OpenAPI specs


wilsonianb:
https://github.com/interledger/rafiki/blob/e796dedeee7c3aedd43c5c977ce83f420ea8667c/packages/backend/src/config/app.ts#L107-L114
Need to figure out how to use
https://github.com/interledger/rafiki/blob/main/packages/auth/src/openapi/resource-server.yaml

PR #799, opened 2022-11-29
chore(backend): update OpenAPI specs

wilsonianb:

- 

## Context

- fixes https://github.com/interledger/rafiki/issues/740
## Checklist

- [ ] Tests added/updated
- [x] Documentation added


njlie (approved):

sabineschaller (commented):
I'm confused, why are we fetching the specs now instead of updating URLs in env files?

wilsonianb:
Here's the previous two related pr's:
- https://github.com/interledger/rafiki/pull/684
- https://github.com/interledger/rafiki/pull/692

The Rafiki specific specs use the shared Open Payments schemas file
https://github.com/interledger/rafiki/tree/main/packages/auth/src/openapi
https://github.com/interledger/open-payments/blob/main/openapi/schemas.yaml

So it is fetched instead of updating the `$ref` urls in the local files, and to support local dev using modified (pending changes to) specs/schemas:
- https://github.com/interledger/rafiki/pull/684#discussion_r1009661953

I suggested also fetching the relevant Open Payments spec(s)
- https://github.com/interledger/rafiki/pull/692#discussion_r1013470271

packages/backend/scripts/get-op-schema.sh
```diff
@@ -0,0 +1,13 @@
...
+SOURCE=${BASH_SOURCE[0]}
+while [ -L "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR=$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )
+  SOURCE=$(readlink "$SOURCE")
+  [[ $SOURCE != /* ]] && SOURCE=$DIR/$SOURCE # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+OUTDIR=$( cd -P "$( dirname "$SOURCE" )/../src/openapi" >/dev/null 2>&1 && pwd )
+
+curl -o "$OUTDIR/resource-server.yaml" https://raw.githubusercontent.com/interledger/open-payments/b363d33038fe789e5388f04f80ddd06a4fa97093/openapi/resource-server.yaml
+curl -o "$OUTDIR/schemas.yaml" https://raw.githubusercontent.com/interledger/open-payments/b363d33038fe789e5388f04f80ddd06a4fa97093/openapi/schemas.yaml
```
mkurapov, line 13: Can we also update the `open-payments` commit version in this PR

_or_ would it make sense to export these from `open-payments`? so we can be sure both are using the same spec? This does seem vaguely familiar though... have we discussed this before?

packages/backend/src/index.ts
```diff
@@ -108,13 +109,17 @@ export function initIocContainer(
...
-  container.singleton('authOpenApi', async (deps) => {
-    const config = await deps.use('config')
-    return await createOpenAPI(config.authServerSpec)
+  container.singleton('openApi', async () => {
+    const authServerSpec = await createOpenAPI(
+      path.resolve(__dirname, './openapi/auth-server.yaml')
+    )
+    const resourceServerSpec = await createOpenAPI(
+      path.resolve(__dirname, './openapi/resource-server.yaml')
+    )
```
mkurapov, line 118: Could we add a good error message here, for when the files can't be found?

https://interledgerfoundation.slack.com/archives/C0302LMKBT7/p1670413450794209?thread_ts=1670407497.730339&cid=C0302LMKBT7

packages/backend/scripts/get-op-schema.sh
mkurapov: We could even do it for both of `backend` and `auth` projects since `open-payments` contains both, which would eliminate the need for the `packages/backend/src/openapi/auth-server.yaml` dependency

README.md
```diff
@@ -81,8 +81,9 @@ pnpm --filter open-api test
 # run all tests
 pnpm -r --workspace-concurrency=1 test
 
-# pull in latest openapi specs for auth server:
+# pull in latest openapi specs:
 pnpm --filter auth fetch-schemas
+pnpm --filter backend fetch-schemas
```
mkurapov, line 86: can we add a root level command?
`"localenv:fetch-op-schemas": "pnpm --filter auth --filter backend fetch-schemas"` , and then include it as part of the local env setup readme?


https://interledgerfoundation.slack.com/archives/C0302LMKBT7/p1670414356297499?thread_ts=1670407497.730339&cid=C0302LMKBT7

packages/backend/scripts/get-op-schema.sh
wilsonianb: >would it make sense to export these from open-payments?

I'm favor of having defining the (default) commit.
It seems like that could happen while still allowing subsequent local tinkering of the specs
https://github.com/interledger/rafiki/pull/684#discussion_r1009661953

>which would eliminate the need for the packages/backend/src/openapi/auth-server.yaml dependency

Did you mean `resource-server.yaml`?

packages/backend/src/index.ts
wilsonianb: As long we remember not to change the command :sweat_smile:

packages/auth/src/index.ts
```diff
@@ -111,20 +111,28 @@ export function initIocContainer(
...
-    const idpSpec = await createOpenAPI(
-      path.resolve(__dirname, './openapi/id-provider.yaml')
-    )
-    return {
-      authServerSpec,
-      resourceServerSpec,
-      idpSpec
+  container.singleton('openApi', async (deps: IocContract<AppServices>) => {
+    try {
+      const authServerSpec = await createOpenAPI(
```
mkurapov, line 116: what do you think of `openPaymentsAuthServerSpec`? Likewise with `openPaymentsResourceServerSpec` in the `backend`?

Trying to see if there is a better way of distinguishing the several different spec we have

mkurapov (approved):
Once the specs stabilize it would be good to have them be defined in a singular place, even like with the exported commit hash as you suggested  - I am a bit worried about specs diverging accidentally

packages/auth/src/index.ts
wilsonianb: This one here is actually `authServerSpec` (from https://github.com/interledger/rafiki/blob/main/packages/auth/src/openapi/resource-server.yaml`)
I think `openPaymentsAuthServerSpec` would be https://github.com/interledger/open-payments/blob/main/openapi/schemas.yaml

What if we just called the first one introspection/tokenIntrospection for both the file name (make it consistent in both `auth` and `backend`) and the variable name?

wilsonianb: I'm gonna change https://github.com/interledger/rafiki/blob/main/packages/auth/src/openapi/resource-server.yaml to `token-introspection.yaml` with
```suggestion
      const tokenIntrospectionSpec = await createOpenAPI(
```
and not
```suggestion
      const tokenIntroSPECtion = await createOpenAPI(
```
and keep the other file and variable names as is.

mkurapov: The `tokenIntrospectionSpec` change makes sense, but it should be replacing`resourceServerSpec` variable (the one that contains the `introspection` route), right?

wilsonianb: Ha, yeah. I was thinking these comments were for `packages/backend/src/index.ts` :upside_down_face:

PR #797, opened 2022-11-29
feat(http-signature-utils): move http signature related code to this package

sabineschaller:
## Changes proposed in this pull request

- moves signature generation, jwk generation, and private key loading to here
- adds app with endpoint for postman to consume pre-request to generate signature headers

## Context

- fixes #665 
- fixes #818 

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [x] Documentation added
- [x] Make sure that all checks pass

PR #804, opened 2022-11-30
feat(auth): add content type and length to signature headers

### LINKED ISSUE #713: Include content type and length in signature input
[`generateSigHeaders`](https://github.com/interledger/rafiki/blob/11170b04bd77562019964f18e77704459ef037ec/packages/auth/src/tests/signature.ts#L57) should include `Content-Type` and `Content-Length` headers in the HTTP signature input.

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-7.3.1

njlie:

-  Adds `Content-Type` and `Content-Length` to the signature input & signature components of a generated httpsig signature.

## Context

- Closes #713 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/tests/signature.ts
+++ b/packages/auth/src/tests/signature.ts
@@ -69,18 +69,33 @@ export async function generateSigHeaders({
     body?: unknown
     authorization?: string
   }
-}): Promise<{ sigInput: string; signature: string; contentDigest?: string }> {
+}): Promise<{
+  sigInput: string
+  signature: string
+  contentDigest?: string
+  contentLength?: string
+  contentType?: string
+}> {
   let sigInputComponents = 'sig1=("@method" "@target-uri"'
   const { body, authorization } = optionalComponents ?? {}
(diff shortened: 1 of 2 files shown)


packages/auth/src/tests/signature.ts
```diff
@@ -69,18 +69,33 @@ export async function generateSigHeaders({
     body?: unknown
     authorization?: string
   }
-}): Promise<{ sigInput: string; signature: string; contentDigest?: string }> {
```
wilsonianb, line 72: Are we going to be able to just replace this with `http-signature-utils`' `createSignatureHeaders`?
- https://github.com/interledger/rafiki/pull/797

wilsonianb (approved):

packages/auth/src/tests/signature.ts
njlie: Yeah, pretty easily I think.

PR #807, opened 2022-12-01
663/mk/create outgoing payment

mkurapov:

- Adding `outgoingPayment.create` to open payments client

## Context

- Progress towards #663 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment/outgoing/service.ts
+++ b/packages/backend/src/open_payments/payment/outgoing/service.ts
@@ -419,10 +419,10 @@ async function addSentAmount(
 ): Promise<OutgoingPayment> {
   const fundingZeroOrUndefined =
     payment.state === OutgoingPaymentState.Funding ? BigInt(0) : undefined
-  let sent = value || (await deps.accountingService.getTotalSent(payment.id))
-  if (sent === undefined) {
-    sent = fundingZeroOrUndefined
-  }
+  const sent =
+    value ??
+    (await deps.accountingService.getTotalSent(payment.id)) ??
+    fundingZeroOrUndefined
 
   if (sent !== undefined) {
     payment.sentAmount = {
--- a/packages/open-payments/src/client/outgoing-payment.test.ts
+++ b/packages/open-payments/src/client/outgoing-payment.test.ts
@@ -1,4 +1,5 @@
 import {
+  createOutgoingPayment,
   createOutgoingPaymentRoutes,
   getOutgoingPayment,
   validateOutgoingPayment
@@ -12,6 +13,7 @@ import {
   silentLogger
 } from '../test/helpers'
 import nock from 'nock'
+import { v4 as uuid } from 'uuid'
 
 describe('outgoing-payment', (): void => {
   let openApi: OpenAPI
@@ -37,6 +39,20 @@ describe('outgoing-payment', (): void => {
       expect(openApi.createResponseValidator).toHaveBeenCalledWith({
         path: '/outgoing-payments/{id}',
         method: HttpMethod.GET
+      })
+    })
+
+    test('creates createOutgoingPaymentOpenApiValidator properly', async (): Promise<void> => {
+      jest.spyOn(openApi, 'createResponseValidator')
+
+      createOutgoingPaymentRoutes({
+        axiosInstance,
+        openApi,
+        logger
+      })
+      expect(openApi.createResponseValidator).toHaveBeenCalledWith({
+        path: '/outgoing-payments',
+        method: HttpMethod.POST
       })
     })
   })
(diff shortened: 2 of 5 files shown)


packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -419,10 +419,10 @@ async function addSentAmount(
 ): Promise<OutgoingPayment> {
   const fundingZeroOrUndefined =
     payment.state === OutgoingPaymentState.Funding ? BigInt(0) : undefined
-  let sent = value || (await deps.accountingService.getTotalSent(payment.id))
```
mkurapov, line 422: _This is a guess about the desired behaviour_

the main issue here was using `||` instead of `??`. When we call this function just above:
https://github.com/interledger/rafiki/blob/cbbc41613fc020ee5c2e6d974e7fc95755dec408/packages/backend/src/open_payments/payment/outgoing/service.ts#L160

we pass in `value: BigInt(0)` . `BigInt(0)` is falsy, so we end up having calling `deps.accountingService.getTotalSent(payment.id))` when we made value a defined value.

packages/open-payments/src/client/outgoing-payment.ts
```diff
@@ -39,11 +58,35 @@ export const getOutgoingPayment = async (
...
+    return validateOutgoingPayment(outgoingPayment)
+  } catch (error) {
+    const errorMessage = 'Could not validate outgoing payment'
+    logger.error({ url, validateError: error?.message }, errorMessage)
+
+    throw new Error(errorMessage)
+  }
+}
+
+export const createOutgoingPayment = async (
```
mkurapov, line 79: Here the only other validation I could think of doing on a newly created outgoing payment (that wasn't already covered in `validateOutgoingPayment`) was checking whether `sentAmount.value === 0`. Feels like a weird behaviour to already start processing a payment before returning it, but does the caller care?

packages/open-payments/src/client/requests.ts
```diff
@@ -84,7 +85,13 @@ export const post = async <TRequest, TResponse>(
   const url = requestUrl.href
 
   try {
-    const { data, status } = await axiosInstance.post<TResponse>(url, body)
+    const { data, status } = await axiosInstance.post<TResponse>(url, body, {
+      headers: accessToken
+        ? {
+            Authorization: `GNAP ${accessToken}`
+          }
+        : {}
```
mkurapov, line 93: was missing previously

packages/open-payments/src/client/outgoing-payment.test.ts
```diff
@@ -113,6 +129,109 @@ describe('outgoing-payment', (): void => {
...
   })
 
+  describe('createOutgoingPayment', (): void => {
+    test.each`
+      quoteId   | description           | externalRef
+      ${uuid()} | ${'Some description'} | ${'#INV-1'}
+      ${uuid()} | ${undefined}          | ${undefined}
+    `(
+      'creates outgoing payment',
+      async ({ quoteId, description, externalRef }): Promise<void> => {
```
wilsonianb, line 139: I think `quoteId` could be defined outside the `test.each` table?

packages/open-payments/src/test/helpers.ts
```diff
@@ -91,6 +91,8 @@ export const mockOutgoingPayment = (
   },
   quoteId: uuid(),
```
wilsonianb, line 92: `quoteId` and `id` are supposed to be `uri`s
https://github.com/interledger/open-payments/blob/02a6d505d79124cce776054df89e28f02c8e5d59/openapi/resource-server.yaml#L1072-L1084
Am I correct that we shouldn't be concerned about the quality of our OpenAPI validation because this package is only testing against `successfulValidator`/`failedValidator`?

packages/open-payments/src/client/requests.ts
wilsonianb: The `properly POSTs request` test is failing with the addition of `headers: {}`

packages/open-payments/src/client/outgoing-payment.ts
wilsonianb: I agree that does somehow seem less concerning than a non-zero `receivedAmount` on a new incoming payment.

packages/open-payments/src/test/helpers.ts
mkurapov: Good point, I'll make the test case a bit more realistic. But yes, in general, IMO the actual OpenAPI spec validation is a bit out of scope of the package, that responsibility is `openapi`s

wilsonianb (approved):

wilsonianb (approved):

PR #823, opened 2022-12-06
(open-payments): list outgoing payments

### LINKED ISSUE #817: (open-payments): list outgoing payments
https://docs.openpayments.guide/reference/list-outgoing-payments

mkurapov:

- Add list outgoing payments functionality to open payments client

## Context

- Fixes #817 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/open-payments/src/client/outgoing-payment.ts
+++ b/packages/open-payments/src/client/outgoing-payment.ts
@@ -14,12 +14,6 @@ interface GetArgs {
   accessToken: string
 }
 
-interface ListArgs {
-  url: string
-  accessToken: string
-  pagination?: PaginationArgs
-}
-
 interface PostArgs<T> {
   url: string
   body: T
@@ -28,7 +22,10 @@ interface PostArgs<T> {
 
 export interface OutgoingPaymentRoutes {
   get(args: GetArgs): Promise<OutgoingPayment>
-  list(args: ListArgs): Promise<OutgoingPaymentPaginationResult>
+  list(
+    args: GetArgs,
+    pagination?: PaginationArgs
+  ): Promise<OutgoingPaymentPaginationResult>
   create(args: PostArgs<CreateOutgoingPaymentArgs>): Promise<OutgoingPayment>
 }
 
@@ -62,11 +59,12 @@ export const createOutgoingPaymentRoutes = (
         args,
         getOutgoingPaymentOpenApiValidator
       ),
-    list: (args: ListArgs) =>
+    list: (getArgs: GetArgs, pagination?: PaginationArgs) =>
       listOutgoingPayments(
         { axiosInstance, logger },
-        args,
-        listOutgoingPaymentOpenApiValidator
+        getArgs,
+        listOutgoingPaymentOpenApiValidator,
+        pagination
       ),
     create: (args: PostArgs<CreateOutgoingPaymentArgs>) =>
       createOutgoingPayment(
@@ -127,20 +125,19 @@ export const createOutgoingPayment = async (
 
 export const listOutgoingPayments = async (
   deps: BaseDeps,
-  args: ListArgs,
-  validateOpenApiResponse: ResponseValidator<OutgoingPaymentPaginationResult>
+  getArgs: GetArgs,
+  validateOpenApiResponse: ResponseValidator<OutgoingPaymentPaginationResult>,
+  pagination?: PaginationArgs
 ) => {
   const { axiosInstance, logger } = deps
-  const { url, accessToken } = args
+  const { url, accessToken } = getArgs
 
   const outgoingPayments = await get(
     { axiosInstance, logger },
     {
       url,
       accessToken,
-      ...(args.pagination
-        ? { queryParams: { pagination: args.pagination } }
-        : {})
+      ...(pagination ? { queryParams: { ...pagination } } : {})
     },
     validateOpenApiResponse
   )
--- a/packages/open-payments/src/client/outgoing-payment.ts
+++ b/packages/open-payments/src/client/outgoing-payment.ts
@@ -1,6 +1,11 @@
 import { HttpMethod, ResponseValidator } from 'openapi'
 import { BaseDeps, RouteDeps } from '.'
-import { getRSPath, OutgoingPayment } from '../types'
+import {
+  getRSPath,
+  OutgoingPayment,
+  OutgoingPaymentPaginationResult,
+  PaginationArgs
+} from '../types'
 import { get } from './requests'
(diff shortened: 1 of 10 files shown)


packages/open-payments/src/client/outgoing-payment.ts
```diff
@@ -100,6 +128,47 @@ export const createOutgoingPayment = async (
...
 
+export const listOutgoingPayments = async (
+  deps: BaseDeps,
+  getArgs: ListGetArgs,
+  validateOpenApiResponse: ResponseValidator<OutgoingPaymentPaginationResult>,
+  pagination?: PaginationArgs
+) => {
+  const { axiosInstance, logger } = deps
+  const { accessToken, paymentPointer } = getArgs
+  const url = `${paymentPointer}${getRSPath('/outgoing-payments')}`
```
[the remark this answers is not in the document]
mkurapov: @wilsonianb

packages/open-payments/src/types.ts
```diff
@@ -18,6 +18,20 @@ export type ILPStreamConnection =
...
 export type CreateOutgoingPaymentArgs =
   RSOperations['create-outgoing-payment']['requestBody']['content']['application/json']
+type PaginationResult<T> = {
+  pagination: RSComponents['schemas']['page-info']
+  result: T[]
+}
+export type OutgoingPaymentPaginationResult = PaginationResult<OutgoingPayment>
+export type ForwardPagination =
+  RSComponents['schemas']['forward-pagination'] & {
+    last?: never
```
wilsonianb, line 28: I was surprised this was necessary because we have `additionalProperties: false`
https://github.com/interledger/open-payments/blob/main/openapi/resource-server.yaml#L1214

Would strict type checking make the `never` unnecessary?
If so, would that be something we couldn't enforce/guarantee on clients of this library?

packages/open-payments/src/client/outgoing-payment.ts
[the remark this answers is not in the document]
mkurapov: yeah exactly what I was thinking. For collection requests/creates its pretty straightforward, but for the GETs, we could allow either full `url` to be passed in as an argument, or just an `id`

packages/open-payments/src/types.ts
mkurapov: So this isn't an Open API validation issue, this is for the TS types. Without the `never` you can provide `first` `last` and `cursor` and TS won't complain:

Screen Shot 2022-12-13 at 11 26 15 AM

while with `never`, it'll be an error:

Screen Shot 2022-12-13 at 11 26 30 AM


https://stackoverflow.com/a/61350902

packages/open-payments/src/client/outgoing-payment.ts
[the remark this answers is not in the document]
wilsonianb: For `GET`s and `/complete`s, I vote we at least do the full url.
I was going to suggest we even call the full url `id`, since that's the name of the field in responses. But we also refer to an `id` path parameter
https://docs.openpayments.guide/reference/get-incoming-payment

packages/open-payments/src/types.ts
wilsonianb: Right, I'm just realizing/thinking through (and I guess trying to avoid) the implications of `additionalProperties: false` not being enforced in the generated types without specifying `never`s.

mkurapov: Isn't that mostly a concern for these `oneOf` union types?

packages/open-payments/src/client/requests.test.ts
```diff
@@ -114,6 +114,43 @@ describe('requests', (): void => {
...
+    test.each`
+      title                      | queryParams
+      ${'all defined values'}    | ${{ first: 5, cursor: 'id' }}
+      ${'some undefined values'} | ${{ first: 5, cursor: undefined }}
+    `(
+      'properly sets query params with $title',
+      async ({ queryParams }): Promise<void> => {
+        const cleanedQueryParams: any = Object.fromEntries(
+          Object.entries(queryParams).filter(([_, v]) => v != null)
+        )
```
mkurapov, line 126: So for some reason, including an undefined param in 
`nock.get().query({ first: 5, cursor: undefined })` will always cause an `Nock: No match for request` error, even though, when you inspect the query params via a function:
Screen Shot 2022-12-13 at 7 13 38 PM

the parameter isn't there.

packages/open-payments/src/client/outgoing-payment.ts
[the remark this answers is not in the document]
wilsonianb: If we just used `id` (no url or payment pointer url) for `GET`s and `/complete`, we could do things like
```
const payment = await client.incomingPayment.create({
  url: paymentPointerUrl
})
const currentPayment = await client.incomingPayment.get({
  id: payment.id
})
await client.incomingPayment.complete({
  id: payment.id
})
```

packages/open-payments/src/client/requests.test.ts
```diff
@@ -114,6 +114,46 @@ describe('requests', (): void => {
...
+          Object.entries(queryParams).filter(([_, v]) => v != null)
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ) as any
+
+        const scope = nock(baseUrl)
+          .matchHeader('Signature', (sig) => sig === undefined)
+          .matchHeader('Signature-Input', (sigInput) => sigInput === undefined)
+          .get('/incoming-payments')
+          .query(cleanedQueryParams)
+          .reply(200)
```
mkurapov, line 135: So for some reason, including an undefined param in 
`nock.get().query({ first: 5, cursor: undefined })` will always cause an `Nock: No match for request` error, even though, when you inspect the query params via a function:

Screen Shot 2022-12-13 at 7 13 38 PM

the parameter isn't there. This isn't great but at least we can assert that the `axios.get` is called with the `cleanedQueryParams`

packages/open-payments/src/types.ts
wilsonianb: Yeah, I don't think there's any concerns here with the OpenAPI validation itself.
But I hadn't considered that a generated type from an object with `additionalProperties: false` could accept an object with additional properties (unless the specified properties are manual defined as `never`)

wilsonianb (approved):
We _could_ change `url` to `id` for `GET` and `/complete` and to `paymentPointerUrl`(?) for collection requests.
However, I doubt that's what we land on If/when we switch to non-payment pointer-based resource urls.

Issue #815, opened 2022-12-05
(open-payments): complete incoming payment


mkurapov:
Add functionality to the open payments client to complete an incoming payment: https://docs.openpayments.guide/reference/complete-incoming-payment

PR #833, opened 2022-12-09
feat(open-payments): add complete incoming payment

raducristianpopa:

- add the completion of an incoming payment to OP Client

## Context
- fixes https://github.com/interledger/rafiki/issues/815

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/open-payments/src/client/incoming-payment.ts
+++ b/packages/open-payments/src/client/incoming-payment.ts
@@ -17,6 +17,7 @@ interface PostArgs<T = undefined> {
 export interface IncomingPaymentRoutes {
   get(args: GetArgs): Promise<IncomingPayment>
   create(args: PostArgs<CreateIncomingPaymentArgs>): Promise<IncomingPayment>
+  complete(args: PostArgs): Promise<IncomingPayment>
 }
 
 export const createIncomingPaymentRoutes = (
@@ -36,6 +37,12 @@ export const createIncomingPaymentRoutes = (
       method: HttpMethod.POST
     })
 
+  const completeIncomingPaymentOpenApiValidator =
+    openApi.createResponseValidator<IncomingPayment>({
+      path: getRSPath('/incoming-payments/{id}/complete'),
+      method: HttpMethod.POST
+    })
+
   return {
     get: (args: GetArgs) =>
       getIncomingPayment(
@@ -48,6 +55,12 @@ export const createIncomingPaymentRoutes = (
         { axiosInstance, logger },
         args,
         createIncomingPaymentOpenApiValidator
+      ),
+    complete: (args: PostArgs) =>
+      completeIncomingPayment(
+        { axiosInstance, logger },
+        args,
+        completeIncomingPaymentOpenApiValidator
       )
   }
 }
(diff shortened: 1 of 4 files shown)


raducristianpopa:
When creating an incoming payment, if the `expiresAt` property is not set, it will automatically set it (in the backend).
https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/backend/src/open_payments/payment/incoming/service.ts#L27
https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/backend/src/open_payments/payment/incoming/service.ts#L98-L114

At the moment, in the Open Payments API Specs, `expiresAt` is optional.
https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/open-payments/src/generated/resource-server-types.ts#L174-L178

Shouldn't every incoming payment that is going to get **retrieved** have the `expiresAt` property ?
Does `expiresAt` need to be set to `required` in the Open Payments API specs ?

raducristianpopa:
I know this is not the scope of this PR, but can this error be fixed here too ?
```
'"../types"' has no exported member named 'ContinuationRequest'. Did you mean 'GrantContinuationRequest'?ts(2724)
```

https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/open-payments/src/test/helpers.ts#L4-L12

https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/open-payments/src/types.ts#L42-L44

mkurapov:
Good catch on the `GrantContinuationRequest` type. Can fix the error on this PR :+1:

wilsonianb:
>Shouldn't every incoming payment that is going to get retrieved have the expiresAt property ?
Does expiresAt need to be set as required in the Open Payments API specs ?

Every incoming payment that is going to get retrieved **from a Rafiki RS/backend** will have `expiresAt`, but that is an implementation detail of Rafiki.
Open Payments doesn't require it.

packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -151,3 +188,19 @@ export const validateCreatedIncomingPayment = (
 
   return validateIncomingPayment(payment)
 }
+
+export const validateCompletedIncomingPayment = (
+  payment: IncomingPayment
+): IncomingPayment => {
+  const { completed, expiresAt } = payment
+
+  if (new Date(expiresAt).getTime() <= Date.now()) {
```
mkurapov, line 197:
```suggestion
  if (expiresAt && new Date(expiresAt).getTime() <= Date.now()) {
```
Since it's possible `expiresAt` can be undefined in Open Payments. Otherwise, if its undefined, it'll be an `Invalid Date`. 

I'm going to enable strict type checking in `open-payments` so TypeScript complains about this next time

mkurapov: I would argue this seems to be like a check mostly necessary at the server receiving this request (which we don't explicitly do based on `expiresAt`, we use the implicit `state`: https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/backend/src/open_payments/payment/incoming/service.ts#L286-L309).

This check could throw even if the non-expired incoming payment was completed successfully, when the `expiryDate` was right at the time of the request

mkurapov (commented):
Looks good, I would consider removing the `expiresAt` check

packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -151,3 +188,19 @@ export const validateCreatedIncomingPayment = (
...
+    throw new Error('Can not complete an expired incoming payment.')
+  }
```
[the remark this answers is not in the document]
mkurapov: I like the idea of adding 409 code as an option to the spec 👍

mkurapov (approved):

wilsonianb (approved):

Issue #813, opened 2022-12-05
(open-payments): rotate token 


mkurapov:
Add functionality to rotate a token in the open-payments client: https://docs.openpayments.guide/reference/post-token. Will need to make a function `createTokenRoutes` to the `createAuthenticatedClient`. This lives in the authorization server spec

Note: make sure this is using the auth server Open API spec

domin191013:
Completed

PR #843, opened 2022-12-13
Fix/issue 813

domin191013:
- (open-payments): rotate token
[Issue Url](https://github.com/interledger/rafiki/issues/813)

## Context
Fixed issue #813

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/open-payments/src/client/token.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export interface TokenRoutes {
+  rotate(
+    args: RotateRequestArgs
+  ): Promise<InteractiveGrant | NonInteractiveGrant>
+}
+
+export const createTokenRoutes = (deps: TokenRouteDeps): TokenRoutes => {
+  const revokeTokenValidator = deps.openApi.createResponseValidator<
+    InteractiveGrant | NonInteractiveGrant
```
mkurapov, line 22:
```suggestion
    AccessToken
```

I think we should create a `AccessToken` type to export, and have that be the return type here. This is what we get back in the response. You can see it in the open api spec: https://github.com/interledger/open-payments/blob/ae0e924a56fe9f12834f65da109a1042760c1605/openapi/auth-server.yaml#L247-L253

Similarly, if you take a look at the incoming request, there is no `body` parameter. Just the `url` provided in the `post` call is sufficient for now.

packages/open-payments/src/client/token.ts
```diff
@@ -0,0 +1,35 @@
...
+    })
+
+  return {
+    rotate: (args: RotateRequestArgs) =>
+      post(
+        deps,
+        {
+          url: args.url
+        },
+        rotateTokenValidator
```
wilsonianb, line 32: I was going to suggest validating the reponse's `access_token.manage` matches `args.url`, but
>The response MUST include an access token management URI, and the value of this URI MAY be different from the URI used by the client instance to make the rotation call.

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6-1

packages/open-payments/src/client/token.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      jest.spyOn(openApi, 'createResponseValidator')
+
+      createTokenRoutes({ axiosInstance, openApi, logger, client })
+      expect(openApi.createResponseValidator).toHaveBeenCalledTimes(1)
+      expect(openApi.createResponseValidator).toHaveBeenCalledWith({
+        path: '/token/{id}',
+        method: HttpMethod.POST
+      })
+    })
+  })
```
wilsonianb, line 28: Add `rotate` tests with  `openApiValidators.successfulValidator`/`openApiValidators.failedValidator`
See:
https://github.com/interledger/rafiki/blob/1124707caf2f6ed9f7ddae777898973f9107b749/packages/open-payments/src/client/outgoing-payment.test.ts#L132

packages/open-payments/src/client/index.ts
```diff
@@ -141,6 +143,12 @@ export const createAuthenticatedClient = async (
       logger,
       client: args.paymentPointerUrl
     }),
+    token: createTokenRoutes({
+      axiosInstance,
+      openApi: authServerOpenApi,
+      logger,
+      client: args.paymentPointerUrl
```
mkurapov, line 150:
```suggestion
```
I don't think we use the `client` in the route

packages/open-payments/src/client/token.ts
domin191013: Yeah, I see. So this doesn't need any update for now I guess.

packages/open-payments/src/client/token.test.ts
domin191013: Well, to be honest I don't quite understand about this.
In open-payments.test.ts, it's testing this function.

export const getOutgoingPayment = async (
  deps: BaseDeps,
  args: GetArgs,
  validateOpenApiResponse: ResponseValidator<OutgoingPayment>
) => {
}

But in token.ts, it doesn't have that function format, except post() function which will be tested in requests.test.ts

domin191013: The only function in token.ts is rotate() function, which only takes url as parameber not validators like successfulValidators or failed Validators.

Could you give me some more details on this?

wilsonianb: You would need to move the post() to an exported function (like `getOutgoingPayment`).

Maybe the route specific tests should be having `nock` check the request body, to distinguish from the general `post` tests in `requests.test.ts`
https://github.com/interledger/rafiki/blob/1124707caf2f6ed9f7ddae777898973f9107b749/packages/open-payments/src/client/outgoing-payment.test.ts#L148-L150

mkurapov: Right, we usually don't have those validator tests for routes/handlers that directly call `get` or `post`

wilsonianb: The exported rotate fucntion should look similar to
https://github.com/interledger/rafiki/blob/1124707caf2f6ed9f7ddae777898973f9107b749/packages/open-payments/src/client/outgoing-payment.ts#L79-L91

packages/open-payments/src/client/token.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  describe('rotateToken', (): void => {
+    test('returns accessToken if passes validation', async (): Promise<void> => {
+      const accessToken = mockAccessToken()
+      const token = 'accessToken'
+
+      nock(baseUrl).post(`/token/${token}`).reply(200, accessToken)
```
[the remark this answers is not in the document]
wilsonianb: You could actually do:
```suggestion
      const manageUrl = new URL(accessToken.access)

      nock(manageUrl.origin).post(manageUrl.pathname).reply(200, accessToken)
```

packages/open-payments/src/client/token.test.ts
mkurapov: @wilsonianb we could probably test like I was doing in the other PR https://github.com/interledger/rafiki/pull/848/files#diff-4f318c2a374cc145f8825fecf95d77b29a728927b5a64ce393897b01becb2b93R95-R118

packages/open-payments/src/test/helpers.ts
```diff
@@ -189,3 +190,20 @@ export const mockContinuationRequest = (
   interact_ref: uuid(),
   ...overrides
 })
+
+export const mockAccessToken = (
+  overrides?: Partial<AccessToken>
+): AccessToken => ({
+  access_token: {
+    value: '99C36C2A4DB5BEBC',
+    manage: 'http://example.com/token/',
```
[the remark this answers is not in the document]
domin191013: Yup. For testing, is it okay to simply put 'accessTokenId' at the end?

wilsonianb (dismissed):

packages/open-payments/src/client/token.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { url } = args
+
+  return post(
+    {
+      axiosInstance,
+      logger
+    },
+    {
+      url
+    },
```
mkurapov, line 29: Based on the auth Open Payments spec we also need to pass in `accessToken` I believe

packages/open-payments/src/client/token.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          openApi,
+          logger
+        },
+        {
+          url: accessToken.access_token.manage,
+          accessToken: accessToken.access_token.value
+        },
+        openApiValidators.successfulValidator
+      )
+      expect(result).toStrictEqual(accessToken)
```
mkurapov, line 55:
```suggestion
      const scope = nock(manageUrl.origin).post(manageUrl.pathname).reply(200, accessToken)

      const result = await rotateToken(
        {
          axiosInstance,
          openApi,
          logger
        },
        {
          url: accessToken.access_token.manage,
          accessToken: accessToken.access_token.value
        },
        openApiValidators.successfulValidator
      )
      expect(result).toStrictEqual(accessToken)
      scope.done()
```

packages/open-payments/src/client/token.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      const result = await rotateToken(
+        {
+          axiosInstance,
+          openApi,
+          logger
+        },
+        {
+          url: accessToken.access_token.manage,
+          accessToken: accessToken.access_token.value
```
mkurapov, line 51: It doesn't really matter since we are throwing anyway, but I would put a different "fake" accessToken here because "accessToken.access_token.value" is the "new", rotated value

packages/open-payments/src/client/token.test.ts
mkurapov: Same for the test below

packages/open-payments/src/client/token.test.ts
domin191013: Yeah, I got it.

PR #848, opened 2022-12-14
812/mk/cancel grant

### LINKED ISSUE #812: (open-payments): cancel grant
Add functionality to cancel a grant in the open-payments client: https://docs.openpayments.guide/reference/delete-continue

mkurapov:

- Adding open payments client support for cancel grant call: https://docs.openpayments.guide/reference/delete-continue
- Fixing grant continuation call to include `accessToken`

## Context

- Fixes #812 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/open-payments/src/client/grant.test.ts
+++ b/packages/open-payments/src/client/grant.test.ts
@@ -1,7 +1,19 @@
 import { createGrantRoutes } from './grant'
 import { OpenAPI, HttpMethod, createOpenAPI } from 'openapi'
 import config from '../config'
-import { defaultAxiosInstance, silentLogger } from '../test/helpers'
+import {
+  defaultAxiosInstance,
+  mockGrantRequest,
+  silentLogger
+} from '../test/helpers'
+import * as requestors from './requests'
+import { v4 as uuid } from 'uuid'
+
+jest.mock('./requests', () => ({
+  ...jest.requireActual('./requests.ts'),
+  deleteRequest: jest.fn(),
+  post: jest.fn()
+}))
 
 describe('grant', (): void => {
   let openApi: OpenAPI
@@ -10,16 +22,18 @@ describe('grant', (): void => {
     openApi = await createOpenAPI(config.OPEN_PAYMENTS_AS_OPEN_API_URL)
   })
 
-  const axiosInstance = defaultAxiosInstance
-  const logger = silentLogger
-  const client = 'https://example.com/.well-known/pay'
+  const deps = {
+    axiosInstance: defaultAxiosInstance,
+    logger: silentLogger,
+    client: 'https://example.com/.well-known/pay'
+  }
 
   describe('createGrantRoutes', (): void => {
     test('creates response validator for grant requests', async (): Promise<void> => {
       jest.spyOn(openApi, 'createResponseValidator')
 
-      createGrantRoutes({ axiosInstance, openApi, logger, client })
-      expect(openApi.createResponseValidator).toHaveBeenCalledTimes(2)
+      createGrantRoutes({ openApi, ...deps })
+      expect(openApi.createResponseValidator).toHaveBeenCalledTimes(3)
       expect(openApi.createResponseValidator).toHaveBeenCalledWith({
         path: '/',
         method: HttpMethod.POST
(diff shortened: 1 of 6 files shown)


packages/open-payments/src/client/grant.ts
```diff
@@ -42,29 +42,50 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
...
+    continue: (
+      { url, accessToken }: PostArgs,
+      args: GrantContinuationRequest
+    ) =>
       post(
         deps,
         {
-          url: args.url,
+          url,
+          accessToken,
```
mkurapov, line 74: added `accessToken` here

packages/open-payments/src/client/requests.test.ts
```diff
@@ -365,4 +365,162 @@ describe('requests', (): void => {
...
+      scope.done()
+
+      expect(axiosInstance.delete).toHaveBeenCalledWith(`${baseUrl}/grant/`, {
+        headers: {
+          Authorization: `GNAP ${accessToken}`
+        }
+      })
+    })
+
+    test('throws if non-succesful status', async (): Promise<void> => {
```
mkurapov, line 468: more so testing axios behaviour here

packages/open-payments/src/client/requests.ts
```diff
@@ -129,6 +134,58 @@ export const post = async <TRequest, TResponse>(
...
+        ? {
+            Authorization: `GNAP ${accessToken}`
+          }
+        : {}
+    })
+
+    try {
+      openApiResponseValidator({
+        status,
+        body: {}
```
mkurapov, line 164: I'm not entirely sure if `body: {}` is ok here. Going to dig some more into the validation lib

mkurapov: Screen Shot 2022-12-14 at 7 21 16 PM

So it looks like for the grant cancellation request which doesn't have a response body, the validator prefers `null | undefined` over anything else.

packages/open-payments/src/client/requests.ts
```diff
@@ -129,6 +134,58 @@ export const post = async <TRequest, TResponse>(
...
+        ? {
+            Authorization: `GNAP ${accessToken}`
+          }
+        : {}
+    })
+
+    try {
+      openApiResponseValidator({
+        status,
+        body: data || undefined
```
mkurapov, line 164: Screen Shot 2022-12-14 at 7 21 16 PM

So it looks like for the grant cancellation request which doesn't have a response body, the validator will support `null | undefined`. When I was testing, `nock` would insert empty string as a the response body, so I'm effectively allowing the AS to return empty string for an empty response here.

packages/open-payments/src/client/grant.ts
```diff
@@ -42,29 +42,50 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
...
           body: {
-            ...args.request
+            ...args
           }
```
wilsonianb, line 77:
```
body: args
```

packages/open-payments/src/client/requests.test.ts
```diff
@@ -365,4 +365,171 @@ describe('requests', (): void => {
...
+            'nextTick',
+            'setImmediate',
+            'clearImmediate',
+            'setInterval',
+            'clearInterval',
+            'setTimeout',
+            'clearTimeout'
+          ]
+        })
+        .setSystemTime(new Date())
```
wilsonianb, line 398: I think this is only necessary if/when we validate the `Signature-Input` header's `created`

packages/open-payments/src/client/requests.test.ts
```diff
@@ -365,4 +365,171 @@ describe('requests', (): void => {
...
+
+      const scope = nock(baseUrl)
```
wilsonianb, line 400:
```suggestion
      const scope = nock(baseUrl)
        .matchHeader('Signature', /sig1=:([a-zA-Z0-9+/]){86}==:/)
        .matchHeader(
          'Signature-Input',
          `sig1=("@method" "@target-uri");created=${Math.floor(
            Date.now() / 1000
          )};keyid="${keyId}";alg="ed25519"`
        )
```

packages/open-payments/src/client/requests.test.ts
```diff
@@ -365,4 +365,171 @@ describe('requests', (): void => {
...
+      scope.done()
+
+      expect(axiosInstance.delete).toHaveBeenCalledWith(`${baseUrl}/grant/`, {
+        headers: {
+          Authorization: `GNAP ${accessToken}`
+        }
+      })
+    })
+
+    test('throws if non-succesful status', async (): Promise<void> => {
```
wilsonianb, line 468:
```suggestion
    test('throws if non-successful status', async (): Promise<void> => {
```

packages/open-payments/src/client/grant.test.ts
```diff
@@ -10,23 +23,98 @@ describe('grant', (): void => {
...
+    const accessToken = 'someAccessToken'
+
+    describe('request', () => {
+      test('calls post method with correct validator', async (): Promise<void> => {
+        const mockResponseValidator = ({ path, method }) =>
+          path === '/' && method === HttpMethod.POST
+
+        jest
+          .spyOn(openApi, 'createResponseValidator')
+          .mockImplementation(mockResponseValidator as never)
```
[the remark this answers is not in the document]
wilsonianb: Why the `as never`?

Are you thinking the same can be done for other route tests?

packages/open-payments/src/client/grant.test.ts
```diff
@@ -10,23 +23,98 @@ describe('grant', (): void => {
...
+
+        const postSpy = jest.spyOn(requestors, 'post').mockResolvedValueOnce({})
```
wilsonianb, line 45: Do we need the `mockResolvedValueOnce` here if we have the mocked `post: jest.fn()` above?

packages/open-payments/src/client/requests.test.ts
mkurapov: Is this necessary if there is no `Authorization` header? Do we want to update our interceptor to include `DELETE`s ?

packages/open-payments/src/client/grant.test.ts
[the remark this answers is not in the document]
mkurapov: > Why the `as never`?
I was just treating it as `any` without disabling the `@typescript-eslint/no-explicit-any` rule 🙃 

> Are you thinking the same can be done for other route tests?
Yes, exactly

packages/open-payments/src/client/grant.test.ts
mkurapov: Yup, don't need it. Thanks

packages/open-payments/src/client/requests.test.ts
wilsonianb: I guess not.
I think our only `DELETE`s will be for grants and tokens, both of which will have an authorization header access token.

packages/open-payments/src/client/grant.test.ts
```diff
@@ -10,23 +24,98 @@ describe('grant', (): void => {
...
+    const accessToken = 'someAccessToken'
+
+    describe('request', () => {
+      test('calls post method with correct validator', async (): Promise<void> => {
+        const mockResponseValidator = ({ path, method }) =>
+          path === '/' && method === HttpMethod.POST
+
+        jest
+          .spyOn(openApi, 'createResponseValidator')
+          .mockImplementation(mockResponseValidator as any)
```
wilsonianb, line 44: Should we be clearing these mocks?
I'm guessing we can't just do `jest.clearAllMocks()` in an `afterEach` because of the `jest.mock('./requests'`

wilsonianb (approved):

packages/open-payments/src/client/grant.test.ts
mkurapov: Because we have `clearMocks: true` in `jest.config.js`, they are cleared after every test already

https://jestjs.io/docs/configuration#clearmocks-boolean

Issue #816, opened 2022-12-05
(open-payments): list incoming payments


mkurapov:
Add functionality to list incoming payments in the the open payments client: https://docs.openpayments.guide/reference/list-incoming-payments

mkurapov:
Refer to https://github.com/interledger/rafiki/issues/817 for similar functionality

PR #851, opened 2022-12-15
feat(open-payments): add list incoming payments

raducristianpopa:

- add list incoming payments to OP Client

## Context

- fixes https://github.com/interledger/rafiki/issues/816

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -322,6 +340,156 @@ describe('incoming-payment', (): void => {
...
+          scope.done()
+        }
+      )
+    })
+
+    describe('backward pagination', (): void => {
+      test.each`
+        last         | cursor
+        ${undefined} | ${uuid()}
+        ${5}         | ${uuid()}
```
mkurapov, line 394: This test is setup correctly, but in general makes me confused about the backward pagination parameters. 

If we only provide a `cursor`, like (backwards-pagination allows) then by default we will end up forwards paginating, based on this code in the `backend`:

https://github.com/interledger/rafiki/blob/b19919a68d1ff782193d171d83053579b5abdff2/packages/backend/src/shared/pagination.ts#L9-L20

would this mean that we should make `last` required as well in Open Payments API? Only `cursor` is required:
 
image

wilsonianb: :grimacing: 
related:
- https://github.com/interledger/open-payments/issues/233
- https://github.com/interledger/open-payments/issues/136

packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -137,6 +165,48 @@ export const completeIncomingPayment = async (
...
+      url,
+      accessToken,
+      ...(pagination ? { queryParams: { ...pagination } } : {})
+    },
+    validateOpenApiResponse
+  )
+
+  for (const incomingPayment of incomingPayments.result) {
+    try {
+      validateIncomingPayment(incomingPayment)
```
wilsonianb, line 191: For list responses, we _could_ validate that `pagination.hasPreviousPage` is `false` if `cursor` was not specified (which should be forward pagination from the start).

packages/open-payments/src/client/incoming-payment.test.ts
mkurapov: I'm in favour of _not_ supporting both parameters (for now) to make the Open Payments pagination type similar to what you describe in [comment](https://github.com/interledger/open-payments/pull/129#discussion_r887840847):

`after + count` || `before + count`

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -74,6 +77,21 @@ describe('incoming-payment', (): void => {
         method: HttpMethod.POST
       })
     })
+
+    test('creates listIncomingPaymentsOpenApiValidator', async (): Promise<void> => {
```
mkurapov, line 81: @raducristianpopa can we try testing that the `list` route calls the proper validator, similar to this test? https://github.com/interledger/rafiki/pull/848/files#diff-4f318c2a374cc145f8825fecf95d77b29a728927b5a64ce393897b01becb2b93R37-R38

If it takes too long, I'm going to approve the PR anyway so we can merge it later and working on these route tests at a separate time

packages/open-payments/src/client/incoming-payment.ts
mkurapov: Mostly agree, but I'm not sure it's worth failing the whole request over it, however.

mkurapov (dismissed):

wilsonianb (dismissed):

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -13,10 +14,22 @@ import {
...
 import nock from 'nock'
+import { v4 as uuid } from 'uuid'
+import * as requestors from './requests'
+import { getRSPath } from '../types'
+
+jest.mock('./requests', () => {
+  return {
+    // https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options
+    __esModule: true,
+    ...jest.requireActual('./requests')
```
mkurapov, line 30: I think we can mock out the `get` here, so we don't need to `.mockResolvedValueOnce(incomingPaymentPaginationResult)` on the `getSpy` below

mkurapov: I also think we can remove `__esModule: true,` because we aren't using a default export

raducristianpopa: > I think we can mock out the `get` here, so we don't need to `.mockResolvedValueOnce(incomingPaymentPaginationResult)` on the `getSpy` below

If we mock the `get`, all the tests that are using it will fail:
```
Test Suites: 1 failed, 1 total
Tests:       8 failed, 26 passed, 34 total
```
image

---
> I also think we can remove __esModule: true, because we aren't using a default export

When `__esModule: true` is not specified, we get the following error:

```
incoming-payment › routes › list › calls get method with correct validator

    Error when making Open Payments GET request: Nock: No match for request {
      "method": "GET",
      "url": "http://localhost:1000/.well-known/pay/incoming-payments",
      "headers": {
        "accept": "application/json, text/plain, */*",
        "content-type": "application/json",
        "authorization": "GNAP accessToken",
        "signature": "sig1=:VYmcFqf6cSzLN10+dG/TnXQI23FxWY/n2/wVbNW4oX+U4rCBHpbT75AaMZOe3qzoUDiQUxZg0fpXjSBxHkNpDw==:",
        "signature-input": "sig1=(\"@method\" \"@target-uri\" \"authorization\");created=1671199508;keyid=\"default-key-id\";alg=\"ed25519\"",
        "user-agent": "axios/1.1.2",
        "accept-encoding": "gzip, deflate, br"
      }
    }
```

mkurapov (approved):
Looks good, can merge in once the conflict is resolved

wilsonianb (approved):

PR #860, opened 2022-12-16
feat(auth): check that token value matches during rotation & revocation

### LINKED ISSUE #382: Idempotency for internal GraphQL API
to not have to worry about duplicate calls

See e.g.: https://shopify.engineering/building-resilient-graphql-apis-using-idempotency

### LINKED ISSUE #832: Verify provided access token value during token rotation
When a token rotation request is made, the request contains the management id for the token as well as the value of the token itself in the headers.

However, the handler only retrieves the token via the management id. It should also verify that the token value provided also matches the token that was retrieved using the management id.

https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/auth/src/accessToken/routes.ts#L82

njlie:

- Checks the token value during revocation and rotation.
- Adds some in-line issues in the comments.

## Context

Closes #832.

When rotating and revoking a token, the provided management id in the URL was used to retrieve the token, but the token value that's provided during these requests wasn't being verified against the token that was retrieved.

Also, some TODOs did not have an associated Github issue attached to them. These have been created and their URIs have been added to the original associted in-line comments.

## Checklist

- [x] Tests added/updated
- [x] Documentation added

### DIFF
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -91,8 +91,9 @@ async function revokeToken(
   deps: ServiceDependencies,
   ctx: RevokeContext
 ): Promise<void> {
+  const token = (ctx.headers['authorization'] ?? '').replace('GNAP ', '')
   const { id: managementId } = ctx.params
-  await deps.accessTokenService.revoke(managementId)
+  await deps.accessTokenService.revoke(managementId, token)
   ctx.status = 204
 }
 
@@ -100,7 +101,6 @@ async function rotateToken(
   deps: ServiceDependencies,
   ctx: RotateContext
 ): Promise<void> {
-  // TODO: verify Authorization: GNAP ${accessToken} contains correct token value
   const { id: managementId } = ctx.params
   const token = (ctx.headers['authorization'] ?? '').replace('GNAP ', '')
   const result = await deps.accessTokenService.rotate(managementId, token)
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -102,7 +102,8 @@ async function rotateToken(
 ): Promise<void> {
   // TODO: verify Authorization: GNAP ${accessToken} contains correct token value
   const { id: managementId } = ctx.params
-  const result = await deps.accessTokenService.rotate(managementId)
+  const token = ctx.headers['authorization'].replace('GNAP ', '')
+  const result = await deps.accessTokenService.rotate(managementId, token)
   if (result.success == true) {
     ctx.status = 200
     ctx.body = {
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -102,7 +102,7 @@ async function rotateToken(
 ): Promise<void> {
   // TODO: verify Authorization: GNAP ${accessToken} contains correct token value
   const { id: managementId } = ctx.params
-  const token = ctx.headers['authorization'].replace('GNAP ', '')
+  const token = (ctx.headers['authorization'] ?? '').replace('GNAP ', '')
   const result = await deps.accessTokenService.rotate(managementId, token)
   if (result.success == true) {
     ctx.status = 200
--- a/packages/auth/src/accessToken/service.ts
+++ b/packages/auth/src/accessToken/service.ts
@@ -143,9 +143,18 @@ async function revoke(
   id: string,
   tokenValue: string
 ): Promise<void> {
-  const token = await AccessToken.query(deps.knex).findOne({ managementId: id })
-  if (token && token.value === tokenValue) {
-    await token.$query(deps.knex).delete()
+  const trx = await AccessToken.startTransaction()
+  try {
+    await AccessToken.query(trx)
+      .findOne({
+        managementId: id,
+        value: tokenValue
+      })
+      .delete()
+
+    await trx.commit()
+  } catch {
+    await trx.rollback()
   }
 }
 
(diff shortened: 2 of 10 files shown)


packages/auth/src/accessToken/routes.ts
```diff
@@ -91,18 +91,19 @@ async function revokeToken(
   deps: ServiceDependencies,
   ctx: RevokeContext
 ): Promise<void> {
+  const token = (ctx.headers['authorization'] ?? '').replace('GNAP ', '')
   const { id: managementId } = ctx.params
-  await deps.accessTokenService.revoke(managementId)
+  await deps.accessTokenService.revoke(managementId, token)
```
wilsonianb, line 96: Would it be possible to check access tokens in a separate middleware (maybe the signature middleware)?
The GNAP spec talks about binding client keys to access tokens (both **access token** and continue access tokens). Could `client` and `clientKeyId` ([on its way out](https://github.com/interledger/rafiki/issues/737)) and maybe even continue details be moved off of grants, which would avoid doing grant service lookups in signature middleware
https://github.com/interledger/rafiki/blob/83a6a546f9d4f2e830db261973e4cffeeaedd828/packages/auth/src/signature/middleware.ts#L151
https://github.com/interledger/rafiki/blob/83a6a546f9d4f2e830db261973e4cffeeaedd828/packages/auth/src/signature/middleware.ts#L84-L89

njlie: Moving `client` & `clientKeyId` into a new table would make sense since they may be bound to multiple grants anyways. In the case of continue details it can be done but I think it's a little trickier to justify because it'd punt the call to a different, more narrow service instead since they're one-to-one with grants.

What service calls are "in-scope" for the signature (or some other) middleware? As I understand it it seems like the idea is to just lookup via the client service if the token/grant continuation is associated with the client via its bound key.

wilsonianb: Maybe there can be path specific auth middlewares (ahead of the signature middleware) that store `client` on `ctx` like the `backend` currently does (via the grant)
https://github.com/interledger/rafiki/blob/ec15723671d951d6ea0657176b4c8d664be97eca/packages/backend/src/open_payments/auth/middleware.ts#L63
Some of these could store other route specific data like the access token or the grant on the `ctx`. The backend does this with payment pointer routes
https://github.com/interledger/rafiki/blob/ec15723671d951d6ea0657176b4c8d664be97eca/packages/backend/src/open_payments/payment_pointer/middleware.ts#L20
https://github.com/interledger/rafiki/blob/ec15723671d951d6ea0657176b4c8d664be97eca/packages/backend/src/open_payments/payment_pointer/routes.ts#L26-L37
Then there can be a single signature middleware.

njlie: Going to capture this for now in https://github.com/interledger/rafiki/issues/882, in case I don't get to it before the holidays.

packages/auth/src/accessToken/service.ts
```diff
@@ -159,10 +164,11 @@ async function createAccessToken(
...
   deps: ServiceDependencies,
-  managementId: string
+  managementId: string,
+  tokenValue: string
 ): Promise<Rotation> {
   let token = await AccessToken.query(deps.knex).findOne({ managementId })
-  if (token) {
+  if (token && token.value === tokenValue) {
     await token.$query(deps.knex).delete()
     token = await AccessToken.query(deps.knex).insertAndFetch({
```
wilsonianb, line 173: I think the `findOne` and `delete` should otherwise use a lock.
```
  const oldToken = await AccessToken.query(deps.knex)
    .delete()
    .returning('*')
    .findOne({
      managementId,
      value: tokenValue
    })
  if (oldToken) {
    const token = await AccessToken.query(deps.knex).insertAndFetch({
```
The order of the `delete`/`returning`/`findOne` is a workaround for:
- https://github.com/Vincit/objection.js/issues/1649#issuecomment-575698815

packages/auth/src/accessToken/service.ts
```diff
@@ -137,9 +138,13 @@ async function introspect(
...
+async function revoke(
+  deps: ServiceDependencies,
+  id: string,
+  tokenValue: string
+): Promise<void> {
   const token = await AccessToken.query(deps.knex).findOne({ managementId: id })
-  if (token) {
+  if (token && token.value === tokenValue) {
     await token.$query(deps.knex).delete()
   }
```
wilsonianb, line 149:
```
  await AccessToken.query(deps.knex)
    .findOne({
      managementId: id,
      value: tokenValue
    })
    .delete()
```

packages/auth/src/accessToken/service.ts
```diff
@@ -159,10 +164,11 @@ async function createAccessToken(
 
 async function rotate(
   deps: ServiceDependencies,
-  managementId: string
+  managementId: string,
+  tokenValue: string
 ): Promise<Rotation> {
```
wilsonianb, line 169: What about
```suggestion
): Promise<Token | undefined> {
```
instead of the `success`, and let the calling route format the response body

packages/auth/src/accessToken/service.ts
wilsonianb: :thinking: 
Maybe this should all happen inside a transaction though.
Both the delete and insert.

packages/auth/src/accessToken/service.ts
```diff
@@ -159,31 +173,49 @@ async function createAccessToken(
...
-      grantId: token.grantId
-    })
-    return {
-      success: true,
-      access,
-      value: token.value,
-      managementId: token.managementId,
-      expiresIn: token.expiresIn
+  const trx = await AccessToken.startTransaction()
+  try {
```
wilsonianb, line 180: I prefer :point_down: so we don't have to manage `trx.commit`/`rollback`
```suggestion
  try {
    return await AccessToken.transaction(async trx => {
```

packages/auth/src/accessToken/service.ts
```diff
@@ -137,10 +138,23 @@ async function introspect(
...
+  tokenValue: string
+): Promise<void> {
+  const trx = await AccessToken.startTransaction()
+  try {
+    await AccessToken.query(trx)
+      .findOne({
+        managementId: id,
+        value: tokenValue
+      })
+      .delete()
```
wilsonianb, line 153: I think this is a single query and doesn't need a transaction

wilsonianb (approved):

PR #877, opened 2022-12-19
feat(auth): delete grant (request)

### LINKED ISSUE #849: Support Open Payment grant cancellation
While working on #812, I noticed we currently do not support grant deletion: https://docs.openpayments.guide/reference/delete-continue

Tasks:
- add a handler for the `DELETE https://openpayments.guide/auth/continue/{id}` route in the `auth` package

sabineschaller:

- add `DELETE /continue/{id}` to auth server

## Context

- fixes #849 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -231,6 +231,19 @@ export class App {
       grantRoutes.continue
     )
 
+    // Grant Cancel
+    this.publicRouter.delete(
+      '/continue/:id',
+      createValidatorMiddleware<ContinueContext>(openApi.authServerSpec, {
+        path: '/continue/{id}',
+        method: HttpMethod.POST
+      }),
+      this.config.bypassSignatureValidation
+        ? (ctx, next) => next()
+        : grantContinueHttpsigMiddleware,
+      grantRoutes.delete
+    )
+
     // Token Rotation
     this.publicRouter.post(
       '/token/:id',
--- a/packages/auth/src/grant/routes.ts
+++ b/packages/auth/src/grant/routes.ts
@@ -48,13 +48,12 @@ interface GrantContinueBody {
   interact_ref: string
 }
 
-interface GrantContinueParams {
+interface GrantParams {
   id: string
 }
-export type ContinueContext = GrantContext<
-  GrantContinueBody,
-  GrantContinueParams
->
+export type ContinueContext = GrantContext<GrantContinueBody, GrantParams>
+
+export type DeleteContext = GrantContext<null, GrantParams>
 
 type InteractionRequest<
   BodyT = never,
@@ -104,6 +103,7 @@ export interface GrantRoutes {
     details(ctx: GetContext): Promise<void>
   }
   continue(ctx: ContinueContext): Promise<void>
+  delete(ctx: DeleteContext): Promise<void>
 }
 
 export function createGrantRoutes({
@@ -134,7 +134,8 @@ export function createGrantRoutes({
       acceptOrReject: (ctx: ChooseContext) => handleGrantChoice(deps, ctx),
       details: (ctx: GetContext) => getGrantDetails(deps, ctx)
     },
-    continue: (ctx: ContinueContext) => continueGrant(deps, ctx)
+    continue: (ctx: ContinueContext) => continueGrant(deps, ctx),
+    delete: (ctx: DeleteContext) => deleteGrant(deps, ctx)
   }
 }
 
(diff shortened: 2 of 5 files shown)


packages/auth/src/grant/service.ts
```diff
@@ -22,6 +22,7 @@ export interface GrantService {
     interactRef: string
   ): Promise<Grant | null>
   rejectGrant(grantId: string): Promise<Grant | null>
+  deleteGrant(continueId: string): Promise<number>
```
mkurapov, line 25:
```suggestion
  deleteGrant(continueId: string): Promise<boolean>
```
Since `number` is just an ORM thing, we should have the service throw or return a boolean depending on the result of the delete

packages/auth/src/app.ts
```diff
@@ -231,6 +231,19 @@ export class App {
...
+    // Grant Cancel
+    this.publicRouter.delete(
+      '/continue/:id',
+      createValidatorMiddleware<ContinueContext>(openApi.authServerSpec, {
+        path: '/continue/{id}',
+        method: HttpMethod.POST
+      }),
+      this.config.bypassSignatureValidation
+        ? (ctx, next) => next()
+        : grantContinueHttpsigMiddleware,
```
wilsonianb, line 243:
```suggestion
      grantContinueHttpsigMiddleware,
```
if :point_down: gets merged first
- https://github.com/interledger/rafiki/pull/879

packages/auth/src/grant/routes.ts
```diff
@@ -413,6 +414,17 @@ async function continueGrant(
...
+async function deleteGrant(
+  deps: ServiceDependencies,
+  ctx: ContinueContext
+): Promise<void> {
+  const { id: continueId } = ctx.params
+  const deletion = await deps.grantService.deleteGrant(continueId)
+  if (deletion === 0) {
+    ctx.throw(404, { error: 'unknown_continue_id' })
+  }
+  ctx.status = 202
```
wilsonianb, line 426:
```suggestion
  ctx.status = 204
```
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-5.4

Looks like oiur spec needs to be updated:
https://github.com/interledger/open-payments/blob/f2f84371778b61ec9481733b06d8d20a26ee9184/openapi/auth-server.yaml#L225

packages/auth/src/grant/routes.ts
```diff
@@ -413,6 +414,17 @@ async function continueGrant(
     accessToken
   })
 }
+async function deleteGrant(
```
njlie, line 417: This should also check that the `Authorization` header contains the correct continue access token. [It's in the form of](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-5.4) `GNAP $accessToken`, and `$accessToken` should be compared against a grant's `continueToken` field.

For future reference when https://github.com/interledger/rafiki/issues/882 gets worked on, grant deletion should also fall under the scope of this issue

packages/auth/src/grant/service.ts
sabineschaller: I had that at first but then decided to keep the number such that I can distinguish between the different scenarios on the route as well. Or would you do `true` for successfully deleted and `false` for not found?

packages/auth/src/grant/routes.ts
sabineschaller: Shouldn't this be done by a middleware?

packages/auth/src/grant/routes.ts
sabineschaller: * https://github.com/interledger/open-payments/pull/238

packages/auth/src/grant/routes.ts
wilsonianb: Yeah :point_down: would move the token checks to a middleware
- #882

packages/auth/src/grant/service.ts
wilsonianb: >Or would you do true for successfully deleted and false for not found?

:+1:

packages/auth/src/grant/routes.ts
```diff
@@ -412,6 +413,28 @@ async function continueGrant(
...
+  )[1]
+  if (!continueId || !continueToken) {
+    ctx.throw(401, { error: 'invalid_request' })
+  }
+  const grant = await deps.grantService.getByContinue(continueId, continueToken)
+  if (!grant) {
+    ctx.throw(404, { error: 'unknown_request' })
+  }
+
+  const deletion = await deps.grantService.deleteGrant(continueId)
```
wilsonianb, line 432: I'm not crazy about the separate lookup and delete, but it looks like that's also how we currently handle grant choices above.

wilsonianb (approved):

packages/auth/src/grant/routes.ts
sabineschaller: I think this will change when we have the middleware that checks that the token belongs to the grant. 
* #882

Issue #834, opened 2022-12-09
Enforce create quote args with types


wilsonianb:
Update
https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/backend/src/open_payments/quote/service.ts#L64-L70

and remove
https://github.com/interledger/rafiki/blob/33f146f0c793bc8915e50c2b3f785d2ae6020742/packages/backend/src/open_payments/quote/service.ts#L76-L78

- depends on https://github.com/interledger/open-payments/issues/224

Issue #810, opened 2022-12-05
(open-payments): create quote


mkurapov:
- Add functionality to create quote in the open-payments client: https://docs.openpayments.guide/reference/create-quote

mkurapov:
@wilsonianb
I was thinking it would make sense to do validation on the request body, before calling the API, checking that `receiveAmount` xor `sendAmount` are defined. 

However, does it make sense more sense to enforce that on the schema level: https://github.com/interledger/open-payments/issues/224

PR #888, opened 2022-12-20
feat(open-payments): add create quote

raducristianpopa:

- add create quote functionality to OP Client
- update types for the quote creation args in `backend`

## Context

- fixes https://github.com/interledger/rafiki/issues/810
- fixes https://github.com/interledger/rafiki/issues/834
## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -61,36 +61,45 @@ async function getQuote(
...
+
+interface QuoteOptionsWithReceiveAmount extends QuoteOptionsBase {
+  receiveAmount: Amount
+}
+
+export type CreateQuoteOptions =
+  | QuoteOptionsBase
+  | QuoteOptionsWithSendAmount
+  | QuoteOptionsWithReceiveAmount
+
```
raducristianpopa, line 82: Should this be done in another way ?

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -61,36 +61,45 @@ async function getQuote(
...
+}
+
+interface QuoteOptionsWithReceiveAmount extends QuoteOptionsBase {
+  receiveAmount: Amount
+}
+
+export type CreateQuoteOptions =
+  | QuoteOptionsBase
+  | QuoteOptionsWithSendAmount
+  | QuoteOptionsWithReceiveAmount
```
wilsonianb, line 81:
```suggestion
interface QuoteOptionsWithSendAmount extends QuoteOptionsBase {
  receiveAmount?: never
  sendAmount?: Amount
}

interface QuoteOptionsWithReceiveAmount extends QuoteOptionsBase {
  receiveAmount?: Amount
  sendAmount?: never
}

export type CreateQuoteOptions =
  | QuoteOptionsWithSendAmount
  | QuoteOptionsWithReceiveAmount
```

packages/backend/src/open_payments/quote/service.ts
wilsonianb: see suggested edit :point_up_2:

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -61,36 +61,45 @@ async function getQuote(
...
-    return QuoteError.InvalidAmount
-  }
   const paymentPointer = await deps.paymentPointerService.get(
     options.paymentPointerId
   )
   if (!paymentPointer) {
     return QuoteError.UnknownPaymentPointer
   }
-  if (options.sendAmount) {
+  if ('sendAmount' in options) {
```
wilsonianb, line 93: I don't think these are necessary with the suggested type edit
```suggestion
  if (options.sendAmount) {
```

packages/backend/src/graphql/resolvers/quote.ts
```diff
@@ -32,8 +33,15 @@ export const getQuote: QueryResolvers<ApolloContext>['quote'] = async (
 export const createQuote: MutationResolvers<ApolloContext>['createQuote'] =
   async (parent, args, ctx): Promise<ResolversTypes['QuoteResponse']> => {
     const quoteService = await ctx.container.use('quoteService')
+    const options: CreateQuoteOptions = {
```
wilsonianb, line 36: Because graphql doesn't do mutually exclusive types (:disappointed:), I think we need to move the old check from the quote service here
https://github.com/interledger/rafiki/blob/4db6f8e8ee6e4812af452af379855e4c3ff8d2f6/packages/backend/src/open_payments/quote/service.ts#L76-L78
```suggestion
    if (args.input.sendAmount && args.input.receiveAmount) {
      return {
        code: errorToCode[QuoteError.InvalidAmount].toString(),
        success: false,
        message: errorToMessage[QuoteError.InvalidAmount]
      }
    }
    const options: CreateQuoteOptions = {
```

packages/backend/src/open_payments/quote/routes.ts
```diff
@@ -54,14 +54,16 @@ async function createQuote(
...
+  if (body.receiveAmount)
+    options.receiveAmount = parseAmount(body.receiveAmount)
   try {
-    const quoteOrErr = await deps.quoteService.create({
-      paymentPointerId: ctx.paymentPointer.id,
-      receiver: body.receiver,
-      sendAmount: body.sendAmount && parseAmount(body.sendAmount),
-      receiveAmount: body.receiveAmount && parseAmount(body.receiveAmount),
-      clientId: ctx.grant?.clientId
-    })
```
wilsonianb, line 64: I _think_ you should be able to make `sendAmount` and `receiveAmount` mutually exclusive in `CreateBody` (like with `CreateQuoteOptions`) and leave this as it was before.
This is enforced by the OpenAPI request validator thanks to:
- https://github.com/interledger/open-payments/pull/230

packages/backend/src/tests/outgoingPayment.ts
```diff
@@ -15,7 +15,15 @@ export async function createOutgoingPayment(
     'quoteId'
   >
```
wilsonianb, line 16: If you change `options` to :point_down: you won't need the changes below
```
  options: CreateTestQuoteOptions &
    Omit<CreateOutgoingPaymentOptions, 'quoteId'>
```
but then it looks like there's some calls to `createOutgoingPayment` that would need to updated.

packages/backend/src/graphql/resolvers/quote.ts
raducristianpopa: Initially I was looking at the `@oneOf` directive ([RFC](https://github.com/graphql/graphql-spec/pull/825)). Even though it's available in a couple of packages, is still a feature from the future and the behavior could change slightly. It would be nice if we can enforce the types at the GraphQL schema later on.

While playing around with it, I got to this point:
```ts
// schema.graphql

input QuoteAmount @oneOf {
  sendAmount: AmountInput
  receiveAmount: AmountInput
}

input CreateQuoteInput {
  paymentPointerId: String!
  amount: QuoteAmount
  receiver: String!
}
```
```ts
// graphql.generated.ts

export type QuoteAmount =
  { sendAmount: AmountInput; receiveAmount?: never; }
  |  { sendAmount?: never; receiveAmount: AmountInput; };

export type CreateQuoteInput = {
  paymentPointerId: Scalars['String'];
  amount?: InputMaybe<QuoteAmount>;
  receiver: Scalars['String'];
};
```

I _think_ it partially fulfills our needs for the `CreateQuoteInput`. 🤔

packages/backend/src/open_payments/quote/routes.ts
raducristianpopa: With the changes to the `CreateBody` type, it's still complaining:
`Types of property 'sendAmount' are incompatible. Type 'Amount' is not assignable to type 'never'`

But I think we should keep the changes to `CreateBody`.

packages/backend/src/tests/outgoingPayment.ts
raducristianpopa: 🤔 I _think_ we should keep the types as it was before, since our changes are affecting only the `createQuote` function ?

raducristianpopa: What do you think if we refactor `createOutgoingPayment` later on, to something like:
```ts
function createOutgoingPayment(
  deps: IocContract<AppServices>,
  quoteOptions: CreateTestQuoteOptions,
  outgoingPaymentOptions: Omit<CreateOutgoingPaymentOptions, 'quoteId'>
): Promise<OutgoingPayment>
```

packages/open-payments/src/types.ts
```diff
@@ -38,6 +38,8 @@ export type PaymentPointer = RSComponents['schemas']['payment-pointer']
 export type JWK = RSComponents['schemas']['json-web-key']
 export type JWKS = RSComponents['schemas']['json-web-key-set']
 export type Quote = RSComponents['schemas']['quote']
+export type CreateQuoteArgs =
+  RSOperations['create-quote']['requestBody']['content']['application/json']
```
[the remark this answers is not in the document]
mkurapov: We could even use the `XOR` helper type, if you'd like: https://stackoverflow.com/a/53229567

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -61,21 +61,30 @@ async function getQuote(
...
+  | QuoteOptionsWithSendAmount
+  | QuoteOptionsWithReceiveAmount
+
 async function createQuote(
   deps: ServiceDependencies,
   options: CreateQuoteOptions
 ): Promise<Quote | QuoteError> {
-  if (options.sendAmount && options.receiveAmount) {
-    return QuoteError.InvalidAmount
-  }
```
mkurapov, line 78: @raducristianpopa @wilsonianb 
I'm leaning on keeping this business logic sanity check here (even though the OpenAPI validator would prevent it). If there are some issues with validator we would still be ok in the service

mkurapov (approved):
Looks good, approving with some comments

packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -646,18 +645,22 @@ describe('QuoteService', (): void => {
...
+          ).url
+        }
+        if (sendAmount) options.sendAmount = sendAmount
+        if (receiveAmount) options.receiveAmount = receiveAmount
+        await expect(quoteService.create(options)).resolves.toEqual(
+          QuoteError.InvalidAmount
+        )
+        await expect(quoteService.create(options)).resolves.toEqual(
+          QuoteError.InvalidAmount
+        )
```
wilsonianb, line 663:
```suggestion
```

packages/backend/src/open_payments/quote/service.ts
wilsonianb: An alternative could be to move the check to `routes.ts` (like we had to do with the graphql resolver), but it'd make more sense to just keep the one check here.
So I vote we drop the resolver check in favor of this, and maybe we revisit if/when we have graphql `@oneOf`.

wilsonianb: :thinking: do we only change `CreateBody` then, and revert `CreateQuoteOptions`?
I think that'd avoid having to change the tests' `createOutgoingPayment`

raducristianpopa:
🤔 i think we should keep the changes for the `CreateQuoteOptions` type for consistency.

packages/backend/src/open_payments/quote/service.ts
raducristianpopa: 🤔 i think we should keep the changes for the CreateQuoteOptions type for consistency.

wilsonianb: My only reservation about that is we'd also need to check in the resolver.

raducristianpopa: So we will need to keep
```ts
 if (args.input.sendAmount && args.input.receiveAmount) {
      return {
        code: errorToCode[QuoteError.InvalidAmount].toString(),
        success: false,
        message: errorToMessage[QuoteError.InvalidAmount]
      }
 }
```
in the resolver even though we check for `sendAmount` and `receiveAmount` in the service ?

wilsonianb: right, if we keep the changes to `CreateQuoteOptions`.

wilsonianb: I think the resolver will need the check in order to be able to call `quoteService.create` if the `options` type is:
```
export type CreateQuoteOptions =
  | QuoteOptionsWithSendAmount
  | QuoteOptionsWithReceiveAmount
```

raducristianpopa: With the actual changes to the resolver:
https://github.com/interledger/rafiki/blob/1b5c01a4a984400406a81f3498bb5ed7ccaed452/packages/backend/src/graphql/resolvers/quote.ts#L33-L46
i was able to make a GraphQL request with both `sendAmount` and `receiveAmount`.
`quoteService.create` gets called properly even with the `options` type being
```ts
export type CreateQuoteOptions =
  | QuoteOptionsWithSendAmount
  | QuoteOptionsWithReceiveAmount
  ```

image

File: packages/open-payments/src/types.ts [hunk 10]
```diff
@@ -38,6 +38,20 @@ export type PaymentPointer = RSComponents['schemas']['payment-pointer']
 export type JWK = RSComponents['schemas']['json-web-key']
 export type JWKS = RSComponents['schemas']['json-web-key-set']
 export type Quote = RSComponents['schemas']['quote']
+type QuoteArgsBase = {
+  receiver: RSOperations['create-quote']['requestBody']['content']['application/json']['receiver']
+}
+type QouteArgsWithSendAmount = QuoteArgsBase & {
```
mkurapov, line 44:
```suggestion
type QuoteArgsWithSendAmount = QuoteArgsBase & {
```

mkurapov (dismissed):
Looks good with the latest changes, just need to resolve merge conflicts and good to go

packages/backend/src/open_payments/quote/service.ts
raducristianpopa: I've just enabled strict type checking for the `backend` package, but TypeScript it's still not complaining about it. 🤔

mkurapov: I think it's not complaining because both `receiveAmount` and `sendAmount` are optional in their respective types

mkurapov (dismissed):

wilsonianb (commented):
Sorry, I gave you more conflicts to resolve...

raducristianpopa:
Resolved the conflicts.

My final thoughts:
* I tend to think that we should keep both changes to `CreateBody` and `CreateQuoteOptions` in order to be type consistent (from schema to service level)
* the only drawback at the current moment is that we can not enforce the type changes for the GraphQL schema
* when and if the `@oneOf` directive will be available, I think it will fulfill our needs for `CreateQuoteInput`

mkurapov (approved):

PR #893, opened 2022-12-20
Bump `@types/koa-bodyparser`

mkurapov:

The new version of `@types/koa-bodyparser` changes the body type from `any` to `unknown`. [(renovate PR)](https://github.com/interledger/rafiki/pull/760) There is a few ways to fix some conflicts as a result of this, but this PR tries to address those with little amount of changes as possible in `auth`.

Also, enables strict type checking in `http-signature-utils` 

If we want, we can also switch to [`koa-body`](https://github.com/koajs/koa-body), since it:
- requires no additional @types lib
- seems to be quite a bit more active
- keeps the `body?: any` type 😆 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/http-signature-utils/src/app.ts
```diff
@@ -6,24 +6,46 @@ import json from 'koa-json'
...
-// Router
-router.post('/', async (ctx): Promise<void> => {
-  const { request, keyId } = ctx.request.body
-  if (!keyId || !request.headers || !request.method || !request.url) {
+type GenerateSignatureRequestBody = {
+  request: RequestLike
+  keyId: string
+}
+
+router.post('/', async (ctx: AppContext<Headers>): Promise<void> => {
```
mkurapov, line 29: With `koa-bodyparser` using `request.body: unknown`, it's not very straightforward to type out the `request.body` as we would like (especially with a route handler _without_ additional middleware, oddly enough. This is because of some weirdness with the `router.post` types & overrides).

I ended up creating a typeguard to cast from `unknown` to `GenerateSignatureRequestBody`.

wilsonianb (commented):
I'd be open to switching to koa-body.
Have you looked if the typings would be more friendly?

packages/auth/src/app.ts
```diff
@@ -206,7 +206,7 @@ export class App {
     const openApi = await this.container.use('openApi')
     /* Back-channel GNAP Routes */
     // Grant Initiation
-    this.publicRouter.post(
+    this.publicRouter.post<DefaultState, CreateContext>(
```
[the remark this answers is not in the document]
mkurapov: We need these for when there is a `body` defined on `Context`s request, since that is the case where the `body: unknown` clashes with the given `Context`'s. e.g. `ManagementContext` just has `params`.

Going to do the `backend` types in separate PR

packages/auth/src/signature/middleware.ts
```diff
@@ -69,7 +70,8 @@ export async function grantContinueHttpsigMiddleware(
     'GNAP ',
     ''
   ) as string
-  const { interact_ref: interactRef } = ctx.request.body
+  const { interact_ref: interactRef } = ctx.request
+    .body as ContinueContext['request']['body']
```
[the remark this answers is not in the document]
mkurapov: We can, except passing in the context to the `contextToRequestLike(ctx: AppContext)` method requires some extra finessing: `Argument of type 'ContinueContext' is not assignable to parameter of type 'AppContext'.` I'll see if I can come up with a decent approach

mkurapov:
>  Have you looked if the typings would be more friendly?

They are friendly in the fact that `koa-body` keeps the status-quo of using `body?: any`

packages/auth/src/grant/routes.ts
```diff
@@ -26,15 +26,15 @@ interface ServiceDependencies extends BaseService {
   config: IAppConfig
 }
 
-type GrantRequest<BodyT = never, QueryT = ParsedUrlQuery> = Omit<
+type GrantRequest<BodyT = never, QueryT = ParsedUrlQuery> = Exclude<
```
mkurapov, line 29: `Exclude` here because `AppContext` is a union. Using `Omit` here gives back just an empty type, minus the defined `body` and `query` fields

packages/auth/src/index.ts
```diff
@@ -140,7 +140,7 @@ export function initIocContainer(
   container.singleton(
     'accessTokenRoutes',
     async (deps: IocContract<AppServices>) => {
-      return await createAccessTokenRoutes({
```
mkurapov, line 143: extraneous

packages/auth/src/app.ts
```diff
@@ -76,6 +77,7 @@ export interface AppServices {
   clientService: Promise<ClientService>
   grantService: Promise<GrantService>
   accessTokenRoutes: Promise<AccessTokenRoutes>
+  grantRoutes: Promise<GrantRoutes>
```
mkurapov, line 80: This is so that `grantRoute` resolution below gives back proper type, not just `any`

packages/auth/src/grant/routes.ts
wilsonianb: :confused: afaict `AppContext` is an intersection
```
type ParameterizedContext<StateT = Koa.DefaultState, ContextT = Koa.DefaultContext, ResponseBodyT = unknown> = Koa.ExtendableContext & {
    state: StateT;
} & ContextT & {
    body: ResponseBodyT;
    response: {
        body: ResponseBodyT;
    };
}
```

wilsonianb: Or is it that `AppContext['request']` is a union type via the double `extends`?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/koa/index.d.ts#L651
```
    interface BaseContext extends ContextDelegatedRequest, ContextDelegatedResponse {
```

packages/auth/src/app.ts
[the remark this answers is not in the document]
wilsonianb: I was trying to understand why `backend` seems fine.
It appears to be due to:
- the generic route definition in the magic `for` loop
https://github.com/interledger/rafiki/blob/e773fb7cd533a1a8495e056987a772154651130b/packages/backend/src/app.ts#L331-L341
- the use of `ts-ignore`
https://github.com/interledger/rafiki/blob/e773fb7cd533a1a8495e056987a772154651130b/packages/backend/src/app.ts#L369-L370

I got errors similar to what `<DefaultState, CreateContext>` here avoids by explicitly defining a route outside of the magic `for` loop as well as removing a `ts-ignore`

I've opened:
- https://github.com/interledger/rafiki/issues/916
- https://github.com/interledger/rafiki/issues/917

packages/http-signature-utils/src/utils/headers.ts
```diff
@@ -29,7 +29,7 @@ export const createHeaders = async ({
   privateKey,
   keyId
 }: SignOptions): Promise<Headers> => {
-  let contentHeaders: ContentHeaders
+  let contentHeaders: ContentHeaders | undefined
   if (request.body) {
     contentHeaders = createContentHeaders(request.body as string)
     request.headers = { ...request.headers, ...contentHeaders }
```
wilsonianb, line 35:
```suggestion
  const contentHeaders = request.body && createContentHeaders(request.body as string)
  if (contentHeaders) {
    request.headers = { ...request.headers, ...contentHeaders }
```

wilsonianb (approved):

packages/auth/src/grant/routes.ts
mkurapov: @wilsonianb it is an intersection indeed -> but I'm having a hard time understanding why exactly Omit isn't working. I'm trying to come up with a simpler example

mkurapov: Looks like its the `DefaultContext` with the `[key: string]: any` that's causing issues:


[Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQghgZwgYQPYDtgQB7CgXigG8BYAKCkqgCNEIAuKBYAJwEt0BzcqqVYABYQW8JI2bsuPKiwgBHAK4Rm41h25kAvuXKhIUAIJgwaTDjyFYdU1ly7w0IyYy3gAdTaDkC5qgC2AAosqGAIBFZINuZQAGTEUADaANYQIKqSnAC6jHDoIADcUJpQAPQl8cmp6erZULkFRVBsYYLQzQhKOmR60ADyfp5YACbh-Z4APE5RuAA0UABEsorKwPMAfLHE0pRLSipMalJaXQDGGMxNAHIQEEMAogBuwiCC6oxjwMPhpBQy8nvAegAciB5G0ZHs+g+ww8Xh8wH8QRCCAASv8VqMBsBJsZpu5PAJvL5AsFQnNFujmOtNj9eLsVtUjuDyGd0Bc2AApeHXW5o5b7aG3WGE+GI0moykWLa-HaS4Gg44QshlKA9KB3bAnAA2CiGQoJRIRJORfIB4Q12t1EBxzjMuGFhrFyPJ9KpGziPxVdLlBwy5BVzOV5VZ7IMWoQqB59yeLBeAje6s1Or1QwdouNoVNGMInvK3v5gJB-vKmiAA)

Screen Shot 2022-12-22 at 1 27 33 AM

wilsonianb (approved):

PR #910, opened 2022-12-20
refactor(auth): factor out nonce generation

### LINKED ISSUE #887: Factor out nonce generation in the AS
https://github.com/interledger/rafiki/blob/e7e1066f61615bb5b627ddc5b14ddbcc154f645c/packages/auth/src/grant/service.ts#L143
The auth server generates a nonce in several places, in the same way. It should be factored out into a separate function.

njlie:

- Refactors nonce generation into a single utilty function.

## Context

Fixes #887.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/shared/utils.ts
+++ b/packages/auth/src/shared/utils.ts
@@ -1,3 +1,4 @@
+import * as crypto from 'crypto'
 import { Access } from '../access/model'
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
@@ -12,4 +13,8 @@ export function accessToBody(access: Access) {
         k != 'updatedAt'
     )
   )
+}
+
+export function generateNonceOrToken(): string {
+  return crypto.randomBytes(8).toString('hex').toUpperCase()
 }
--- a/packages/auth/src/shared/utils.ts
+++ b/packages/auth/src/shared/utils.ts
@@ -15,6 +15,10 @@ export function accessToBody(access: Access) {
   )
 }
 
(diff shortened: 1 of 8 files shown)


packages/auth/src/shared/utils.ts
```diff
@@ -13,3 +14,7 @@ export function accessToBody(access: Access) {
     )
   )
 }
+
+export function generateNonceOrToken(): string {
+  return crypto.randomBytes(8).toString('hex').toUpperCase()
```
wilsonianb, line 19: While the GNAP spec only requires token value adhere to :point_down:, the example access tokens are larger than example nonce or continue tokens, and none of them appear to be hex-encoded
https://www.rfc-editor.org/rfc/rfc9110#section-11.2

wilsonianb (approved):

packages/auth/src/shared/utils.ts
njlie: I use the hex encoding to convert into a series of alphanumeric characters, but the length is definitely longer so I'll make a version for tokens.

wilsonianb: GNAP Java appears to use:
https://github.com/bspk/oauth.xyz-java/blob/283c52f49d916e3b77a9d1744fd28d82f8c45af4/lib/src/main/java/io/bspk/oauth/xyz/data/AccessToken.java#L44-L45

wilsonianb (approved):

Issue #844, opened 2022-12-13
Add token introspection package


wilsonianb:
- https://github.com/interledger/rafiki/pull/797#discussion_r1046284040

In order for both `auth` and `backend` to use generated OpenAPI types from :point_down:, generate the types in a new `token-introspection` package.
https://github.com/interledger/rafiki/blob/d73f7d65b245815df46cc4cfd06409c66ead7081/packages/auth/src/openapi/token-introspection.yaml

Like the existing [`open-payments`](https://github.com/interledger/rafiki/tree/main/packages/open-payments) package, this could include a client for sending the token introspection request and validating the response. This client could replace `backend`'s existing [`AuthService`](https://github.com/interledger/rafiki/blob/d73f7d65b245815df46cc4cfd06409c66ead7081/packages/backend/src/open_payments/auth/service.ts).

PR #911, opened 2022-12-20
chore(backend): do not store entire grant on context

### LINKED ISSUE #688: Do not store entire grant on context
- https://github.com/interledger/rafiki/pull/647#discussion_r994615515

wilsonianb:

- Make access limits specific to outgoing payment creation.

## Context

- fixes #688 
- setup for #844 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -37,26 +38,45 @@ export function createTokenIntrospectionMiddleware({
...
-      })
+      // TODO
+      // https://github.com/interledger/rafiki/issues/835
+      const access = tokenInfo.access.find(
+        (access) =>
+          access.type === type &&
+          (!access['identifier'] ||
+            access['identifier'] === ctx.paymentPointer.url) &&
+          access.actions.find((tokenAction) => {
+            if (tokenAction == action) {
```
mkurapov, line 53:
```suggestion
            if (tokenAction === action) {
```
I don't think a triple-equals will cause issues here, it'll be just a string comparison at runtime

same below

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -37,26 +38,45 @@ export function createTokenIntrospectionMiddleware({
...
-        action,
-        identifier: ctx.paymentPointer.url
-      })
+      // TODO
+      // https://github.com/interledger/rafiki/issues/835
+      const access = tokenInfo.access.find(
+        (access) =>
+          access.type === type &&
+          (!access['identifier'] ||
+            access['identifier'] === ctx.paymentPointer.url) &&
```
mkurapov, line 51:
```suggestion
          (access.identifier === ctx.paymentPointer.url) &&
```
if its optional

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -37,26 +38,45 @@ export function createTokenIntrospectionMiddleware({
...
+          access.actions.find((tokenAction) => {
+            if (tokenAction == action) {
+              // Unless the relevant grant action is ReadAll/ListAll add the
+              // clientId to ctx for Read/List filtering
```
mkurapov, line 55: I'm not sure if I understand the comment fully, but it reads as though

```ts
return (
    (action === AccessAction.Read &&
      tokenAction == AccessAction.ReadAll) ||
     (action === AccessAction.List &&
       tokenAction == AccessAction.ListAll)
)
```

should be before 
```
if (tokenAction == action) { ... }
```

mkurapov (commented):
Going to come back to this tomorrow (getting late here), just a few minor comments for now

packages/backend/src/open_payments/auth/middleware.ts
wilsonianb: Oops, I originally had this based on a branch using generated OpenAPI types where it _was_ an issue.
I tripled the access type check but missed the actions.

packages/backend/src/open_payments/auth/middleware.ts
wilsonianb: `action` will never be `ReadAll` or `ListAll`
https://github.com/interledger/rafiki/blob/4db6f8e8ee6e4812af452af379855e4c3ff8d2f6/packages/backend/src/app.ts#L279-L296
so this should be fine, but I'll update the types to make that explicit.

wilsonianb: :thinking: Actually this is still dependent on `Read` and `ReadAll` (and `List`/`ListAll`) to be mutually exclusive, which is no longer the case (in OpenAPI)
- https://github.com/interledger/open-payments/pull/215

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -92,7 +98,7 @@ async function createOutgoingPayment(
   deps: ServiceDependencies,
   options: CreateOutgoingPaymentOptions
 ): Promise<OutgoingPayment | OutgoingPaymentError> {
-  const grantId = options.grant ? options.grant.grant : undefined
+  const grantId = options.grant?.id
```
mkurapov, line 101: Will this be required in the future? i.e. will this end up throwing if there is no grant on the ctx?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -32,6 +32,11 @@ import {
 import { Interval } from 'luxon'
 import { knex } from 'knex'
 
+export interface Grant {
```
mkurapov, line 35: Maybe this interface first better in the `auth/middleware.ts` file?

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -105,83 +108,214 @@ describe('Auth Middleware', (): void => {
...
-      const introspectSpy = jest
-        .spyOn(authService, 'introspect')
-        .mockResolvedValueOnce(tokenInfo)
 
-      await expect(middleware(ctx, next)).resolves.toBeUndefined()
-      expect(introspectSpy).toHaveBeenCalledWith(token)
-      expect(next).toHaveBeenCalled()
-      expect(ctx.grant).toEqual(tokenInfo)
+      if (identifierOption !== IdentifierOption.Conflicting) {
+        test('sets the context client info and calls next', async (): Promise<void> => {
```
mkurapov, line 176: More for my own understanding, why are we ok to allow `tokenInfo` to _not_ have an identifier?

mkurapov (dismissed):
LGTM, non-blocking comments

packages/backend/src/open_payments/payment/outgoing/service.ts
wilsonianb: I think `grant` is mainly optional for creating outgoing payments via the admin api
https://github.com/interledger/rafiki/blob/e773fb7cd533a1a8495e056987a772154651130b/packages/backend/src/graphql/schema.graphql#L97-L99

packages/backend/src/open_payments/auth/middleware.test.ts
wilsonianb: - https://github.com/interledger/open-payments/issues/124

but then I'm not sure why it was subsequently re-required for outgoing payments :grimacing: 
- https://github.com/interledger/open-payments/pull/144

mkurapov: >  identifier (string):
     A string identifier indicating a specific resource at the RS. For example, a patient identifier for a medical API or      a bank account number for a financial API.
     
Based on the description, this feels like it should be reference to specific sub resource, and _not_ just a payment pointer (of how we are currently treating it). I'm ok with keeping it optional across the board, thoughts?

mkurapov (dismissed):

packages/backend/src/open_payments/auth/middleware.test.ts
wilsonianb: :thinking: Our own `description`s also seem to suggests it would refer to a specific (sub)resource...
https://github.com/interledger/open-payments/blob/main/openapi/schemas.yaml#L56-L59
>description: A string identifier indicating a specific resource at the RS.

mkurapov (approved):

packages/backend/src/open_payments/auth/middleware.test.ts
wilsonianb: I think we do want to be able to refer to a payment pointer, especially for `create` and `list` action grants.
If/when we switch to Adrian's grant requests to the payment pointer itself, the client will no longer need to specify the payment pointer. And he's proposed just having a `paymentPointer` field in the token introspection response for the grant.

PR #915, opened 2022-12-21
feat(backend): adds a name field for the peer model

### LINKED ISSUE #827: Add a name field for the Peer model in the GraphQL schema
Currently the GraphQL schema model for a Peer is:
```
type Peer implements Model {
  id: ID!
  maxPacketAmount: UInt64
  http: Http!
  asset: Asset!
  staticIlpAddress: String!
  createdAt: String!
}
```
An optional `name` field needs to be added so that there is a field which is easy to use and remember when looking at, or searching for specific peers.

dragosp1011:

- adds a name field for the peer model

## Context

fixes #827 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/peer/model.ts
+++ b/packages/backend/src/peer/model.ts
@@ -46,6 +46,8 @@ export class Peer
 
   public staticIlpAddress!: string
 
+  public name?: string
+
   $formatDatabaseJson(json: Pojo): Pojo {
     if (json.http?.outgoing) {
       json.outgoingToken = json.http.outgoing.authToken
--- a/packages/backend/src/peer/service.test.ts
+++ b/packages/backend/src/peer/service.test.ts
@@ -35,7 +35,8 @@ describe('Peer Service', (): void => {
       }
     },
     maxPacketAmount: BigInt(100),
-    staticIlpAddress: 'test.' + uuid()
+    staticIlpAddress: 'test.' + uuid(),
+    name: faker.name.fullName()
   })
 
   beforeAll(async (): Promise<void> => {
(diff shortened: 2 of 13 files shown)


packages/backend/src/peer/model.ts
```diff
@@ -46,6 +46,8 @@ export class Peer
 
   public staticIlpAddress!: string
 
+  public name?: string
```
wilsonianb, line 49: The column needs to be added to `peers` table.
You can add it to the migration file:
https://github.com/interledger/rafiki/blob/4db6f8e8ee6e4812af452af379855e4c3ff8d2f6/packages/backend/migrations/20211026133430_create_peers_table.js#L2
The create peer insert should also include `name`
https://github.com/interledger/rafiki/blob/4db6f8e8ee6e4812af452af379855e4c3ff8d2f6/packages/backend/src/peer/service.ts#L103-L109

wilsonianb (commented):
Add `name` to peer service and resolver tests
https://github.com/interledger/rafiki/blob/main/packages/backend/src/peer/service.test.ts
https://github.com/interledger/rafiki/blob/main/packages/backend/src/graphql/resolvers/peer.test.ts

mkurapov (approved):
looks good!

packages/backend/src/peer/service.test.ts
```diff
@@ -66,7 +67,8 @@ describe('Peer Service', (): void => {
             endpoint: faker.internet.url()
           }
         },
-        staticIlpAddress: 'test.' + uuid()
+        staticIlpAddress: 'test.' + uuid(),
+        name: faker.name.fullName()
```
wilsonianb, line 71: You could remove `name` here and just test it in the `all settings` test below.

wilsonianb (approved):

Issue #547, opened 2022-08-25
Fix rates service to use currency pairs rather than relative prices of currencies.


cairin:
Currently the `/prices` endpoint in the `mock-account-provider` returns an object of currency prices relative to a single currency:

```
{
  "USD": 1.0,
  "EUR": 1.2
}
```

Instead, we should return currency pairs instead:

```
{
  "USD:USD": 1.0,
  "USD:EUR": 1.2,
  "EUR:USD": 0.8
}
```

This `/prices` endpoint gets called in the `backend` project `RatesService`. This service is a dependency of the `ConnectorService`, the service responsible for orchestrating the actual ILP packet transfers within rafiki.

PR #955, opened 2023-01-11
feat(map/backend): use currency pairs

raducristianpopa:

- use currency pairs instead of a currency object that is relative to a single one

## Context

- fixes https://github.com/interledger/rafiki/issues/547

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


raducristianpopa:
While trying to change the `RatesService` to use currency pairs, some of the tests are failing.
The main reason is that the Interledger `Pay` is expecting the price to be an object relative to a single currency (like it's currently returned from `/prices` route) and not be stored as pairs (ie: `USD:USD`).

```
● QuoteService › create › incomingPayment.incomingAmount › receiver.incomingAmount › creates a Quote with a clientId

    thrown: "ExternalRateUnavailable"
```
https://github.com/interledgerjs/interledgerjs/blob/fdfd4e638399e40b675f75be01eb7c3e08e9545c/packages/pay/src/index.ts#L323-L340

How should the currency pairs that are returned by the `/prices` route be stored, so that the Interledger `Pay` can be happy again ?

wilsonianb:
We could either have two endpoints, one for for all currency pairs, one for prices relative to a single currency.
Or we have something that converts the currency pairs to prices relative to a single currency.

Either way, would it be worth adding a `rates`/`prices` package a la:
- https://github.com/interledger/rafiki/issues/844

with OpenAPI definition, client to do the actual request(s), and possibly the currency pairs to `Pay` prices conversion

mkurapov:
I like the idea of the latter, having a single endpoint returning prices as such:
```json
{
  "base": "USD",
  "rates": {
     "CAD": 1.26,
      ...
  }
}
```
Later, we can add an option to be able to pass in the `base` as a query param.
Given this, the only change this would require in the PR (other than the `mock-account-provider` route), would just have to return `res.rates` when making the `/prices` API call in the `RatesService`.

@wilsonianb are you thinking moving the bulk of `RatesService` into a package? I'm hesitant, since I'm not sure we would we need to use it across multiple places just yet.

wilsonianb:
I just had the idea cause I had `token-introspection` package on my mind.
I don't have a strong opinion.
fwiw, the rates service used to be its own package:
- https://github.com/interledger/rafiki/pull/139

mkurapov:
Makes sense! I think we can just update the `/prices` endpoint in the MAP for now, and go from there. Once we have a better idea of how a real-life account provider wants to support and use multiple currencies, we can update as necessary.

Thoughts @raducristianpopa ?

raducristianpopa:
> I think we can just update the /prices endpoint in the MAP for now, and go from there. 

I _think_ that updating _only_ the `/prices` route to return an object where the base and rates are provided (like you mentioned above) or an object with currency pairs, will probably brake the local demo. 🤔

mkurapov:
@raducristianpopa 

If we make the API as suggested above, we would just have to return `res.rates` when making the `/prices` API call in the `RatesService`.

packages/backend/src/rates/service.ts
```diff
@@ -107,13 +112,26 @@ class RatesServiceImpl implements RatesService {
...
+    if (!base) {
+      const errorMessage = 'Base currency was not provided'
+      this.deps.logger.warn({ errorMessage })
+      throw new Error(errorMessage)
+    }
+
+    const data = {
+      [base]: 1.0,
+      ...(rates ? rates : {})
+    }
```
raducristianpopa, line 130: In order to make same asset transfers, the base asset needs to be added to the `Prices` object that `loadNow` is returning.

packages/backend/src/rates/service.ts
```diff
@@ -107,13 +112,26 @@ class RatesServiceImpl implements RatesService {
...
       this.deps.logger.warn({ err: err.message }, 'price request error')
       throw err
     })
-    this.pricesRequest = Promise.resolve(res.data)
+
+    const { base, rates } = res.data
+    if (!base) {
+      const errorMessage = 'Base currency was not provided'
+      this.deps.logger.warn({ errorMessage })
+      throw new Error(errorMessage)
```
mkurapov, line 124: We should make this method public, and add a small test for this error

raducristianpopa:
Should the `checkBaseAsset` be a separate function outside of the `RatesService`? The build is failing because `checkBaseAsset` is not implemented in the `RafikiServicesFactory` in the `ConnectorService`.

"Implementing" it like it is done for `prices` would be ok?
https://github.com/interledger/rafiki/blob/dc4a698b46af14e66501858da262919fb2d83387/packages/backend/src/connector/core/factories/rafiki-services.ts#L48-L53

mkurapov:
@raducristianpopa wouldn't we just able to remove `checkBaseAsset` from `RatesService` interface?

packages/backend/src/rates/service.ts
```diff
@@ -107,13 +113,38 @@ class RatesServiceImpl implements RatesService {
...
+    }
+
+    this.pricesRequest = Promise.resolve(data)
     this.pricesExpiry = new Date(Date.now() + this.deps.pricesLifetime)
-    return res.data
+    return data
+  }
+
+  checkBaseAsset(asset: unknown): void {
+    let errorMessage: string
```
mkurapov, line 135:
```suggestion
    let errorMessage: string | undefined
```
I'm getting some typescript warnings on L144

packages/backend/src/rates/service.ts
```diff
@@ -107,13 +113,38 @@ class RatesServiceImpl implements RatesService {
...
+    if (!asset) {
+      errorMessage = 'Missing base asset'
+    } else if (typeof asset !== 'string') {
+      errorMessage = 'Base asset should be a string'
+    } else if (typeof asset === 'string' && asset.length === 0) {
+      errorMessage = 'Invalid base asset'
+    }
```
mkurapov, line 142:
```suggestion
    } 
```
The first `if (!asset)` check will already handle the empty string case

packages/backend/src/rates/service.test.ts
```diff
@@ -178,4 +180,15 @@ describe('Rates service', function () {
...
     })
   })
+
+  describe('checkBaseAsset', (): void => {
+    it.each`
+      asset        | description
+      ${undefined} | ${'is not provided'}
+      ${['USD']}   | ${'is not a string'}
+      ${''}        | ${'is an empty string'}
+    `(`throws if base asset is $description`, ({ asset }): void => {
```
mkurapov, line 190:
```suggestion
    `(`throws if base asset $description`, ({ asset }): void => {
```

packages/backend/src/rates/service.ts
```diff
@@ -107,13 +113,38 @@ class RatesServiceImpl implements RatesService {
...
+      ...(rates ? rates : {})
+    }
+
+    this.pricesRequest = Promise.resolve(data)
     this.pricesExpiry = new Date(Date.now() + this.deps.pricesLifetime)
-    return res.data
+    return data
+  }
+
+  checkBaseAsset(asset: unknown): void {
```
mkurapov, line 134:
```suggestion
  checkBaseAsset(asset: unknown): asset is string {
```
Even though `PricesResponse` already gives you the proper type, this (optionally) could be a good addition too

raducristianpopa:
```suggestion
  checkBaseAsset(asset: unknown): asset is string {
```

Isn't this covered by this check ? 👇
https://github.com/interledger/rafiki/blob/dc4a698b46af14e66501858da262919fb2d83387/packages/backend/src/rates/service.ts#L138-L139
By typing the return type as `asset is string` we will have to return and an extra check will be required: 
```ts
if(!this.checkBaseAsset(base)){
  // log & throw
}
```

mkurapov (approved):

Issue #798, opened 2022-11-29
Update token introspection request body


wilsonianb:
https://github.com/interledger/rafiki/blob/6dfc5dc49e266e53210486364452a0789f10e9b2/packages/auth/src/openapi/resource-server.yaml#L80-L96

Remove `proof` and `resource_server`.
Either remove `access` or enforce it at the AS (return ~404~ `active: false` if the grant does not satisfy the specified `access`)
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#section-3.3
>   access (array of strings/objects):  OPTIONAL.  The minimum access
      rights required to fulfill the request.

wilsonianb:
I'm going to include this with:
- https://github.com/interledger/rafiki/issues/737

I'm undecided on starting to require specifying the RS request's `access` in the introspection request and returning only the single relevant access in the response's access list, which would then really only be useful for conveying the outgoing payment access limits.

wilsonianb:
With :point_up_2: would we disallow the RS from specifying `limits` in introspection request access (using the amounts of a create outgoing payment request. :thinking: which the RS won't have until it looks up the specified quote...)

wilsonianb:
Moving AS `access` matching to:
- https://github.com/interledger/rafiki/issues/835

PR #956, opened 2023-01-11
feat(types): replace access type/action enums with generated unions

dragosp1011:
## Changes proposed in this pull request

- replace access type and action enums with generated unions from open payments

## Context

fixes #808 
## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass

PR #960, opened 2023-01-12
feat(token-introspection): add token introspection package

wilsonianb:

- 

## Context

- #844 
- #798 
## Checklist

- [x] Tests added/updated
- [ ] Documentation added


sabineschaller:
I'm a bit confused. `/auth/accessToken` still has introspection methods that don't use the package, do they? What is the package for?

wilsonianb:
Like our `open-payments` package, it includes:
- OpenAPI spec and types for the token introspection request/response (for both the client (RS) and server (AS))
- client for performing requests (for the client)

It's similar to how the `backend` still has `outgoingPaymentService.create`, etc for serving requests for the `open-payments` client's `outgoingPayment.create`, etc

packages/token-introspection/src/openapi/token-introspection.yaml
```diff
@@ -0,0 +1,175 @@
...
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - properties:
+                      active:
+                        enum:
+                          - false
```
mkurapov, line 33: Is this ok to have as an `enum`?

packages/auth/src/accessToken/routes.ts
```diff
@@ -64,26 +66,25 @@ async function introspectToken(
...
-      access: result.access?.map((a: Access) => accessToBody(a)),
-      key: result.key,
-      client_id: result.clientId
-    }
+function introspectionToBody({ grant, jwk }: Introspection): ActiveTokenInfo {
+  return {
+    active: true,
+    grant: grant.id,
+    access: grant.access.map((a: Access) =>
+      accessToBody(a)
```
mkurapov, line 81: Not related to this PR, but `accessToBody()` seems a bit odd to me:
```ts
export function accessToBody(access: Access) {
  return Object.fromEntries(
    Object.entries(access.toJSON()).filter(
      ([k, v]) =>
        v != null &&
        k != 'id' &&
        k != 'grantId' &&
        k != 'createdAt' &&
        k != 'updatedAt'
    )
  )
}
```
I feel like it could be updated to just take the keys that it needs, instead of filtering most of them out?

packages/auth/src/accessToken/routes.ts
```diff
@@ -64,26 +66,25 @@ async function introspectToken(
...
-    ctx.throw(404, {
-      error: 'invalid_request',
-      message: 'token not found'
-    })
-  }
+  ctx.body = introspectionResult
+    ? introspectionToBody(introspectionResult)
+    : {
+        active: false
+      }
```
mkurapov, line 73: Maybe we can move this ternary into `introspectionToBody` as well, so that we are a bit more safe with the types? (i.e. have `introspectionToBody` return just `TokenInfo`)

packages/token-introspection/src/openapi/token-introspection.yaml
```diff
@@ -0,0 +1,175 @@
...
+                    key:
+                      proof: httpsig
+                      jwk:
+                        alg: EdDSA
+                        kid: 20f24ce2-a5f6-4f28-bf7d-ed52d0490187
+                        kty: OKP
+                        use: sig
+                        crv: Ed25519
+                        x: AAAAC3NzaC1lZDI1NTE5AAAAIK0wmN/Cr3JXqmLW7u+g9pTh+wyqDHpSQEIQczXkVx9q
+                    client_id: 581a1dd9-aef5-4b1a-9d78-c1afc4a5fdbf
```
mkurapov, line 64: Not a big deal since it's an example, but would it be more accurate if this was a hash instead of a uuid?

.github/workflows/lint_test_build.yml
```diff
@@ -90,6 +96,22 @@ jobs:
...
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/rafiki/env-setup
+      - run: pnpm --filter token-introspection build:deps
+      - run: pnpm --filter token-introspection test
+      - name: AsyncAPI extension
+        run: |
+          echo "{\"extends\":[\"spectral:oas\",\"spectral:asyncapi\"]}" >> .spectral.json
+      - name: Validate Open API specs
```
mkurapov, line 111: 👍

mkurapov (dismissed):
Looks solid, just a few comments

packages/token-introspection/src/openapi/token-introspection.yaml
wilsonianb: :+1: but `client_id` is living on borrowed time
- https://github.com/interledger/rafiki/issues/737

packages/token-introspection/src/openapi/token-introspection.yaml
wilsonianb: Do you have a specific concern about `enum`?
I think I got the boolean as enum from this:
https://stackoverflow.com/a/16826238/12196980

Also, I think the [validation](https://github.com/kogosoftwarellc/open-api) [libraries](https://github.com/openapi-library/OpenAPIValidators) we're using don't fully support our OpenAPI 3.1 (just 3.0, which doesn't support `const`).

packages/auth/src/accessToken/routes.ts
wilsonianb: - https://github.com/interledger/rafiki/issues/973

Postponing until after :point_down: is merged:
- https://github.com/interledger/rafiki/pull/956

mkurapov (dismissed):

packages/token-introspection/src/openapi/token-introspection.yaml
mkurapov: No specific concern, just wanted to double check this would validate properly if given the correct `false` boolean :+1:

mkurapov (approved):

PR #986, opened 2023-01-17
938/mk/create receiver

### LINKED ISSUE #938: Allow creating remote incoming payments via GraphQL API
Alice is a customer at Wallet1 and Bob is a customer at Wallet2. Alice wants to send a P2P payment to Bob on Wallet1's UI. 
The UI uses the GraphQL API to create payments and quotes. However, using the GraphQL API, the UI can only create incoming payments at Wallet1, not at Wallet2. Hence, we need to allow creating incoming payments on Wallet2 using the GraphQL API. Ideally, this uses the same `CreateIncomingPayment` endpoint we already have and under the hood it figures out whether the API call wants to create a local or a remote incoming payment.

mkurapov:

- Adding `CreateReceiver` mutation to GraphQL admin API, which allows creating a remote (or local) incoming payment  

## Context

- Fixes #938 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+async function create(
+  deps: ServiceDependencies,
+  args: CreateRemoteIncomingPaymentArgs
+): Promise<OpenPaymentsIncomingPayment> {
+  const { paymentPointerUrl } = args
+  const grant = await getGrant(deps, paymentPointerUrl, [
+    AccessAction.Create,
+    AccessAction.ReadAll
```
mkurapov, line 55: Do we want to have a `GetRemoteIncomingPayment` query in the Admin API in addition to this "Create" one?

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,147 @@
...
+  ): Promise<OpenPaymentsIncomingPayment>
+}
+
+interface ServiceDependencies extends BaseService {
+  grantService: GrantService
+  openPaymentsUrl: string
+  openPaymentsClient: AuthenticatedClient
+}
+
+export async function createRemoteIncomingPaymentService(
```
mkurapov, line 32: In a separate PR, I want the `ReceiverService` to use this service to fetch remote incoming payments, since it has some similar logic currently. (`ReceiverService` determines whether an incoming payment (or connection) is local and remote, and fetches either)

packages/backend/src/tests/openPaymentsMocks.ts
```diff
@@ -0,0 +1,92 @@
+import { v4 as uuid } from 'uuid'
```
mkurapov, line 1: I took this from the `open-payments` package, since I wanted to mock the results from the `open-payments`client calls. I could also export these mocks directly from the library?

wilsonianb: I'm good with exported mocks

packages/backend/src/open_payments/payment/incoming_remote/service.ts
wilsonianb: What if the mutation was instead `CreateReceiver`?
I like the idea of making it clear what the purpose of this is.
We could add a `ReceiverService.create` (which could create locally or remotely) but still move remote management to `RemoteIncomingPaymentService` if we wanted.

mkurapov: 🤔 

- Are you suggesting we also remove `CreateIncomingPayment` and consolidate both remote and incoming payment creation with `CreateReceiver`? 
- If not, is it confusing having both `CreateIncomingPayment` and `CreateReceiver`? Could someone get confused why are they different at first glance?

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  if (!paymentPointer) {
+    const errorMessage = 'Could not get payment pointer'
+    deps.logger.error({ paymentPointerUrl }, errorMessage)
+    throw new Error(errorMessage)
+  }
+
+  const grantOptions = {
+    authServer: paymentPointer.authServer,
+    accessType: 'incoming-payment' as const,
```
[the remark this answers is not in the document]
mkurapov: It does allow `AccessType.IncomingPayment`, but I still need to do `as `incoming-payment` when making the call for the grant request below

packages/backend/src/open_payments/payment/incoming_remote/service.ts
wilsonianb: I was thinking we have both
- `CreateIncomingPayment` for a user to get paid
- `CreateReceiver` to pay someone

mkurapov: It does fit in well with the quote creation, `receiver` attribute as well.

`RemoteIncomingPayment` GraphQL type can also be also a `Receiver` instead, going to make that change

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -245,7 +245,7 @@ export class IncomingPayment
     ilpStreamConnection: Connection
   }): OpenPaymentsIncomingPayment {
     return {
-      id: this.id,
+      id: this.url,
```
mkurapov, line 248: (Copying comment from other [PR](https://github.com/interledger/rafiki/pull/998) for visibility):

> OpenPayment API IncomingPayments return the full url instead of just the uuid. This reflects what is done in the IncomingPaymentRoutes, but standardizes it when converting between IncomingPayment model and what Open Payments expects

packages/backend/src/open_payments/payment/incoming_remote/service.ts
mkurapov: https://github.com/interledger/rafiki/issues/1005

CC @wilsonianb

packages/backend/src/app.ts
```diff
@@ -155,6 +157,8 @@ export interface AppServices {
   paymentPointerKeyRoutes: Promise<PaymentPointerKeyRoutes>
   paymentPointerRoutes: Promise<PaymentPointerRoutes>
   incomingPaymentService: Promise<IncomingPaymentService>
+  remoteIncomingPaymentService: Promise<RemoteIncomingPaymentService>
+  receiverService: Promise<ReceiverService>
```
mkurapov, line 161: Not new, just adding so we get proper typing when we resolve this dependency

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -45,10 +61,74 @@ export async function createReceiverService(
...
+  if (!connection) {
+    const errorMessage = 'Could not get connection for local incoming payment'
+    deps.logger.error({ incomingPaymentOrError }, errorMessage)
+
+    throw new Error(errorMessage)
+  }
+
+  return incomingPaymentOrError.toOpenPaymentsType({
+    ilpStreamConnection: connection
+  })
```
mkurapov, line 129: Once I split out the `IncomingPayment` types in `open-payments` with and without the ilpStreamConnection
- https://github.com/interledger/rafiki/issues/662

The corresponding connection fetching will (should?) be optional

mkurapov:
Will also update the Postman collection once this is merged in

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,150 @@
...
+  deps: ServiceDependencies,
+  args: CreateRemoteIncomingPaymentArgs
+): Promise<OpenPaymentsIncomingPayment> {
+  const { paymentPointerUrl } = args
+  const grant = await getGrant(deps, paymentPointerUrl, [
+    AccessAction.Create,
+    AccessAction.ReadAll
+  ])
+
+  if (!grant?.accessToken) {
```
wilsonianb, line 59:
```suggestion
  if (!grant.accessToken) {
```

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,150 @@
...
+  description?: string
+  expiresAt?: Date
+  incomingAmount?: Amount
+  externalRef?: string
+}
+
+export interface RemoteIncomingPaymentService {
+  create(
+    args: CreateRemoteIncomingPaymentArgs
+  ): Promise<OpenPaymentsIncomingPayment>
```
wilsonianb, line 24:
```suggestion
  ): Promise<OpenPaymentsIncomingPayment | IncomingPaymentError>
```
to be able to return `IncomingPaymentError.UnknownPaymentPointer` (instead of throwing) and return the corresponding code and message in the `createReceiver` resolver
https://github.com/interledger/rafiki/blob/752efd71a7ee4cfa3200a9757db7a09242ba7606/packages/backend/src/open_payments/payment/incoming/errors.ts#L1-L2

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    )
+  } catch (error) {
+    const errorMessage = 'Error creating remote incoming payment'
+    deps.logger.error({ error, paymentPointerUrl }, errorMessage)
+    throw new Error(errorMessage, { cause: error })
+  }
+}
+
+async function getGrant(
```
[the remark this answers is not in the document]
mkurapov: Yup, for sure. I was going to do that as a follow up so the PR was manageable

packages/backend/src/open_payments/payment/incoming_remote/service.ts
mkurapov: Good idea

mkurapov: Actually, do you mean
```ts
// ReceiverService
async function createLocalIncomingPayment(
  deps: ServiceDependencies,
  args: CreateReceiverArgs,
  paymentPointer: PaymentPointer
): Promise<OpenPaymentsIncomingPayment | IncomingPaymentError>
```
instead? `RemotePaymentService` can only return the `OpenPaymentsIncomingPayment` since it uses the client

mkurapov: I wonder if having a custom Error object with something like a `userMessage` property would be a better approach

wilsonianb: No, I think it needs to be in `RemotePaymentService`.
The payment pointer look up in `getGrant` should somehow bubble up an error to allow the resolver to return `4XX`.

wilsonianb: > I wonder if having a custom Error object with something like a `userMessage` property would be a better approach

That'd probably work too. I'm just thinking through the lens of the other services' `*Error`s.

mkurapov: Makes sense, I'll give it a try :+1:

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -45,8 +63,76 @@ export async function createReceiverService(
...
+    return incomingPaymentOrError
+  }
+
+  const connection = deps.connectionService.get(incomingPaymentOrError)
+
+  if (!connection) {
+    const errorMessage = 'Could not get connection for local incoming payment'
+    deps.logger.error({ incomingPaymentOrError }, errorMessage)
+
+    throw new Error(errorMessage)
```
mkurapov, line 130: Maybe I can also return an additional type of `ReceiverError` here as well, together with L93?

packages/backend/src/open_payments/receiver/errors.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const receiverErrorToCode = (error: ReceiverError): number =>
+  isIncomingPaymentError(error)
+    ? incomingPaymentErrorToCode[error]
+    : remoteIncomingPaymentErrorToCode[error]
+
+export const receiverErrorToMessage = (error: ReceiverError): string =>
+  isIncomingPaymentError(error)
+    ? incomingPaymentErrorToMessage[error]
+    : remoteIncomingPaymentErrorToMessage[error]
```
mkurapov, line 31: I could have also just spread the two `errorToMessage`/`errorToCode` objects into a new `receiver` `errorToMessage`/`errorToCode` object, but thought this was easier to follow

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -67,11 +68,11 @@ describe('Remote Incoming Payment Service', (): void => {
...
         })
 
-      await expect(
-        remoteIncomingPaymentService.create({
+      expect(
+        await remoteIncomingPaymentService.create({
           paymentPointerUrl: paymentPointer.id
         })
-      ).rejects.toThrow('Could not get payment pointer')
+      ).toBe(RemoteIncomingPaymentError.UnknownPaymentPointer)
```
wilsonianb, line 75:
```
      await expect(
        remoteIncomingPaymentService.create({
          paymentPointerUrl: paymentPointer.id
        })
      ).resolves.toEqual(RemoteIncomingPaymentError.UnknownPaymentPointer)
```

packages/backend/src/open_payments/receiver/errors.ts
```diff
@@ -0,0 +1,31 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+export const isReceiverError = (o: any): o is ReceiverError =>
+  Object.values(ReceiverError).includes(o)
+
+export const receiverErrorToCode = (error: ReceiverError): number =>
+  isIncomingPaymentError(error)
+    ? incomingPaymentErrorToCode[error]
+    : remoteIncomingPaymentErrorToCode[error]
+
+export const receiverErrorToMessage = (error: ReceiverError): string =>
```
wilsonianb, line 28:
```suggestion
export const errorToCode = (error: ReceiverError): number =>
  isIncomingPaymentError(error)
    ? incomingPaymentErrorToCode[error]
    : remoteIncomingPaymentErrorToCode[error]

export const errorToMessage = (error: ReceiverError): string =>
```

packages/backend/src/graphql/resolvers/receiver.test.ts
```diff
@@ -0,0 +1,254 @@
...
+            id: `${paymentPointer.id}/incoming-payments/${uuid()}`,
+            paymentPointer: paymentPointer.id,
+            description,
+            externalRef,
+            expiresAt: expiresAt?.toISOString(),
+            incomingAmount: incomingAmount
+              ? serializeAmount(incomingAmount)
+              : undefined
+          })
+        ) as Receiver
```
wilsonianb, line 62: Would it be worth adding a `createReceiver` (here or in `tests/receiver`) that returns a guaranteed `Receiver`?

wilsonianb (dismissed):

packages/backend/src/graphql/resolvers/receiver.test.ts
mkurapov: Maybe it would be better to have `Receiver.fromIncomingPayment/fromConnection`to throw errors instead of returning undefined?

wilsonianb: :+1: that or return `Receiver | ReceiverError` :grimacing:

mkurapov: Ended up going with just throwing, will likely be changing out how the Receiver class works anyway if we want to go with `type Receiver = IncomingPayment | Connection`

packages/backend/src/open_payments/receiver/model.test.ts
```diff
@@ -64,42 +64,61 @@ describe('Receiver Model', (): void => {
...
+    test('throws if stream connection has invalid ILP address', async () => {
+      const paymentPointer = await createPaymentPointer(deps)
+      const incomingPayment = await createIncomingPayment(deps, {
+        paymentPointerId: paymentPointer.id
+      })
 
-      expect(receiver).toBeUndefined()
+      const connection = connectionService.get(incomingPayment) as Connection
+
+      ;(connection.ilpAddress as string) = 'not base 64 encoded'
```
wilsonianb, line 113: is this you or `pnpm format`?

wilsonianb (dismissed):

packages/backend/src/open_payments/receiver/model.test.ts
mkurapov: `pnpm format`

wilsonianb (approved):

PR #998, opened 2023-01-18
938/mk/create receiver

mkurapov:

- Rename `createRemoteIncomingPayment` mutation to `createReceiver`
- Add `receiverService.create` method to be used in the resolver

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -245,7 +245,7 @@ export class IncomingPayment
     ilpStreamConnection: Connection
   }): OpenPaymentsIncomingPayment {
     return {
-      id: this.id,
+      id: this.url,
```
mkurapov, line 248: OpenPayment API IncomingPayments return the full `url` instead of just the uuid. This reflects what is done in the IncomingPaymentRoutes, but standardizes it when converting between `IncomingPayment` model and what Open Payments expects

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -39,45 +67,31 @@ export class Receiver extends ConnectionBase {
...
+      }
     )
   }
 
   private constructor(
     connection: OpenPaymentsConnectionWithIlpAddress,
-    private readonly incomingAmountValue?: bigint,
-    private readonly receivedAmountValue?: bigint,
-    public readonly expiresAt?: Date
+    public incomingPayment?: ReceiverIncomingPayment
```
mkurapov, line 94: I decided to place the full `incomingPayment` on the Receiver object. This makes sure we don't lose all of the information when creating a receiver from a payment & and as a result, allows me to keep the `CreateReceiverResponse` GraphQL type filled with the same payment information as before (`paymentPointer` `description` `expiresAt` etc)

I think ideally, I would prefer to have `type Receiver = Connection | IncomingPayment` and then just use typeguards to determine what kind of Receiver it is without needing a class. We would lose the logic in the `Receiver.fromIncomingPayment` static method, but maybe we could convert those to a few simple functions if we need to use them.

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -45,8 +57,30 @@ export async function createReceiverService(
...
+
+async function createReceiver(
+  deps: ServiceDependencies,
+  args: CreateReceiverArgs
+): Promise<Receiver> {
+  const remoteIncomingPayment = await deps.remoteIncomingPaymentService.create(
+    args
+  )
+
+  const receiver = Receiver.fromIncomingPayment(remoteIncomingPayment)
```
mkurapov, line 73: As the comment above, do we want to keep the validation in `Receiver.fromIncomingPayment` ? https://github.com/interledger/rafiki/blob/07461933be2f4e90c350d99396bbdcf09944401a/packages/backend/src/open_payments/receiver/model.ts#L23-L48

Would this be better as separate functions, and have the caller of `Receiver.fromIncomingPayment` decide if a Receiver is "legit" or not?

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -248,14 +268,12 @@ describe('Receiver Service', (): void => {
...
-              value: BigInt(5),
-              assetCode: paymentPointer.asset.code,
-              assetScale: paymentPointer.asset.scale
-            }
+          paymentPointer = mockPaymentPointer({
+            authServer
+          })
+          incomingPayment = mockIncomingPayment({
+            id: `${paymentPointer.id}/incoming-payments/${uuid()}`,
+            paymentPointer: paymentPointer.id
```
mkurapov, line 276: I thought it made more sense to mock the payment pointer and incoming payment here, since we are fetching remote incoming payments, and do not need create any locally to test this

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -106,9 +120,9 @@ export class Receiver extends ConnectionBase {
   }
 
   public get receivedAmount(): Amount | undefined {
-    if (this.receivedAmountValue !== undefined) {
+    if (this.incomingPayment?.receivedAmount) {
       return {
-        value: this.receivedAmountValue,
+        value: this.incomingPayment.receivedAmount.value,
         assetCode: this.assetCode,
         assetScale: this.assetScale
```
mkurapov, line 127: Should I add validation in this class to make sure `assetCode` and `assetScale` are the same across `receivedAmount` and `incomingAmount`?

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -45,8 +57,30 @@ export async function createReceiverService(
   }
 
   return {
-    get: (url) => getReceiver(deps, url)
+    get: (url) => getReceiver(deps, url),
+    create: (url) => createReceiver(deps, url)
+  }
+}
+
+async function createReceiver(
```
mkurapov, line 65: Should this allow creating a local incoming payment as well? When we ever want to use this over `incomingPaymentService.create`?

packages/backend/src/open_payments/payment/incoming/model.ts
mkurapov: Although I'm not sure how this plays with https://github.com/interledger/rafiki/issues/662, I'll have to take a look at how the `IncomingPaymentsJSON` type is used

wilsonianb: :+1:

packages/backend/src/open_payments/receiver/model.ts
wilsonianb: `type Receiver = Connection | IncomingPayment` also sounds appealing to me.
Maybe some of the `fromIncomingPayment` could be moved to the `isReceiverIncomingPayment` type guard.

wilsonianb (commented):
This is tricky to review at the moment, I think due to the two branches having different amounts of `main` merged in.
What do you think about going ahead and consolidating these PRs?

packages/backend/src/open_payments/receiver/service.ts
wilsonianb: The `createReceiver` mutation resolver should support creating local or remote incoming payments.
The decision/attempt to create locally would either need to happen in the resolver or here.
I vote here.

mkurapov:
@wilsonianb oh that's right, I think I merged but didn't push to the remote on the other branch

packages/backend/src/open_payments/receiver/model.ts
mkurapov: Since `receiverService` will require some changes (in order to replace some of those methods it has to call `remoteIncomingPayment` service) I could make an attempt using `type Receiver = Connection | IncomingPayment` instead.

packages/backend/src/open_payments/receiver/service.ts
mkurapov: Makes sense, I can add that in.=

mkurapov:
@wilsonianb merged into https://github.com/interledger/rafiki/pull/986, we can move the conversation there

PR #1006, opened 2023-01-20
feat(backend): consolidate tigerbeetle testcontainer code

### LINKED ISSUE #982: Consolidate tigerbeetle testcontainer management
`backend` test setup creates a tigerbeetle testcontainer
https://github.com/interledger/rafiki/blob/04698a2d19ef442c9fc481ec005ee3a675066e57/packages/backend/jest.setup.js#L72

Some tigerbeetle-heavy test files start their own tigerbeetle testcontainer
https://github.com/interledger/rafiki/blob/04698a2d19ef442c9fc481ec005ee3a675066e57/packages/backend/src/tests/tigerbeetle.ts#L9
https://github.com/interledger/rafiki/blob/04698a2d19ef442c9fc481ec005ee3a675066e57/packages/backend/src/accounting/service.test.ts#L39

Common code doesn't appear to be reused between the two.
This means changes like :point_down: would need to duplicated in both
- https://github.com/interledger/rafiki/issues/963

sabineschaller:

- fixes #982

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/jest.setup.js
+++ b/packages/backend/jest.setup.js
@@ -1,9 +1,9 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { knex } = require('knex')
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const { GenericContainer, Wait } = require('testcontainers')
+const { GenericContainer } = require('testcontainers')
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const tmp = require('tmp')
+const tigerbeetle = require('./dist/tests/tigerbeetle')
 
 const POSTGRES_PORT = 5432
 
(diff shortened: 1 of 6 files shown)


packages/backend/src/config/app.ts
```diff
@@ -18,6 +18,11 @@ function envFloat(name: string, value: number): number {
   return envValue == null ? value : +envValue
 }
 
+function envBool(name: string, value: boolean): boolean {
```
wilsonianb, line 21: It's back! :sweat_smile:

.github/workflows/lint_test_build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/workflows/rafiki/env-setup
-      - run: pnpm --filter backend build:deps
+      - run: pnpm --filter backend build
```
wilsonianb, line 28: Why this?

packages/backend/jest.setup.js
```diff
@@ -76,67 +76,12 @@ module.exports = async (globalConfig) => {
...
-      // Give TB a chance to startup (no message currently to notify allocation is complete):
-      await new Promise((f) => setTimeout(f, 1000))
-
-      const tbContStart = await new GenericContainer(
-        'ghcr.io/coilhq/tigerbeetle:dj-request-dirty-prepare@sha256:c312832a460e7374bcbd4bd4a5ae79b8762f73df6363c9c8106c76d864e21303'
+      const tbContStart = await tigerbeetle.startTigerbeetleContainer(
+        TIGERBEETLE_DIR,
+        TIGERBEETLE_PORT,
+        TIGERBEETLE_CLUSTER_ID,
+        TIGERBEETLE_CONTAINER_LOG
```
wilsonianb, line 83: Should this also use values from `Config`?

packages/backend/src/tests/tigerbeetle.ts
```diff
@@ -3,56 +3,73 @@ import tmp from 'tmp'
...
 
-const TIGERBEETLE_DIR = '/var/lib/tigerbeetle'
-export const TIGERBEETLE_PORT = 3004
-
 export async function startTigerbeetleContainer(
-  clusterId: number = Config.tigerbeetleClusterId
+  dir: string = Config.tigerbeetleDir,
+  port: number = Config.tigerbeetlePort,
+  clusterId: number = Config.tigerbeetleClusterId,
+  logsOn: boolean = Config.tigerbeetleContainerLogs
```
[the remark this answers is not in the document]
wilsonianb: Or what if just do configuration via `Config`
(Don't specify anything when calling `startTigerbeetleContainer`?

.github/workflows/lint_test_build.yml
sabineschaller: Because I need backend to be compiled in order to use `./dist/tests/tigerbeetle` in `backend/jest.setup.js`

packages/backend/src/tests/tigerbeetle.ts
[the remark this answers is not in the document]
sabineschaller: I like option 2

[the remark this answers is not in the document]
sabineschaller: I think we can only do option one if we want to be able to configure it differently in `jest.setup.js`

packages/backend/jest.setup.js
sabineschaller: You mean importing that from `dist` too?

.github/workflows/lint_test_build.yml
wilsonianb: What do you think about one of these alternatives?
https://stackoverflow.com/questions/48318230/configure-jest-global-tests-setup-with-ts-file

packages/backend/jest.setup.js
sabineschaller: Is there ever a case where we want to have different configs?

wilsonianb: As is, it seems like we'd be fine just using `Config` and not passing anything to `startTigerbeetleContainer`

wilsonianb: :thinking: or does that work if `jest.setup.ts` is defining environment variables used by `Config` (used by `startTigerbeetleContainer`)?

packages/backend/src/accounting/service.test.ts
```diff
@@ -36,9 +33,9 @@ describe('Accounting Service', (): void => {
   }
 
   beforeAll(async (): Promise<void> => {
-    tigerbeetleContainer = await startTigerbeetleContainer()
+    tigerbeetleContainer = await startTigerbeetleContainer({})
     Config.tigerbeetleReplicaAddresses = [
-      tigerbeetleContainer.getMappedPort(TIGERBEETLE_PORT)
+      tigerbeetleContainer.getMappedPort(Config.tigerbeetlePort)
```
wilsonianb, line 38: Could `startTigerbeetleContainer` also return this, so that callers don't need to use `Config.tigerbeetlePort`?
Alternatively `tests/tigerbeetle.ts` could also export a
```
getAddress(container: StartedTestContainer): string
```

packages/backend/jest.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
   clearMocks: true,
   testTimeout: 15000,
   roots: [`<rootDir>/packages/${packageName}`],
-  globalSetup: `<rootDir>/packages/${packageName}/jest.setup.js`,
+  globalSetup: `<rootDir>/packages/${packageName}/jest.setup.ts`,
```
wilsonianb, line 12: :sparkles:

packages/backend/src/config/app.ts
sabineschaller: And it's gone again 😅

packages/backend/jest.setup.js
sabineschaller: I ended up only having those in Config that index.ts uses and having all others in `src/tests/tigerbeetle.ts`, i.e. where `startTigerbeetleContainer` is defined. What do you think?

packages/backend/jest.setup.ts
```diff
@@ -0,0 +1,100 @@
...
+    global.__BACKEND_KNEX__ = db
+  }
+
+  const setupTigerbeetle = async () => {
+    if (!process.env.TIGERBEETLE_REPLICA_ADDRESSES) {
+      const { container, port } = await startTigerbeetleContainer(
+        TIGERBEETLE_CLUSTER_ID
+      )
+
+      process.env.TIGERBEETLE_CLUSTER_ID = TIGERBEETLE_CLUSTER_ID.toString()
```
wilsonianb, line 78: Do we need to specify the cluster id if here and `Config` both have `0`?
```suggestion
      const { container, port } = await startTigerbeetleContainer()
```

wilsonianb (approved):

Issue #988, opened 2023-01-17
Replace deprecated apollo-server-koa


wilsonianb:
https://www.npmjs.com/package/apollo-server-koa

We should update to Apollo Server 4
https://www.apollographql.com/docs/apollo-server/previous-versions/

>Three features of Apollo Server 2 have an earlier end-of-life date: December 31st, 2022. These three features are:
>- [Subscriptions](https://www.apollographql.com/docs/apollo-server/v2/data/subscriptions/)
>- [File uploads](https://www.apollographql.com/docs/apollo-server/v2/data/file-uploads/)
>- [**GraphQL Playground**](https://www.apollographql.com/docs/apollo-server/v2/testing/graphql-playground/).

PR #1014, opened 2023-01-23
feat(backend): migrate to `@apollo/server`

raducristianpopa:

- Migrate to `@apollo/server`

## Context

- fixes https://github.com/interledger/rafiki/issues/988

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/app.ts
```diff
@@ -224,21 +228,31 @@ export class App {
...
+      '/graphql',
+      koaMiddleware(this.apolloServer, {
+        context: async (): Promise<ApolloContext> => {
+          return {
+            container: this.container,
+            logger: await this.container.use('logger')
+          }
+        }
+      })
+    )
```
raducristianpopa, line 251: We need to specify a route for GraphQL, otherwise any endpoint (eg: `localhost:3001/foo/bar`) will be treated as a GraphQL request.

Without this route, this was failing: 
https://github.com/interledger/rafiki/blob/f8d41ddbc970272e94fc4794c907ef16ef7ddf47/packages/backend/src/graphql/resolvers/index.test.ts#L21

packages/backend/src/app.ts
```diff
@@ -224,21 +228,31 @@ export class App {
...
+    // Setup Apollo
     this.apolloServer = new ApolloServer({
       schema: schemaWithResolvers,
-      context: async (): Promise<ApolloContext> => {
-        return {
-          container: this.container,
-          logger: await this.container.use('logger')
-        }
-      }
+      plugins: [ApolloServerPluginDrainHttpServer({ httpServer })]
```
raducristianpopa, line 234: Apollo highly recommends using this plugin. From Apollo docs:

```
We highly recommend using this plugin to ensure your server shuts down gracefully.

Specifically, it will:

    Stop listening for new connections
    Close idle connections (i.e., connections with no current HTTP request)
    Close active connections whenever they become idle
    Wait for all connections to be closed
    After a grace period, if any connections remain active, forcefully close them.
```

Although, do we still need it since we have graceful shutdown for the container ?

packages/backend/src/graphql/resolvers/asset.test.ts
```diff
@@ -1,4 +1,4 @@
-import { gql } from 'apollo-server-koa'
+import { gql } from '@apollo/client'
```
raducristianpopa, line 1: `@apollo/server` does not export the `gql` tag. 

Importing this from `@apollo/client` since it's only used for tests, so we don't have to install another dependency (`graphl-tag`).

packages/backend/src/app.ts
raducristianpopa: At the moment the GraphQL playground is not working since we are only expecting the `POST` method for this endpoint. 

Should we do `router.all` to match any method or remove the router? Removing the router means that any endpoint will match the GraphQL request and we can remove the test mentioned above.

wilsonianb: I wouldn't expect GraphQL Playground to be working. It sounds like we'd need to instead switch to Apollo Sandbox.
Deciding what method(s) to have here would depend on what Apollo Sandbox needs.

https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages/#graphql-playground-landing-page

mkurapov (approved):
Was able to make the same Postman GraphQL calls with this checked out. Not sure why Sandbox isn't working, however.

packages/backend/src/app.ts
raducristianpopa: > I wouldn't expect GraphQL Playground to be working. It sounds like we'd need to instead switch to Apollo Sandbox.

Sorry for the confusion. The "playground" in the newer versions of Apollo is an embedded version of Sandbox.

> It seems as though Apollo Sandbox should be working out of the box with the current setup as it stands... nothing jumps to mind based on the docs 🤔

The Sandbox is not working at the moment because I did this:
```ts
koa.use(router.routes()).use(router.allowedMethods())
```
and it was only expecting `POST` requests for that route.

raducristianpopa:
I've removed the router because I find it a bit confusing in this context. The main problem was that `router.allowedMethods` was used and only the `POST` request was defined for `/graphql`. Out of the box, Apollo throws a `405 Method Not Allowed` if other request methods are used, since it's only supporting `GET` and `POST` methods.

Apollo Sandbox is working. 👍

packages/backend/package.json
```diff
@@ -41,6 +41,8 @@
   "dependencies": {
     "@adonisjs/fold": "^8.2.0",
     "@apollo/client": "^3.7.4",
```
wilsonianb, line 43: I think `@apollo/clien` can be moved to `devDependencies`

wilsonianb (commented):
Conflict needs resolving.

I tried out the Sandbox, and I don't think docs need to be updated :rocket: 
https://github.com/interledger/rafiki/blob/main/infrastructure/local/peer-setup.md

mkurapov:
Sandbox works for me as well at http://localhost:3001/graphql and http://localhost:4001/graphql now :+1:

mkurapov:
@raducristianpopa I can approve once you let me know :+1:

PR #1024, opened 2023-01-24
662/mk/use open payments types

### LINKED ISSUE #662: Use generated Open Payments OpenAPI types

mkurapov:

- Exporting additional IncomingPayment types from `open-payments` package
- Adding `toOpenPaymentsType()` method for subresources to convert from model to the corresponding `open-payments` type, removing `toJSON` types and method where necessary
- use `toOpenPaymentsType()` methods in the route handlers when creating response 

## Context

- Fixes #662 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


mkurapov:
Currently running into this challenge: when calling `.toOpenPayments` on a model, I need the corresponding `paymentPointer.url`. However, other than the  IncomingPayment model, `Model.paymentPointer` is optional, since it needs to be fetched from the DB. IncomingPayment gets around this via `.withGraphFetched(..., 'paymentPointer')` existing on every IncomingPayment query, but I feel like it could be easy to miss when making those calls in the service. A few options exist here:

- Go with the `IncomingPayment` method, add a non-null assertion (`paymentPointer!`) to each of the PaymentPointerSubresources, and add `.withGraphFetched(..., 'paymentPointer')` everytime when requesting a model from the DB.
- Make `toOpenPaymentsType` async, and have a helper method `getPaymentPointer/getWithPaymentPointer` that fetches the paymentPointer if its not already on the model instance
- Make `paymentPointer` non-nullable on `PaymentPointerSubresources` by always eager fetching it (there's a few ways of doing this, one of them being as such
```ts
class SubresourceQueryBuilder {
  ...
  execute() {
    this.withGraphFetched('paymentPointer')
    return super.execute()
  }
}
```

wilsonianb:
Reading option 1 made think of doing something like option 3.

Adrian's idea :point_down: might eventually help with this.
- https://github.com/interledger/open-payments/issues/206

mkurapov:
I was trying it out with the example above of overriding `execute()`, it seemed to work. I was testing out with the debugger, seeing all of the different SQL calls that were made, and even though its straightforward request, I'm a bit concerned that every single request, insert, get, update, it would do this DB call.

Instead of solving it programmatically, what if we just stored a `paymentPointerUrl` foreign key on the PaymentPointerSubresource models? It is already a non-nullable unique index. 

Otherwise, I'm leaning toward making `toOpenPaymentsType` async, since we fully control when we need that data, and there will never be a possibility of us not adding a field to `withGraphFetched`

wilsonianb:
I vote async `toOpenPaymentsType` over storing `paymentPointerUrl` on each subresource
~but only if we rename every synchronous function `*Sync` to make it clear that `toOpenPaymentsType` is async~ :stuck_out_tongue_winking_eye:

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -84,7 +84,6 @@ export class IncomingPayment
     }
   }
 
-  public paymentPointer!: PaymentPointer
```
mkurapov, line 83: no non-null assertion, will need to use `getPaymentPointer()` if necessary

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -131,8 +130,15 @@ export class IncomingPayment
     this.receivedAmountValue = amount.value
   }
 
-  public get url(): string {
-    return `${this.paymentPointer.url}${IncomingPayment.urlPath}/${this.id}`
+  public async getUrl(fetchedPaymentPointer?: PaymentPointer): Promise<string> {
```
mkurapov, line 133: Using `fetchedPaymentPointer` in-case you already have the paymentPointer and don't want to grab it again, like in `toOpenPaymentsType`

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -168,6 +183,29 @@ export class OutgoingPayment
...
+    return this.paymentPointer ?? (await this.$relatedQuery('paymentPointer'))
+  }
+
+  public async toOpenPaymentsType(): Promise<OpenPaymentsOutgoingPayment> {
+    const paymentPointer = await this.getPaymentPointer()
+
+    return {
+      id: await this.getUrl(paymentPointer),
+      paymentPointer: paymentPointer.url,
+      quoteId: (await this.quote?.getUrl()) ?? undefined,
```
mkurapov, line 197: this is optional in the Open Payments API -> should we keep it?

packages/backend/src/open_payments/payment_pointer/routes.ts
```diff
@@ -66,10 +68,7 @@ export const listSubresource = async <M extends PaymentPointerSubresource>({
   )
   const result = {
     pagination: pageInfo,
-    result: page.map((item: M) => {
-      item.paymentPointer = ctx.paymentPointer
```
mkurapov, line 70: We are fetching paymentPointer on the fly (whenever it's missing), this line isn't necessary

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -239,39 +249,56 @@ export class IncomingPayment
...
   }
 }
-
-// TODO: disallow undefined
-// https://github.com/interledger/rafiki/issues/594
-export type IncomingPaymentJSON = {
-  id: string
-  paymentPointer: string
-  incomingAmount?: AmountJSON
-  receivedAmount: AmountJSON
```
mkurapov, line 269: @wilsonianb Keeping `AmountJSON` over `OpenPaymentsAmount` makes sense right? The Amount interface is a more generic thing we use across the code, not necessarily being an "open payments" concept

mkurapov:
Going to update the `auth` package with a few of it's types in a follow-up

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -149,19 +149,23 @@ async function listIncomingPayments(
     await listSubresource({
       ctx,
       getPaymentPointerPage: deps.incomingPaymentService.getPaymentPointerPage,
-      toBody: (payment) =>
-        incomingPaymentToBody(payment, deps.connectionService.getUrl(payment))
+      toBody: async (payment) =>
+        await incomingPaymentToBody(
+          payment,
+          deps.connectionService.getUrl(payment)
```
wilsonianb, line 155: Could we either make this `getUrl` async for consistency or rename it?

packages/backend/src/open_payments/payment/outgoing/model.ts
wilsonianb: :thinking: I wonder if we should make it required in the spec

packages/backend/src/open_payments/quote/routes.ts
```diff
@@ -80,7 +80,7 @@ async function createQuote(
     }
 
     ctx.status = 201
-    ctx.body = quoteToBody(quoteOrErr)
+    ctx.body = await quoteToBody(quoteOrErr)
```
wilsonianb, line 83:
```suggestion
    ctx.body = await quoteOrErr.toOpenPaymentsType()
```
Remove `quoteToBody`

packages/backend/src/open_payments/payment/incoming/model.ts
wilsonianb: What if callers just assigned `incomingPayment.paymentPointer = fetchedPaymentPointer`, so that `getUrl`+`getPaymentPointer` only dealt with whether `this.paymentPointer` is defined or not?

packages/backend/src/open_payments/payment/outgoing/model.ts
mkurapov: If you can only create an Outgoing Payment with a required `quoteId`, it would make sense that this would be required here

packages/backend/src/open_payments/payment/incoming/routes.ts
mkurapov: I _think_ it should be ok as it is, since it's a service method, not a model?

packages/backend/src/open_payments/payment/incoming/model.ts
mkurapov: I'm hesitant to do that, since 
a) we either have to change the type definitions of `getUrl/toOpenPaymentsType` to possibly return `undefined`, or throw inside the method if `paymentPointer` isn't defined
b) it's not entirely clear that the caller has to do that in order to make these methods work.

Maybe we can 
1) have `toOpenPaymentsType` take in a `paymentPointer` (this would also make it sync), it would require passing in `ctx.paymentPointer` in most places but I think that's ok
2) A bit more complicated, but we make a separate `<subresource>WithPaymentPointer` class that constructs a new model with the `paymentPointer` defined, although I'm not sure the additional boilerplate is worth it

wilsonianb (commented):
Oops, here's some several day old comments that I didn't realize were pending in a review

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -239,39 +242,60 @@ export class IncomingPayment
...
+    ilpStreamConnection: string
+  ): OpenPaymentsIncomingPaymentWithConnectionUrl
+  public toOpenPaymentsType(
+    paymentPointer: PaymentPointer,
+    ilpStreamConnection?: Connection | string
+  ):
+    | OpenPaymentsIncomingPaymentWithConnection
+    | OpenPaymentsIncomingPaymentWithConnectionUrl
+
+  public toOpenPaymentsType(
```
wilsonianb, line 263: What if we added some `model.test.ts`s at least to test `toOpenPaymentsType`?
Or are existing response body checks in routes tests good enough?

packages/backend/src/open_payments/payment/incoming/routes.test.ts
```diff
@@ -82,31 +82,33 @@ describe('Incoming Payment Routes', (): void => {
...
-              id: `${config.openPaymentsUrl}/connections/${incomingPayment.connectionId}`,
-              ilpAddress: expect.stringMatching(
-                /^test\.rafiki\.[a-zA-Z0-9_-]{95}$/
-              ),
-              sharedSecret: expect.stringMatching(/^[a-zA-Z0-9-_]{43}$/),
-              assetCode: incomingPayment.receivedAmount.assetCode,
-              assetScale: incomingPayment.receivedAmount.assetScale
-            }
-      }),
+      getBody: (incomingPayment, list) => {
```
wilsonianb, line 85: Could this instead be
```suggestion
      getBody: (incomingPayment, list) => incomingPayment.toOpenPaymentsType(paymentPointer)
```
with some possible special assignment for `ilpStreamConnection`?

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -168,6 +180,29 @@ export class OutgoingPayment
       updatedAt: json.updatedAt
     }
   }
```
wilsonianb, line 182: Can we not get rid of the `$formatJson` :point_up_2: ?
Ditto for incoming payments but not quotes
https://github.com/interledger/rafiki/blob/4df802bb059097546f21178505aff35d89a36757/packages/backend/src/open_payments/quote/service.ts#L296

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -168,6 +180,29 @@ export class OutgoingPayment
...
+
+  public async getPaymentPointer(): Promise<PaymentPointer> {
+    return this.paymentPointer ?? (await this.$relatedQuery('paymentPointer'))
+  }
+
```
wilsonianb, line 187:
```suggestion
```

packages/backend/src/open_payments/payment/incoming/routes.test.ts
mkurapov: In terms of the connection, we would be able to do something like:
```ts
getBody: (incomingPayment, list) => ({
        ...incomingPayment.toOpenPaymentsType(paymentPointer),
        ilpStreamConnection: list
          ? `${config.openPaymentsUrl}/connections/${incomingPayment.connectionId}`
          : {
              id: `${config.openPaymentsUrl}/connections/${incomingPayment.connectionId}`,
              ilpAddress: expect.stringMatching(
                /^test\.rafiki\.[a-zA-Z0-9_-]{95}$/
              ),
              sharedSecret: expect.stringMatching(/^[a-zA-Z0-9-_]{43}$/),
              assetCode: incomingPayment.receivedAmount.assetCode,
              assetScale: incomingPayment.receivedAmount.assetScale
            }
      }),
```

Since `incomingPayment.toOpenPaymentsType` expect a `Connection`, we wouldn't be able to use `expect.stringMatching(...)`, but would have to create a `Connection` instance with proper `credentials` etc

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -68,10 +72,18 @@ export class OutgoingPayment
...
 
+  public getUrl(paymentPointer: PaymentPointer): string {
+    return `${paymentPointer.url}${OutgoingPayment.urlPath}/${this.id}`
+  }
+
   public get asset(): Asset {
     return this.quote.asset
   }
 
+  public get failed(): boolean {
```
wilsonianb, line 83: This could be updated to `grantPayment.failed`
https://github.com/interledger/rafiki/blob/4df802bb059097546f21178505aff35d89a36757/packages/backend/src/open_payments/payment/outgoing/service.ts#L296

packages/backend/src/open_payments/payment/incoming/model.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    jest.useRealTimers()
+    await truncateTables(appContainer.knex)
+  })
+
+  afterAll(async (): Promise<void> => {
+    await appContainer.shutdown()
+  })
+
+  describe('toOpenPaymentsType', () => {
+    test('returns incoming payment without connection provided', async () => {
```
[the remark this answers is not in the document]
mkurapov: I was thinking this, but having a test case where the `connection` var in the `test.each` case was either
```
const connection = `${config.openPaymentsUrl}/connections/${incomingPayment.connectionId}` 
```
or 
```
const connection = Connection.fromPayment({
        payment: incomingPayment,
        openPaymentsUrl: config.openPaymentsUrl,
        credentials: {
          ilpAddress: 'test.ilp' as IlpAddress,
          sharedSecret: Buffer.from('')
        }
      })
```
didn't feel nice

packages/backend/src/open_payments/payment/incoming/routes.test.ts
mkurapov: i.e. I'm not sure if that's a much better solution

wilsonianb (approved):

PR #1028, opened 2023-01-25
feat(backend): rotate expired access token

### LINKED ISSUE #795: Rotate expired grant access token for incoming payment lookup from one Rafiki to another
https://github.com/interledger/rafiki/pull/779#discussion_r1033736498

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-6.1
>   If the client instance has an access token and that access token
   expires, the client instance might want to rotate the access token.
   Rotating an access token consists of issuing a new access token in
   place of an existing access token, with the same rights and
   properties as the original token, apart from an updated expiration
   time.

sabineschaller:

- adds management url to stored `Grant`
- implements update functionality for grant (to updated access token)
- rotates expired access tokens and updates `Grant`

## Context

- fixes #795 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -298,10 +298,24 @@ async function getIncomingPaymentGrant(
...
+        return undefined
+      }
+      try {
+        const rotatedToken = await deps.openPaymentsClient.token.rotate({
+          url: existingGrant.managementUrl,
+          accessToken: existingGrant.accessToken
+        })
+        return deps.grantService.update(existingGrant, {
+          accessToken: rotatedToken.access_token.value,
+          managementUrl: rotatedToken.access_token.manage,
```
wilsonianb, line 312: Since this `url` should adhere to the Open Payments spec, could we just store the `managementId`?
https://docs.openpayments.guide/reference/post-token
The url can be reconstructed when rotating with
```
`${grant.authServer.url}/token/${grant.managementId}`
```
(could be a `managementUrl` getter on the `Grant` model)

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -280,7 +289,11 @@ describe('Receiver Service', (): void => {
             .mockResolvedValueOnce(undefined)
         })
 
-        test('resolves incoming payment', async () => {
+        test.each`
+          rotate   | description
+          ${false} | ${''}
+          ${true}  | ${'- after rotating access token'}
+        `('resolves incoming payment $description', async ({ rotate }) => {
```
wilsonianb, line 296: I wonder if we can skip for `!existingGrant && rotate`

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -298,10 +298,24 @@ async function getIncomingPaymentGrant(
...
-      // TODO
-      // https://github.com/interledger/rafiki/issues/795
-      deps.logger.warn({ grantOptions }, 'Grant access token expired')
-      return undefined
+      if (!existingGrant.managementUrl || !existingGrant.accessToken) {
+        deps.logger.warn({ existingGrant }, 'Grant token cannot be rotated.')
+        return undefined
+      }
+      try {
+        const rotatedToken = await deps.openPaymentsClient.token.rotate({
```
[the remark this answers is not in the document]
mkurapov: @sabineschaller would you like to consolidate this in this PR? Or I can do it once this is finished?

packages/backend/src/open_payments/receiver/service.ts
mkurapov: is it possible `${grant.authServer.url}/token/${grant.managementId}` !== `rotatedToken.access_token.manage` in any scenario? If not then `managementId` with a `managementUrl` getter makes sense to me

packages/backend/migrations/20221012013413_create_grants_table.js
```diff
@@ -6,6 +6,7 @@ exports.up = function (knex) {
     table.string('continueId').nullable()
     table.string('continueToken').nullable()
     table.string('accessToken').nullable().unique()
+    table.string('managementUrl').nullable()
```
mkurapov, line 9: Can we make this nonnullable, since `access_token.manage` is always defined?

packages/backend/src/open_payments/receiver/service.ts
wilsonianb: The fact that we've got the rotate and revoke token routes defined in the Open Payments spec makes me think `access_token.manage` should always match that format.
I'm even wondering if `open-payments` client should be validating that in relevant response bodies.

packages/backend/src/open_payments/receiver/service.test.ts
sabineschaller: I was wondering the same thing. I meant to ask you guys.

packages/backend/src/open_payments/receiver/service.ts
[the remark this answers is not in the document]
sabineschaller: Let's try to not blow up the PRs too much. Either you @mkurapov or I can do it once this is merged, I suggest.

packages/backend/migrations/20221012013413_create_grants_table.js
sabineschaller: I think the same goes for `accessToken`

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -324,12 +334,27 @@ async function getIncomingPaymentGrant(
...
   deps.logger.warn({ grantOptions }, 'Grant request required interaction')
   return undefined
 }
+
+function retrieveManagementId(managementUrl: string): string {
+  const managementUrlParts = managementUrl.split('/')
+  const managementId = managementUrlParts.pop() || managementUrlParts.pop() // handle trailing slash
+  if (!managementId) {
+    throw new Error('invalid management id')
+  }
```
sabineschaller, line 358: If we did response checking of the management url in the client, I wouldn't have to do that.

packages/backend/src/open_payments/receiver/service.ts
sabineschaller: I think we should do that 👆

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -3,14 +3,15 @@ import { faker } from '@faker-js/faker'
...
-import { GrantOptions, GrantService } from './service'
+import { CreateOptions, GrantService } from './service'
 import { AuthServer } from '../authServer/model'
 import { initIocContainer } from '../..'
 import { AppServices } from '../../app'
 import { Config } from '../../config/app'
 import { createTestApp, TestContainer } from '../../tests/app'
 import { truncateTables } from '../../tests/tableManager'
 import { AccessType, AccessAction } from 'open-payments'
+import { uuid } from '../../connector/ilp-routing/lib/utils'
```
wilsonianb, line 14:
```suggestion
import { v4 as uuid } from 'uuid'
```

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -84,7 +85,9 @@ describe('Grant Service', (): void => {
       `(
         'Grant can be created and fetched ($description)',
         async ({ expiresIn }): Promise<void> => {
-          const options: GrantOptions = {
+          const options: CreateOptions = {
+            accessToken: 'aq1sw2de3fr4',
+            managementId: 'gt5hy6ju7ki8',
```
wilsonianb, line 90:
```suggestion
            managementId: uuid(),
```

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -133,4 +138,42 @@ describe('Grant Service', (): void => {
...
+
+    test('can update grant', async (): Promise<void> => {
+      const updateOptions = {
+        accessToken: uuid(),
+        managementId: 'gt5hy6ju7ki8',
+        expiresIn: 3000
+      }
+      const updatedGrant = await grantService.update(grant, updateOptions)
+      console.log(grant)
+      console.log(updatedGrant)
```
wilsonianb, line 169:
```suggestion
```

packages/backend/src/open_payments/grant/model.ts
```diff
@@ -36,8 +38,7 @@ export class Grant extends BaseModel {
     return !!this.expiresAt && this.expiresAt <= new Date()
   }
 
-  $afterFind(queryContext: QueryContext): void {
-    super.$afterFind(queryContext)
-    delete this['authServer']
+  public get managementUrl(): string {
+    return `${this.authServer?.url}/token/${this.managementId}`
```
wilsonianb, line 42:
```suggestion
    return `${this.authServer.url}/token/${this.managementId}`
```
since we have `public authServer!: AuthServer` required above

If `authServer` is actually optional, then I think `managementUrl` should throw or return `undefined` if it's missing

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -133,4 +138,42 @@ describe('Grant Service', (): void => {
...
+      authServerId = (await authServerService.getOrCreate(options.authServer))
+        .id
+    })
+
+    test('can update grant', async (): Promise<void> => {
+      const updateOptions = {
+        accessToken: uuid(),
+        managementId: 'gt5hy6ju7ki8',
+        expiresIn: 3000
+      }
```
wilsonianb, line 166: Can you add a test with `expiresIn: undefined` to confirm that the existing value is still there vs being set to `null` (or should it be set to `null`, maybe if it's `null` in `updateOptions`)?

wilsonianb (commented):
Looks like there's some type errors

sabineschaller:
> Looks like there's some type errors

I asked for the review too quickly. I've been going back and forth on where to parse the management id and have some testing issues with 'packages/backend/src/open_payments/payment/incoming_remote/service.ts`

packages/backend/src/open_payments/grant/model.ts
sabineschaller: In my opinion, `authServer` should be required.

packages/backend/src/open_payments/grant/service.test.ts
sabineschaller: This service really updates the token values, not the grant details. So after rotating a token, if the return value has an `expiresIn` of `null`, I think the`expiresAt` should be set to `undefined` meaning that it does not expire, right?

I'll add a test for that.

sabineschaller: But for some reason, it does not set `expiresAt` to `undefined`, even when I set `expiresIn` to `undefined` 🤔. I checked the `BaseModel` but we don't set defaults anywhere, do we?

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from 'open-payments'
 import { GrantService } from '../../grant/service'
 import { RemoteIncomingPaymentError } from './errors'
+import { AccessToken } from 'open-payments/dist/types'
```
wilsonianb, line 21:
```suggestion
import { AccessToken } from 'open-payments'
```

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -85,69 +100,42 @@ describe('Remote Incoming Payment Service', (): void => {
...
+      test('returns error if rotated token does not have accessToken', async () => {
         await grantService.create({
           ...grantOptions,
-          accessToken: undefined
+          expiresIn: -10
         })
+        const clientRequestRotateTokenSpy = jest
+          .spyOn(openPaymentsClient.token, 'rotate')
+          .mockResolvedValueOnce({
+            access_token: { ...newToken.access_token, value: '' }
```
wilsonianb, line 130: I don't think this is possible.
https://github.com/interledger/open-payments/blob/f1b51c3c051c9030cea1f3d2542cfb492c4330d3/openapi/auth-server.yaml#L258-L261
https://github.com/interledger/open-payments/blob/f1b51c3c051c9030cea1f3d2542cfb492c4330d3/openapi/auth-server.yaml#L388-L406

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -41,7 +45,7 @@ describe('Grant Service', (): void => {
       existingAuthServer | description
       ${false}           | ${'new auth server'}
       ${true}            | ${'existing auth server'}
-    `('$description', ({ existingAuthServer }): void => {
+    `('description', ({ existingAuthServer }): void => {
```
wilsonianb, line 48:
```suggestion
    `('$description', ({ existingAuthServer }): void => {
```

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -82,9 +83,11 @@ describe('Grant Service', (): void => {
         ${undefined} | ${'without expiresIn'}
         ${600}       | ${'with expiresIn'}
       `(
-        'Grant can be created and fetched ($description)',
+        'Grant can be created and fetched (description)',
```
wilsonianb, line 86:
```suggestion
        'Grant can be created and fetched ($description)',
```

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -132,5 +137,71 @@ describe('Grant Service', (): void => {
...
+    })
+
+    test('can update grant', async (): Promise<void> => {
+      const updateOptions = {
+        accessToken: uuid(),
+        managementUrl: `${faker.internet.url()}/${uuid()}`,
+        expiresIn: 6000
+      }
+      const updatedGrant = await grantService.update(grant, updateOptions)
+      expect(updatedGrant).toMatchObject({
```
wilsonianb, line 181: This could be a `toEqual` compared against `grant` (instead of `options`)

packages/backend/src/open_payments/grant/service.ts
```diff
@@ -34,24 +36,30 @@ export interface GrantOptions {
   accessActions: AccessAction[]
 }
 
-export interface CreateOptions extends GrantOptions {
-  accessToken?: string
+export interface UpdateOptions {
+  accessToken: string
+  managementUrl: string
   expiresIn?: number
 }
```
wilsonianb, line 43:
```suggestion
export type UpdateOptions = Omit<AccessToken, 'access'>
```
Import `AccessToken` from `open-payments`.
(This would have `expires_in` instead of `expiresIn`).
Maybe we could also update `CreateOptions` to be based on `open-payments` types.

wilsonianb: and `manage` instead of `managementUrl`
and `value` instead of `accessToken`
:thinking: 
Nevermind

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -164,6 +152,69 @@ describe('Remote Incoming Payment Service', (): void => {
...
+                  ? serializeAmount(args.incomingAmount)
+                  : undefined
+              }
+            )
+            if (grantExpired) {
+              expect(clientRequestRotateTokenSpy).toHaveBeenCalledWith({
+                url: grantOptions.managementUrl,
+                accessToken: grantOptions.accessToken
+              })
+            }
```
wilsonianb, line 214:
```suggestion
            } else {
              expect(clientRequestRotateTokenSpy).not.toHaveBeenCalled()
            }
```

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
sabineschaller: `AccessToken` is currently not exported by `open-payments` but I could change that.

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
sabineschaller: Then let's get rid of the test

packages/backend/src/open_payments/grant/service.test.ts
sabineschaller: But then we'll also have to check the entire authServer details plus createdAt, updatedAt. I think it's enough to check the crucial details.

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -137,6 +137,19 @@ describe('Grant Service', (): void => {
         })
       ).resolves.toBeUndefined()
     })
+
+    test('cannot store grant with misformatted management url', async (): Promise<void> => {
+      const options: CreateOptions = {
+        accessToken: 'aq1sw2de3fr4',
+        managementUrl: '',
```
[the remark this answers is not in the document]
sabineschaller: Do we want all of these to fail? Because if so, I'll have to rewrite the checking function. Right now, it only checks that it can parse some sort of id but does not care about the path. 
Actually, it only checks that there is any string, which definitely needs updating.

packages/backend/src/open_payments/grant/service.test.ts
wilsonianb:
```
      expect(updatedGrant).toEqual({
        ...grant,
        accessToken: updateOptions.accessToken,
        managementId: updateOptions.managementUrl.split('/').pop(),
        expiresAt: new Date(Date.now() + updateOptions.expiresIn * 1000),
        updatedAt: updatedGrant.updatedAt
      })
```

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
mkurapov: I think would be better to export from `index` in `open-payments` 👍

packages/backend/src/open_payments/receiver/service.ts
mkurapov: I'm guessing that check can be done anytime we return an access_token:
- grant creation
- grant continuation
- token rotation
right?

wilsonianb: :heavy_check_mark:  right

packages/backend/src/open_payments/grant/service.test.ts
sabineschaller: Ok, you win :joy:

packages/backend/src/open_payments/receiver/service.ts
sabineschaller: https://github.com/interledger/rafiki/issues/1063

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -296,10 +296,20 @@ async function getIncomingPaymentGrant(
   const existingGrant = await deps.grantService.get(grantOptions)
```
mkurapov, line 296: L241 in this service we could update from
`accessToken: grant.accessToken || ''` -> `accessToken: grant.accessToken`

packages/backend/src/open_payments/grant/model.ts
```diff
@@ -25,19 +25,20 @@ export class Grant extends BaseModel {
   }
 
   public authServerId!: string
+  public authServer!: AuthServer
```
mkurapov, line 28: To be on the safe side, we _could_ keep this as optional, just in case there is no `withGraphFetched` on a `Grant` query, and make `getManagementUrl` an `async` call. 

Similar discussion here: https://github.com/interledger/rafiki/pull/1024#issuecomment-1405432584

mkurapov (approved):

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -132,5 +137,70 @@ describe('Grant Service', (): void => {
...
+      const updatedGrant = await grantService.update(grant, updateOptions)
+      expect(updatedGrant).toMatchObject({
+        authServerId,
+        accessType: options.accessType,
+        accessActions: options.accessActions,
+        accessToken: updateOptions.accessToken,
+        managementId: updateOptions.managementUrl.split('/').pop(),
+        expiresAt: null
+      })
+    })
```
wilsonianb, line 204: The meaningful change is the `describe.each`
```suggestion
  describe.each`
    expiresIn    | description
    ${undefined} | ${'without prior expiresIn'}
    ${3000}      | ${'with prior expiresIn'}
  `('Update Grant ($description)', ({ expiresIn }): void => {
    let grant: Grant
    beforeEach(async (): Promise<void> => {
      const options = {
        authServer: faker.internet.url(),
        accessType: AccessType.IncomingPayment,
        accessActions: [AccessAction.ReadAll],
        accessToken: uuid(),
        managementUrl: `${faker.internet.url()}/gt5hy6ju7ki8`,
        expiresIn
      }
      grant = await grantService.create(options)
    })

    test.each`
      expiresIn    | description
      ${undefined} | ${'without expiresIn'}
      ${6000}      | ${'with expiresIn'}
    `(
      'can update grant ($description)',
      async ({ expiresIn }): Promise<void> => {
        const updateOptions = {
          accessToken: uuid(),
          managementUrl: `${faker.internet.url()}/${uuid()}`,
          expiresIn
        }
        const updatedGrant = await grantService.update(grant, updateOptions)
        expect(updatedGrant).toEqual({
          ...grant,
          accessToken: updateOptions.accessToken,
          managementId: updateOptions.managementUrl.split('/').pop(),
          expiresAt: expiresIn ? new Date(Date.now() + expiresIn * 1000) : null,
          updatedAt: updatedGrant.updatedAt
        })
      }
    )
  })
```

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -85,6 +100,24 @@ describe('Remote Incoming Payment Service', (): void => {
...
+      test('returns error if grant expired and token cannot be rotated', async () => {
+        await grantService.create({
+          ...grantOptions,
+          expiresIn: -10
+        })
+        const clientRequestRotateTokenSpy = jest
+          .spyOn(openPaymentsClient.token, 'rotate')
+          .mockImplementationOnce(() => {
+            throw new Error('Error in rotating client')
+          })
```
wilsonianb, line 112:
```suggestion
          .mockRejectedValueOnce(new Error('Error in rotating client'))
```

packages/backend/src/open_payments/payment/incoming_remote/service.test.ts
```diff
@@ -164,6 +171,71 @@ describe('Remote Incoming Payment Service', (): void => {
           })
         ).resolves.toEqual(RemoteIncomingPaymentError.InvalidRequest)
       })
+
+      describe.each`
```
wilsonianb, line 175: The `newToken` defiinition could be moved inside this describe or test block

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -120,10 +120,20 @@ async function getGrant(
...
+          accessToken: existingGrant.accessToken
+        })
+        return deps.grantService.update(existingGrant, {
+          accessToken: rotatedToken.access_token.value,
+          managementUrl: rotatedToken.access_token.manage,
+          expiresIn: rotatedToken.access_token.expires_in
+        })
+      } catch (err) {
+        deps.logger.error({ grantOptions }, 'Grant token rotation failed.')
+        return RemoteIncomingPaymentError.ExpiredGrant
```
wilsonianb, line 135: I'm wondering if we should instead throw or if this should be renamed.
If I'm a GNAP/Open Payments-aware client and was told a relevant grant was expired, I would think "well, that grant's access token should be rotated". But rotation was already attempted and failed.

packages/backend/src/open_payments/receiver/service.test.ts
wilsonianb: I vote yes but I'm not sure how other than wrap the inside of the test with `if (!existingGrant && rotate) {`?

packages/backend/src/open_payments/grant/model.ts
wilsonianb: I vote
```suggestion
  public authServer?: AuthServer

  public getManagementUrl(authServerUrl: string): string {
    return `${authServerUrl}/token/${this.managementId}`
  }
```

afaict, all the places we're calling this have the `paymentPointer.authServer` url
which might let us remove some or all `.withGraphFetched('authServer')

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -120,10 +120,20 @@ async function getGrant(
 
   if (existingGrant) {
```
mkurapov, line 121: I also noticed: since `grant.accessToken` is now always defined, we can remove the `if (!grantOrError.accessToken) {` check on L67 in this service

packages/backend/src/open_payments/payment/incoming_remote/service.ts
mkurapov: that makes sense to me - I wouldn't mind logging out the error here in the log on 134, and just rethrowing the error. Since it's mostly an Admin API, we'll return a error message 500

sabineschaller: :+1:

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -120,10 +113,23 @@ async function getGrant(
...
+          url: existingGrant.getManagementUrl(existingGrant.authServer.url),
+          accessToken: existingGrant.accessToken
+        })
+        return deps.grantService.update(existingGrant, {
+          accessToken: rotatedToken.access_token.value,
+          managementUrl: rotatedToken.access_token.manage,
+          expiresIn: rotatedToken.access_token.expires_in
+        })
+      } catch (err) {
+        deps.logger.error({ grantOptions }, 'Grant token rotation failed.')
```
wilsonianb, line 130:
```suggestion
        deps.logger.error({ err, grantOptions }, 'Grant token rotation failed.')
```

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -120,10 +113,23 @@ async function getGrant(
 
   if (existingGrant) {
     if (existingGrant.expired) {
-      // TODO https://github.com/interledger/rafiki/issues/795
-      const errorMessage = 'Grant access token expired'
-      deps.logger.error({ grantOptions }, errorMessage)
-      return RemoteIncomingPaymentError.ExpiredGrant
+      if (!existingGrant.authServer) {
+        return RemoteIncomingPaymentError.UnknownAuthServer
```
wilsonianb, line 117: This _shouldn't_ happen
https://github.com/interledger/rafiki/blob/bbebbdf987b00b6256e239ef30497a21dca3ceb2/packages/backend/src/open_payments/grant/service.ts#L65-L73
Should we instead throw? Or have `grantService.get` return a type where `authServer` is required?

mkurapov: > Or have `grantService.get` return a type where authServer is required?

I like this as a general pattern

mkurapov: _maybe_ we could just have a separate method even?

sabineschaller: Returning a different type where `authServer` is required is not that straight forwards due to the `PaginationQueryBuilder` on the `BaseModel`. I think we should just throw.

wilsonianb (approved):

mkurapov (approved):

PR #1029, opened 2023-01-25
feat(backend): make asset scale uint8 in admin API

### LINKED ISSUE #957: Disallow negative asset scales throughout the backend code
Currently the GraphQL API accepts negative asset scale values when creating a new asset. Instead the backend code should not allow the asset scale to be negative. The [STREAM spec](https://interledger.org/rfcs/0029-stream/#5-packet-and-frame-specification) says asset scale is UInt8.

We may need to add a UInt8 scalar to our graphql schema.
https://github.com/interledger/rafiki/blob/main/packages/backend/src/graphql/schema.graphql#L608
https://www.npmjs.com/package/graphql-type-ints

sabineschaller:

- adds UInt8 scalar to GraphQL schema and resolvers to make sure that asset scale is non-negative (and not too large)
- (not needed for Open Payments API since the spec validation checks that asset scale UInt8)

## Context

- fixes #957 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/scalars/UInt8.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        a: mirror(input: $input1) { result }
+        b: mirror(input: $input2) { result }
+        d: mirror(input: $input4) { result }
+      }`
+
+  const validVariables = {
+    input1: { num: 5 },
+    input2: { num: '' },
+    input4: { num: '1' }
+  }
```
wilsonianb, line 95: Can these be moved inside their corresponding tests?
(Are these tests based on an example you found?)

wilsonianb (commented):
The linked issue('s title) suggests enforcing uint8 throughout `backend`.
I'm ok confining this pr to the admin api but maybe it either shouldn't close the issue or we should open a separate issue.

sabineschaller:
> The linked issue('s title) suggests enforcing uint8 throughout backend.
I'm ok confining this pr to the admin api but maybe it either shouldn't close the issue or we should open a separate issue.

Where else do we need to enforce it? In the Open Payments API, it's enforced via the request/response validation.

packages/backend/src/graphql/scalars/UInt8.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  it.each`
+    query                      | description
+    ${`{a: mirror(num: -1)}`}  | ${'is negative'}
+    ${`{a: mirror(num: 256)}`} | ${'is larger than 255'}
```
mkurapov, line 100: Maybe a test for `0` would be a good sanity check

wilsonianb: it's in `validQuery`

wilsonianb:
> Where else do we need to enforce it? In the Open Payments API, it's enforced via the request/response validation.

At least in postgres
https://github.com/interledger/rafiki/blob/52b6bbfe32f37ed05a62ca9fe0de2bdf702d0530/packages/backend/migrations/20210422194130_create_assets_table.js#L8

sabineschaller:
>> Where else do we need to enforce it? In the Open Payments API, it's enforced via the request/response validation.

> At least in postgres

Good point

packages/backend/migrations/20210422194130_create_assets_table.js
```diff
@@ -5,7 +5,7 @@ exports.up = function (knex) {
     // TigerBeetle account 2 byte ledger field representing account's asset
     table.specificType('ledger', 'smallserial').notNullable()
     table.string('code').notNullable()
-    table.integer('scale').notNullable()
+    table.tinyint('scale').unsigned().notNullable()
```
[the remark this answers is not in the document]
mkurapov: I wonder if `knex` would automatically convert this to a `smallint`, since `tinyint` does not seem to be supported by psql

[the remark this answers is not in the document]
mkurapov: Can confirm `table.tinyint('tinyinttest').nullable()` shows up as `smallint`:
Screenshot 2023-02-01 at 7 14 30 PM

packages/backend/src/graphql/scalars/UInt8.test.ts
mkurapov: 👓 

With the new changes it's easier to see now as well

mkurapov (dismissed):
Looks good, I wouldn't mind updating the data type of `scale`, however

packages/backend/migrations/20210422194130_create_assets_table.js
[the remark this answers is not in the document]
sabineschaller: So that's not the answer. I'll see if I can enforce min and max.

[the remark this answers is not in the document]
sabineschaller: Or I guess max is enough since it's unsigned.

mkurapov (approved):

packages/backend/src/asset/service.test.ts
```diff
@@ -115,6 +116,16 @@ describe('Asset Service', (): void => {
...
       )
     })
+
+    test('Cannot create asset with scale > 255', async (): Promise<void> => {
+      const options = {
+        code: 'ABC',
+        scale: 256
+      }
+      await expect(assetService.create(options)).rejects.toThrow(
+        CheckViolationError
```
wilsonianb, line 126: I think I'm ok with throwing instead of returning an `AssetError` (https://github.com/interledger/rafiki/pull/1029#discussion_r1093353141) since this should already have been checked at all our entry points (Open Payments, admin api).
Unless there's some other route here I'm overlooking.

packages/backend/src/asset/service.test.ts
```diff
@@ -115,6 +116,16 @@ describe('Asset Service', (): void => {
         AssetError.DuplicateAsset
       )
     })
+
+    test('Cannot create asset with scale > 255', async (): Promise<void> => {
+      const options = {
+        code: 'ABC',
+        scale: 256
```
wilsonianb, line 123: You could also test with a negative scale

packages/backend/src/asset/service.test.ts
sabineschaller: That is what I was thinking. I just wrote the test to make sure that my check works.

packages/backend/src/asset/service.test.ts
sabineschaller: I was sure the `unsigned` constraint would work but I wasn't sure the `check` would so that is why I included this test.

Issue #916, opened 2022-12-21
Define Open Payments routes explicitly


wilsonianb:
Replace the existing magic `for` loop
https://github.com/interledger/rafiki/blob/e773fb7cd533a1a8495e056987a772154651130b/packages/backend/src/app.ts#L309-L367

wilsonianb:
While looking into resolving type issues related to this, I discovered `ctx.state`
>The recommended namespace for passing information through middleware and to your frontend views.

https://koajs.com/#context

wilsonianb:
> While looking into resolving type issues related to this, I discovered `ctx.state`
> 
> > The recommended namespace for passing information through middleware and to your frontend views.
> 
> https://koajs.com/#context

meanwhile in trpc land
https://trpc.io/docs/middlewares#context-swapping

wilsonianb:
limited progress
https://github.com/wilsonianb/rafiki/tree/bw-routes

PR #1032, opened 2023-01-25
fix(backend): support SPSP queries

### LINKED ISSUE #439: Support SPSP queries to stream connection endpoints
https://docs.openpayments.guide/reference/get-ilp-stream-connection

To allow web monetization payments to incoming payments.

### LINKED ISSUE #1030: SPSP query fails OpenAPI validation
running localenv:
```
$ curl  'http://localhost:3000/accounts/gfranklin' -H 'accept: application/spsp4+json' -H 'host: backend' 
must accept json
```
That error is presumably from `createValidatorMiddleware`, which shouldn't be used when handling an SPSP query
https://github.com/interledger/rafiki/blob/3968ae347743e1c05ae19c7bcd0f559655cbd5d6/packages/backend/src/app.ts#L381-L400

related to:
- https://github.com/interledger/rafiki/issues/439

wilsonianb:

- allow SPSP queries to both payment pointers and Open Payments connections

## Context

- fixes #1030 
- fixes #439 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -21,6 +21,10 @@ import { HttpTokenService } from './httpToken/service'
 import { AssetService, AssetOptions } from './asset/service'
 import { AccountingService } from './accounting/service'
 import { PeerService } from './peer/service'
+import {
+  connectionMiddleware,
+  ConnectionContext
+} from './open_payments/connection/middleware'
 import { createPaymentPointerMiddleware } from './open_payments/payment_pointer/middleware'
 import { PaymentPointer } from './open_payments/payment_pointer/model'
 import { PaymentPointerService } from './open_payments/payment_pointer/service'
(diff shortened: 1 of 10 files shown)


packages/backend/src/spsp/routes.ts
```diff
@@ -46,18 +45,15 @@ async function getPay(
 
   try {
     const { ilpAddress, sharedSecret } = deps.streamServer.generateCredentials({
```
wilsonianb, line 47: Existential question:
Do `routes.ts` files need `ServiceDependencies` if they can just access things via
```suggestion
    const streamServer = await ctx.container.use('streamServer')
    const { ilpAddress, sharedSecret } = streamServer.generateCredentials({
```

packages/backend/src/app.ts
```diff
@@ -329,6 +334,30 @@ export class App {
...
+                    const incomingPayment =
+                      await incomingPaymentService.getByConnection(
+                        ctx.params.id
+                      )
+                    if (!incomingPayment) return ctx.throw(404)
+                    ctx.paymentTag = incomingPayment.id
+                    ctx.asset = {
+                      code: incomingPayment.asset.code,
+                      scale: incomingPayment.asset.scale
+                    }
```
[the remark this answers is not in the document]
wilsonianb: A little better
https://github.com/interledger/rafiki/pull/1032/commits/8ef3153a12954c3685840c91ec8ec07683523746

packages/backend/src/app.ts
```diff
@@ -329,6 +338,23 @@ export class App {
               route = connectionRoutes.get
               router[method](
                 toRouterPath(path),
+                connectionMiddleware,
```
mkurapov, line 341: this will be nicer once the magic for loop is gone (ie https://github.com/interledger/rafiki/issues/916)

packages/backend/src/spsp/routes.ts
mkurapov: I wouldn't mind keeping `ServiceDependencies` for consistency IMO

packages/backend/src/open_payments/connection/middleware.ts
```diff
@@ -0,0 +1,23 @@
+import { AppContext } from '../../app'
+import { IncomingPayment } from '../payment/incoming/model'
+
+export interface ConnectionContext extends AppContext {
+  incomingPayment: IncomingPayment
+}
+
+export const connectionMiddleware = async (
```
mkurapov, line 8: 🤔 
since we would only need this if we are dealing with `application/spsp4+json`, maybe if we had the SPSP middleware in a function (ie the comment above), we could just do this incoming payment lookup directly in that method, within the `if (application/spsp4+json)` block?

mkurapov (dismissed):
Looks good, a few considerations

packages/backend/src/open_payments/connection/middleware.ts
wilsonianb: `ctx.incomingPayment` is also used in the (non-SPSP) `ConnectionRoutes.get`

mkurapov (approved):

PR #1056, opened 2023-02-01
chore(backend): make toOpenPaymentType sync

mkurapov:

- The sync version of https://github.com/interledger/rafiki/pull/1024
- https://github.com/interledger/rafiki/pull/1024#discussion_r1091032439

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -75,7 +76,11 @@ async function getIncomingPayment(
   }
   if (!incomingPayment) return ctx.throw(404)
   const connection = deps.connectionService.get(incomingPayment)
-  ctx.body = await incomingPaymentToBody(incomingPayment, connection)
+  ctx.body = incomingPaymentToBody(
```
wilsonianb, line 79: Could these `*toBody` calls be replaced with calling `*.toOpenPaymentsType` directly?

wilsonianb: :thinking: 
...or we do the `*withPaymentPointer` subresource type idea and make `*toBody` the limited places where we're doing the `*.paymentPointer = ctx.paymentPointer` assignment.

wilsonianb (commented):
I think I like it.
I bet this would even let us remove some `.withGraphFetched['paymentPointer']`s in subresource service queries.

packages/backend/src/open_payments/payment/incoming/routes.ts
mkurapov: >  I think I like it.
>
> I bet this would even let us remove some .withGraphFetched['paymentPointer']s in subresource service queries.

It's a little more straightforward than the alternatives, and will indeed allow us to  remove the `withGraphFetched` calls. I can do that as a separate follow-up PR

wilsonianb: I don't feeling too strongly about `toBody` or `*withPaymentPointer`.
The latter would let `getUrl` be a `url` getter but it'd probably only be called within `toOpenPaymentType`.

wilsonianb: I guess `url` getter would also be used in tests...

wilsonianb (approved):
:+1: 
if you're wanting to merge this into your OG pr

Issue #1016, opened 2023-01-23
add postgres as an option for balance tracking


matdehaast:
There is a need to have Rafiki configurable to allow just using the Postgres datastore to track the balances.

* Configurable on startup
* Use an interface that can either use TB or Postgres

Issues:
- [x] #1086
- [x] #1087
- [x] #1088
- [x] #1089
- [x] #1090
- [x] #1091
- [x] #1092
- [x] #1093

wilsonianb:
There should be benefits to using postgres in the majority of `backend` tests (including being able to truncate)

wilsonianb:
I'm wondering if this could also replace the connector existing
https://github.com/interledger/rafiki/blob/d05a80d6bbb268a5ec65dad06d79583fba55b785/packages/backend/src/connector/core/test/mocks/accounting-service.ts#L40

wilsonianb:
The connector could benefit from support for single phase transfers
https://github.com/interledger/rafiki/blob/f7017172b576185b37e885c513555c5704a24ceb/packages/backend/src/connector/core/middleware/balance.ts#L78-L79

Not sure if it makes to include as a part of this. Otherwise we should belatedly open an issue for it.

PR #1057, opened 2023-02-01
1016/mk/psql accounting

mkurapov:

- Pulling tigerbeetle `accountingService` into own `accounting/tigerbeetle` directory, (while keeping `AccountingService` interface basically the same)
- Making `LiquidityAccountType` enum, allowing less coupling to Tigerbeetle account codes
- Making errors thrown from the tigerbeetle `accountingService` more generic, so the PSQL version can also throw the same errors (without coupling to the Tigerbeetle errors)

## Context

- Progress towards #1016 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -39,7 +40,7 @@ export async function createAccounts(
       ledger: account.ledger,
       code: account.code,
       flags:
-        account.type === AccountType.Debit
+        account.type === TigerbeetleAccountType.Debit
```
wilsonianb, line 43: I wonder if we can remove `TigerbeetleAccountType` and instead use the `AccountType` from `accounting/service` and do a settlement or liquidity check here

packages/backend/src/accounting/utils.ts
```diff
@@ -1,21 +1,5 @@
...
-  if (!validateId(id)) {
-    throw new Error()
-  }
-
-  return uuidToBigInt(id)
-}
-
 export function uuidToBigInt(id: string): bigint {
   return BigInt(`0x${id.replace(/-/g, '')}`)
 }
```
wilsonianb, line 5: Why not move all of this to `accounting/tigerbeetle/utils`?

packages/backend/src/accounting/service.ts
```diff
@@ -1,45 +1,16 @@
...
-  LiquidityPeer = 3,
-  LiquidityIncoming = 4,
-  LiquidityOutgoing = 5,
-  Settlement = 101
+export enum AccountType {
+  LIQUIDITY = 'LIQUIDITY',
+  LIQUIDITY_ASSET = 'LIQUIDITY_ASSET',
+  LIQUIDITY_PEER = 'LIQUIDITY_PEER',
+  LIQUIDITY_INCOMING = 'LIQUIDITY_INCOMING',
+  LIQUIDITY_OUTGOING = 'LIQUIDITY_OUTGOING',
```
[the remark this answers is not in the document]
mkurapov: What if we have 
```
LIQUIDITY_PAYMENT
LIQUIDITY_ASSET
LIQUIDITY_PEER
LIQUIDITY_WEBMON
SETTLEMENT
```

Would this also make sense to remove `LIQUIDITY_` though?

I do like the idea of having `accountType` required in the `createLiquidityAccount` call in general, however.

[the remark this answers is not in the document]
wilsonianb: I'd lean towards `*OUTGOING` & `*INCOMING` over `*PAYMENT`

re: `LIQUIDITY_`
I just prefer `LiquidityAccountType.ASSET` over `LiquidityAccountType.LIQUIDITY_ASSET`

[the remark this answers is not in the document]
mkurapov: makes sense. I was thinking we can keep `LIQUIDITY_ ` on the `AccountType`, because the `LiquidityAccountType` is just a subset of `AccountType` so you can't pass in `AccountType.SETTLEMENT` for `createLiquidityAccount`. 

However, I can just make a separate `AccountType` on the `LedgerAccount` Model _only_, and cast it from `LiquidityAccountType` if necessary

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -38,8 +27,13 @@ export async function createAccounts(
...
       ledger: account.ledger,
       code: account.code,
+      // Credit and debit accounts can both send and receive
+      // but are restricted by their respective Tigerbeetle flags.
+      // In Rafiki transfers:
+      // - the source account's debits increase
+      // - the destination account's credits increase
       flags:
-        account.type === AccountType.Debit
+        account.code === TigerbeetleAccountCode.SETTLEMENT
```
mkurapov, line 36: @wilsonianb I think this is what you were referring to

packages/backend/src/connector/core/middleware/account.ts
```diff
@@ -73,7 +75,10 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
         )
         if (paymentPointer) {
           if (!paymentPointer.totalEventsAmount) {
-            await createLiquidityAccount(paymentPointer)
+            await createLiquidityAccount(
+              paymentPointer,
+              LiquidityAccountType.WEB_MONETIZATION
```
mkurapov, line 80: I believe `WEB_MONETIZATION` is the correct one here

packages/backend/src/tests/accountFactory.ts
```diff
@@ -37,13 +42,19 @@ export class AccountFactory {
...
+      )
     }
     const account = {
       id: options.id || uuid(),
       asset
     }
-    await this.accounts.createLiquidityAccount(account)
+    await this.accounts.createLiquidityAccount(
+      account,
+      options?.type || LiquidityAccountType.INCOMING
```
mkurapov, line 56: default for the test factory, thought this was ok for now

wilsonianb (approved):

PR #1061, opened 2023-02-02
662/mk/use open payments types auth

### LINKED ISSUE #662: Use generated Open Payments OpenAPI types

### LINKED ISSUE #973: Update `accessToBody`
https://github.com/interledger/rafiki/pull/960#discussion_r1068819370

To something like
```
export const accessToBody = (access: Access): ActiveTokenInfo['access'] => ({
  type: access.type,
  actions: access.actions,
  identifier: access.identifier,
  limits: access.limits
})
```
Hopefully the types will be easier after :point_down: is merged:
- https://github.com/interledger/rafiki/pull/956

mkurapov:

- Adding `toOpenPaymentsType` functions  to `auth` pacakge to convert from Model types to `open-payments` types
- Updating logic in some services & routes
- Fixing typing issues in tests

## Context

- Officially fixes #662  (recently merged PR was updating the types in `backend`)
- Also fixes #973 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/access/model.ts
```diff
@@ -28,3 +32,15 @@ export class Access extends BaseModel {
...
+
+export function toOpenPaymentsAccess(
+  accessItem: Access
+): OpenPaymentsAccessItem {
+  return {
+    actions: accessItem.actions,
+    identifier: accessItem.identifier as string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    type: accessItem.type as any,
+    limits: accessItem.limits ?? undefined
```
mkurapov, line 44: Should we include this?

packages/auth/src/access/model.ts
```diff
@@ -28,3 +32,15 @@ export class Access extends BaseModel {
...
 }
+
+export function toOpenPaymentsAccess(
+  accessItem: Access
+): OpenPaymentsAccessItem {
+  return {
+    actions: accessItem.actions,
+    identifier: accessItem.identifier ?? undefined,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    type: accessItem.type as any,
```
mkurapov, line 43: This was a bit of a tough one. Without `any` it complains about including `identifier`, since it only exists on the `outgoing-payments` access item type:
Screenshot 2023-02-02 at 7 44 54 PM

By checking whether or not identifier exists, it complains about `limits`:

Screenshot 2023-02-02 at 7 49 02 PM
Screenshot 2023-02-02 at 7 49 42 PM

Since we don't need to do any operations after `toOpenPaymentsAccess`, I left it as `any`.

packages/auth/src/accessToken/routes.ts
```diff
@@ -1,26 +1,27 @@
...
 import { IAppConfig } from '../config/app'
 import { AccessTokenService } from './service'
-import { accessToBody } from '../shared/utils'
 import { ClientService } from '../client/service'
 import { Grant } from '../grant/model'
+import { toOpenPaymentsAccessToken } from './model'
+import { AccessService } from '../access/service'
 
-type TokenRequest<BodyT> = Omit<AppContext['request'], 'body'> & {
+type TokenRequest<BodyT> = Exclude<AppContext['request'], 'body'> & {
```
mkurapov, line 12: Similar to the `backend` package, using `Exclude` instead of `Omit` properly gives us typing for any of the context types that extended `AppContext`

packages/auth/src/accessToken/routes.ts
```diff
@@ -102,18 +102,26 @@ async function rotateToken(
...
+  try {
+    newToken = await deps.accessTokenService.rotate(managementId, token)
+  } catch (error) {
+    const errorMessage = 'Could not rotate token'
+    deps.logger.error({ error: error && error['message '] }, errorMessage)
+    ctx.throw(400, { message: errorMessage })
+  }
+
+  if (!newToken) {
+    ctx.throw(404, { message: 'Token not found' })
```
mkurapov, line 116: because we use `Extends<AppContext>` now, we properly have the type on `ctx` and it will know that `ctx.throw` has a `never` return type, and as a result, will flag any code under `ctx.throw` as unreachable

packages/auth/src/accessToken/service.ts
```diff
@@ -7,15 +7,17 @@ import { Grant, GrantState } from '../grant/model'
...
   get(token: string): Promise<AccessToken | undefined>
   getByManagementId(managementId: string): Promise<AccessToken | undefined>
   introspect(token: string): Promise<Grant | undefined>
   revoke(id: string, tokenValue: string): Promise<void>
   create(grantId: string, opts?: AccessTokenOpts): Promise<AccessToken>
-  rotate(managementId: string, tokenValue: string): Promise<Rotation>
+  rotate(
+    managementId: string,
+    tokenValue: string
+  ): Promise<AccessToken | undefined>
```
mkurapov, line 20: Allows this service to exclusively deal with `AccessToken` types

packages/auth/src/grant/model.ts
```diff
@@ -61,6 +66,70 @@ export class Grant extends BaseModel {
...
+      authServerUrl: args.authServerUrl
+    }),
+    continue: {
+      access_token: {
+        value: grant.continueToken
+      },
+      uri: `${args.authServerUrl}/continue/${grant.continueId}`
+    }
+  }
+}
```
mkurapov, line 131: I decided to not call this `toOpenPaymentsNonInteractiveGrant`, as I _think_ I like having just `Grant` or `InteractiveGrant`, thoughts? 

Depending on the decision, I will also update the exported types out of `open-payments`

packages/auth/src/grant/model.ts
```diff
@@ -61,6 +66,70 @@ export class Grant extends BaseModel {
...
+interface ToOpenPaymentsGrantArgs {
+  authServerUrl: string
+}
+
+export function toOpenPaymentsGrant(
+  grant: Grant,
+  args: ToOpenPaymentsGrantArgs,
+  accessToken: AccessToken,
+  accessItems: Access[]
+): OpenPaymentsGrant {
```
mkurapov, line 119: I kind of like the approach of having this as a standalone function that isn't a method on the `XModel` itself, but takes in a `Model` instance itself. It felt better to do this because of the amount of different arguments we pass in to construct the resulting OpenPayments object. NABD, but would be curious to hear opinion on this.

I could update the `backend` `toOpenPaymentsX` methods as well to this

packages/auth/src/grant/model.ts
```diff
@@ -61,6 +66,70 @@ export class Grant extends BaseModel {
...
+      wait: waitTimeSeconds
+    }
+  }
+}
+
+interface ToOpenPaymentsGrantArgs {
+  authServerUrl: string
+}
+
+export function toOpenPaymentsNonInteractiveGrant(
```
mkurapov, line 114: I'm not sure I'm fully convinced of having `NonInteractiveGrant` and `InteractiveGrant`.
Maybe we could just have `Grant` and `InteractiveGrant`, it still could be clear but less wordy...thoughts?

packages/open-payments/src/types.ts
```diff
@@ -111,13 +116,18 @@ export type AccessAction = (
...
+} = {
   IncomingPayment: 'incoming-payment',
   OutgoingPayment: 'outgoing-payment',
   Quote: 'quote'
-})
+}
 
-export const AccessAction: Record<string, AccessAction> = Object.freeze({
+export const AccessAction: Record<
+  'Create' | 'Read' | 'ReadAll' | 'Complete' | 'List' | 'ListAll',
```
mkurapov, line 128: So the keys are typed

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -25,7 +25,7 @@ describe('Access Token Routes', (): void => {
   let accessTokenRoutes: AccessTokenRoutes
 
   beforeAll(async (): Promise<void> => {
-    deps = await initIocContainer(Config)
+    deps = initIocContainer(Config)
```
wilsonianb, line 28: :eyes: are we pointlessly `await`ing this _everywhere_?

packages/auth/src/accessToken/service.test.ts
```diff
@@ -116,6 +117,7 @@ describe('Access Token Service', (): void => {
 
     test('Can get an access token by its value', async (): Promise<void> => {
       const fetchedToken = await accessTokenService.get(accessToken.value)
+      assert.ok(fetchedToken)
       expect(fetchedToken.value).toEqual(accessToken.value)
       expect(fetchedToken.managementId).toEqual(accessToken.managementId)
       expect(fetchedToken.grantId).toEqual(accessToken.grantId)
```
wilsonianb, line 123:
```suggestion
      await expect(accessTokenService.get(accessToken.value)).resolves.toEqual(
        accessToken
      )
```

packages/auth/src/accessToken/service.test.ts
```diff
@@ -251,29 +254,29 @@ describe('Access Token Service', (): void => {
...
+        managementId: result.managementId
       })
-      expect(rotatedToken).toBeDefined()
+      assert.ok(rotatedToken)
       expect(rotatedToken?.value).not.toBe(originalTokenValue)
     })
     test('Cannot rotate token with incorrect management id', async (): Promise<void> => {
       const result = await accessTokenService.rotate(v4(), token.value)
-      expect(result.success).toBe(false)
+      expect(result).toBeUndefined()
```
wilsonianb, line 275:
```
      await expect(
        accessTokenService.rotate(v4(), token.value)
      ).resolves.toBeUndefined()
```

packages/auth/src/grant/model.ts
wilsonianb: We could be inconsistent with everywhere else in rafiki and have `Grant` and `PendingGrant`
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-1.5

mkurapov: I wouldn't mind updating to `Grant` and `PendingGrant` everywhere

packages/auth/src/accessToken/routes.test.ts
mkurapov: yes, but doesn't really impact anything

packages/auth/src/grant/model.ts
mkurapov: @wilsonianb i'm kind of leaning toward using this method in `backend` too 😬

mkurapov:
Follow-up PR will be
- Converting `NonInteractiveGrant` to `PendingGrant`
- Possibly `backend` changes re: https://github.com/interledger/rafiki/pull/1061#discussion_r1095135540

packages/auth/src/grant/model.ts
wilsonianb: I agree it makes sense in `auth`
I'm still pondering `backend`.

packages/auth/src/access/model.ts
wilsonianb: Yes. If an outgoing payment access has `limits` it should be included.

packages/auth/src/accessToken/service.test.ts
```diff
@@ -251,29 +248,31 @@ describe('Access Token Service', (): void => {
         token.managementId,
         token.value
       )
-      expect(result.success).toBe(true)
-      expect(result.success && result.value).not.toBe(originalTokenValue)
+      assert.ok(result)
+      expect(result.value).not.toBe(originalTokenValue)
```
wilsonianb, line 252:
```suggestion
      expect(result.value).not.toBe(originalTokenValue)
      await expect(accessTokenService.get(token)).resolves.toBeUndefined()
```
Not sure if it's worth also checking `getByManagementId`, `introspect`, `revoke`, or `rotate`

packages/auth/src/grant/routes.test.ts
```diff
@@ -427,14 +439,16 @@ describe('Grant Routes', (): void => {
 
         ctx.request.body = {}
 
-        await expect(grantRoutes.continue(ctx)).rejects.toMatchObject({
+        await expect(
+          grantRoutes.continue(ctx as ContinueContext)
```
wilsonianb, line 443: Why not `const ctx = createContext<ContinueContext>(` above?

packages/auth/src/accessToken/routes.ts
```diff
@@ -102,18 +102,26 @@ async function rotateToken(
...
+  }
+
+  const accessItems = await deps.accessService.getByGrant(newToken.grantId)
```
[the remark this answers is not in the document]
mkurapov: We could even lookup `accessItems` first to check if they are non-empty, before rotating, thoughts?

packages/auth/src/accessToken/service.test.ts
mkurapov: Do you mean checking whether the old one can be fetched by `getByManagementId`/`introspect`/`revoke`/`rotate` ?

packages/auth/src/accessToken/routes.ts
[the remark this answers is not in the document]
wilsonianb: I would only expect `accessItems` to be empty if the grant was deleted.
I want to say we should lock the grant here, but we'd need to at least look up the access token first (given we're starting with `managementId` + `token`).

packages/auth/src/accessToken/service.test.ts
wilsonianb: correct

wilsonianb: Alternatively, each of those look up functions could have a `cannot _____  rotated token` test

packages/auth/src/accessToken/routes.ts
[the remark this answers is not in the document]
wilsonianb: yeah, that's the best steps I can think of for locking here

packages/auth/src/accessToken/service.ts
```diff
@@ -140,45 +129,37 @@ async function rotate(
...
     }
-  }
+
+    // lock grant
+    await deps
+      .knex<Grant>('grants')
+      .select()
+      .where('id', oldToken.grantId)
+      .forNoKeyUpdate()
+      .timeout(5000)
```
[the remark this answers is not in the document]
wilsonianb: Would this help with the grant being deleted between generating the new token and querying the the access items?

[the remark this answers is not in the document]
mkurapov: Got it now - I see what you mean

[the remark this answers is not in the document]
mkurapov: Let me take care of this in a follow-up PR

wilsonianb (approved):

PR #1073, opened 2023-02-07
chore(auth): use trx for token rotation

mkurapov:

- Wrap token rotation in transaction, lock grant during it
- Clean up `auth` test logging:

Before:

Screenshot 2023-02-07 at 2 27 25 PM

After:
Screenshot 2023-02-07 at 2 38 42 PM

## Context

- Follow up to https://github.com/interledger/rafiki/pull/1061#issuecomment-1416268765
## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/access/service.test.ts
```diff
@@ -49,56 +50,87 @@ describe('Access Service', (): void => {
...
-    const access = await accessService.createAccess(grant.id, [
-      outgoingPaymentAccess
-    ])
-
-    expect(access.length).toEqual(1)
-    expect(access[0].grantId).toEqual(grant.id)
-    expect(access[0].type).toEqual(AccessType.OutgoingPayment)
-    expect(access[0].limits).toEqual(outgoingPaymentLimit)
+  describe('getByGrant', (): void => {
+    test('gets access', async () => {
```
mkurapov, line 109: added new test

packages/auth/src/accessToken/routes.ts
```diff
@@ -101,23 +103,36 @@ async function rotateToken(
...
     const errorMessage = 'Could not rotate token'
     deps.logger.error({ error: error && error['message'] }, errorMessage)
-    ctx.throw(400, { message: errorMessage })
-  }
-
-  if (!newToken) {
-    ctx.throw(404, { message: 'Token not found' })
+    ctx.throw((error && error['status']) ?? 400, {
+      message: (error && error['message']) ?? errorMessage
+    })
```
mkurapov, line 133: Handling the L121 case this way

packages/auth/src/access/service.test.ts
```diff
@@ -21,9 +22,9 @@ describe('Access Service', (): void => {
   let trx: Knex.Transaction
 
   beforeAll(async (): Promise<void> => {
-    deps = await initIocContainer(Config)
-    accessService = await deps.use('accessService')
+    deps = initIocContainer(Config)
     appContainer = await createTestApp(deps)
```
mkurapov, line 26: order matters: if resolving a dependency before `createTestApp`, if will use a normal logger instead of a test logger, since `createTestApp` overrides the logger.

packages/auth/src/grant/service.ts
```diff
@@ -244,3 +250,16 @@ async function getGrantsPage(
...
+  _: ServiceDependencies,
+  grantId: string,
+  trx: Transaction,
+  timeoutMs?: number
+): Promise<void> {
+  await trx<Grant>('grants')
+    .select()
+    .where('id', grantId)
+    .forNoKeyUpdate()
+    .timeout(timeoutMs ?? 5000)
```
mkurapov, line 264: https://github.com/Vincit/objection.js/pull/2319 🙏

packages/auth/src/index.ts
```diff
@@ -47,6 +47,20 @@ export function initIocContainer(
...
+        },
+        error(message) {
+          logger.error(message)
+        },
+        deprecate(message) {
+          logger.warn(message)
+        },
+        debug(message) {
+          logger.debug(message)
+        }
```
mkurapov, line 63: use our `pino` logger for knex messages

packages/auth/src/grant/service.ts
```diff
@@ -244,3 +250,16 @@ async function getGrantsPage(
...
+
+async function lock(
+  _: ServiceDependencies,
+  grantId: string,
+  trx: Transaction,
+  timeoutMs?: number
+): Promise<void> {
+  await trx<Grant>('grants')
+    .select()
+    .where('id', grantId)
```
wilsonianb, line 262: What if this grant doesn't exist?

packages/auth/src/accessToken/routes.ts
```diff
@@ -101,22 +105,37 @@ async function rotateToken(
...
 
-  if (!newToken) {
-    ctx.throw(404, { message: 'Token not found' })
-  }
+    if (!newToken) {
+      ctx.throw(404, { message: 'Token not found' })
+    }
+
+    await deps.grantService.lock(newToken.grantId, trx)
```
wilsonianb, line 126: :thinking: does this need to happen before `rotate`?
(it'd require an initial `accessTokenService.get` or `getByManagementId` to get the `grantId`)

wilsonianb (commented):
Would it be easier to just add a `withGraphFetched` with everything in `rotate`?
https://github.com/interledger/rafiki/blob/6a709dc92e8bc48ed5e173ac9fa0d908019cbab3/packages/auth/src/accessToken/service.ts#L150-L155

packages/auth/src/accessToken/routes.ts
mkurapov: I wonder if we should `accessToken` with its `grant` it in the context via `tokenHttpsigMiddleware`:

https://github.com/interledger/rafiki/blob/6a709dc92e8bc48ed5e173ac9fa0d908019cbab3/packages/auth/src/signature/middleware.ts#L110-L144

wilsonianb: yes :rocket:

mkurapov: In general, do you think it's worth it to fetch the grantId again after first checking it in the middleware?

mkurapov: In that case, since we will have the `grantId`, first thing would be to lock, then rotate (I don't mind just keeping the `accessItems = await deps.accessService.getByGrant(newToken.grantId, trx)` lookup)

mkurapov: > Didn't refresh before commenting again
I did the same again 😆

mkurapov: > Could rotate even be replaced with a delete and create?

I like this

packages/auth/src/accessToken/service.ts
```diff
@@ -127,31 +139,19 @@ async function createAccessToken(
...
-  })
+  const isRevoked = await revoke(deps, args)
+
+  if (!isRevoked) {
+    deps.logger.warn(
+      { managementId, tokenValue, grantId },
+      'Could not revoke access token'
+    )
+    return undefined
+  }
```
mkurapov, line 154: Should we stop if we can't revoke the old token?

packages/auth/src/accessToken/routes.ts
```diff
@@ -90,34 +100,47 @@ async function revokeToken(
...
   ctx: RevokeContext
 ): Promise<void> {
-  const token = (ctx.headers['authorization'] ?? '').replace('GNAP ', '')
-  const { id: managementId } = ctx.params
-  await deps.accessTokenService.revoke(managementId, token)
+  await deps.accessTokenService.revoke({
+    managementId: ctx.accessToken.managementId,
+    tokenValue: ctx.accessToken.value
+  })
+
```
mkurapov, line 107: Do we want to actually check if we weren't able to revoke the token here? I'm guessing, like the old behaviour we should continue returning 204 since it could have been revoked previously for example

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -200,86 +206,34 @@ describe('Access Token Routes', (): void => {
...
       grant = await Grant.query(trx).insertAndFetch(BASE_GRANT)
       token = await AccessToken.query(trx).insertAndFetch({
         grantId: grant.id,
         ...BASE_TOKEN
       })
-      managementId = token.managementId
-      url = `/token/${managementId}`
-    })
-
-    test('Returns status 204 even if management id does not exist', async (): Promise<void> => {
```
mkurapov, line 218: Not relevant, we have the token in the context

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -303,63 +256,13 @@ describe('Access Token Routes', (): void => {
...
-        { id: managementId }
-      )
-
-      await expect(accessTokenRoutes.rotate(ctx)).rejects.toMatchObject({
-        status: 404,
-        message: 'Token not found'
-      })
-    })
-
-    test('Cannot rotate nonexistent token', async (): Promise<void> => {
```
mkurapov, line 332: Can't happen, we already have it in the ctx

packages/auth/src/accessToken/routes.ts
sabineschaller: We only would not have been able to revoke the token if we couldn't find it, right? In that case, it either never existed or was revoked a while ago and cleaned up. I think you are right, we should always return the 204 and I don't think what value it adds saying "token was revoked before / never existed"

packages/auth/src/accessToken/service.test.ts
```diff
@@ -1,7 +1,7 @@
 import { faker } from '@faker-js/faker'
 import nock from 'nock'
 import { Knex } from 'knex'
-import { v4 } from 'uuid'
+import { v4 as uuid } from 'uuid'
```
sabineschaller, line 4: Consistency matters! 😂

packages/auth/src/accessToken/service.ts
sabineschaller: Agreed

packages/auth/src/accessToken/service.test.ts
mkurapov: I regretted this change as soon as I saw it in the diff 😆

packages/auth/src/grant/service.ts
mkurapov: ... we log it out 😀

packages/auth/src/accessToken/service.ts
```diff
@@ -8,17 +8,25 @@ import { ClientService } from '../client/service'
...
 }
 
 export interface AccessTokenService {
   get(token: string): Promise<AccessToken | undefined>
   getByManagementId(managementId: string): Promise<AccessToken | undefined>
   introspect(token: string): Promise<Grant | undefined>
-  revoke(id: string, tokenValue: string): Promise<void>
+  revoke(args: RevokeTokenArgs): Promise<boolean>
   create(grantId: string, opts?: AccessTokenOpts): Promise<AccessToken>
   rotate(args: RotateTokenArgs): Promise<AccessToken | undefined>
```
wilsonianb, line 31:
```suggestion
  revoke(id: string, trx?: TransactionOrKnex): Promise<boolean>
  create(grantId: string, opts?: AccessTokenOpts): Promise<AccessToken>
  rotate(id: string, trx?: TransactionOrKnex): Promise<AccessToken | undefined>
```
We have the tokens

mkurapov: I still needed `grantId` for the `rotate` but made the suggested changed

mkurapov: I still needed `grantId` for the `rotate` but made the suggested changed

packages/auth/src/accessToken/service.ts
```diff
@@ -8,15 +8,26 @@ import { ClientService } from '../client/service'
...
 
+interface RotateTokenArgs {
+  id: string
+  grantId: string
+  expiresIn?: number
+}
+
+interface CreateTokenArgs {
+  grantId: string
+  expiresIn?: number
```
wilsonianb, line 19: I don't think `expiresIn` needs to be specified and we can use `deps.config.accessTokenExpirySeconds` internally for both of these.
(You could also replace `config` in `ServiceDependencies` with `expirySeconds`)

packages/auth/src/accessToken/service.ts
```diff
@@ -102,56 +107,42 @@ async function introspect(
...
-      managementId: id,
-      value: tokenValue
-    })
+  trx?: TransactionOrKnex
+): Promise<boolean> {
+  const deletedCount = await AccessToken.query(trx || deps.knex)
+    .findById(id)
     .delete()
+
+  return deletedCount === 1
```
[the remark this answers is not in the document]
mkurapov: 🧠

wilsonianb (approved):

PR #1065, opened 2023-02-03
feat(backend): adding migrations for new tables

mkurapov:
## Changes proposed in this pull request

- Adding two new tables (migrations) for #1016 

| ***LedgerAccounts*** |                                 |                                  |
|----------------------|---------------------------------|----------------------------------|
| **Field**            | **Data type**                   | **Comments**                     |
| id                   | uuid                            |                                  |
| accountRef           | uuid                            | Used as the external tracking id |
| ledger              | smallint                            | FK on Assets.ledger                  |
| type                 | LIQUIDITY, LIQUIDITY_ASSET...   |                                  |
| createdAt            | Date                            |                                  |
| updatedAt            | Date                            |                                  |

| ***LedgerTransfers*** |                                    |                                  |
|-----------------------|------------------------------------|----------------------------------|
| **Field**             | **Data type**                      | **Comments**                     |
| id                    | uuid                               |                                  |
| transferRef           | uuid                               | Used as the external tracking id |
| ledger               | smallint                               | FK on Assets.ledger                  |
| debitAccountId        | uuid                               | FK on LedgerAccounts.id          |
| creditAccountId       | uuid                               | FK on LedgerAccounts.id          |
| amount                | bigint                             | Positive amount only             |
| state                 | PENDING, POSTED, VOIDED   |                                  |
| expiresAt             | Date                               |                                  |
| createdAt             | Date                               |                                  |
| updatedAt             | Date                               |                                  |

Notes:
- Balances calculated via 
```
(SELECT SUM(amount) FROM LedgerTransfers where LedgerAccount.id = LedgerTransfers.creditAccountId)  -
(SELECT SUM(amount) FROM LedgerTransfers where LedgerAccount.id = LedgerTransfers.debitAccountId)
```
No flags like `credits_must_exceed_debits` for now
- The `ids` of `ledgerAccounts` and `ledgerTransfers` are only used internally, between these two tables. This means when creating accounts for `incomingPayment` or `outgoingPayment`s, the `incoming/outgoingPayment.ids` will be actually the `accountRef`s instead. LIkewise for `LedgerTransfer.transferRef`, these will be the `WithdrawalId/DepositId`. ie `transferRef` is the caller-assigned reference, while id is just used internally

## Context

- Fixes #1087 

## Checklist

- [x] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass

PR #1079, opened 2023-02-07
1016/mk/psql accounting 3

### LINKED ISSUE #1088: Decouple accountingService account creation from Tigerbeetle
Currently, `accountingService.createSettlementAccount` method takes in a `ledger: number`, however, based on the psql accounting table structure (in #1087), we need to have account identifiers be _only_ strings

mkurapov:

- Updating `accountingService.createSettlementAccount` method to take in a `SettlementAccount` instead of a `ledger: number`, to decouple from Tigerbeetle's `ledger`. The new PSQL accounting service will keep track settlement accounts by `accountRef: asset.id, type: AccountType.SETTLEMENT` instead. More context [here](https://github.com/interledger/rafiki/pull/1065#discussion_r1096202049).
- Updating `accountingService.getSettlementAccount` method to take in a `id: string` instead of `ledger: number`. This allows us to look up the settlement account by a string id/accountRef for the PSQL accounting service and `ledger.toString()` in the Tigerbeetle accounting service.

## Context

- fixes #1088

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/tigerbeetle/utils.ts
```diff
@@ -9,6 +9,9 @@ export function toTigerbeetleId(id: AccountId): bigint {
   if (typeof id === 'bigint') {
     return id
   }
+  if (typeof id === 'string' && !isNaN(Number(id))) {
+    return BigInt(id)
+  }
```
mkurapov, line 14: Allows numerical, non-UUID strings to be used as an accountId. Basically, this allows `asset.ledger` to be passed in as an id

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -29,10 +30,10 @@ describe('Accounting Service', (): void => {
...
   const timeout = BigInt(10_000) // 10 seconds
 
-  let ledger = 1
-  function newLedger() {
-    return ledger++
-  }
+  const newLedger = (() => {
+    let id = 1
+    return () => id++
+  })()
```
mkurapov, line 36: so there is no `ledger` variable hanging out

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -29,10 +30,10 @@ describe('Accounting Service', (): void => {
...
   const timeout = BigInt(10_000) // 10 seconds
 
-  let ledger = 1
-  function newLedger() {
-    return ledger++
-  }
+  const newLedger = (() => {
+    let i = 1
+    return () => i++
+  })()
```
mkurapov, line 36: so there is no `ledger` variable hanging out

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -245,32 +246,45 @@ describe('Accounting Service', (): void => {
...
+        asset: {
+          id: uuid(),
+          ledger: newLedger()
+        }
+      }
+
+      await accountingService.createSettlementAccount(account)
       await expect(
-        accountingService.getSettlementBalance(ledger)
+        accountingService.getSettlementBalance(account.asset.ledger.toFixed())
```
sabineschaller, line 281:
```suggestion
        accountingService.getSettlementBalance(account.asset.ledger.toString())
```

packages/backend/src/accounting/service.ts
```diff
@@ -8,6 +8,14 @@ export enum LiquidityAccountType {
   WEB_MONETIZATION = 'WEB_MONETIZATION'
 }
 
+export interface SettlementAccount {
+  id: string
+  asset: {
+    id: string
+    ledger: number
+  }
+}
```
sabineschaller, line 17: Why are there two `id`s and isn't the id not just the ledger as string?

mkurapov: 🤔 yeah I think that's just better & simpler. I think originally I liked that [this](https://github.com/interledger/rafiki/pull/1095/files#diff-db7f903acbbd222c73bf85a026fc87e5dfb271fede799b42a5a163e8bc62fd29R69-R73) was explicit in showing that accountRef = assetId for settlement accounts, but it'll be just as clear on the caller side with a single `id: string` argument
```
  await deps.ledgerAccountService.create({
    accountRef: account.asset.id,
    assetId: account.asset.id,
    type: LedgerAccountType.SETTLEMENT
  })
```

mkurapov: Oh, I know why now, one moment

mkurapov: I think I prefer `getSettlementAccount` taking in a `SettlementAccount` argument, thoughts?

packages/backend/src/accounting/tigerbeetle/service.ts
```diff
@@ -118,24 +120,26 @@ export async function createLiquidityAccount(
...
-  ledger: number
-): Promise<void> {
+  account: SettlementAccount
+): Promise<SettlementAccount> {
   try {
     await createAccounts(deps, [
       {
-        id: ledger,
-        ledger,
+        id: account.asset.ledger.toString(),
```
wilsonianb, line 128: We're now going number -> string -> bigint
How bad would it be to use `AccountId` in the base `AccountService? :grimacing: :grimacing: :grimacing:
https://github.com/interledger/rafiki/blob/d8800feba065bcd875db363d1d1472f146f08c4c/packages/backend/src/accounting/tigerbeetle/utils.ts#L3

mkurapov: 🤔 hMMM I might be complicating it. Maybe I can just store `ledger` on the `LedgerAccount` and then do an `Access` lookup to get the information I need without changing the function signature

mkurapov:
Yep, I overthought this. Going to keep methods the same, but update the structure of `LedgerAccounts` table. Going to store `ledger` instead of `assetId`, and whenever I need to `createSettlementAccount/getSettlementAmount`, I'll just do a lookup for the actual `Asset` via the `ledger` field.

@wilsonianb @sabineschaller

PR #1095, opened 2023-02-09
Account creation for psql accounting

### LINKED ISSUE #1089: Implement account creation for PSQL accounting
Implement `createLiquidityAccount` and `createSettlement` methods in the `psqlAccountingService`

mkurapov:

- Add `USE_TIGERBEETLE` config flag, instantiate different accountingService based on it
- Create `ledgerAccount` model and service
- Implement `accountingService.createLiquidityAccount` and `createSettlementAccount` methods

## Context

- Fixes #1089 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/index.ts
```diff
@@ -155,16 +157,39 @@ export function initIocContainer(
       accountingService: await deps.use('accountingService')
     })
   })
+
+  container.singleton('ledgerAccountService', async (deps) => {
```
mkurapov, line 161: I was thinking of registering this singleton conditionally after L185, but it felt wrong to have an optional singleton resolution

packages/backend/migrations/20230201133042_create_ledger_accounts_table.js
```diff
@@ -5,7 +5,7 @@
 exports.up = function (knex) {
   return knex.schema.createTable('ledgerAccounts', function (table) {
     table.uuid('id').notNullable().primary()
-    table.uuid('accountRef').notNullable().unique().index()
+    table.uuid('accountRef').notNullable().index()
```
mkurapov, line 8: This won't be unique because Assets will use the same `accountRef` (will equal `asset.id`), but a different `type` (`LIQUIDITY_ASSET` or `SETTLEMENT`)

packages/backend/migrations/20230201133042_create_ledger_accounts_table.js
```diff
@@ -5,7 +5,7 @@
 exports.up = function (knex) {
   return knex.schema.createTable('ledgerAccounts', function (table) {
     table.uuid('id').notNullable().primary()
-    table.uuid('accountRef').notNullable().unique().index()
```
mkurapov, line 8: This won't be unique because Assets will use the same `accountRef` (will equal `asset.id`), but a different `type` (`LIQUIDITY_ASSET` or `SETTLEMENT`). 

However, we can set a unique constraint (like below)

packages/backend/src/accounting/psql/ledger-account/model.test.ts
```diff
@@ -0,0 +1,25 @@
+import { LiquidityAccountType } from '../../service'
+import {
+  LedgerAccountType,
+  mapLiquidityAccountTypeToLedgerAccountType
+} from './model'
+
+describe('Ledger Account Model', (): void => {
+  describe('mapLiquidityAccountTypeToLedgerAccountType', (): void => {
```
mkurapov, line 8: just a sanity

packages/backend/src/config/app.ts
```diff
@@ -53,6 +58,7 @@ export const Config = {
     ? Buffer.from(process.env.STREAM_SECRET, 'base64')
     : crypto.randomBytes(32),
 
+  useTigerbeetle: envBool('USE_TIGERBEETLE', true),
```
wilsonianb, line 61: :thinking: would it be worth supporting a sanity checking mode where both are used?

wilsonianb (approved):

packages/backend/src/accounting/psql/ledger-account/model.ts
```diff
@@ -0,0 +1,47 @@
...
+  LIQUIDITY_WEB_MONETIZATION = 'LIQUIDITY_WEB_MONETIZATION',
+  SETTLEMENT = 'SETTLEMENT'
+}
+
+export class LedgerAccount extends BaseModel {
+  public static get tableName(): string {
+    return 'ledgerAccounts'
+  }
+
+  public readonly id!: string
```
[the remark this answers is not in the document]
mkurapov: Can't make it `readonly` unfortunately bc of `baseModel.$beforeInsert`, but good flag

packages/backend/src/accounting/psql/service.ts
```diff
@@ -0,0 +1,162 @@
...
+  })
+
+  return account
+}
+
+export async function createSettlementAccount(
+  deps: ServiceDependencies,
+  ledger: number
+): Promise<void> {
+  const asset = await Asset.query(deps.knex).findOne({ ledger })
```
mkurapov, line 69: I _dont_ like using the `Asset` model here directly but if I were to pull in `assetService`, it would be a circular dependency 😞 

This isn't the ideal solution, but at least allows us to not have to change any of the AccountingService function signatures & allows us to support both the psql tables and tigerbeetle service

packages/backend/src/config/app.ts
mkurapov: How would that work? Interesting idea, we could maybe turn it into a follow-up issue?

mkurapov:
Trying to figure out why the tests aren't running ...

packages/backend/src/config/app.ts
wilsonianb: Put a check in each function of the "base" account service directing to the correct db service (instead of just setting up a particular account service based on `useTigerbeetle` in index.ts)
If it's both, check the results and blow up if they don't match.

I wonder if that'd even let you reduce the two db-specific account services to more direct db interfaces (instead of the full account service). Like just have `accounts.ts` and `transfers.ts` for each of them.

mkurapov: https://github.com/interledger/rafiki/issues/1105

packages/backend/src/accounting/psql/service.ts
mkurapov: Yup, good call

wilsonianb (approved):

Issue #1091, opened 2023-02-08
Implement transfer creation for PSQL accounting


mkurapov:

wilsonianb:
I'm guessing this should support linked/atomic transfers a la tigerbeetle.
https://github.com/tigerbeetledb/tigerbeetle/blob/main/docs/reference/transfers.md#flagslinked

PR #1111, opened 2023-02-14
PSQL accounting service: createAccountDeposit

mkurapov:

- Implements `psqlAccountingService.createAccountDeposit`

## Context

- Progress towards #1091 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/psql/service.ts
```diff
@@ -211,9 +243,66 @@ export async function createTransfer(
...
+      return TransferError.TransferExists
+    }
+
+    deps.logger.error(
+      { errorMessage: error && error['message'], transferRef, accountRef },
+      'Could not create deposit'
+    )
+
+    throw error
+  }
```
mkurapov, line 305: It feels weird to not have to do any balance checks here. In tigerbeetle speak, the settlement (debit) account that ensures `credits_must_not_exceed_debits` is being debited, so the 
```
account.credits_pending + account.credits_posted + transfer.amount > account.debits_posted
```
check does not apply. Similarily, the liquidity (credit) account that ensures `debits_must_not_exceed_credits`, is being credited, so the
```
account.debits_pending + account.debits_posted + transfer.amount > account.credits_posted
``` 
check does not apply.


For now, are we just assuming the settlement account always has enough $$$ to fund deposits?

mkurapov: I also might move more of this into `ledger-transfers/index.ts` once I implement more transfer methods

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -38,3 +50,43 @@ export async function getAccountTransfers(
...
+  return {
+    amount: transfer.amount,
+    transferRef: transfer.transferRef,
+    creditAccountId: transfer.creditAccountId,
+    debitAccountId: transfer.debitAccountId,
+    ledger: transfer.ledger,
+    state:
+      transfer.timeoutMs != null
+        ? LedgerTransferState.PENDING
+        : LedgerTransferState.POSTED,
```
mkurapov, line 85: Tigerbeetle-esque code here, single phase if no timeout

packages/backend/src/accounting/psql/service.ts
wilsonianb: what if we did check balances (in `createTransfers`)?

mkurapov: Yep, we can do that, I was just trying to see once I have the other methods in what could stay in the service itself, vs go into `createTransfers`.

For this case with deposits, is
> For now, are we just assuming the settlement account always has enough $$$ to fund deposits?

the right assumption?

wilsonianb: correct

wilsonianb (dismissed):
might need to merge `main` here and the target branch
there's currently terraform stuff in the changeset

packages/backend/src/accounting/psql/service.ts
mkurapov: Balance checks moved to: https://github.com/interledger/rafiki/pull/1111/files#diff-66a9f0afadd8190ee2c83380f68156a747f881b15ec41c47dda14c229eeb26adR143-R156

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -38,3 +65,135 @@ export async function getAccountTransfers(
...
+    return {
+      results: createdTransfers,
+      errors: []
+    }
+  } catch (error) {
+    if (error instanceof UniqueViolationError) {
+      return {
+        results: [],
+        errors: [{ index: 0, error: TransferError.TransferExists }]
+      }
```
mkurapov, line 116: We _could_ add a proper index for the `TransferError.TransferExists` error, by parsing the `UniqueViolationError` details to get the `transferRef` that violated the constraint, but I felt like it's not necessary (yet)

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -38,3 +65,135 @@ export async function getAccountTransfers(
...
 }
+
+export async function createTransfers(
+  deps: ServiceDependencies,
+  transfers: CreateTransferArgs[],
+  trx?: TransactionOrKnex
+): Promise<CreateTransfersResult> {
+  const validationResults = await Promise.all(
+    transfers.map((transfer) =>
+      validateTransfer(deps, {
```
wilsonianb, line 76: I think we'd need some sort of locking to really enforce balance limits.

I'd previously had the thought that it might be worth considering storing the balance(s) on the accounts for easier lookup but it'd also make locking easier here

wilsonianb (commented):
balance limit error tests for `createTransfers` and `createDeposit`

packages/backend/src/accounting/psql/ledger-transfer/index.ts
mkurapov: Yeah I've been thinking about this, I think we will need to provide a `FOR KEY UPDATE` lock on the credit& debitAccountIds

mkurapov: I'm cautious about de-normalizing and storing balances on the accounts since having a discrepancy between the sum(transfers) vs the calculated balances is scary

mkurapov: We can always move to calculated balances with a lot of scale as well

wilsonianb: > I'm cautious about de-normalizing and storing balances on the accounts since having a discrepancy between the sum(transfers) vs the stored calculated balances is scary

:100:

mkurapov: Going to leave locking for follow-up, going to focus on getting tests right for the balances first

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -38,3 +72,156 @@ export async function getAccountTransfers(
...
+
+  if (
+    !hasEnoughLiquidity({
+      account: debitAccount,
+      balances: debitAccountBalance,
+      transferAmount: amount
+    })
+  ) {
+    return TransferError.InsufficientLiquidity
+  }
```
mkurapov, line 187: Should this just be `TransferError.InsufficientBalance`?

mkurapov: Since `InsufficientLiquidity` is more referring to the asset liquidity account?

packages/backend/src/accounting/psql/ledger-transfer/index.ts
wilsonianb: What about doing `index: -1` to maybe avoid confusion?

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -49,6 +69,322 @@ describe('Ledger Transfer', (): void => {
...
+        errors: [{ index: 0, error: TransferError.TransferExists }]
+      })
+    })
+
+    test('throws if unhandled error during insert', async (): Promise<void> => {
+      const transfer: CreateTransferArgs = {
+        ...baseTransfer
+      }
+
+      transfer.creditAccount.id = ''
```
wilsonianb, line 170: Should we be handling invalid account ids?

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -49,6 +69,322 @@ describe('Ledger Transfer', (): void => {
...
+        LedgerTransfer.query(knex).findOne({
+          transferRef: failTransfer.transferRef
+        })
+      ).resolves.toBeUndefined()
+
+      await expect(
+        LedgerTransfer.query(knex).findOne({
+          transferRef: transfer.transferRef
+        })
+      ).resolves.toBeUndefined()
```
wilsonianb, line 272: Alternatively
```suggestion
      await expect(
        getAccountTransfers(serviceDeps, account.id, knex)
      ).resolves.toEqual({
        credits: [depostTransfer],
        debits: []
      })
```
`depostTransfer` would need to be stored from the `beforeEach`

packages/backend/src/accounting/psql/ledger-transfer/index.ts
mkurapov: Yup, makes sense to me

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -49,6 +69,322 @@ describe('Ledger Transfer', (): void => {
...
+    })
+
+    test('returns error if not enough liquidity', async (): Promise<void> => {
+      const transfer: CreateTransferArgs = {
+        ledger: peerAccount.ledger,
+        debitAccount: peerAccount,
+        creditAccount: account,
+        transferRef: uuid(),
+        amount: accountStartingBalance + 1n
+      }
```
wilsonianb, line 282:
```suggestion
      const transfer: CreateTransferArgs = {
        ...baseTransfer,
        amount: accountStartingBalance + 1n
      }
```

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -49,6 +69,322 @@ describe('Ledger Transfer', (): void => {
...
+    })
+
+    test('returns error if not enough debit balance', async (): Promise<void> => {
+      const transfer: CreateTransferArgs = {
+        ledger: peerAccount.ledger,
+        debitAccount: account,
+        creditAccount: settlementAccount,
+        transferRef: uuid(),
+        amount: totalAssetSettlementBalance + 1n
+      }
```
wilsonianb, line 299:
```suggestion
      const transfer: CreateTransferArgs = {
        ...baseTransfer,
        creditAccount: settlementAccount,
        amount: totalAssetSettlementBalance + 1n
      }
```

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -49,6 +69,322 @@ describe('Ledger Transfer', (): void => {
...
+      ${'passes if non-settlement account'}                                      | ${false}            | ${10n}         | ${0n}         | ${0n}          | ${0n}        | ${0n}         | ${true}
+      ${'passes if posted debits surpass transfer amount'}                       | ${true}             | ${10n}         | ${0n}         | ${0n}          | ${11n}       | ${0n}         | ${true}
+      ${'passes if posted debits equal transfer amount'}                         | ${true}             | ${10n}         | ${0n}         | ${0n}          | ${10n}       | ${0n}         | ${true}
+      ${'passes if posted debits equal transfer amount and all credits'}         | ${true}             | ${6n}          | ${2n}         | ${2n}          | ${10n}       | ${0n}         | ${true}
+      ${'fails if posted debits below transfer amount'}                          | ${true}             | ${11n}         | ${0n}         | ${0n}          | ${10n}       | ${0n}         | ${false}
+      ${'fails if posted debits below transfer amount, ignores pending credits'} | ${true}             | ${11n}         | ${0n}         | ${0n}          | ${10n}       | ${10n}        | ${false}
```
wilsonianb, line 357:
```suggestion
      ${'fails if posted debits below transfer amount, ignores pending debits'}  | ${true}             | ${11n}         | ${0n}         | ${0n}          | ${10n}       | ${10n}        | ${false}
```

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -59,9 +395,9 @@ describe('Ledger Transfer', (): void => {
       async ({ accountType }): Promise<void> => {
         const transfer = await createLedgerTransfer(
           {
-            ledger: creditAccount.ledger,
-            creditAccountId: creditAccount.id,
-            debitAccountId: debitAccount.id,
+            ledger: account.ledger,
+            creditAccountId: account.id,
+            debitAccountId: peerAccount.id,
```
wilsonianb, line 400: I think `creditAccount`/`debitAccount` might be clearer.
What if you defined them in a `beforeEach` from `account`/`peerAccount`?

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
mkurapov: This shouldn't happen since we fetch them directly from the DB first, but this would fail anyway on the insert

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
mkurapov: Did something a little different without needing to store `depositTransfer`

packages/backend/src/accounting/psql/ledger-transfer/index.ts
wilsonianb: I'm not attached to the name, but I think it's still worth having separate errors for `hasEnoughDebitBalance` vs `hasEnoughLiquidity`

wilsonianb (dismissed):

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -38,3 +72,156 @@ export async function getAccountTransfers(
...
+
+function prepareTransfer(
+  transfer: CreateTransferArgs
+): Partial<LedgerTransfer> {
+  return {
+    amount: transfer.amount,
+    transferRef: transfer.transferRef,
+    creditAccountId: transfer.creditAccount.id,
+    debitAccountId: transfer.debitAccount.id,
+    ledger: transfer.ledger,
```
wilsonianb, line 218:
```suggestion
    ledger: transfer.creditAccount.ledger,
```
and remove `ledger` from `CreateTransferArgs`

wilsonianb (approved):

PR #1100, opened 2023-02-10
Balance calculation for PSQL accounting

mkurapov:
## Changes proposed in this pull request

- Adding `LedgerTransfer` model
- Adding balance calculations for `get*Balance` methods in the psqlAccountingService

## Context

- fixes #1090 

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass

PR #1113, opened 2023-02-14
1016/mk/refactor services

mkurapov:

- Removed `ledgerAccountService` `ledgerTransferService` 
- Everything now under `AccountingService` instead, except separated into functions
- New structure:

Screenshot 2023-02-14 at 2 10 51 PM

## Context
 
- Comment: https://github.com/interledger/rafiki/pull/1100#discussion_r1105479593

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/psql/balance.ts
```diff
@@ -0,0 +1,46 @@
...
+import { getAccountTransfers } from './ledger-transfer'
+
+interface AccountBalance {
+  creditsPosted: bigint
+  creditsPending: bigint
+  debitsPosted: bigint
+  debitsPending: bigint
+}
+
+export async function getAccountBalances(
```
wilsonianb, line 13: could test this in `balance.test.ts`

packages/backend/src/accounting/psql/ledger-transfer/index.test.ts
```diff
@@ -67,7 +65,8 @@ describe('Ledger Transfer Service', (): void => {
         )
 
         await expect(
-          ledgerTransferService.getAccountTransfers(
+          getAccountTransfers(
+            serviceDeps,
```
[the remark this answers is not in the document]
mkurapov: It doesn't really matter, since in `container.start()` the objection models will be instantiated with `Model.knex(knex)`, so they don't need to pass in `knex` directly. 

I updated it to be explicit, however

packages/backend/src/accounting/psql/balance.test.ts
```diff
@@ -0,0 +1,186 @@
...
+          POSTED: 40n,
+          PENDING: 30n
+        },
+        debit: {
+          POSTED: 20n,
+          PENDING: 10n
+        }
+      }
+
+      for (const type of ['credit', 'debit']) {
```
wilsonianb, line 98: I think some these tests would fail if we started enforcing balance limits
https://github.com/interledger/rafiki/pull/1111#discussion_r1105683510

wilsonianb (approved):

packages/backend/src/accounting/psql/balance.test.ts
mkurapov: Since I'm updating the `createTransfers` method to do balance checks itself, I don't think these tests should be converted to use the actual `createTransfers` method, but continue on creating transfers directly with the model.

TL;DR, I don't think these tests should necessarily care/need to know about balance checks.

wilsonianb: :thinking: how bad would it be to enforce balance limits in 
https://github.com/interledger/rafiki/blob/b9719023dff1d585667145bf899bcc73e9a40143/packages/backend/src/accounting/psql/ledger-transfer/model.ts#L39-L42

mkurapov: 🤔 
Should be a relatively straightforward change, would just need to have the balance calculation live as a method on the `LedgerTransfer` model... it does prevent errors _if_ someone were to use the model directly to create transfers, but seems a bit hidden at first to see where the balance checks are done, makes serving up errors a bit more convoluted.

Thinking if it's worth it...

PR #1139, opened 2023-02-19
enable psql accounting by default

### LINKED ISSUE #1093: Use PSQL accounting for backend tests
Use PSQL accounting as a default for tests, only require Tigerbeetle for the `tigerbeetleAccountingService`

mkurapov:

- Enable PSQL as accounting by default (in tests and at startup)
- Bump `backend` tests node memory, logHeapUsage for debugging
- Clean up test logging

## Context

- fixes #1093 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -232,6 +227,10 @@ async function validateTransfer(
     return TransferError.DifferentAssets
   }
 
+  if (await LedgerTransfer.query(trx).findOne({ transferRef })) {
+    return TransferError.TransferExists
+  }
```
mkurapov, line 232: Making two identical withdrawals/transfers with the second one putting an account into overbalance would result in a "insufficient balance error" instead of a "transfer exists error". This would cause the liquidity resolver tests to fail. It does feel "tidier" to skip balance checks if a transfer happens to exist, but does add one additional lookup.

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -254,7 +253,7 @@ async function validateBalances(
       transferAmount: amount
     })
   ) {
-    return TransferError.InsufficientDebitBalance
```
mkurapov, line 257: So interestingly, the `peer liquidity withdrawal` and `asset liquidity withdrawal` resolver tests in `liqudity.test.ts` expected a `InsufficientBalance` error instead of a `InsufficientDebitBalance` here.

_However_, `InsufficientDebitBalance` seems to make more sense here, since we don't have enough debits to withdraw from the settlement/debit account. Not sure what is best here, `InsufficientBalance` does keep it simple

packages/backend/src/config/app.ts
```diff
@@ -58,7 +58,7 @@ export const Config = {
     ? Buffer.from(process.env.STREAM_SECRET, 'base64')
     : crypto.randomBytes(32),
 
-  useTigerbeetle: envBool('USE_TIGERBEETLE', true),
+  useTigerbeetle: envBool('USE_TIGERBEETLE', false),
```
wilsonianb, line 61: Can we remove the tigerbeetle testcontainer from the jest global setup?
https://github.com/interledger/rafiki/blob/d33d4d9b5c3423e7990d0151647062e78942d53b/packages/backend/jest.setup.ts#L74-L95

packages/backend/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "backend",
   "scripts": {
-    "test": "jest --passWithNoTests --maxWorkers=50%",
+    "test": "NODE_OPTIONS=--max-old-space-size=4096 jest --passWithNoTests --maxWorkers=50% --logHeapUsage",
```
wilsonianb, line 4: Could we confine these to GitHub Actions? 
https://github.com/interledger/rafiki/blob/d33d4d9b5c3423e7990d0151647062e78942d53b/.github/workflows/lint_test_build.yml#L29
especially `--max-old-space-size`

mkurapov: Yeah I was going back and forth on this, but `--max-old-space-size` is pretty device dependent.

wilsonianb (approved):

PR #1186, opened 2023-03-03
docs: integration - payment pointers

### LINKED ISSUE #238: Document Payment Pointers/User Accounts
Rafiki has a unique idea of what a Payment Pointer is and how that interacts/relates to accounts at a Wallet. We need to document this for Wallets to understand how they could issue Payment Pointers for their accounts.

sabineschaller:

- short info about payment pointers (plus link to other resources)
- how to create payment pointer (and asset since it's a prerequisite)
- how to add key to payment pointer (and why you want to do that)

## Context

- fixes #238 
- 
## Checklist

- [ ] Tests added/updated
- [x] Documentation added


docs/integration.md
```diff
@@ -2,7 +2,14 @@
 
 **❗ Rafiki is intended to be run by [Account Servicing Entities](./glossary.md#account-servicing-entity) only and should not be used in production by non-regulated entities.**
 
-Account Servicing Entities provide and maintain payment accounts. In order to make these accounts Interledger-enabled via Rafiki, they need to provide the following endpoints and services.
+Account Servicing Entities provide and maintain payment accounts. In order to make these accounts Interledger-enabled via Rafiki, they need to provide the following endpoints and services:
+
+- prices (exchange rates)
+- fees
+- [webhook events listener](#webhook-events-listener)
+- (optional) [Open Payments Auth Server](#open-payments-auth-server)
```
wilsonianb, line 10: If we're considering the AS optional, should we make the public Open Payments backend endpoints optional?

docs/integration.md
```diff
@@ -2,7 +2,14 @@
...
+
+Furthermore, each payment account managed by the Account Servicing Entity needs to be issued at least one [payment pointer](#issuing-payment-pointers).
```
wilsonianb, line 12: ... in order to be serviced by Rafiki / send or receive Interledger payments

docs/integration.md
sabineschaller: Which ones are you referring to?

wilsonianb: The Open Payments RS
https://docs.openpayments.guide/reference/get-payment-pointer
Rafiki accounts are technically Interledger-enabled without them

sabineschaller: Valid point. In the spirit of making everything pluggable, we could also make it optional. But in that case, I think we should pull the OP server out of `backend` or rather have the connector and the OP server as 2 different packages. I think that is something for v2. Since the auth server is already a separate package, we can already mark it as optional. But I'm not saying we have to.
@AlexLakatos @mkurapov thoughts?

mkurapov: Pulling OP out of `backend` makes sense to me, but for a v2. 

For now though, IMO I don't think we should mark the AS it as optional: we should just ship Open Payments by default. If an implementer has more questions about the AS/Open Payments, then we could have that discussion then, since we would need to understand what their use case is, and why they wouldn't want to support Open Payments right out of the box.

mkurapov (commented):
All the doc look good :+1:, but I think I'm leaning towards not marking AS as optional for now

mkurapov (approved):

PR #1207, opened 2023-03-09
docs(open-payments): Add readme docs

mkurapov:
## Changes proposed in this pull request

- Adds initial README to `open-payments`
- Adds some TSDoc comments for explaining the arguments in the client

## Context

- Progress towards #970 

## Checklist

- [ ] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass

PR #1223, opened 2023-03-14
1215/mk/non interactive quotes

### LINKED ISSUE #1215: Make quote grants non-interactive
Currently, grant creation for quotes is interactive, so if we break up the quote and outgoing payment quote requests into two steps, a user (RO), would need to go through an interaction flow twice to complete a payment.

We would want to break up the grant requests into two, since we can then use amounts from the quote in order to inform the limits for the outgoing payment grant request, and provide the actual payment amount to the user who is completing the interaction. 

Tasks:
- Create `QUOTE_INTERACTION` flag, disable by default in `auth` package
- Allow grants for quotes to be created without interaction in `auth` package
- Edit eCommerce demo example in Postman to

mkurapov:

- Make quote grant creation non-interactive (behind `QUOTE_INTERACTION` flag) & to unblock https://github.com/interledger/rafiki/pull/1207#discussion_r1135203368
- A Postman PR splitting out the quote & outgoing payment grant requests has been created as well

## Context

- Fixes #1215 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/auth/src/grant/routes.test.ts
```diff
@@ -170,54 +170,57 @@ describe('Grant Routes', (): void => {
...
-          {
-            headers: {
-              Accept: 'application/json',
-              'Content-Type': 'application/json'
+      test.each`
+        accessType
+        ${AccessType.IncomingPayment}
+        ${AccessType.Quote}
+      `(
+        'Can get an grant without interaction for $accessType grants',
```
mkurapov, line 178: we don't generally have a good pattern for testing different ENV variable configurations. Should I recreate a new container with different Config variables to test the cases where interaction for incoming payments and quotes _is_ enabled?

packages/auth/src/grant/routes.ts
```diff
@@ -146,23 +146,29 @@ async function createGrant(
...
+  const skipIncomingPaymentInteraction =
+    ctx.request.body.access_token.access.every(
+      isIncomingPaymentAccessRequest
+    ) && !deps.config.incomingPaymentInteraction
+
+  const skipQuoteInteraction =
+    ctx.request.body.access_token.access.every(isQuoteAccessRequest) &&
+    !deps.config.quoteInteraction
+
+  return skipIncomingPaymentInteraction || skipQuoteInteraction
```
wilsonianb, line 169: As is, this would return false for a grant request with only an incoming payment and quote access

packages/auth/src/grant/routes.test.ts
wilsonianb: We dabbled with that in
- https://github.com/interledger/rafiki/pull/683

but ended up relocating the env var / config check to outside the scope of the relevant tests
- https://github.com/interledger/rafiki/pull/735

Could we similarly move this check outside the route? To a prior middleware?

packages/auth/src/grant/routes.ts
mkurapov: I originally decided against this because I wasn't sure of situations where you'd want to create an incoming payment and quote on the same payment pointer at a time, but in hindsight it shouldn't matter since both should be non-interactive... I'll update it

mkurapov:
@wilsonianb what do you think of something like this for the app config object:

https://github.com/interledger/rafiki/blob/80071d3508422320430f94219c6e7cbf5ec3d82d/packages/open-payments/src/test/helpers.ts#L37-L52

mkurapov:
@wilsonianb because of:

```ts
config.prop = 'foo'
const config2 = await deps.use('config')
config.prop === config2.prop // true
```

I was thinking of overwriting the `config` object, calling the test (which now uses the new values), and reverting the config back to the original value. I did a small test of this, and it does seem to work.

packages/auth/src/grant/routes.test.ts
```diff
@@ -127,6 +129,91 @@ describe('Grant Routes', (): void => {
...
+                () => config,
+                interactionFlagsEnabled
+                  ? {
+                      incomingPaymentInteraction: true,
+                      quoteInteraction: true
+                    }
+                  : {
+                      incomingPaymentInteraction: false,
+                      quoteInteraction: false
+                    },
```
wilsonianb, line 157:
```suggestion
                {
                  incomingPaymentInteraction: interactionFlagsEnabled,
                  quoteInteraction: interactionFlagsEnabled
                },
```

wilsonianb (approved):

wilsonianb (approved):

PR #1231, opened 2023-03-15
chore(openapi): use tagged spec version v1.0

### LINKED ISSUE #1200: use Open Payments spec tag within Rafiki instead of commit hash

sabineschaller:

- 

## Context

- fixes #1200 

## Checklist

- [ ] Tests added/updated
- [x] Documentation added

### DIFF
--- a/docs/open-payments.md
+++ b/docs/open-payments.md
@@ -4,8 +4,6 @@ Open Payments is an API standard that allows third-parties (with the account hol
 
 Rafiki implements the Open Payments APIs. Every request and response is validated against the [Open Payments API specification](https://github.com/interledger/open-payments/tree/main/openapi).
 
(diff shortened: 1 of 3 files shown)


mkurapov:
Unfourtunately, this (originating from https://github.com/interledger/open-payments/pull/243) actually breaks token-introspection's type generation, since its OpenAPI spec was relying on the shared `access` schema:

https://github.com/interledger/rafiki/blob/5341d7600cea3192180695117b5ad55336f5ec34/packages/token-introspection/src/openapi/token-introspection.yaml#L89-L90

I'm thinking we can just have token-introspection also take the auth-server.yaml and reference the schemas as it needs locally: 
- #1239 

Additionally, as part of the package versioning, I'm cleaning up some dependencies, and I think we should remove symlinks for the OpenAPI yaml files within the project, since each package need to be self-contained.

PR #1239, opened 2023-03-16
fix(token-introspection): fix type generation

mkurapov:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/token-introspection/src/openapi/schemas.yaml
```diff
@@ -1 +0,0 @@
-../../../open-payments/src/openapi/schemas.yaml
```
mkurapov, line 1: To be a self-contained package, `token-introspection` (and similarly, `open-payments`) should fetch its own schemas

PR #1246, opened 2023-03-19
chore(docs): document connector

### LINKED ISSUE #1179: Document Rafiki connector
The connector code was imported from here: https://github.com/interledgerjs/rafiki/

We should document what it is comprised of and how Rafiki Quoting and Sending works.

wilsonianb:

- 

## Context

- fixes #1179 

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


docs/connector.md
```diff
@@ -0,0 +1,51 @@
+# Connector
+
+The [`backend`](./architecture.md#backend) includes an [Interledger](./glossary.md#interledger-protocol) connector for sending and receiving [STREAM](./glossary.md#stream) packets.
+
+The amounts of these packets are used to update account balances in [TigerBeetle](./glossary.md#tigerbeetle). (See [Accounts and Transfers](./accounts-and-transfers.md))
```
sabineschaller, line 5: I think we have used "Tigerbeetle" rather than "TigerBeetle" within the docs. I think we should make that consistent.

docs/connector.md
```diff
@@ -0,0 +1,51 @@
...
+
+Amounts are adjusted based on the destination/outgoing account's asset and Rafiki's configured exchange [rates service](./integration.md#rates-prices).
+
+## Packet Origination
+
+The connector receives incoming ILP packets via:
+
+- HTTP
+- direct calls from within the `backend`
```
sabineschaller, line 14: That is when the outgoing payment sends to an incoming payment on the same Rafiki, right? Can we maybe add that in a half-sentence?

docs/connector.md
```diff
@@ -0,0 +1,51 @@
...
+
+These calls to the connector are actually made by calling `makeIlpPlugin` and passing the plugin to the [`@interledger/pay`](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay) library.
+
+## Packet Destination
+
+An ILP packet may either terminate at the local Rafiki's STREAM server or continue on to a peer via HTTP.
+
+### Local STREAM Server
+
+The connector attempts to extract and decode the payment tag from a received ILP packet's destination address. If it is successfully able to match the tag with a locally managed [Open Payments](./glossary.md#open-payments) [payment pointer](./glossary.md#payment-pointer) or [incoming payment](https://docs.openpayments.guide/reference/create-outgoing-payment), it will not forward the packet. The connector will credit the corresponding balance as well as track the total amount received for the STREAM connection in [Redis](./architecture.md) in order to support [STREAM receipts](https://interledger.org/rfcs/0039-stream-receipts/).
```
sabineschaller, line 34: Do we still want to mention STREAM receipts now that we are deprecating them? Or if we keep the mentioning of it, but say that they are deprecated?

docs/connector.md
```diff
@@ -0,0 +1,51 @@
...
+## Packet Rejection
+
+The connector may reject a packet with a corresponding [ILP error code](https://interledger.org/rfcs/0027-interledger-protocol-4/#error-codes) due to a number of reasons including:
+
+- invalid packet
+- insufficient liquidity
+- rate limit exceeded
+- amount exceeding `maxPacketAmount`
+- expired packet
+- and more!
```
sabineschaller, line 51: Should we link to the spec that includes all the error codes and reasons?

wilsonianb: Which spec?
There's a link to https://interledger.org/rfcs/0027-interledger-protocol-4/#error-codes above

docs/connector.md
wilsonianb: I vote we match their capitalization

docs/connector.md
wilsonianb: I think outgoing payments send packets to the connector for both local and remote incoming payment stream connections

docs/connector.md
wilsonianb: To what extend are STREAM receipts deprecated?
This is primarily mentioned here to explain why we have Redis.

sabineschaller: To the extend that they are marked as deprecated in the WM spec.

sabineschaller: To the extend that they are marked as deprecated in the WM spec. But it makes sense to mention them here as that is the only reason we have Redis.

docs/connector.md
wilsonianb: Both HTTP and local backend originating packets can be addressed to either local destination or outgoing peer.
So "direct calls from within the `backend`" includes but isn't limited to intra-Rafiki payments.

docs/connector.md
wilsonianb: :sweat_smile: 
Screenshot from 2023-03-20 15-25-14

sabineschaller: To the extend that they are marked as deprecated in the WM spec. But it makes sense to mention them here as that is the only reason we have Redis.

sabineschaller: 😂
I need to ask Alex when we finally release the new WM spec.

docs/connector.md
```diff
@@ -0,0 +1,51 @@
+# Connector
+
+The [`backend`](./architecture.md#backend) includes an [Interledger](./glossary.md#interledger-protocol) connector for sending and receiving [STREAM](./glossary.md#stream) packets.
```
mkurapov, line 3: In there: https://github.com/interledger/rafiki/blob/65e316b6f786b9d0acd041cd0778944cf6a18613/docs/architecture.md?plain=1#L24 we should link this connector document

docs/connector.md
mkurapov, line 34:
```suggestion
The connector attempts to extract and decode the payment tag from a received ILP packet's destination address. If it is successfully able to match the tag with a locally managed [Open Payments](./glossary.md#open-payments) [payment pointer](./glossary.md#payment-pointer) or [incoming payment](https://docs.openpayments.guide/reference/create-incoming-payment), it will not forward the packet. The connector will credit the corresponding balance as well as track the total amount received for the STREAM connection in [Redis](./architecture.md) in order to support [STREAM receipts](https://interledger.org/rfcs/0039-stream-receipts/).
```

mkurapov (approved):

PR #1255, opened 2023-03-21
fix(dockerbuild): architectures

rluckom-coil:
fixes [#1263](https://github.com/interledger/rafiki/issues/1263)

Also uses base images specific to the architectures which seems to speed up the builds a lot

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -13,12 +13,8 @@ jobs:
     strategy:
       matrix:
         package: [auth, backend]
-        platform: [linux/amd64, linux/arm64]
         push:
           - ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
-        exclude:
-          - platform: linux/arm64
-            push: false
     steps:
       - name: Docker meta
         id: meta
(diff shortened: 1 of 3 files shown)


sabineschaller:
While you are working on this, could you include this?
- #1263

.github/workflows/docker.yml
```diff
@@ -13,22 +13,14 @@ jobs:
...
     steps:
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v4
         with:
           images: ghcr.io/${{ github.repository_owner }}/rafiki-${{ matrix.package }}
-          tags: |
-            type=raw,enable={{is_default_branch}},value=latest
-            type=raw,enable=true,value={{ sha }}
-            type=raw,enable=true,value={{ refs }}
```
sabineschaller, line 31: Does the removal of this ensure that the image tag =  gh tag?

sabineschaller (commented):
Docker images work for me. 

And I'm assuming that the docker image tagging with the gh tag also works since I see a new tag in the list of tags?

sabineschaller:
Also, can you add the 
- fixes #1263 

to your description if it is actually working?

.github/workflows/docker.yml
rluckom-coil: yeah, the default behavior is to add the git tag as a label and the other defaults do what we were already doing (branch name, commit sha, etc)

sabineschaller (approved):
Thank you!

Issue #626, opened 2022-09-21
Validate backend <> account provider requests


wilsonianb:
Write OpenAPI spec for quote request, prices request, and webhook events, to allow the backend and account provider to validate requests and responses.

See: [integration docs](https://github.com/interledger/rafiki/blob/main/docs/integration.md)

Graphql already handles admin api validation.

wilsonianb:
- related to #571

sabineschaller:
Webhooks spec was added with
- #1210

sabineschaller:
Prices (exchange rates) request responses are already validated:

https://github.com/interledger/rafiki/blob/main/packages/backend/src/rates/service.ts#L115

mkurapov:
- Quote route no longer exists
- We have a webhook OpenAPI spec, but no need to validate the request itself
- We validate rate response in route

PR #1351, opened 2023-04-19
refactor: rename prices -> exchange rates

sabineschaller:

- rename prices -> exchange rates because the prices term was confusing
- add OpenAPI spec for exchange rates

## Context

- part of #626 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -53,7 +53,7 @@ services:
       PUBLIC_HOST: http://cloud-nine-wallet-backend
       OPEN_PAYMENTS_URL: http://cloud-nine-wallet-backend
       WEBHOOK_URL: http://cloud-nine-wallet/webhooks
-      PRICES_URL: http://cloud-nine-wallet/prices
+      EXCHANGE_RATES_URL: http://cloud-nine-wallet/prices
```
mkurapov, line 56: should we rename the `/prices` route in the ASE?

packages/backend/src/openapi/exchange-rates.yaml
```diff
@@ -0,0 +1,59 @@
+openapi: 3.1.0
```
mkurapov, line 1: will validation using this spec be added later when receiving the response from the /rates endpoint?

packages/backend/src/rates/service.ts
```diff
@@ -96,23 +96,23 @@ class RatesServiceImpl implements RatesService {
...
     }
-    return this.pricesRequest
+    return this.ratesRequest
   }
 
-  private async loadNow(): Promise<Prices> {
-    const url = this.deps.pricesUrl
+  private async loadNow(): Promise<Rates> {
+    const url = this.deps.exchangeRatesUrl
     if (!url) return {}
```
mkurapov, line 113: when no url, should we return 
```ts
    const data = {
      [base]: 1.0,
      rates: {} 
    }
```
?

mkurapov (commented):
a few comments

packages/backend/src/rates/service.ts
raducristianpopa: Shouldn't it be just:
```ts
const data = {
  [base]: 1.0,
}
```
?

packages/backend/src/openapi/exchange-rates.yaml
sabineschaller: I started adding it but then I realized that responses are already validated. See [here](https://github.com/interledger/rafiki/blob/main/packages/backend/src/rates/service.ts#L115). Hence, the spec is rather for documentation than for validation.

packages/backend/src/rates/service.ts
sabineschaller: @raducristianpopa is right, but it is not necessary. If `{}` is returned, `{ [base]: 1.0 }` is assumed by the pay library (which I think is the only code that consumes that response, but I'm still digging through the code to understand that part better).

localenv/cloud-nine-wallet/docker-compose.yml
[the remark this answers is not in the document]
mkurapov: I think just to be consistent (to have the MASE as a good example) IMO it's worth it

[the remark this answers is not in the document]
mkurapov: I think just to be consistent (to have the MASE as a good example) IMO it's worth it

mkurapov (commented):
Just the MASE change, otherwise LGTM

mkurapov (approved):

PR #1357, opened 2023-04-20
chore: update .dockerignore

### LINKED ISSUE #447: update dockerignore file
check what can be added to the dockerignore file. Specifically the infra folder

sabineschaller:

- 

## Context

- fixes #447 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


.dockerignore
```diff
@@ -3,3 +3,27 @@ node_modules
...
+localenv
+
+.github
+.husky
+.dockerignore
+.eslintignore
+.eslintrc.yml
+.gitattributes
+.gitignore
+.lintstagedrc.js
```
raducristianpopa, line 20: I think `.lintstagedrc.js` can be deleted, since it's not used anymore.

packages/frontend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18.13.0-slim as builder
```
mkurapov, line 1: :+1:

.dockerignore
```diff
@@ -3,3 +3,26 @@ node_modules
...
+.gitignore
+.prettierignore
+commitlint.config.js
+jest.config.base.js
+renovate.json
+
+**/jest.config.js
+**/jest.setup.js
+**/jest.teardown.js
+**/docker-compose.yml
```
mkurapov, line 28: maybe we can also ignore test files in here?

.dockerignore
```diff
@@ -3,3 +3,26 @@ node_modules
 **/node_modules
 **/dist
 dist
+tmp
+**/tmp
```
mkurapov, line 7:
```suggestion
```
I think `tmp` already covers this

.dockerignore
```diff
@@ -3,3 +3,26 @@ node_modules
 **/node_modules
 **/dist
```
mkurapov, line 4:
```suggestion
```
same comment as below

.dockerignore
sabineschaller: Yes!

.dockerignore
sabineschaller: I have tmp folders in the root and in sub-directories.

sabineschaller: Never mind, I read up on the matter.

mkurapov: I think just `tmp` would also cover root and sub-directories, but `**/tmp` works as well

mkurapov (approved):

.dockerignore
sabineschaller: From what I read, `tmp` only covers the root directory but `**/tmp` covers root and sub-directories. But I did find conflicting information.

PR #1376, opened 2023-04-25
chore: fix releases

sabineschaller:

- set package version to 1.0.0
- will re-tag after this is merged (v1.0.0-alpha.1 instead of v1.1.0-alpha) and delete wrong tags and releases
- updated renovate config and release.yaml for better release notes

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -1,22 +1,5 @@
 changelog:
-  exclude:
-    labels:
-      - ignore-for-release
   categories:
-    # - title: Breaking Changes 🛠
-    #   labels:
-    #     - Semver-Major
-    #     - breaking-change
-    #   exclude:
-    #     labels:
-    #       - dependencies
-    # - title: Exciting New Features 🎉
-    #   labels:
-    #     - Semver-Minor
-    #     - enhancement
-    #   exclude:
-    #     labels:
-    #       - dependencies
     - title: General Changes
       labels:
         - "*"
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -3,21 +3,21 @@ changelog:
     labels:
       - ignore-for-release
   categories:
-    - title: Breaking Changes 🛠
-      labels:
-        - Semver-Major
-        - breaking-change
-      exclude:
-        labels:
-          - dependencies
-    - title: Exciting New Features 🎉
-      labels:
-        - Semver-Minor
-        - enhancement
-      exclude:
-        labels:
-          - dependencies
-    - title: Fixes and Patches
+    # - title: Breaking Changes 🛠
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interledger/rafiki",
-  "version": "1.1.1-alpha",
+  "version": "1.0.0",
```
mkurapov, line 3: not `-alpha.1` or similar?

.github/release.yaml
```diff
@@ -2,17 +2,27 @@ changelog:
...
       labels:
-        - "*"
\ No newline at end of file
+        - "*"
+      exclude:
+        labels:
+          - dependencies
+    - title: Dependency Updates
+      labels:
+        - "dependencies"
```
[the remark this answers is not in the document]
mkurapov: Is there a difference between `"dependencies"` and just `dependencies`? just sanity checking

.github/release.yaml
```diff
@@ -2,17 +2,27 @@ changelog:
   exclude:
     labels:
       - ignore-for-release
-    authors:
-      - app/renovate
   categories:
     - title: Breaking Changes 🛠
       labels:
         - Semver-Major
         - breaking-change
```
mkurapov, line 9: Should we add these labels?

.github/release.yaml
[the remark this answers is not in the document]
sabineschaller: copy-paste error

package.json
sabineschaller: Yes, you are right. We should actually have a script that does that automatically when tagging

.github/release.yaml
sabineschaller: I actually decided to cut these categories for now while we are in a pre-release state. Since we have the semantic PRs now, it should be sufficient to just split general changes and dependency updates.

.github/release.yaml
```diff
@@ -2,17 +2,27 @@ changelog:
   exclude:
     labels:
       - ignore-for-release
-    authors:
-      - app/renovate
   categories:
-    - title: Breaking Changes 🛠
+    # - title: Breaking Changes 🛠
```
mkurapov, line 6: I would usually prefer to just delete commented out code, but in this case, we'll be using it soon enough

mkurapov (approved):

.github/release.yaml
sabineschaller: Yes, usually I prefer deleting, too. I can also just delete it. We'll find it in the github docs

mkurapov (approved):

PR #1398, opened 2023-05-01
feat: update mutations to use input object type, add idempotencyKey

mkurapov:

- Add optional idempotencyKey to non-liquidity mutations
- Convert mutations to all use input objects (some of them were just using single primitive parameters, like `deletePeer` `revokePaymentPointerKey` `triggerPaymentPointerEvents`) 

## Context

- Related Postman [PR](https://interledger.postman.co/workspace/My-Workspace~7aa3e484-5e2e-4154-8a45-0cd200e000c0/pull-request/5c55fba9-459d-4c77-af6c-301de9e05f20)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/schema.graphql
```diff
@@ -231,13 +235,6 @@ input HttpOutgoingInput {
   endpoint: String!
 }
 
-input AssetInput {
```
mkurapov, line 234: wasn't used

packages/frontend/app/lib/validate.server.ts
```diff
@@ -73,8 +73,7 @@ export const createAssetSchema = z
         invalid_type_error: 'Max packet amount is expected to be a number.'
       })
       .int()
-      .positive()
```
mkurapov, line 76: we can have zero-decimal currencies (like JPY)

sabineschaller: But does `.positive()` exclude `0`? If it does, we should add a `.min(0, { message: 'Scale should be between 0 and 255' })` because isn't `int()` a signed integer?

mkurapov: Good catch, I will update this

sabineschaller (approved):
LGTM

Issue #1408, opened 2023-05-04
Merge rafiki.dev documentation into rafiki


mkurapov:
As discussed in-person, we would like to consolidate our documentation into a single repository. The documentation & docusaurus setup will now live in Rafiki instead of the rafiki.dev repo.

PR #1411, opened 2023-05-08
feat(documentation): scaffold docusaurus

mkurapov:

- Moving [rafiki.dev](https://github.com/interledger/rafiki.dev) docusaurus setup into new `documentation` directory/project
    - Setup is primarily the same, minus a few simplifications for our use case
- Autogenerating graphql API docs directly from the sibling `backend` project folder instead (since the schema and docs now live in the same repo)

https://user-images.githubusercontent.com/15069181/236885248-0bc08afa-5f89-44a8-883d-495002a78558.mp4

## Context

- Progress towards #1408 

This is just scaffolding right now, more changes will be coming in follow-up PRs, ie, properly structuring docs from existing `/docs` folder, adding netlify integration similar to rafiki.dev for previewing any changes to the docs.

**Questions**:
- Do we want this to be a sibling project? I though _yes_ this since this is a thing "we ship" along with the other project folders
- How do we want to call this sub-folder? A few suggestions other than `documentation` are `website` `docusaurus` and simply `docs`
- Where should we put/how should we call the GraphQL auth API? Do we want to at this point?

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


BlairCurrey:
> Do we want this to be a sibling project? I though yes this since this is a thing "we ship" along with the other project folders

I find myself agreeing. My understanding is that we are trying to consolidate everything on rafiki.dev and use that as the primary source for documentation. I think putting it at the root level kinda suggests they are like the current /docs where they can be consumed raw from the repository. I mean it's trivial to link to rafiki.dev in either scenario, but conceptually it makes sense to put it with everything else that get's built and scope the commands/packages to their own project IMO.

> How do we want to call this sub-folder? A few suggestions other than documentation are website docusaurus and simply docs

I dont think I have a strong opinion... I guess I like docs, documentation, and docusaurus. I think its less obvious what 'website' its.

BlairCurrey (approved):

sabineschaller:
Did you restore the glossary?

mkurapov:
> Did you restore the glossary?

That is for a follow-up PR, this one just adds the current rafiki.dev pages

PR #1417, opened 2023-05-10
feat: set payment pointer inactive

### LINKED ISSUE #1362: Add ability to mark payment pointers as inactive
- Add GraphQL mutation to mark payment pointer as inactive
- Check whether payment pointer is active in payment pointer middleware

BlairCurrey:

- adds `updatePaymentPointer` mutation where the status can be changed to inactive or active (`publicName` can be set too).
- adds `deactivatesAt` timestamp field to payment pointer table
- adds update method on payment pointer service which generates a new `deactivatesAt` timestamp based on the update input
- adds `isActive` and `status` to the payment pointer model which are abstractions on the `deactivatesAt` field. `null` or future timestamp means active, otherwise its inactive (any non-future timestamp)
- throws 404 in middleware if payment pointer is inactive
- postman pr: https://interledger.postman.co/workspace/blair-interledger~54a25937-6992-4b7f-829f-91ac2142b7c0/pull-request/da718db4-2538-42b1-bdab-bf09aec862ee

## Context

fixes #1362 

That issue also contains some discussion on the implementation details

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/open_payments/payment_pointer/model.ts
```diff
@@ -98,12 +105,21 @@ export class PaymentPointer
       authServer
     }
   }
+
+  public isActive() {
+    return this.status === PaymentPointerStatus.Active
+  }
```
BlairCurrey, line 111: having both `status` and `isActive()` is a bit redundant. I could remove one of them. Originally I had `isActive` but then added the status when adding the new payment pointer update method.

I'm also just noticing that isActive could probably just be a getter instead of a regular method, like status.

packages/backend/src/open_payments/payment_pointer/middleware.ts
```diff
@@ -14,7 +14,8 @@ export function createPaymentPointerMiddleware() {
       const paymentPointer = await paymentPointerService.getByUrl(
         ctx.paymentPointerUrl
       )
-      if (!paymentPointer) {
+
+      if (!paymentPointer || !paymentPointer.isActive()) {
```
BlairCurrey, line 18: Wasn't entirely sure if throwing the 404 was the right behavior (or best error code). If it is, I also considered not adding this condition and instead changing the query to get the payment pointer to filter out inactive ones (either new arg on `getByUrl` or new method like `getActiveByUrl`). Seems like fetching the pointer would be unnecessary unless we were actually going to handle it differently than an unfound one.

packages/backend/src/open_payments/payment_pointer/model.test.ts
```diff
@@ -353,4 +366,63 @@ export const getRouteTests = <M extends PaymentPointerSubresource>({
   }
 }
 
-test.todo('test suite must contain at least one test')
+describe('Payment Pointer Model', (): void => {
```
BlairCurrey, line 369: I wasn't completely sure what was going on in this existing test file. It looks like this file just defines and exports some test infrastructure that is used in `open_payments/incoming`, 
 ` open_payments/outgoing`, `open_payments/quote` (hence the original test.todo). 
 
I don't necessarily think there is a problem here but wanted to note that.

BlairCurrey:
It appears anyone can update any payment pointer if they have the id. I tested sending updates to existing payment pointers in the database and I was able to. Is it currently possible to limit this?  I believe I remember hearing something about the gql api not having authentication yet.

packages/backend/src/graphql/resolvers/payment_pointer.test.ts
```diff
@@ -272,6 +272,7 @@ describe('Payment Pointer Resolvers', (): void => {
       assert.ok(updatedPaymentPointer)
 
       const {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
BlairCurrey, line 275: I added the ignore rule because I am destructuring `deactivatesAt` and then not using it on purpose to make a new object from `paymentPointer` without that property. I thought maybe `deactivatesAt: _` would work but it doesnt. 

I preferred this over refactoring to copy and then delete properties. I think this is just a case where we actually dont want the rule to apply so I added the ignore.

BlairCurrey:
It looks like I may need to update the mutation in light of recent changes here: https://github.com/interledger/rafiki/pull/1398

Also, I need to look into the postman workflow more. I opened a PR but the link appears broken. https://www.postman.com/workspace/fd712d53-6058-4110-8449-0e160ea62129/pull-request/c8f19f73-52c7-412d-a178-aca5163c1a56

packages/backend/src/open_payments/payment_pointer/model.test.ts
```diff
@@ -353,4 +366,61 @@ export const getRouteTests = <M extends PaymentPointerSubresource>({
...
+
+  afterEach(async (): Promise<void> => {
+    await truncateTables(appContainer.knex)
+  })
+
+  afterAll(async (): Promise<void> => {
+    await appContainer.shutdown()
+  })
+
+  describe(`deactivatedAt cases @ ${new Date().toISOString()}`, () => {
```
sabineschaller, line 386:
```suggestion
  describe(`deactivatesAt cases @ ${new Date().toISOString()}`, () => {
```

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -102,6 +119,29 @@ async function createPaymentPointer(
...
+        status === PaymentPointerStatus.Inactive ? new Date() : null
+    }
+
+    return await PaymentPointer.query(deps.knex)
+      .patchAndFetchById(id, update)
+      .withGraphFetched('asset')
+      .throwIfNotFound()
+  } catch (err) {
+    if (err instanceof NotFoundError) {
+      return PaymentPointerError.UnknownAsset
```
sabineschaller, line 139: We need to define a new error for that case. `UnknownAsset` is for when you are trying to create a payment pointer with an asset (e.g. USD, scale 2), and that asset does not exist.

packages/backend/src/open_payments/payment_pointer/middleware.ts
sabineschaller: I think that is the correct behavior. The middleware is used when fetching payment pointer details prior to another action. If the payment pointer is inactive, the other action should not be possible. But @mkurapov correct me if I'm wrong.

packages/backend/src/graphql/schema.graphql
```diff
@@ -400,6 +405,15 @@ type PaymentPointer implements Model {
   ): OutgoingPaymentConnection
   "Date-time of creation"
   createdAt: String!
+  "Status of the payment pointer"
+  status: PaymentPointerStatus
```
sabineschaller, line 409: If we get rid of `status` in the model, this could potentially just be `isActive` with a boolean. No need to define another enum then.

mkurapov: This was my suggestion, I typically would prefer an enum over a boolean for APIs like this

packages/backend/src/open_payments/payment_pointer/model.test.ts
mkurapov: I would go towards using `deactivatedAt` for everything IMO, since `deactivatesAt` seems like it should be used for the future (ie, for scheduled jobs or something)

packages/backend/src/open_payments/payment_pointer/model.ts
mkurapov: Agree with keeping just `isActive`

packages/backend/src/open_payments/payment_pointer/middleware.ts
mkurapov: This makes behaviour is what I was thinking of 👍

mkurapov:
```suggestion
      if (!paymentPointer?.isActive()) {
```
some sugar if you'd like

packages/backend/src/open_payments/payment_pointer/model.test.ts
mkurapov: I'm not sure if we need to have the `{new Date().toISOString()` here, I think just `describes("isActive")` is sufficient since that is what you are testing

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -133,6 +134,112 @@ describe('Open Payments Payment Pointer Service', (): void => {
...
+        // ensure all cases are checked
+        assert.ok(statusChecked)
+
+        await expect(
+          paymentPointerService.get(paymentPointer.id)
+        ).resolves.toEqual(updatedPaymentPointer)
+      }
+    )
+
+    test('Payment pointer fields can be updated independently', async (): Promise<void> => {
```
mkurapov, line 196: I would break this test up into two tests/two cases

mkurapov (commented):
good start, these are the main things I noticed, submitting review but will go over this in a bit again

mkurapov (commented):
good start, these are the main things I noticed, submitting review but will go over this in a bit again

mkurapov:
> It looks like I may need to update the mutation in light of recent changes here: https://github.com/interledger/rafiki/pull/1398

Yes, good catch :+1:

mkurapov:
Main questions (we can move this to Slack if we'd like):
- What should the behaviour be for fetching resources of deactivated payment pointer: e.g. should we still be able to fetch the incoming payments? (I'm guessing no)
- If we have an incoming payment that gets created, then a payment pointer is deactivated for it, should I still be able to create the corresponding outgoing payment (or quote for that matter)? In the current code this is still possible, if the payment pointer is a "local" payment pointer to Rafiki, whenever we fetch a local receiver, we whenever we "get" a receiver, we would still resolve the payment pointer properly: https://github.com/interledger/rafiki/blob/516ed0d4bb7cb6818836a75feb02012e6fb6a3d6/packages/backend/src/open_payments/payment/outgoing/lifecycle.ts#L22


CC @sabineschaller

packages/backend/src/open_payments/payment_pointer/model.test.ts
BlairCurrey: > I would go towards using deactivatedAt for everything IMO, since deactivatesAt seems like it should be used for the future (ie, for scheduled jobs or something)

Currently the timestamp will always be in the past after its set. But should it always? Currently there is nothing preventing it from being set to a future date which is why I went with `deactivatesAt`. 

I realize this isnt a current use-case but If we wanted to we could set a `deactivatesAt`to a future value and then when that time comes it will be deactivated. I feel like `deactivatesAt` is truer to what the value could be. If it should not be a value in the future I wonder if I should add something to the model preventing it, because obviously the database will accept any timestamp.

> I'm not sure if we need to have the {new Date().toISOString() here, I think just describes("isActive") is sufficient since that is what you are testing

I included the date in the describe name because otherwise its not clear from the test names whether `deactivatesAt` being tested is in the future or the past: 

```
  Payment Pointer Model
    deactivatedAt cases @ 2023-05-11T12:08:21.637Z
      ✓ For a deactivatesAt of null, status should be ACTIVE and isActive should be true (109 ms)
      ✓ For a deactivatesAt of 2023-05-12T12:08:21.637Z, status should be ACTIVE and isActive should be true (70 ms)
      ✓ For a deactivatesAt of 2023-05-10T12:08:21.637Z, status should be INACTIVE and isActive should be false (97 ms)
```

Although now I'm starting to think I could just set an explicit name for each case like "Future deactivatesAt isActive should be true" and "Past deactivatesAt isActive should be false".

packages/backend/src/graphql/schema.graphql
BlairCurrey: This is the root of having both `isActive` and `status`.

Whatever abstractions we want for the `deactivatesAt` field, I think it needs to be on the model. So if we return a status in gql then I think it needs to be on the model. Otherwise we would have to derive the status from either `deactivatesAt` (or isActive if exists) everywhere we are returning a payment pointer in gql (create, update, get, and maybe other resolvers).

So I think the fundamental question is what we want to return in the API. My initial assumptions was a boolean because that's all that was necessary. Max suggested an enum and I was receptive to points about it being more readable/explicit. Most importantly I want to be consistent with our other responses. Less importantly, I think it could handle (possibly very unlikely) situations like adding another status like "PENDING".

For these reasons I was actually leaning towards ditching `isActive` for `status`, if anything.

packages/backend/src/open_payments/payment_pointer/model.ts
BlairCurrey: I think this is ultimately about what we want to return in the API, see my reply here: https://github.com/interledger/rafiki/pull/1417#discussion_r1190885248

packages/backend/src/graphql/schema.graphql
BlairCurrey: > So if we return a status in gql then I think it needs to be on the model. Otherwise we would have to derive the status from either deactivatesAt (or isActive if exists) everywhere we are returning a payment pointer in gql (create, update, get, and maybe other resolvers).

Thinking about this more, I guess I could add a status resolver on our PaymentPointer resolver here:
```
  PaymentPointer: {
    incomingPayments: getPaymentPointerIncomingPayments,
    outgoingPayments: getPaymentPointerOutgoingPayments,
    quotes: getPaymentPointerQuotes
  }
  ```
If we want to keep returning the status on the API but use `isActive` on the backend.

**Comment by sabineschaller** [comment]
> What should the behaviour be for fetching resources of deactivated payment pointer: e.g. should we still be able to fetch the incoming payments? (I'm guessing no)

If it is a fetch via Open Payments, no. If it is a fetch via the admin APIs, I'd say yes. 

> If we have an incoming payment that gets created, then a payment pointer is deactivated for it, should I still be able to create the corresponding outgoing payment (or quote for that matter)? 

I'd say no. Maybe we should complete all open incoming payments when a payment pointer is deactivated.

File: packages/backend/src/open_payments/payment_pointer/model.test.ts (hunk 5 above)
Comment (sabineschaller) [reply packages/backend/src/open_payments/payment_pointer/model.test.ts:386]: I actually liked the forward thinking here and that you could deactivate a payment pointer at a given date and time in the future. That's what I can do with my bank account, too.

Comment (BlairCurrey) [reply packages/backend/src/open_payments/payment_pointer/model.test.ts:386]: I think i'm changing my mind after realizing setting a future date only works out-of-the-box for some cases. It works when we read payment pointers then doing something based on its active/inactive state. It wont work if we want to perform some action at the time it becomes invalid - we'd need to build something additional and it probably wouldn't be driven by a future `deactivatesAt` field.

Consider the possible direction @sabineschaller's offered for dealing with invalid payment pointers and incoming payemnts: 

> Maybe we should complete all open incoming payments when a payment pointer is deactivated.

With respect to setting a future date on `deactivatesAt`, that would require something to constantly watch and do something with dates that switch from being in the future to in the past. But this is probably not the way we'd want to do it. Instead I imagine we'd schedule some event to happen in the future, at which point a `deactivatedAt` field goes from `null` to the current time.

File: packages/backend/src/open_payments/payment_pointer/service.ts (hunk 6 above)
Comment (BlairCurrey) [reply packages/backend/src/open_payments/payment_pointer/service.ts]: added a more appropriate error

File: packages/backend/src/open_payments/payment_pointer/service.test.ts (hunk 8 above)
Comment (BlairCurrey) [reply packages/backend/src/open_payments/payment_pointer/service.test.ts]: i split these

File: packages/backend/src/open_payments/payment_pointer/model.test.ts (hunk 5 above)
Comment (mkurapov) [reply packages/backend/src/open_payments/payment_pointer/model.test.ts]: > It works when we read payment pointers then doing something based on its active/inactive state. It wont work if we want to perform some action at the time it becomes invalid - we'd need to build something additional and it probably wouldn't be driven by a future deactivatesAt field.

I agree. Until we do some sort of scheduling (that will require adding addtional flags, table, etc), `deactivatedAt` field is the simplest solution at the moment. I don't think we need to validate this at the DB layer either IMO, as long as the application layer is setting `deactivatedAt = now` as it is doing now.

File: packages/backend/src/graphql/schema.graphql (hunk 7 above)
Comment (mkurapov) [reply packages/backend/src/graphql/schema.graphql:464]: > Will go with isActive if we prefer that but I do think it's tied to what we want to return in the API.

i.e. we keep everything as is, with `status` in the GraphQL API and `isActive` as a "helper" method in the application layer model?

File: packages/backend/src/open_payments/payment_pointer/model.ts [hunk 9]
```diff
@@ -48,6 +48,11 @@ export class PaymentPointer
   // select sum(`withdrawalAmount`) from `webhookEvents` where `withdrawalAccountId` = `paymentPointer.id`
   public totalEventsAmount!: bigint
   public processAt!: Date | null
+  public deactivatesAt!: Date | null
+
+  public get isActive() {
+    return !this.deactivatesAt || new Date(this.deactivatesAt) > new Date()
```
mkurapov, line 54: I'm understanding we need to wrap `new Date(this.deactivatesAt)` in `new Date()` since `objection` doesn't properly cast the value into a javascript date when retrieving from the DB?

BlairCurrey: Unwrapped may work, not sure if I tried it. I will test that use it if it does.

BlairCurrey: yeah I don't need to construct a date for deactivates, as it's already a date. good catch

packages/backend/src/graphql/schema.graphql
BlairCurrey: I removed status from the model and kept `isActive`. I left status in the graphql layer as a field on paymentPointer and an option in the update input.

packages/backend/src/open_payments/payment_pointer/model.test.ts
BlairCurrey: i renamed back to `deactivatedAt`

BlairCurrey:
I think I have addressed everything here. To recap:

- renamed `deactivatesAt` to `deactivatedAt`
- removed `status` from model, keeping `isActive`. still using `status` enum in graphql layer (update input and field on `paymentPointer`)
- have an issue lined up for handling open payments tied to a payment pointer being deactivated. will submit it when merging these changes in.

I also opened a postman PR after (mostly) figuring out that workflow.

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -12,15 +16,27 @@ import {
 } from './model'
 import { BaseService } from '../../shared/baseService'
 import { AccountingService } from '../../accounting/service'
+import { PaymentPointerStatus } from '../../graphql/generated/graphql'
```
mkurapov, line 19: The layers (GraphQL & business logic layer) are being coupled here (since its a GraphQL import).
ie, other services that would want to use this method shouldn't have to import a GraphQL type.

We can update the `UpdateOptions` type to fix this, see below:

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -12,15 +16,27 @@ import {
...
-export interface CreateOptions {
+export interface CreateOptions extends Options {
   url: string
   assetId: string
-  publicName?: string
 }
 
+export interface UpdateOptions extends Options {
+  id: string
+  status?: PaymentPointerStatus
```
mkurapov, line 32:
```suggestion
  status?: 'ACTIVE' | 'INACTIVE'
```
_should_ do it

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -133,6 +134,107 @@ describe('Open Payments Payment Pointer Service', (): void => {
...
+          id: paymentPointer.id,
+          status,
+          publicName
+        })
+        assert.ok(!isPaymentPointerError(updatedPaymentPointer))
+
+        expect(updatedPaymentPointer.publicName).toEqual(publicName)
+
+        // check status. should be whatever the update input is, or the original value if not updated
+        let statusChecked = false
```
mkurapov, line 163: I feel like we can simplify this test by:
- having an `expectedStatus` column in the table
- removing `publicName` since IMO it's tested sufficiently in "Payment pointer fields can be updated independently > publicName"

packages/backend/src/open_payments/payment_pointer/service.ts
BlairCurrey: yup, that'll do. tested locally.

packages/backend/src/open_payments/payment_pointer/service.test.ts
BlairCurrey: There isn't really a `status` to expect at this level but there is an `isActive`. I think having `expectedIsActive` is good all the same instead of determining the output based off the combination of `status`/`initialIsActive`. And I take your point with `publicName` - it's not really relevant to what we want to test here.

As for not having a status to expect here, we are getting a status from the gql input and passing that into the update. What we get from the database just has an isActive (via the model). Maybe we should pass in `isActive` instead? For the current use case that seemed unnecessary because we're just passing in the gql input and not getting it from some payment pointer record with `paymentPointer.isActive` or something. But I guess we might want to. Additionally, the way it is now also makes the test a bit more confusing to read (and maybe makes the update method more confusing to think about).

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -102,6 +119,29 @@ async function createPaymentPointer(
...
 
+async function updatePaymentPointer(
+  deps: ServiceDependencies,
+  { id, status, publicName }: UpdateOptions
+): Promise<PaymentPointer | PaymentPointerError> {
+  try {
+    const update: UpdateInput = { publicName }
+    if (status) {
+      update.deactivatedAt =
+        status === PaymentPointerStatus.Inactive ? new Date() : null
```
mkurapov, line 130:
```suggestion
        status === 'INACTIVE' ? new Date() : null
```
so we can remove `PaymentPointerStatus` enum from here

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -133,6 +134,86 @@ describe('Open Payments Payment Pointer Service', (): void => {
...
 
+  describe('Update Payment Pointer', (): void => {
+    test.each`
+      initialIsActive | status        | expectedIsActive
+      ${true}         | ${undefined}  | ${true}
+      ${true}         | ${'INACTIVE'} | ${false}
+      ${false}        | ${'ACTIVE'}   | ${true}
+      ${false}        | ${undefined}  | ${false}
+    `(
+      'Payment pointer  with initial isActive of $initialIsActive can be updated with status of $status and publicName of $publicName',
```
mkurapov, line 145:
```suggestion
      'Payment pointer with initial isActive of $initialIsActive can be updated with $status status ',
```

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -133,6 +134,86 @@ describe('Open Payments Payment Pointer Service', (): void => {
...
+        })
+        assert.ok(!isPaymentPointerError(updatedPaymentPointer))
+        expect(updatedPaymentPointer.deactivatedAt).toBeDefined()
+        expect(updatedPaymentPointer.isActive).toEqual(false)
+        expect(updatedPaymentPointer.publicName).toEqual(initialName)
+        await expect(
+          paymentPointerService.get(paymentPointer.id)
+        ).resolves.toEqual(updatedPaymentPointer)
+      })
+    })
```
mkurapov, line 204: I think we're already testing this with the test above IMO. We can just remove the `describe('Payment pointer fields can be updated independently'` block, to have a simple `test('can update publicName', ...)` maybe?

mkurapov (commented):
Final pass

packages/backend/src/open_payments/payment_pointer/service.test.ts
BlairCurrey: yeah you're right, it's pure overlap now. I refactored as you described.

packages/backend/src/open_payments/payment_pointer/service.ts
BlairCurrey: good catch, should have done this with the previous suggestion on this topic. removed along with import and usage in the service tests here: https://github.com/interledger/rafiki/pull/1417/commits/c973590975e09c4c0a6088456779f5f79151bfc5

mkurapov (approved):
🚀

Issue #1423, opened 2023-05-12
Fix broken docker-compose link


BlairCurrey:
The link to the compose plugin in `localenv/README.md` is broken: https://docs.docker.com/compose/install/compose-plugin/

It is listed as a dependency for running the local environment but it doesn't seem necessary on newer versions of docker, which includes compose support by default, so I think it can simply be removed. https://docs.docker.com/compose/migrate/

Issue #1430, opened 2023-05-15
Add documentation for assets


mkurapov:
We should add a document describing what assets are in Rafiki:
- What is `assetScale` and `assetCode`
- How they can be managed
- Give some examples 
- Link to the liquidity document (#1429) for managing assets

PR #1434, opened 2023-05-16
1408/mk/arrange docs

mkurapov:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/documentation/docs/concepts/asset.md
```diff
@@ -0,0 +1,8 @@
+# Asset
```
mkurapov, line 1: just took from the old glossary, still needs more additions. issue created: https://github.com/interledger/rafiki/issues/1430

packages/documentation/docs/playground/overview.md
```diff
@@ -0,0 +1,176 @@
+# Overview
```
mkurapov, line 1: Taken from the `localenv` README. Maybe we just change that README to just link to this doc? So they don't ever diverge

packages/documentation/docusaurus.config.js
```diff
@@ -58,7 +64,7 @@ const config = {
             type: 'docSidebar',
             position: 'left',
             sidebarId: 'reference',
-            label: 'API Reference'
+            label: 'API'
```
mkurapov, line 62: since itll also contain the auth admin API as well

packages/documentation/docs/integration/management.md
```diff
@@ -0,0 +1,8 @@
+# Management
```
mkurapov, line 1: was the `admin-api.md` file before

BlairCurrey:
It looks like the `README.md` links need to be updated. I started checking them and the first 4 (Account Servicing Entity, Interledger, SPSP, Open Payments) still link to docs/glossary.md.

packages/documentation/docs/playground/overview.md
BlairCurrey: My gut instinct is to have the setup instructions in the repo but I guess if it's linked then I would have found it in the same way (visit repo -> read readme -> proceed to localenv link). I agree we shouldn't really have multiple versions of the same thing. 

I was thinking about this with the README.md and docusaurus's docs/introduction/overview but that's a bit different because we really need to answer "What is rafiki" at the beginning of the readme (as we currently do).

I do feel like the setup instructions are a bit buried in the `playground/overview` document but if we link specifically to the running-the-local-environment section that's probably not an issue.

BlairCurrey (commented):
> The root level docs folder only contains the renovate.md as well as excalidraw - should we keep them in here or move them out?

Any thoughts on where you might move them? It seems like the excalidraw assets could go into documentation project since (I think) the images produced from them are just used in the `documentation` project.

I guess I'm unsure about the `renovate.md` doc and the directory more generally. Doesn't seem like it belongs in docusaurus. I'm trying to think about other docs we might generate that fall into the same category but it seems like a fine line. Repo-level docs like contribution guidelines or code of conduct might be in the same category but I could also see us wanting those in the website.

BlairCurrey:
> Maybe just `documentation/excalidraw` is the most straightforward for storing them.

That makes the most sense to me
 
> I'm kind of leaning towards just placing it at the root level, thoughts?

I wouldn't really object but I feel like adding stray docs to the root might pollute it a bit. I wonder about these alternatives:

- just adding it to the readme. 
- in `.github` with code of conduct and contribution guidelines. I'm not sure why those docs are there (general github workflow I guess), but renovate seems to fall in the same category. I don't find that location great in terms of discovery, but it does fit in terms of scope. I only found these docs from the link in the readme when I was thinking about where we could potentially insert the `renovate.md` content 

Ultimately I dont have a strong opinion on this and wouldn't object to just dropping it in the root if that makes sense to you.

BlairCurrey (dismissed):

mkurapov:
@BlairCurrey I remembered I needed to update the README links as well

BlairCurrey (dismissed):

packages/documentation/docs/playground/overview.md
```diff
@@ -74,33 +74,27 @@ When clicking on the Account Name, a list of Transactions appears.
 
 - [Rafiki local environment setup](https://github.com/interledger/rafiki/blob/main/README.md#environment-setup)
 - [docker](https://docs.docker.com/get-docker/)
-- [compose plugin](https://docs.docker.com/compose/install/compose-plugin/)
```
BlairCurrey: fixes https://github.com/interledger/rafiki/issues/1423

BlairCurrey: oops, see you already linked.

BlairCurrey (approved):

PR #1454, opened 2023-05-17
feat(backend): add webhook event query

### LINKED ISSUE #234: Add `webhookEvents` query resolver
As discussed in https://github.com/interledger/rafiki/pull/228#issuecomment-1036244056

Support wallets handling (and/or reconciling missed) events via an `events` query resolver.
~Support pagination as well as [querying by event type](https://stripe.com/docs/api/events/list#list_events-types).~ #411 

This can allow failed webhook event delivery attempts to not be retried.

Query by event type

BlairCurrey:

- adds `webhookEvents` gql query with filtering
- adds query to postman collection: https://interledger.postman.co/workspace/blair-interledger~54a25937-6992-4b7f-829f-91ac2142b7c0/pull-request/720e0624-10a9-46b3-9ac6-8195b2e43bc1

Also, I mocked up some changes for mapping gql filters to knex where clauses on a seperate branch (not sure we want to go that far here but wanted to get the idea down on "paper"). https://github.com/interledger/rafiki/compare/bc-events-query...bc-filter-to-knex-where-mapper?expand=1

The gist of it is we parse the info/filter arg in the resolver to create a config for `mapFilterToKnexWhere`. We make a mapper for each filter type (string, float, date, etc.) and use the config to call these accordingly in `mapFilterToKnexWhere`.  Then we use `mapFilterToKnexWhere` to build the query in the service. 

The idea behind this is that it makes a standardized way to filter certain field types (string, float, date, etc.). This would make adding new filters easy, ensure consistency, and lets us test the mapping logic once instead of separately in each service/resolver. Kinda overkill without much filtering but my hope is we can at least stick to a consistent structure in our gql filter types so that we could easily move to something like this in the future if we want.

## Context

fixes #234

## Checklist

- [X] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/schema.graphql
```diff
@@ -770,6 +773,41 @@ type PaymentPointerWithdrawal {
...
+type WebhookEvent implements Model {
+  "Event id"
+  id: ID!
+  "Type of event"
+  type: String!
+  "Stringified JSON data"
+  data: String!
+}
+
+type WebhookEventsConnection {
```
mkurapov, line 785: I'm not sure why `Assets` and `Peers` pagination results were called `AssetsConnection` and `PeersConnection` originally... something like `WebhookEventsPaginationResults` make more sense to me IMO

packages/backend/src/graphql/schema.graphql
```diff
@@ -40,6 +40,9 @@ type Query {
 
   "Fetch an Open Payments incoming payment"
   incomingPayment(id: String!): IncomingPayment
+
+  "Fetch a page of webhook events"
+  webhookEvents(input: GetWebhookEventsInput): WebhookEventsConnection
```
mkurapov, line 45: `GetWebhookEventsInput` non-required because it'll just return the default pagination parameters, right?

BlairCurrey: Yes that's the intention

packages/backend/src/graphql/schema.graphql
```diff
@@ -770,6 +773,41 @@ type PaymentPointerWithdrawal {
...
+  edges: [WebhookEventsEdge!]!
+}
+
+type WebhookEventsEdge {
+  node: WebhookEvent!
+  cursor: String!
+}
+
+input GetWebhookEventsInput {
+  type: String
```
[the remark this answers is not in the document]
BlairCurrey: I wondered about this - I didn't really see them defined in a central place. It looks like currently they could be any string, according to the model. I expected an enum in `WebhookEvents` model.

packages/backend/src/graphql/schema.graphql
BlairCurrey: My understanding is that it's tied to the way we do pagination: https://relay.dev/graphql/connections.htm in our [baseModel.ts](https://github.com/interledger/rafiki/blob/210b201f8261a46f9df83f2dd649bfa791beb5c9/packages/backend/src/shared/baseModel.ts#LL41C1-L47C6) 

I'm not that familiar with cursor based pagination and the relay spec, so `Connection` was not immediately obvious to me. But it looks like the correct name for the specification we implemented.

mkurapov: I see, thanks for the article, now I remember!
There are a few related posts we can consider when adding additional filters:
- https://github.com/graphql/graphql-relay-js/issues/20
- https://medium.com/@conrardy/expanding-relay-cursor-connections-cb6b84fee1e8

BlairCurrey: great resources on this topic, thank you for sharing. I will read these and revisit the `GetWebhookEventsInput` - I suspect there will be some improvements I can make in light of these.

mkurapov: This makes me think about keeping the Relay connections convention in using multiple parameters instead of a unifying Input Object (GetWebhookEventsInput), as much as I prefer using object as inputs. What do you think?

BlairCurrey: Generally speaking I'd like to follow the convention if there is one. I only did a cursory scan of those links but the "Expanding Relay Cursor Connections" article shows this for querying which I think makes sense:

```gql
type Query {
    users(
        first: Int,
        after: String,
        filterBy: UserFilterBy
        orderBy: UserOrderBy
    ): UserConnection
}
```

So the pagination args are all seperate top-level args and then filterting and ordering is scoped to different input objects. This also remains consistent with our other paginated endpoints. If we ever added filtering we could just add a filterBy input and not have to refactor the pagination stuff.

packages/backend/src/graphql/schema.graphql
```diff
@@ -770,6 +784,31 @@ type PaymentPointerWithdrawal {
...
+}
+
+type WebhookEventsEdge {
+  node: WebhookEvent!
+  cursor: String!
+}
+
+input WebhookEventFilter {
+  type: String
+}
```
BlairCurrey, line 810: I am trying to determine the pattern to set with this initial filter. I've been looking at this version but I'm not sure if we want or need the full extent of this filtering: https://medium.com/@conrardy/expanding-relay-cursor-connections-cb6b84fee1e8

It seems pretty comprehensive but I'm not sure to what extent we would want to implement this in this task. Im trying to think about simplified versions (which could be extended further if needed).

Currently the `WebhookEventFilter` assumes an "and" comparison with each filter. For `type`, it assumes an equality comparison. Additional filters wouldn't need to be strictly equal.

I'm not sure if this is extendable enough.  If we wanted to also get events with withdrawalAmount > 0, for example, we could do:

**A)** (current way)
```gql
input WebhookEventFilter {
  type: String
  withdrawalAmount_gt: UInt64
}
```

or maybe

**B)**
```gql
input WebhookEventFilter {
  type: {eq: String, contains: String, startsWith: String etc.}
  withdrawalAmount: {gt: UInt64, lt: UInt64, eq: Uint64 etc.}
}
```

Although in either **A** or **B**, nothing is stopping someone from making a request with all these comparisons. In that case it seems they are assumed to all apply (treating them as "and"). So someone could request all withdrawalAmounts > 0 and equal to -1. Shame on them I guess, but this makes me wonder if there is a better pattern. This is basically 1 step short of the referenced medium article, which takes the additional leap of implementing and/or with a filter like this:

**C)**
```gql
input WebhookEventFilter {
    and: [WebhookEventFilter!]
    or: [WebhookEventFilter!]
    type: {eq: String, contains: String, startsWith: String etc.}
    withdrawalAmount: {gt: UInt64, lt: UInt64, eq: Uint64 etc.}
}
```

Currently I'm leaning towards **B** (without the example extensions), which would look something like:

```gql
input WebhookEventFilter {
  type: {eq: String}
}
```

 This would involve creating some utility function to apply the filter to a knex query builder. If we ever wanted to move to **C** we would add the `and` and `or` fields to schema (trivial, non-breaking) and then implement the logic in the utility function (more complicated). I think there is also an argument to be made for **A** because it's very simple and everything we need right now, but if we intend on extending it more I prefer to start with that pattern.

packages/backend/src/webhook/service.ts
```diff
@@ -160,3 +172,18 @@ export function generateWebhookSignature(
...
+  deps: ServiceDependencies,
+  options?: GetPageOptions
+): Promise<WebhookEvent[]> {
+  const { filter, pagination } = options ?? {}
+  const { type } = filter ?? {}
+  const query = WebhookEvent.query(deps.knex)
+
+  if (type) {
+    query.where({ type })
+  }
```
BlairCurrey, line 186: Still thinking through where and how to handle applying the filters to the query.

Either we manually apply each one in the service like I'm doing here, which seems reasonable for very simple filters like this, or we create some utility function that applies them. I think this utility will be necessary if we do something like:

```gql
input WebhookEventFilter {
  type: {eq: String, contains: String, startsWith: String etc.}
  withdrawalAmount: {gt: UInt64, lt: UInt64, eq: Uint64 etc.}
}
```

Seconds, is where to apply them. If doing the utility that can handle any given filter, I wonder about passing the filter into `baseModel.getPage` and call the utility internally. This would save calling the function in every service and ensure it's always applied correctly (before pagination). Additionally, if we need to encode the filter in the cursor (TBD) then we'll need to pass the filter in anyways.

packages/backend/src/graphql/resolvers/webhooks.ts
```diff
@@ -0,0 +1,42 @@
...
+  ): Promise<ResolversTypes['WebhookEventsConnection']> => {
+    const { filter, ...pagination } = args
+    const webhookService = await ctx.container.use('webhookService')
+    const webhookEvents = await webhookService.getPage({ pagination, filter })
+    // TODO: test this... does getPageInfo work with filters?
+    const pageInfo = await getPageInfo(
+      (pagination: Pagination) =>
+        webhookService.getPage({ pagination, filter }),
+      webhookEvents
+    )
```
BlairCurrey, line 25: Still need to implement the `getPageInfo` tests for this and verify that they pass (I think they should) with the filters and meaningfully test the filters.

packages/backend/src/graphql/schema.graphql
BlairCurrey: As of now, query would look like this:

```gql
query WebhookEvents($first: Int, $after: String, $filter: WebhookEventFilter) {
  webhookEvents(first: $first, after: $after, filter: $filter) {
    edges {
      cursor
      node {
        createdAt
        data
        id
        type
      }
    }
    pageInfo { ... }
  }
}
```

with inputs like this (actual filter TBD: https://github.com/interledger/rafiki/pull/1454#discussion_r1202593563):

```json
{
  "first": ,
  "after": ,
  "filter": {
    "type": "some_type"
  }
}
```

packages/backend/src/graphql/schema.graphql
mkurapov: I like the proposed solution for B. Keeps this pretty simple at the beginning, and leaves the door (relatively) open for any additional filters. I might even suggest something like:

```gql
input WebhookEventFilter { 
    type: {in: String[]} 
}
```
Seems pretty useful in the context of the admin API, which could basically have a box for selecting multiple types at once.

BlairCurrey: `in` makes sense instead of `eq`. Works for 1 or many. Good suggestion.

packages/backend/src/webhook/service.ts
mkurapov: I think this is the right place for them for now :+1:. If we start adding more filters to different models, we could see a pattern, and how we could abstract it away

packages/backend/src/graphql/resolvers/webhooks.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+
+export const webhookEventToGraphql = (
+  webhookEvent: WebhookEvent
+): SchemaWebhookEvent => ({
+  id: webhookEvent.id,
+  type: webhookEvent.type,
+  data: JSON.stringify(webhookEvent.data),
+  createdAt: new Date(+webhookEvent.createdAt).toISOString()
```
mkurapov, line 41:
```suggestion
  createdAt: new Date(webhookEvent.createdAt).toISOString()
```

mkurapov: Since it's a timestamp, not a date I believe. Some other *toGraphql functions do the same, but I'm not sure if that makes sense

packages/backend/src/tests/webhook.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const webhookEventTypes = ['event1', 'event2', 'event3'] as const
+
+export function randomWebhookEvent(): EventPayload {
+  return {
+    id: uuid(),
+    type: sample(webhookEventTypes) as string,
+    data: { field1: faker.string.sample() }
+  }
+}
```
mkurapov, line 17:
```suggestion
export function randomWebhookEvent(overrides?: Partial<EventPayload>): EventPayload {
  return {
    id: uuid(),
    type: sample(webhookEventTypes) as string,
    data: { field1: faker.string.sample() },
    ...overrides
  }
}
```

packages/backend/src/shared/pagination.test.ts
```diff
@@ -333,6 +341,65 @@ describe('Pagination', (): void => {
           }
         )
       })
+      describe('webhookEvents', (): void => {
```
mkurapov, line 344: I'm not sure if someone else would agree with me, but I don't see the need to test `getPageInfo` for all of the resources. It's outside the scope of this test to test everything that `getPage` _could_ be, all this function knows/should know about is the type definition. As long as we give it a proper function to test, I think we should reduce this test suite.

As long as we test `getPage` functions in each of the service, (as you do here: https://github.com/interledger/rafiki/pull/1454/files#diff-f6ae8e955305efe9c101fc4a66e9c257713a099b0018dcc5061a08983b854134R99), we are good to remove this IMO

packages/backend/src/webhook/service.test.ts
```diff
@@ -93,6 +96,94 @@ describe('Webhook Service', (): void => {
...
+        id: uuid(),
+        type: 'pagination_filtering.Z',
+        data: {}
+      }
+    ]
+
+    beforeEach(async (): Promise<void> => {
+      for (const eventInsert of eventInserts) {
+        await WebhookEvent.query(knex).insert(eventInsert)
+      }
```
mkurapov, line 150:
```suggestion
      await WebhookEvent.query(knex).insert(eventInsert)
```

packages/backend/src/webhook/service.test.ts
```diff
@@ -93,6 +96,94 @@ describe('Webhook Service', (): void => {
...
+    getPageTests({
+      createModel: () => createWebhookEvent(deps),
+      getPage: (pagination?: Pagination) =>
+        webhookService.getPage({ pagination })
+      // TODO: Make this work with filtering? eg: webhookService.getPage({ pagination, type: 'event1' })
+      // Currently does not work with filters unless you increase the # of models created
+      // in getPageTests such that there are ~20+ samples for the event type when filtering.
+      // Not a great fix, because technically increasing amount of models created doesnt garuntee
+      // enough isntances of each event type. Would probably be best to somehow pass in created models
+      // or override function to create models in beforeEach.
```
mkurapov, line 109: I think just make sure to test the filters sufficiently in this file (like you have below), and we won't need to update `getPageTests` IMO

BlairCurrey: Still thinking about your comment here... 

Originally my concern was "is there some combination of pagination args and filter args where the page/cursor will be incorrect?" So my thinking was adding filtering to the `getPageTests` would be an exhaustive way to test that. It is certainly important that filters are applied before pagination, but beyond that I dont see a way filtering could impact pagination. Which points towards just keeping the filter test below. I'm just trying to validate I am using a good pattern for that. I suppose an additional filters can just be added to the same `Can paginate and filter` test.

Probably moot, but for whatever it's worth, I think there is a better way to accomplish what I describe in my comment here. Instead of changing the `getPageTests` I could make the `createModel` function we pass in here make a new model for each test webhook event type.

packages/backend/src/webhook/service.test.ts
BlairCurrey: I used a loop to ensure `createdAt` timestamps are different.  If they are all the same the pagination doesn't work reliably. I'm not exactly sure what to make of that (turned into a bit of a rabbit-hole trying to root cause it), but that is independent of filtering or any changes I've made.  I verified that this is just how our pagination works by running the `getPageTests` (no filters) where all the created models have the same `createdAt` dates and seeing that the tests failed.

packages/backend/src/shared/pagination.test.ts
BlairCurrey: I happened to find these tests by accident - very easily could have missed them. My initial reaction was kinda similar in that I didn't think these tests should concern themselves with every resource that uses them. It seems analogous to `getPageTests` so I thought (ideally) there should be something like `getPageInfoTests` that are exported and run in whichever resource uses them. But I suppose I can also see an argument for testing `getPageInfo` once and then just ensuring each resource calls it correctly (side note, but wouldn't this also apply to `getPageTests`? does being a method on the inherited `baseModel` instead of a `utility` function make some difference there?).

In fact that is really what I am interested in with `getPageInfo` is ensuring the right function is passed in (it needs the `pagination` from the `getPageInfo` scope but still needs the filter received by the resolver, which should also be the filter passed into the service's `getPage`). But this doesn't test that.

> As long as we test getPage functions in each of the service

I think it's actually the resolver tests that need it (that's where it's called - service doesn't know about it) but I think I agree in principle.

packages/backend/src/graphql/resolvers/webhooks.ts
BlairCurrey: I tested your suggestion and it works fine. I dont think we need to cast to a number either.

mkurapov:
Some of the `packages/backend/src/graphql/resolvers/webhooks.test.ts` are failing for me locally, is this happening to you?

mkurapov:
So, the tests were failing sporadically for me, and I'm guessing would be hard to replicate on another machine. The TL;DR of it was that models were being created with the same `createdAt` timestamp, which ended up resulting in a different order of models when they were stored in memory: `let modelsCreated: Type[]` in `baseModel.test.ts` vs when they were fetched from the database. The tests would fail when the cursor fell on one of these models with the same `createdAt`. 

To make sure this wouldn't happen, we could either _not_ use `modelsCreated`, but always fetch all of the results from the DB directly to avoid this discrepancy _or_ add some sort of timeout to ensure all of the models would have a different `createdAt`. In a real world scenario, this shouldn't happen since the cursor that is used for pagination across multiple queries would always come from the DB anyway, meaning, the order would be correct, regardless of whether `createdAt` is the same. (In addition, it would be very unlikely for `createdAt` to be the same in the DB, anyway)

mkurapov (approved):
Looks good, approving postman PR as well

BlairCurrey:
> So, the tests were failing sporadically for me, and I'm guessing would be hard to replicate on another machine. The TL;DR of it was that models were being created with the same `createdAt` timestamp, which ended up resulting in a different order of models when they were stored in memory: `let modelsCreated: Type[]` in `baseModel.test.ts` vs when they were fetched from the database. The tests would fail when the cursor fell on one of these models with the same `createdAt`.
> 
> To make sure this wouldn't happen, we could either _not_ use `modelsCreated`, but always fetch all of the results from the DB directly to compare with pagination results _or_ add some sort of timeout to ensure all of the models would have a different `createdAt`. In a real world scenario, this shouldn't happen since the cursor that is used for pagination across multiple queries would always come from the DB anyway, meaning, the order would be correct, regardless of whether `createdAt` is the same. (In addition, it would be very unlikely for `createdAt` to be the same in the DB, anyway)

@mkurapov Thanks for digging into that and sharing. I guess this means all pagination tests are liable to cause these sporadic errors then. Although I'm surprised it's event possible for them to have the same `createdAt` since they are done in a loop. Guess you just have really fast hardware 😄  

Comparing those two options I think adding a the smallest possible delay makes sense. If we didn't check `modelsCreated` and refetched form the database then I guess we'd need to pass in the tablename everywhere (otherwise we wouldn't know which records to get). Changing to fetch all from the db would also behave different than referencing modelsCreated with regards to records created outside the `getPageTests` (setup in `beforeEach` for example). Although I'm not sure if that is necessarily bad. I often try to avoid adding delays because they can be flakey but in this case any amount should work.

PR #1457, opened 2023-05-18
feat(frontend): add confirmation dialog

### LINKED ISSUE #1446: Add confirmation dialog
To ensure user awareness, it is advisable to add a confirmation dialog for irreversible actions. Currently, we only have a designated "danger" zone without such prompts.

raducristianpopa:

When dealing with irreversible actions, we only had a danger zone without any confirmation. This meant that if a user accidentally clicked on a delete button (e.g., delete peer button), the record would be immediately deleted. In this pull request, a confirmation dialog is introduced, along with a prompt that the user must complete before proceeding with irreversible actions.

<details>
  <summary>Confirmation dialog preview:</summary>
  
  

https://github.com/interledger/rafiki/assets/60383558/8aa672f1-b375-4999-a39f-4dc946a4bff9

</details>

## Context

- closes #1446 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/frontend/app/components/ConfirmationDialog.tsx
+++ b/packages/frontend/app/components/ConfirmationDialog.tsx
@@ -6,7 +6,7 @@ import {
   useRef,
   useState
 } from 'react'
-import { ExclamationTriangle, X } from './icons'
+import { ExclamationTriangle, XIcon } from './icons'
 import { Button, Input } from './ui'
 
 export type ConfirmationDialogRef = {
@@ -35,12 +35,14 @@ export const ConfirmationDialog = forwardRef<
 
     const inputRef = useRef<HTMLInputElement>(null)
 
-    const display = () => {
+    const display = (): void => {
       setIsOpen(true)
     }
 
-    const confirmHandler = () => {
-      if (confirmationPrompt === keyword) {
+    const isCorrectPrompt = (): boolean => confirmationPrompt === keyword
+
+    const confirmHandler = (): void => {
+      if (isCorrectPrompt()) {
         setIsOpen(false)
         setConfirmationPrompt('')
         if (onConfirm) {
@@ -49,7 +51,7 @@ export const ConfirmationDialog = forwardRef<
       }
     }
 
-    const cancelHandler = () => {
+    const cancelHandler = (): void => {
       setIsOpen(false)
       setConfirmationPrompt('')
       if (onCancel) {
@@ -99,7 +101,7 @@ export const ConfirmationDialog = forwardRef<
                         onClick={() => cancelHandler()}
                       >
                         <span className='sr-only'>Close</span>
-                        <X className='h-8 w-8' aria-hidden='true' />
+                        <XIcon className='h-8 w-8' aria-hidden='true' />
                       </button>
                     </div>
                     <div>
@@ -141,7 +143,7 @@ export const ConfirmationDialog = forwardRef<
                             intent='danger'
                             onClick={() => confirmHandler()}
                             className='w-full'
-                            disabled={!(confirmationPrompt === keyword)}
+                            disabled={!isCorrectPrompt()}
                           >
                             {confirmButtonText}
                           </Button>
--- a/packages/frontend/app/components/ConfirmationDialog.tsx
+++ b/packages/frontend/app/components/ConfirmationDialog.tsx
@@ -18,7 +18,7 @@ type ConfirmationDialogProps = {
   keyword: string
   message?: string
   confirmButtonText: string
-  onConfirm?: () => void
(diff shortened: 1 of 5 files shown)


packages/frontend/app/routes/peers.$peerId.tsx
```diff
@@ -217,20 +238,22 @@ export default function ViewPeerPage() {
...
               Delete peer
             </Button>
           </Form>
         </DangerZone>
       </div>
+      <ConfirmationDialog
+        ref={dialogRef}
+        onConfirm={onConfirm}
+        title='Delete Peer'
+        keyword='delete peer'
```
mkurapov, line 254:
```suggestion
        keyword={peer.name || 'delete peer'}
```
I think that kind of follows how archiving & deleting repos in Github (as an example).
Too bad `peer.name` is optional...

packages/frontend/app/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,161 @@
...
+                            }
+                          />
+                        </div>
+                        <div className='mt-2'>
+                          <Button
+                            aria-label={confirmButtonText ?? 'Confirm'}
+                            intent='danger'
+                            onClick={() => confirmHandler()}
+                            className='w-full'
+                            disabled={!(confirmationPrompt === keyword)}
```
mkurapov, line 144:
```suggestion
                            disabled={confirmationPrompt.toLowerCase() !== keyword.toLowerCase()}
```

mkurapov: and maybe we can pull this into a method so we can reuse it in L43

packages/frontend/app/components/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,161 @@
+import { Dialog, Transition } from '@headlessui/react'
+import {
+  forwardRef,
+  Fragment,
+  useImperativeHandle,
+  useRef,
+  useState
+} from 'react'
+import { ExclamationTriangle, X } from './icons'
```
mkurapov, line 9:
```suggestion
import { ExclamationTriangle, X as XIcon } from './icons'
```
Maybe a bit more straightfoward

packages/frontend/app/routes/peers.$peerId.tsx
```diff
@@ -47,10 +53,25 @@ export async function loader({ params }: LoaderArgs) {
...
   const response = useActionData<typeof action>()
+  const [formData, setFormData] = useState<FormData>()
+  const submit = useSubmit()
   const { state } = useNavigation()
+  const dialogRef = useRef<ConfirmationDialogRef>(null)
 
   const isSubmitting = state === 'submitting'
 
+  const submitHandler = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
```
mkurapov, line 64: not having `event.preventDefault()` really destroys everything doesn't it 😆

mkurapov (approved):
LGMT, just a few thoughts

packages/frontend/app/routes/peers.$peerId.tsx
raducristianpopa: Yes, that was the pattern I was going for in the first place, but like you mentioned I noticed that `peer.name` is optional.

packages/frontend/app/components/ConfirmationDialog.tsx
raducristianpopa: It does make sense to extract this into a method. However, I believe transforming the string to lowercase would not be beneficial in this case. If we make `peer.name` a required field and the peer's name is 'My first PEER,' I think it would be reasonable to expect the user to input the exact peer name without case modifications.

packages/frontend/app/routes/peers.$peerId.tsx
raducristianpopa: Indeed, removing `event.preventDefault()` would have significant consequences. The form data is processed through the `submit` function, which is obtained from the `useSubmit` hook provided by Remix. When the user clicks the `Delete peer` button, we capture the necessary data to be sent to the server, but we don't immediately submit the form. The actual submission, which is handled within the `onConfirm` function, takes place only when the user clicks the "Delete this peer" button within the dialog.

packages/frontend/app/components/ConfirmationDialog.tsx
mkurapov: I was just following how github archiving/deleting repository UX worked and it was indeed case insensitive, I don't feel strongly about it

mkurapov (approved):

PR #1479, opened 2023-05-26
feat(backend): get page of payment pointers

### LINKED ISSUE #1344: Add GraphQL admin API query to fetch list of payment pointers
Similar to how we can query peers and assets, we can add a query to the admin API that will use pagination for providing a list of payment pointers.

sabineschaller:

- adds possibility to get page of payment pointers via Admin API

## Context

- fixes #1344 

## Checklist

- [x] Tests added/updated
- [x] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment_pointer/service.test.ts
+++ b/packages/backend/src/open_payments/payment_pointer/service.test.ts
@@ -24,6 +24,8 @@ import { initIocContainer } from '../../'
 import { AppServices } from '../../app'
 import { faker } from '@faker-js/faker'
 import { getPageInfo } from '../../shared/pagination'
+import { getPageTests } from '../../shared/baseModel.test'
+import { Pagination } from '../../shared/baseModel'
 
 describe('Open Payments Payment Pointer Service', (): void => {
   let deps: IocContract<AppServices>
@@ -256,6 +258,13 @@ describe('Open Payments Payment Pointer Service', (): void => {
         })
       }
     )
+    describe('getPage', (): void => {
+      getPageTests({
+        createModel: () => createPaymentPointer(deps),
+        getPage: (pagination?: Pagination) =>
+          paymentPointerService.getPage(pagination)
+      })
+    })
   })
 
   describe('onCredit', (): void => {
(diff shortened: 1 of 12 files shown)


packages/backend/src/shared/pagination.test.ts
```diff
@@ -58,6 +59,52 @@ describe('Pagination', (): void => {
     )
   })
   describe('getPageInfo', (): void => {
+    describe('payment pointers', (): void => {
```
mkurapov, line 62: I'm going to have a similar comment that I left on @BlairCurrey 's PR:
- https://github.com/interledger/rafiki/pull/1454#discussion_r1204616058

I think we should have pagination tests for payment pointers live in the `PaymentPointerService` tests

sabineschaller: Happy to change that. I also found it weird that it was where it was but I put it there for consistency.

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -214,6 +215,49 @@ describe('Open Payments Payment Pointer Service', (): void => {
     })
   })
 
+  describe('Payment Pointer pagination', (): void => {
```
BlairCurrey, line 218: Does this supersede the need for the other `getPageTests` call? In `assets` and `peers` we call  `getPageTests` (with different options) in both the resolver and the service.

I was expecting something like this here too:

```ts
    getPageTests({
      createModel: () => createPaymentPointer(deps),
      getPage: (pagination?: Pagination) =>
        paymentPointerService.getPage(pagination)
    })
```

BlairCurrey: I really should have highlighted the line below. I'm referring the cases in the `test.each`.

sabineschaller: Oh, I missed the ones for the service. Thank you!

These are actually 2 different sets of tests, the ones for the resolvers (packages/backend/src/graphql/resolvers/page.test.ts) and the ones for the services (packages/backend/src/shared/baseModel.test.ts).

I'll add the service tests.

BlairCurrey (approved):

Issue #1464, opened 2023-05-22
Prepare for Remix future changes


raducristianpopa:
```
⚠️ REMIX FUTURE CHANGE: The behaviors of 'CatchBoundary' and 'ErrorBoundary' are changing in v2. You can prepare for this change at your convenience with the 'v2_errorBoundary' future flag. For instructions on making this change see https://remix.run/docs/en/v1.15.0/pages/v2#catchboundary-and-errorboundary
⚠️ REMIX FUTURE CHANGE: APIs that provide 'formMethod' will be changing in v2. All values will be uppercase (GET, POST, etc.) instead of lowercase (get, post, etc.) You can prepare for this change at your convenience with the 'v2_normalizeFormMethod' future flag. For instructions on making this change see https://remix.run/docs/en/v1.15.0/pages/v2#formMethod
⚠️ REMIX FUTURE CHANGE: The route 'meta' export signature is changing in v2. You can prepare for this change at your convenience with the 'v2_meta' future flag. For instructions on making this change see https://remix.run/docs/en/v1.15.0/pages/v2#meta
⚠️ REMIX FUTURE CHANGE: The 'serverModuleFormat' config default option will be changing in v2 from 'cjs' to 'esm'. You can prepare for this change by explicitly specifying 'serverModuleFormat: 'cjs''. For instructions on making this change see https://remix.run/docs/en/v1.16.0/pages/v2#servermoduleformat
⚠️ REMIX FUTURE CHANGE: The route file convention is changing in v2. You can prepare for this change at your convenience with the 'v2_routeConvention' future flag. For instructions on making this change see https://remix.run/docs/en/v1.15.0/pages/v2#file-system-route-convention
```

We should also add a typecheck command to the project, and update the Github Action to call it. 
Typescript compilation should pass since v2: https://github.com/remix-run/remix/issues/4371

PR #1491, opened 2023-06-01
feat(frontend): add/display peers and assets liquidity

### LINKED ISSUE #1490: Add/display liquidity for peers and assets

raducristianpopa:

- Display liquidity for peers and assets
- Ability to add liquidity
- Adds Remix's v2 route convention
- Put back `setMessage` function and create a wrapper around it

## Context

- closes #1490 
- in a previous PR, Max had a suggestion to keep the `setMessage` function and to create a wrapper around `redirect` 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


raducristianpopa:
Did I accidentally mess up the lock file?

packages/frontend/app/routes/assets.$assetId.add-liquidity.tsx
```diff
@@ -0,0 +1,79 @@
+import { type ActionArgs } from '@remix-run/node'
```
mkurapov, line 1: is there a way where we can add folders for the routes? i.e. `routes/assets/*` and `routes/peers/*`?

packages/frontend/app/components/ui/LiquidityDialog.tsx
```diff
@@ -0,0 +1,62 @@
+import { Dialog } from '@headlessui/react'
```
mkurapov, line 1: What should go into `components/ui` vs just `components/`? I'm just thinking whether there is a better spot for this component specifically

packages/frontend/app/lib/message.server.ts
```diff
@@ -23,3 +23,24 @@ export const messageStorage = createCookieSessionStorage<
     secure: process.env.NODE_ENV === 'production'
   }
 })
+
+export function setMessage(session: Session, message: Message) {
```
mkurapov, line 27: is this one used?

packages/frontend/app/routes/assets.$assetId.add-liquidity.tsx
mkurapov: https://remix.run/docs/en/main/file-conventions/route-files-v2#folders-for-organization

packages/frontend/app/components/ui/LiquidityDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+            </div>
+            <div>
+              <Dialog.Title
+                as='h3'
+                className='font-semibold leading-6 text-lg text-center'
+              >
+                {title}
+              </Dialog.Title>
+              <div className='mt-2'>
+                <Form method='post' replace preventScrollReset>
```
mkurapov, line 41: Based on the UI flashing, I think this will trigger the `isSubmitting` on the parent `peers.$peerId.tsx` page. Is there a way to decouple this? If no easy way, NABD at all, just something I noticed

mkurapov (dismissed):
Looks good, works well for me! Just comments, nothing that needs to change necessarily

packages/frontend/app/components/ui/LiquidityDialog.tsx
raducristianpopa: Oh, I misplaced this. Basically, `ui` is similar to `shared` - most reusable code blocks and the components that comprise the UI of the applications (buttons, inputs, etc...).

packages/frontend/app/routes/assets.$assetId.add-liquidity.tsx
raducristianpopa: Yes, I'll reorganize them in a PR that will close https://github.com/interledger/rafiki/issues/1464.

packages/frontend/app/lib/message.server.ts
raducristianpopa: Not currently, but I've kept it in case we will need to set a message without redirecting instantly.

packages/frontend/app/components/ui/LiquidityDialog.tsx
raducristianpopa: Not really direct way to decouple them since the dialog is rendered as a child for `peers.$peerId.tsx`.

mkurapov (dismissed):
Looks good. The only thing I saw was when I wanted to add or withdraw "10.00" liquidity, I ended up getting a server error (I'm guessing the graphql doesn't like the decimal input). We can handle this when the amount input is updated to do the right formatting for different assets.

mkurapov (approved):

mkurapov (dismissed):

mkurapov (approved):

PR #1493, opened 2023-06-01
fix(backend): handle open payments with inactive payment pointer

### LINKED ISSUE #1460: Handle Payments to Deactivated Payment Pointers
We introduced the deactivation of payment pointers in #1417

Currently this does not account for open payments. Currently, it's possible to create the corresponding outgoing payment (and quote). We need to define and implement the correct behavior for deactivating payment pointers with open incoming/outgoing payments.

Should open payments be completed prior to deactivating?

Should we disallow deactivating payment pointers with open incoming payments?

Or something else?

BlairCurrey:

- Adds check for active payment pointer in creation of outgoing payment, quote, and incoming payment. throws new `InactivePaymentPointer` error if not active.
- When deactivating, sets `expiresAt` for related open incoming payments to now + grace period (new env var defaulting to 1 day)

## Context

Fixes #1460 

## Checklist

- [X] Tests added/updated
- [X] Documentation added


packages/backend/src/config/app.ts
```diff
@@ -105,7 +105,12 @@ export const Config = {
   privateKey: parseOrProvisionKey(envString('PRIVATE_KEY_FILE', '')),
 
   graphQLIdempotencyKeyLockMs: envInt('GRAPHQL_IDEMPOTENCY_KEY_LOCK_MS', 2000),
-  graphQLIdempotencyKeyTtlMs: envInt('GRAPHQL_IDEMPOTENCY_KEY_TTL_MS', 86400000)
+  graphQLIdempotencyKeyTtlMs: envInt(
+    'GRAPHQL_IDEMPOTENCY_KEY_TTL_MS',
+    86400000
+  ),
+
+  gracePeriodMs: envInt('GRACE_PERIOD_MS', 86400000)
```
BlairCurrey, line 113: This is the grace period for the incoming payments to complete when a payment pointer is deactivated. On deactivation, the `expiresAt` field is updated to now + grace period.

I tried to think of a more specific name but struggled to come up with something that wasn't very verbose. Like `incomingPaymentExpireDelayOnPaymentPointerDeactivation`. Maybe it's best to leave the specificity to the env var documentation (which I have done).

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -119,6 +119,9 @@ async function createIncomingPayment(
   if (!paymentPointer) {
     return IncomingPaymentError.UnknownPaymentPointer
   }
+  if (!paymentPointer.isActive) {
+    return IncomingPaymentError.InactivePaymentPointer
+  }
```
BlairCurrey, line 124: Do I need to do this same check in the `incoming_remote` create? I'm not quite clear on what a remote incoming payment is. Does that use an external payment pointer by chance (and therefore doesnt need this check)?

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -127,12 +135,28 @@ async function updatePaymentPointer(
...
+      update.deactivatedAt = new Date()
+      const expiresAt = new Date(Date.now() + deps.config.gracePeriodMs)
+
+      await IncomingPayment.query(deps.knex)
+        .patch({ expiresAt })
+        .where('paymentPointerId', id)
+        .whereIn('state', [
+          IncomingPaymentState.Pending,
+          IncomingPaymentState.Processing
+        ])
```
BlairCurrey, line 152: I figured these incoming payment states constitute "open" payments based off the names/comments in the enum. Is that accurate?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -95,8 +99,19 @@ async function createOutgoingPayment(
...
+      const paymentPointer = await deps.paymentPointerService.get(
+        paymentPointerId
+      )
+      if (!paymentPointer) {
+        throw OutgoingPaymentError.UnknownPaymentPointer
+      }
+      if (!paymentPointer.isActive) {
+        throw OutgoingPaymentError.InactivePaymentPointer
+      }
+
```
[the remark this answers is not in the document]
BlairCurrey: confirmed w/ Sabine that the `receiver.incomingPayment` pointer does not need to be checked here (will be handled in incoming payment creation).

packages/backend/src/open_payments/payment_pointer/service.ts
mkurapov: This is accurate:
`IncomingPaymentState.Processing` -> payment is currently being streamed into. Typically doesn't happen in our P2P scenarios, since we just receive a 
`IncomingPaymentState.Pending` -> payment was created, no $ has streamed into it yet.

packages/backend/src/config/app.ts
mkurapov: Oof yeah, I don't think there is a good & short name for this. I do think verbose is better though, maybe `paymentPointerDeactivationPaymentGracePeriodMs`?

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -183,6 +187,37 @@ describe('Open Payments Payment Pointer Service', (): void => {
       ).resolves.toEqual(updatedPaymentPointer)
     })
 
+    test('Deactivating updates expirey dates of existing incoming payments', async (): Promise<void> => {
+      const paymentPointer = await createPaymentPointer(deps)
+
+      const nowMs = new Date('2023-06-01T00:00:00Z').getTime()
+      jest.spyOn(Date, 'now').mockImplementation(() => nowMs)
```
mkurapov, line 194: I think you could use
```js
    jest.useFakeTimers()
    jest.setSystemTime(nowMs)
```

packages/backend/src/open_payments/payment_pointer/service.test.ts
```diff
@@ -183,6 +187,37 @@ describe('Open Payments Payment Pointer Service', (): void => {
...
+      })
+
+      await paymentPointerService.update({
+        id: paymentPointer.id,
+        status: 'INACTIVE'
+      })
+      const incomingPaymentUpdated = await incomingPayment.$query(knex)
+
+      expect(incomingPaymentUpdated.expiresAt.getTime()).toEqual(
+        expiresAt.getTime() + config.gracePeriodMs - duration
```
mkurapov, line 217: If you want control over `config.gracePeriodMs` without consequences for other tests, you could use the `withConfigOverride` function.

packages/backend/src/config/app.ts
BlairCurrey: I think `paymentPointerDeactivationPaymentGracePeriodMs` is a good descriptive name... I dont mind how long it is personally. I will change to that.

packages/backend/src/open_payments/payment_pointer/service.test.ts
BlairCurrey: That is good to know, although I don't think that would apply here. I am just reading the grace period here, not changing the value.

packages/backend/src/open_payments/payment_pointer/service.test.ts
BlairCurrey: changed to use the fake timers

packages/backend/src/open_payments/payment_pointer/service.ts
BlairCurrey: @sabineschaller Could you chime in here? My understanding is that we want to push the `expiresAt` back so that we ensure the payment has at least that amount of time to complete. 

I think this turns into a no-op if we filter out the existing `expiresAt` that are less than the new `expiresAt`. From what I can see the max expiration date for incoming payments created via openapi is the creation time + 1 day. https://github.com/interledger/rafiki/blob/2a3242a355551044230e02597cbf9c98cf211347/packages/backend/src/open_payments/payment/incoming/routes.ts#L32

So if we only get the incoming payments with an `expiresAt` that is less than now + 1 day, we will never have incoming payments because they have at max some past date + 1 day.

mkurapov: Just to confirm from Lisbon: we wanted 90 days in the future as default for expiryDate for incoming payments, correct?

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -127,12 +135,20 @@ async function updatePaymentPointer(
...
     const update: UpdateInput = { publicName }
-    if (status) {
-      update.deactivatedAt = status === 'INACTIVE' ? new Date() : null
+    const paymentPointer = await PaymentPointer.query(deps.knex)
+      .findById(id)
+      .throwIfNotFound()
+
+    if (status === 'INACTIVE' && paymentPointer.isActive) {
+      update.deactivatedAt = new Date()
+      deactivateOpenIncomingPaymentsByPaymentPointer(deps, id)
```
mkurapov, line 144:
```suggestion
      await deactivateOpenIncomingPaymentsByPaymentPointer(deps, id)
```

packages/backend/src/open_payments/payment_pointer/service.ts
```diff
@@ -266,6 +282,23 @@ async function createWithdrawalEvent(
...
+    .patch({
+      expiresAt: new Date(
+        Date.now() + deps.config.paymentPointerDeactivationPaymentGracePeriodMs
+      )
+    })
+    .where('paymentPointerId', paymentPointerId)
+    .whereIn('state', [
+      IncomingPaymentState.Pending,
+      IncomingPaymentState.Processing
+    ])
```
mkurapov, line 299: I think after the 90 day default changes, we _should_ include expiryDate as a `where` clause in here

packages/backend/src/open_payments/payment_pointer/service.ts
BlairCurrey: I think we actually settled on 30 days.

packages/documentation/docs/integration/deployment.md
```diff
@@ -36,51 +36,51 @@ $ helm install ...
...
+| `AUTH_SERVER_GRANT_URL`                             | `http://127.0.0.1:3006`                                     | endpoint to on the Open Payments Auth Server to request a grant                                            |
+| `AUTH_SERVER_INTROSPECTION_URL`                     | `http://127.0.0.1:3007`                                     | endpoint to on the Open Payments Auth Server to introspect an auth token                                   |
+| `CONNECTOR_PORT`                                    | `3002`                                                      | STREAM/ILP connector port                                                                                  |
+| `DATABASE_URL`                                      | `postgresql://postgres:password@localhost:5432/development` | Postgres database URL                                                                                      |
+| `PAYMENT_POINTER_DEACTIVATION_PAYMENT_GRACE_PERIOD` | `86400000`                                                  | Milliseconds into the future to set expiration of open incoming payments when deactivating payment pointer |
```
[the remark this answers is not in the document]
BlairCurrey: yes, missed that. I re-ordered it.

packages/backend/src/open_payments/payment/incoming/routes.test.ts
```diff
@@ -142,19 +146,25 @@ describe('Incoming Payment Routes', (): void => {
...
-      })
-      ctx.request.body['expiresAt'] = new Date(
-        Date.now() + MAX_EXPIRY + 1000
-      ).toISOString()
-      await expect(incomingPaymentRoutes.create(ctx)).rejects.toHaveProperty(
-        'message',
-        'expiry too high'
-      )
-    })
+    test.each(Object.values(IncomingPaymentError).map((err) => [err]))(
```
mkurapov, line 149:
```suggestion
    test.each(Object.values(IncomingPaymentError)(
```

packages/backend/src/config/app.ts
```diff
@@ -111,9 +111,11 @@ export const Config = {
   ),
 
   paymentPointerDeactivationPaymentGracePeriodMs: envInt(
-    'PAYMENT_POINTER_DEACTIVATION_PAYMENT_GRACE_PERIOD',
+    'PAYMENT_POINTER_DEACTIVATION_PAYMENT_GRACE_PERIOD_MS',
     86400000
-  )
+  ),
+
+  paymentExpiryMaxMs: envInt('PAYMENT_EXPIRY_MAX_MS', 2592000000) // 30 days
```
mkurapov, line 118:
```suggestion
  incomingPaymentExpiryMaxMs: envInt('INCOMING_PAYMENT_EXPIRY_MAX_MS', 2592000000) // 30 days
```
💅

packages/backend/src/open_payments/payment_pointer/service.ts
mkurapov: Still need this I believe ^

packages/backend/src/open_payments/payment/incoming/routes.test.ts
BlairCurrey: committed locally, suggested was just missing a closing parenthesis

packages/backend/src/open_payments/payment_pointer/service.ts
BlairCurrey: @mkurapov added

packages/documentation/docs/integration/deployment.md
```diff
@@ -56,7 +56,7 @@ $ helm install ...
 | `OUTGOING_PAYMENT_WORKERS`                             | `4`                                                         | number of workers processing outgoing payment requests                                                                       |
 | `OUTGOING_PAYMENT_WORKER_IDLE`                         | `200`                                                       |
 | milliseconds                                           |
-| `PAYMENT_EXPIRY_MAX_MS`                                | `86400000`                                                  | Maximum milliseconds into the future incoming payments can be set to on creation. Default: 1 day                             |
+| `INCOMING_PAYMENT_EXPIRY_MAX_MS`                       | `86400000`                                                  | Maximum milliseconds into the future incoming payments can be set to on creation. Default: 1 day                             |
```
mkurapov, line 59: so srry, same thing for alphabetization 😅

BlairCurrey: oh geeze, no need to apologize... didn't think about it after renaming :satisfied: Also made me discover that I had the default values swapped for `PAYMENT_POINTER_DEACTIVATION_PAYMENT_GRACE_PERIOD_MS` and `INCOMING_PAYMENT_EXPIRY_MAX_MS`

PR #1504, opened 2023-06-09
feat(frontend): add webhook events page

### LINKED ISSUE #1503: Add webhook events page
The page should contain a table with the columns: `id`, `type` and `data`.

raducristianpopa:

- 

## Context

- closes #1503 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


sabineschaller:
@raducristianpopa I know you started this PR so I don't know if Github will allow you to review it, but could you try to review it anyways? You could just leave comments or something.

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <Button
+              aria-label='go to next page'
+              disabled={!webhooks.pageInfo.hasNextPage}
+              onClick={() => {
+                navigate(
+                  nextPageUrl +
+                    getTypeFilterParams(type, undefined).replace('?', '&')
+                )
+              }}
+            >
```
raducristianpopa, line 151: We can construct the `nextPageUrl` alongside with the type filters in the `loader` function.

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    name:
+                      value.charAt(0).toUpperCase() +
+                      value.slice(1).replace(/[_.]/g, ' '),
+                    value: value,
+                    action: () => {
+                      navigate(`${getTypeFilterParams(type, value)}`)
+                    }
+                  }))
+                ]}
+              />
```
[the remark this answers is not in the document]
raducristianpopa: Yes! You're on the right track with the checkbox approach.

packages/frontend/app/routes/webhooks.tsx
JoblersTune: I'm not sure I'm following exactly what you want me to change here but I have changed the code a bit as well so that this getTypeFilterParams actually sets the filters now so it would need to be in the component.

Let me know if that resolves your concern or if there is something else I can change.

raducristianpopa: Currently, `nextPageUrl` and `previousPageUrl` are sent from the `loader`. By this meaning, we construct them on the server and then send them to the client.

https://github.com/interledger/rafiki/blob/da0263a44dc774caa0bbceb95ce73331a220a7dc/packages/frontend/app/routes/webhooks.tsx#L32-L38

Instead of performing this operation on the client with `getTypeFilterParams`, we can add all the necessary search params in the loader function as well. As an example, something that have just came into my mind, LOC from 32 to 34 (👆) can be changed to something like this:

```ts
if (webhooks.pageInfo.hasPreviousPage && webhooks.pageInfo.startCursor) {
    const searchParams = new URLSearchParams()
    searchParams.set('before', webhooks.pageInfo.startCursor)
    if (type) searchParams.set('type', type.toString())
    previousPageUrl = `/webhooks?${searchParams}`
}
```

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,164 @@
...
+    selectedTypeSet.has(selectedType)
+      ? selectedTypeSet.delete(selectedType)
+      : selectedTypeSet.add(selectedType)
+    if (selectedTypeSet.size > 0) {
+      searchParams.set('type', Array.from(selectedTypeSet).join(','))
+    } else {
+      searchParams.delete('type')
+    }
+    setSearchParams(searchParams)
+  }
```
raducristianpopa, line 69: I don't really know what to say about using a `Set` for this. 🤔 Let me know what you think about this approach:
```suggestion
  const [, setSearchParams] = useSearchParams()
  const navigate = useNavigate()

  function setTypeFilterParams(selectedType: WebhookEventType): void {
    const selected = type

    if (!selected.includes(selectedType)) {
      selected.push(selectedType)
      setSearchParams(new URLSearchParams({ type: selected.join(',') }))
      return
    }

    setSearchParams(() => {
      const newParams = selected.filter((t) => t !== selectedType).join(',')
      if (newParams.length === 0) {
        return ''
      }

      return new URLSearchParams({
        type: newParams
      })
    })
  }
```
```

File: packages/frontend/app/lib/validate.server.ts [hunk 4]
```diff
@@ -15,6 +16,22 @@ export const paginationSchema = z
...
   .partial()
 
+export const paginationSchemaTest = z.object({
+  after: z.string().uuid(),
+  before: z.string().uuid(),
+  first: z.coerce.number().int().positive(),
+  last: z.coerce.number().int().positive()
+})
+
+export const paginationSearchParams = paginationSchemaTest.partial()
```
raducristianpopa, line 26:
```suggestion
```

packages/frontend/app/lib/validate.server.ts
```diff
@@ -15,6 +16,22 @@ export const paginationSchema = z
...
+export const webhooksSearchParams = paginationSearchParams
+  .merge(
+    z.object({
+      type: z.array(z.nativeEnum(WebhookEventType))
+    })
+  )
+  .partial()
+
```
raducristianpopa, line 34:
```suggestion
export const webhooksSearchParams = paginationSchema
  .merge(
    z.object({
      type: z.array(z.nativeEnum(WebhookEventType)).default([])
    })
  )

```

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,164 @@
...
+            <div className='flex-1'>
+              <h3 className='text-2xl leading-10'>Webhook Events</h3>
+            </div>
+          </PageHeader>
+          <div className='p-1'>
+            <h3 className='text-lg font-bold'>Filters</h3>
+            <div className='flex items-center'>
+              <PopoverFilter
+                label='Event type'
+                values={type ? type : ['all']}
```
raducristianpopa, line 85:
```suggestion
                values={type.length > 0 ? type : ['all']}
```

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,164 @@
...
+          </Table>
+          <div className='flex items-center justify-between p-5'>
+            <Button
+              aria-label='go to previous page'
+              disabled={!webhooks.pageInfo.hasPreviousPage}
+              onClick={() => {
+                searchParams.delete('after')
+                searchParams.set('before', previousPageUrl)
+                setSearchParams(searchParams)
+              }}
```
raducristianpopa, line 143: If we construct `previousPageUrl` on the server, we can directly call `navigate(previousPageUrl)`.

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,164 @@
...
+              Previous
+            </Button>
+            <Button
+              aria-label='go to next page'
+              disabled={!webhooks.pageInfo.hasNextPage}
+              onClick={() => {
+                searchParams.delete('before')
+                searchParams.set('after', nextPageUrl)
+                setSearchParams(searchParams)
+              }}
```
raducristianpopa, line 154: Ditto above.

packages/frontend/app/routes/webhooks.tsx
JoblersTune: From what I read my understanding is that it's better to use setSearchParams when you're only changing the search parameters. What is the advantage of using navigate here instead?

packages/frontend/app/routes/webhooks.tsx
JoblersTune: type is passed in from the loader where it is either of type WebhookEventType[] or undefined. The loader gets this value from the Graphql response. So if the array was empty it would simply be omitted. But if you'd rather I check length then we can change this to `values={type && type.length > 0 ? type : ['all']}`

raducristianpopa: Check the above changes for the validation schema. We are now defaulting to an empty array if no `type` is provided.

packages/frontend/app/routes/webhooks.tsx
JoblersTune: Yes, that's much better, thanks.

packages/frontend/app/components/Filters.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const PopoverFilter = ({
+  label,
+  options,
+  values
+}: PopoverFilterProps) => {
+  return (
+    <Popover className='relative mt-2'>
+      <Popover.Button className='relative w-[400px] cursor-default rounded-md bg-white py-1.5 pl-3 pr-10 text-left text-gray-900 shadow-sm ring-1 ring-inset ring-gray-300 focus:outline-none focus:ring-2 focus:ring-indigo-600 sm:text-sm sm:leading-6'>
```
raducristianpopa, line 23:
```suggestion
      <Popover.Button className='relative w-[400px] cursor-default rounded-md bg-white py-1.5 pl-3 pr-10 text-left text-gray-900 shadow-sm ring-1 ring-inset ring-pearl focus:outline-none focus:ring-2 focus:ring-[#F37F64] sm:text-sm sm:leading-6'>
```

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,170 @@
...
+    nextPageUrl = `/webhooks?${nextPageSearchParams.toString()}`
+  }
+
+  return json({ webhooks, previousPageUrl, nextPageUrl, type })
+}
+
+export default function WebhookEventsPage() {
+  const { webhooks, previousPageUrl, nextPageUrl, type } =
+    useLoaderData<typeof loader>()
+  const [, setSearchParams] = useSearchParams()
```
raducristianpopa, line 57:
```suggestion
  const [searchParams, setSearchParams] = useSearchParams()
```

packages/frontend/app/routes/webhooks.tsx
```diff
@@ -0,0 +1,170 @@
...
+              {webhooks.edges.length ? (
+                webhooks.edges.map((webhook) => (
+                  <Table.Row key={webhook.node.id}>
+                    <Table.Cell>{webhook.node.id}</Table.Cell>
+                    <Table.Cell>{webhook.node.type}</Table.Cell>
+                    <Table.Cell>
+                      <Button
+                        aria-label='view webhook data'
+                        state={{ data: webhook.node.data }}
+                        to={'/webhooks/data'}
```
raducristianpopa, line 129: Related to the above suggestion. This way we can maintain the search parameters.
```suggestion
                        to={`/webhooks/data${
                          searchParams ? `?${searchParams}` : null
                        }`}
```

packages/frontend/app/routes/webhooks.data.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export default function WebhookEventData() {
+  const location = useLocation()
+  const navigate = useNavigate()
+  const state = location.state as { data: string }
+  const dismiss = () => navigate('/webhooks', { preventScrollReset: true })
+
+  useEffect(() => {
+    if (!state) dismiss()
+  }, [])
```
raducristianpopa, line 15:
```suggestion
  const location = useLocation()
  const navigate = useNavigate()
  const [searchParams] = useSearchParams()
  const state = location.state as { data: string }
  const dismiss = () =>
    navigate(`/webhooks${searchParams ? `?${searchParams}` : null}`, {
      preventScrollReset: true
    })

  useEffect(() => {
    if (!state) dismiss()
  }, [])
```

sabineschaller (approved):
@raducristianpopa is happy with this 😉

PR #1506, opened 2023-06-12
feat(grant): add filter by identifier

### LINKED ISSUE #1505: Filter grants by payment pointer (identifier)
For grant management user should see only his own grants.
List grants graphql endpoint should allow filtering by identifier (payment pointer)

dragosp1011:

- Add filter of grants by identifier defined in grant access. (As identifier does not exist on grant itself in db)
- Add graphql query to get grant by id

## Context

fixes #1505 
## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/graphql/resolvers/grant.test.ts
+++ b/packages/auth/src/graphql/resolvers/grant.test.ts
@@ -51,12 +51,6 @@ describe('Grant Resolvers', (): void => {
         grants.push(grant)
       }
 
-      const filter = {
-        identifier: {
-          in: ['test']
-        }
-      }
-
       const query = await appContainer.apolloClient
         .query({
           query: gql`
@@ -71,8 +65,57 @@ describe('Grant Resolvers', (): void => {
                 }
               }
             }
+          `
+        })
+        .then((query): GrantsConnection => {
+          if (query.data) {
+            return query.data.grants
+          } else {
+            throw new Error('Data was empty')
+          }
+        })
+      expect(query.edges).toHaveLength(2)
+      query.edges.forEach((edge, idx) => {
+        const grant = grants[idx]
+        expect(edge.cursor).toEqual(grant.id)
+        expect(edge.node).toEqual({
+          __typename: 'Grant',
+          id: grant.id,
+          state: grant.state
+        })
+      })
+    })
+
+    test('Can filter grants', async (): Promise<void> => {
+      const grants: GrantModel[] = []
+      const identifier = 'https://example.com/test'
+      for (let i = 0; i < 2; i++) {
+        const grant = await createGrant(deps, identifier)
+        grants.push(grant)
+      }
+
+      const filter = {
+        identifier: {
+          in: [identifier]
+        }
+      }
+
+      const query = await appContainer.apolloClient
+        .query({
+          query: gql`
+            query grants($filter: GrantFilter) {
+              grants(filter: $filter) {
+                edges {
+                  node {
+                    id
+                    state
+                  }
+                  cursor
+                }
+              }
+            }
           `,
-          variables: filter
+          variables: { filter }
         })
         .then((query): GrantsConnection => {
           if (query.data) {
@@ -116,6 +159,20 @@ describe('Grant Resolvers', (): void => {
                   createdAt
                   actions
                   type
+                  limits {
+                    receiver
+                    sendAmount {
+                      value
+                      assetCode
+                      assetScale
+                    }
+                    receiveAmount {
+                      value
+                      assetCode
+                      assetScale
+                    }
+                    interval
+                  }
                 }
                 createdAt
               }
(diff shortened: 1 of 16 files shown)


packages/auth/src/graphql/schema.graphql
```diff
@@ -40,6 +43,14 @@ input PaginationInput {
   last: Int
 }
 
+input GrantFilter {
+  identifier: FilterString
```
mkurapov, line 47: @sabineschaller what do you think about always requiring `identifier` in any Open Payments Grant Request? Does it make sense to allow issuing grants for any payment pointer?

packages/auth/src/graphql/schema.graphql
```diff
@@ -1,6 +1,9 @@
 type Query {
   "Fetch a page of grants."
-  grants(input: PaginationInput): GrantsConnection!
+  grants(input: PaginationInput, filter: GrantFilter): GrantsConnection!
```
mkurapov, line 3: @BlairCurrey what do you think about the `input: PaginationInput` object? vs like you have for the webhook events:

https://github.com/interledger/rafiki/blob/b96ddf1a865143a5c87b3705594b3cd5c2f48287/packages/backend/src/graphql/schema.graphql#L57-L68

mkurapov (commented):
Looks good, just want to get some thoughts from Blair & Sabine

packages/auth/src/graphql/schema.graphql
BlairCurrey: Although my initial instinct was a pagination input like what we have in the `auth` package, we settled on separate `after`, `before`, `first`, `last` args for consistency with the relay spec and conventions around implementing it. I think that's probably still the right approach and ideally we standardize across packages.

packages/auth/src/graphql/resolvers/grant.ts
```diff
@@ -17,7 +17,8 @@ export const getGrants: QueryResolvers<ApolloContext>['grants'] = async (
   ctx
 ): Promise<ResolversTypes['GrantsConnection']> => {
   const grantService = await ctx.container.use('grantService')
-  const grants = await grantService.getPage(args.input)
+  const { filter, input: pagination } = args
+  const grants = await grantService.getPage(pagination, filter)
   const pageInfo = await getPageInfo(
     (pagination: Pagination) => grantService.getPage(pagination),
```
BlairCurrey, line 23: I think the `pageInfo` will be innacurate here because the actual page is filtering, but the `getPage` call that `pageInfo` uses does not filter. I think we want to pass in `(pagination: Pagination) => grantService.getPage(pagination, filter)` to `pageInfo` instead.

See https://github.com/interledger/rafiki/blob/main/packages/backend/src/graphql/resolvers/webhooks.ts. 

This is beyond the scope of this task, but just as an aside, I think this a good reason for returning `pageInfo` from `getPage` itself IMO. Instead of deriving it from a separate function call. I believe that was what was suggested here https://github.com/interledger/rafiki/pull/342#discussion_r879841635

packages/auth/src/graphql/schema.graphql
BlairCurrey: My opinion isn't very firm here TBH. Other than being consistent. I mostly preferred separate because that's how we saw the relay spec implemented elsewhere and it's also less to break/refactor (but I guess that's not a major consideration). I'm still leaning that way for those reasons but I'm wondering what you think @mkurapov because I think all of our initial instincts were to not use separate args.

mkurapov: Even though I agree w @dragosp1011 (I like having the object type instead of the listed arguments), let's stay consistent and have the 4 variables instead. We can then do a swoop of all of the changes in one go to convert to object at some point in the future.

packages/auth/src/graphql/resolvers/grant.ts
mkurapov: Good catch :+1:

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -48,6 +50,13 @@ describe('Grant Resolvers', (): void => {
         const grant = await createGrant(deps)
         grants.push(grant)
       }
+
+      const filter = {
+        identifier: {
+          in: ['test']
+        }
+      }
```
mkurapov, line 58: This isn't being passed into the actual graphql query

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -84,6 +94,88 @@ describe('Grant Resolvers', (): void => {
     })
   })
 
+  describe('Grant By id Queries', (): void => {
+    let grant: GrantModel
+    beforeEach(async (): Promise<void> => {
+      const createdGrant = await createGrant(deps)
+      grant = (await getGrantById(deps, createdGrant.id)) || createdGrant
```
mkurapov, line 101: Won't we always have `createdGrant` defined?

packages/auth/src/graphql/schema.graphql
```diff
@@ -95,3 +129,6 @@ enum GrantState {
   "grant was rejected"
   REJECTED
 }
+
+scalar UInt8
+scalar UInt64
```
mkurapov, line 134: These scalars would need to be added to the resolvers if you would like to use them. See: `packages/backend/src/graphql/resolvers/index.ts`. Otherwise, you can leave adding limits for a separate PR (if necessary)

PR #1554, opened 2023-07-05
ci: add check for gql docs generation

### LINKED ISSUE #1526: Fix docs:graphql:generate
Currently the documentation `docs:graphql:generate` points to the wrong directory (should be `api-reference` instead of `reference`). Additionally, we should add a CI step to check if anything was generated and fail if so. Similar to our graphql check: https://github.com/interledger/rafiki/blob/7a3078c0ec7bafd3888e7bb49ae5c16d4272c083/.github/workflows/lint_test_build.yml#L21-L42

BlairCurrey:

- adds CI step to check if graphql api doc generation was missed

## Context

Fixes #1526

To verify it works, see:
- commit without gql changes (passes): https://github.com/interledger/rafiki/actions/runs/5465439452/jobs/9948935691
- commit with gql changes but without generating docs (fails): https://github.com/interledger/rafiki/actions/runs/5465866657/jobs/9949978766
- commit with gql changes and generating docs (passes): https://github.com/interledger/rafiki/actions/runs/5466158888/jobs/9950698667?pr=1554
## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/.github/workflows/lint_test_build.yml
+++ b/.github/workflows/lint_test_build.yml
@@ -41,6 +41,22 @@ jobs:
         if: steps.verify-changed-files.outputs.files_changed == 'true'
         run: exit 1
 
+  documenation:
+    runs-on: ubuntu-22.04
+    needs: checkout
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/rafiki/env-setup
+      - name: generate graphql api documentation
+        run: pnpm --filter documentation docs:graphql:generate
+      - name: verify changed files
+        uses: tj-actions/verify-changed-files@v14
+        id: verify-changed-files
+
+      - name: fail if docs were generated
+        if: steps.verify-changed-files.outputs.files_changed == 'true'
+        run: exit 1
+
   backend:
     runs-on: ubuntu-22.04
     needs: [checkout]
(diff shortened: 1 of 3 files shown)


.github/workflows/lint_test_build.yml
```diff
@@ -41,6 +41,22 @@ jobs:
...
         run: exit 1
 
+  documenation:
+    runs-on: ubuntu-22.04
+    needs: checkout
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/workflows/rafiki/env-setup
+      - name: generate graphql api documentation
+        run: pnpm --filter documentation docs:graphql:generate && pnpm format
```
mkurapov, line 51:
```suggestion
        run: pnpm --filter documentation docs:graphql:generate
```
Maybe?

BlairCurrey: I believe we need the format. Without any gql changes, generating the docs will have changes due to not being formatted. Meaning the currently committed api docs have been formatted. This step would always fail because there would always be changes. Additionally, I think committing without formatting would cause the `pnpm check` cmd to fail:

https://github.com/interledger/rafiki/blob/d0a1a47daeb1467376dc97cc56029b57e475b95d/.github/workflows/lint_test_build.yml#L12-L19

mkurapov: I don't think it affects `pnpm checks` since that runs before, but now I understand it _would_ always have changes, regardless.

mkurapov (approved):

Issue #1510, opened 2023-06-20
Soft delete instead of hard delete grants


mkurapov:
Currently, whenever we revoke a grant (via the auth Admin API), we are hard deleting it in the DB. However, we also have the `REVOKED` status on the grant model. We should use the status instead of hard deleting the grant in the DB whenever we revoke a grant.

Issue #1428, opened 2023-05-15
Q2 documentation updates


mkurapov:
README:

- [x] Update Rafiki planning call time
- [ ] Rename “Interledger Docs” to “Interledger Specs” (and update the link)

`architecture.md`

- [ ] Update excalidraw diagram
- [ ] Add GraphQL admin API description to `auth` service section

`admin-api.md`

- [x] Include auth admin API (moved to #1651)

`connector.md`

- [ ] Remove "often for Web Monetization" from Local Stream Server
- [ ] Remove “and more!” from Packet Rejection (and remove “including”)

`accounts-and-transfers.md`

- [ ] Link double-entry accounting to Wiki page
- [ ] Make it clear that PP and IP liquidity accounts get created on the fly, while asset and peer created when they get created via GraphQL calls

`grant-interaction.md`

- [ ] Put the diagram at the beginning
- [x] Endpoints for the API calls should link to the OpenAPI specs (that will be added separately) (moved to #1668)

`key-registry.md`

- [x] Should be moved to the Open Payments website
- [x] Should be turned into a document explaining what it means for an Admin to add keys to a payment pointer (moved to #1667)

PR #1576, opened 2023-07-11
feat(auth): revoke grant instead of hard delete

BlairCurrey:

-  Removed hard deletion of `grants`/`access`/`accessTokens` in favor of a `grant.state` of `GrantState.Revoked`. 
- Updated where `grants`/`access`/`accessTokens` are read to either filter out or handle revoked grants
- Renamed some "delete" verbiage to "revoke" to better reflect this behavior
- added a `notIn` filter to the `getGrants` gql query so that certain states (such as revoked) can be filtered out

## Context

Fixes https://github.com/interledger/rafiki/issues/1510

Some discussion about my thought process: https://github.com/interledger/rafiki/pull/1576#discussion_r1259844570

This PR involves filtering out/handling revoked grants where they previously would not have been returned because they had been deleted. I generally preferred filtering them out at the service level (because that's effectively what was happening before). In a few places grants with a revoked state were already handled so for those methods I just added checks everywhere else they were called.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/service.ts
```diff
@@ -142,12 +146,14 @@ async function rejectGrant(
...
 ): Promise<boolean> {
-  const deletion = await Grant.query(deps.knex).delete().where({ continueId })
-  if (deletion === 0) {
+  const update = await Grant.query(deps.knex)
+    .patch({ state: GrantState.Revoked })
+    .where({ continueId })
+  if (update === 0) {
     deps.logger.info(
       `Could not find grant corresponding to continueId: ${continueId}`
     )
```
BlairCurrey, line 159: Wasn’t sure about changing this one as well. After sitting with it a bit longer I think maybe we should continue hard-deleting here. Or, soft-deleting via a new `deletedAt` field (see note in PR body). This would mean there isnt parity between gql api and the rest api (one is delete, one is state update). Or change the http method from delete to post/put but probably not…

Currently in my PR the delete HTTP method will update the state and then we just handle grants with a `REVOKED` state differently depending on where we get them, which definitely seems wrong.

sabineschaller:
Should we still have a delete function for grant requests that were never granted? I'm thinking about a scenario where a client requests to access my account (maybe even a thief) and I don't know the client so I deny the grant. If this is some sort of attack, I don't want these grants to show up in the list of revoked grants. I do, however, want to see grants that I actively disabled/revoked.

packages/auth/src/grant/service.ts
sabineschaller: A revoke grant REST endpoint should probably be a PATCH, not a DELETE. We actually call the DELETE endpoint "Cancel grant" [in the docs](https://docs.openpayments.guide/reference/delete-continue), which is for a client to withdraw his grant request or to surrender the grant. I think you are right in that in such a scenario, we shouldn't revoke but we should have a `deletedAt` field. And revocation should then only happen via the admin APIs because the user itself would revoke using his ASE's interface.

BlairCurrey: After your comment and looking at the GNAP spec I think a `deletedAt` field for the DELETE method is what I'm going to do for this, but it seems like the spec has changed and our implementation doesn't align on a few things. Then again this is the first I've waded into the auth server and the GNAP spec so correct me if I'm wrong or missing some context.

Here is the specification for this DELETE route: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-14

Of note it specifies this route is for "Revoking a Grant Request" (previously "Canceling" like we use). This suggests there isn't the distinction between deleting and revoking like I thought. Not so simple to just revoke though because it specifies:
   - updating grant state to `finalized` (not `revoked`). Which doesn't exist in our code. It seems like the grant states are now (`processing`, `pending`, `approved`, `finalized`) instead of our (`pending`, `granted`, `revoked`, `rejected`). These cut across many operations and in terms of `revoked` and `finalized`, im not sure they are just interchangable.
   - should disable all token rotation and other token management functions. I think we are _effectively_ doing this because the tokens are on the grant that are deleted (so they cant be rotated). This is why I want to do a soft-delete instead of using grant state (`revoked` or `finalized`) but it doesn't seem like this is ultimately the right way to handle revocation.

BlairCurrey: There may not be too much to make GNAPy after all. I noticed that we are deleting access and access token as well on grant deletion (via a postgres `ON DELETE CASCADE`). Which means I can't just draw the line at a soft delete of grants anyways. I ended up changing both the `DELETE /continue` and `revokeGrant` routes to update the grant state to `REVOKED` and handled or filtered out grants/access/access tokens with a revoked grant state on the read side. All the token and grant openapi requests should remain the same but we can now get the revoked grants via the gql admin api.

I still wonder about the states I see in the spec (`processing`, `pending`, `approved`, `finalized`) vs. what we have (`pending`, `granted`, `revoked`, `rejected`). Perhaps our `GrantStates` are not meant to reflect the states here: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-11`#section-1.5 For example, we have no concept of `finalized` but it was in version 11 of the spec. Still, the spec is pretty explicit about when to update the grant state to these values. We have `pending` and `approved` might map pretty clearly to `granted` but `finalized` doesn't really seem to have an analog.

BlairCurrey:
> Should we still have a delete function for grant requests that were never granted? I'm thinking about a scenario where a client requests to access my account (maybe even a thief) and I don't know the client so I deny the grant. If this is some sort of attack, I don't want these grants to show up in the list of revoked grants. I do, however, want to see grants that I actively disabled/revoked.

We should be able to distinguish between these via the grant state value `GrantState.Revoked` and `GrantState.Rejected`. This makes me wonder more about the states described in the GNAP spec (specifically, not having a rejected state).

packages/auth/src/grant/routes.test.ts
```diff
@@ -927,43 +1032,47 @@ describe('Grant Routes', (): void => {
...
             method: 'GET'
           },
-          { id: v4(), nonce: grant.interactNonce }
+          { id: grant.interactId, nonce: grant.interactNonce }
         )
+
         await expect(
           grantRoutes.interaction.details(ctx)
         ).rejects.toMatchObject({
-          status: 404
```
BlairCurrey, line 966: This diff is hard to follow but basically I removed a duplicate test (one of https://github.com/interledger/rafiki/blob/main/packages/auth/src/grant/routes.test.ts#L910, https://github.com/interledger/rafiki/blob/main/packages/auth/src/grant/routes.test.ts#L950)
and added the `Cannot get grant details for revoked grant` test.

packages/auth/src/grant/service.ts
```diff
@@ -142,12 +146,14 @@ async function rejectGrant(
...
+async function revokeGrant(
   deps: ServiceDependencies,
   continueId: string
 ): Promise<boolean> {
-  const deletion = await Grant.query(deps.knex).delete().where({ continueId })
-  if (deletion === 0) {
+  const update = await Grant.query(deps.knex)
+    .patch({ state: GrantState.Revoked })
+    .where({ continueId })
+  if (update === 0) {
```
BlairCurrey, line 156: At this point, there are probably still access/accessTokens in the db associated with this revoked grant. Previously they would have been deleted via a cascade with the grant. Just like grants, I am filtering out such `access`/`accessTokens` on the read side but also considered keeping a hard-delete here. I decided against it but I do wonder if these records will accumulated to rather large volumes unnecessarily.

packages/auth/src/grant/service.ts
```diff
@@ -16,8 +16,12 @@ import { AccessService } from '../access/service'
...
 
+interface FilterGrantState {
+  notIn?: GrantState[]
+}
 interface GrantFilter {
   identifier?: FilterString
+  state?: FilterGrantState
 }
 
 export interface GrantService {
```
BlairCurrey, line 27: I found the following unused service methods when auditing the reads for access, access tokens, and grants:

- `GrantService.get`
- `GrantService.getByInteraction`
- `AccessTokenService.get`

I left them to be safe but wonder if maybe we should remove them.

packages/auth/src/accessToken/service.ts
BlairCurrey, line 1: I wonder if I should change the revoke function in this file to set `expiresIn` to now instead of hard-deleting?

previously access token and access were hard deleted with the grant delete. Now they are no longer hard deleted, except in this revoke method here. 

https://github.com/interledger/rafiki/pull/1576/files#diff-a2d9fc0dba9cf7d48aae084d4b7ecacd25cc06a0767f06a9a9b83f8b934c5a3fR93-R102

sabineschaller:
Is it ok if we hold off for until Nathan is back?

packages/auth/src/grant/service.ts
njlie: `GrantService.getByInteraction` we can probably get rid of, I believe it has been supplanted by `GrantService.getByInteractionSession`. We could also remove `(GrantService/AccessTokenService).get` too, I wrote those anticipating they would be used but as I wrote out the rest of the GNAP implementation they didn't end up getting used.

packages/auth/src/grant/service.ts
njlie: I think we should add the distinction between the `processing` and `pending` states. Respectively, one indicates the AS is determining what to do next with the grant and the other indicates that the grant is awaiting interaction.

As for `finalized`, I would consider `granted`, `revoked`, and `rejected` to be the AS's (granulized) concept of GNAP's `finalized`. It seems like the concept of a `finalized` state irrespective of _how_ a grant was finalized is all that matters for GNAP, but for our purposes we need more specific reasons why. I would think we should maybe modify the Grant model so that `granted/revoked/rejected` is replaced by `finalized` and add another field indicating how it was finalized. I think it would make this easier to follow as an implementation of GNAP that way, and we could keep information that we need.

I agree with @sabineschaller that this is a separate, larger issue that should be worked on separately from this, though.

packages/auth/src/accessToken/service.ts
njlie: I think we should still hard delete. A client should still be able to rotate expired tokens, so we still need to distinguish expired tokens vs. revoked ones.

packages/auth/src/grant/service.ts
njlie: I think we should still hard delete here. Revoking a grant is supposed to [revoke all associated access tokens](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#name-revoking-a-grant-request).

njlie:
> > Should we still have a delete function for grant requests that were never granted? I'm thinking about a scenario where a client requests to access my account (maybe even a thief) and I don't know the client so I deny the grant. If this is some sort of attack, I don't want these grants to show up in the list of revoked grants. I do, however, want to see grants that I actively disabled/revoked.
> 
> We should be able to distinguish between these via the grant state value `GrantState.Revoked` and `GrantState.Rejected`. This makes me wonder more about the states described in the GNAP spec (specifically, not having a rejected state).

My stance (to summarize [this comment](https://github.com/interledger/rafiki/pull/1576#discussion_r1269956283)) is to bring the `GrantStates` in line with GNAP and augment the grant model with a field describing how the grant was finalized. @sabineschaller let me know if you still would like to have a call on this still to discuss but for now that's the feeling I have towards this

njlie (changes requested):
See https://github.com/interledger/rafiki/pull/1576#discussion_r1262976270, I think hard deleting, or otherwise invalidating the access tokens of a revoked grant, should still be done here.

packages/auth/src/grant/service.ts
BlairCurrey: I'll change it to hard-delete

njlie (dismissed):

packages/auth/src/graphql/schema.graphql
```diff
@@ -45,12 +45,17 @@ type GrantEdge {
...
+  state: FilterGrantState
 }
 
 input FilterString {
   in: [String!]!
 }
 
+input FilterGrantState {
+  notIn: [GrantState!]!
+}
```
mkurapov, line 57: Why `notIn` instead of `in`?

packages/auth/src/grant/routes.ts
```diff
@@ -233,7 +233,7 @@ async function getGrantDetails(
   }
   const { id: interactId, nonce } = ctx.params
   const grant = await grantService.getByInteractionSession(interactId, nonce)
-  if (!grant) {
+  if (!grant || grant.state === GrantState.Revoked) {
```
mkurapov, line 236: Just a suggestion: if we want to keep this check in the service layer, we could pass an optional arg to `getByInteractionSession`, something like `includeRevoked` (false by default). We could also include `.whereNot('state', 'REVOKED')` in cases where it is false in the DB lookup.

packages/auth/src/grant/routes.ts
```diff
@@ -416,12 +416,12 @@ async function deleteGrant(
...
   }
   const grant = await deps.grantService.getByContinue(continueId, continueToken)
-  if (!grant) {
+  if (!grant || grant.state === GrantState.Revoked) {
     ctx.throw(404, { error: 'unknown_request' })
   }
 
-  const deletion = await deps.grantService.deleteGrant(continueId)
-  if (!deletion) {
+  const revoked = await deps.grantService.revokeGrant(continueId)
```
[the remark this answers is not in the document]
BlairCurrey: I suppose we don't need `revokeGrant` (by `continueId`) since we have the grant id here as well. Both `revokeGrant` and `revokeGrantById` existed already as `deleteGrant` and `deleteGrantById`. I dont see the harm in changing it to one `revokeGrant` function that takes the grant id.

packages/auth/src/graphql/schema.graphql
BlairCurrey: This endpoint now returns revoked grants, when previously it did not. So the thinking with `notIn` was giving the client a way to return the same set of results as before.

packages/auth/src/grant/routes.ts
BlairCurrey: That might make more sense. In most cases `GrantState.Revoked` is treated the same as the grant not existing, but there are a few places it's explicitly handled. This would apply to `getByContinue` as well.

BlairCurrey: i implemented this on `getByContinue` and `getByInteractionSession`

packages/auth/src/graphql/schema.graphql
BlairCurrey: i added and `in` filter for good measure

packages/auth/src/grant/routes.ts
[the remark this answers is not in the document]
BlairCurrey: changed to just do `revokeGrant(grantId)`

mkurapov (dismissed):

mkurapov:
At first I was confused why there weren't any documentation failures for not updating the Admin API types, but I realized this is only the auth Admin API...

BlairCurrey:
> At first I was confused why there weren't any documentation failures for not updating the Admin API types, but I realized this is only the auth Admin API...
> 
> Made this as a result:
> 
> * [Add documentation for Auth Admin API #1651](https://github.com/interledger/rafiki/issues/1651)

I had the same experience initially and asked Sabine about it. She pointed out this issue: https://github.com/interledger/rafiki/issues/1428

Wasn't sure how to present it initially but what you've depicted in the issue makes sense.

mkurapov (approved):

PR #1620, opened 2023-07-21
docs: update and add accounting and liquidity docs

### LINKED ISSUE #1429: Document how liquidity moves within Rafiki
We should have a document that gives an overview of how liquidity moves around in Rafiki. 

- What is peer liquidity? What is asset liquidity?
- How ASEs interact with liquidity, how to add it, withdraw it, deposit it, etc
- Diagram similar to the flow of funds

sabineschaller:

- 

## Context

- fixes #1429 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/documentation/docs/concepts/accounting/accounts-and-transfers.md
+++ b/packages/documentation/docs/concepts/accounting/accounts-and-transfers.md
@@ -30,39 +30,206 @@ Rafiki transfers perform double-entry accounting. Every transfer increases both 
 
 #### Deposits
 
+##### Depositing Asset Liquidity
+
 | Debit Account | Credit Account  |
 | ------------- | --------------- |
 | Settlement    | Asset Liquidity |
+
+- Example: depositing 100 USD
+
+<table>
+<tr><th>USD Settlement Acc. </th><th>USD (Asset) Liquidity Acc.</th></tr>
+<tr><td style="text-align: left;">
+
+| Debit | Credit |
+| ----- | ------ |
+| 100   |        |
+
+</td><td style="text-align: right;">
+
+| Debit | Credit |
+| ----- | ------ |
+|       | 100    |
+
+</td></tr>
+</table>
+
+##### Depositing Peer Liquidity
 
 | Debit Account | Credit Account |
 | ------------- | -------------- |
 | Settlement    | Peer Liquidity |
 
+- Example: peering relationship in USD, depositing 100 USD
+
+<table>
+<tr><th>USD Settlement Acc. </th><th>Peer Liquidity Acc.</th></tr>
+<tr><td style="text-align: left;">
+
+| Debit | Credit |
+| ----- | ------ |
+| 100   |        |
+
+</td><td style="text-align: right;">
+
+| Debit | Credit |
+| ----- | ------ |
+|       | 100    |
+
+</td></tr>
+</table>
+
+##### Depositing Outgoing Payment Liquidity
+
 | Debit Account | Credit Account   |
 | ------------- | ---------------- |
 | Settlement    | Outgoing Payment |
 
+- Example: depositing 35 USD
+
+<table>
+<tr><th>USD Settlement Acc. </th><th>Outgoing Payment Liquidity Acc.</th></tr>
+<tr><td style="text-align: left;">
+
+| Debit | Credit |
+| ----- | ------ |
+| 35    |        |
+
+</td><td style="text-align: right;">
+
+| Debit | Credit |
+| ----- | ------ |
+|       | 35     |
+
+</td></tr>
+</table>
+
 #### Withdrawals
+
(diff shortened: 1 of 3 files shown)


njlie (commented):
LGTM

njlie (approved):
LGTM

packages/documentation/docs/concepts/accounting/accounts-and-transfers.md
BlairCurrey, line 1: Not from your changes, but I noticed the `[Account Servicing Entity](./glossary/a ccount-servicing-entity)` link is not rendering correctly. There is an extra space in the filename. See https://rafiki.dev/docs/concepts/accounting/accounts-and-transfers#settlement-account

image

packages/documentation/docs/concepts/accounting/liquidity.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Example
+
+A configured peer _Cloud Nine Wallet_ within Rafiki has a peer liquidity of 100 USD. Rafiki can send packets up to 100 USD to payment pointers issued by _Cloud Nine Wallet_. Once that liquidity is used up, we should settle with _Cloud Nine Wallet_ and then reset their liquidity to 100 USD.
+
+### Event Liquidity
+
+When Open Payments incoming or outgoing payments are created, a liquidity account is created within the accounting database. Liquidity needs to be added to an outgoing payment before the payment can be processed. The Account Servicing Entity is notified to add liquidity via the `outgoing_payment.created` event, hence the name _Event Liquidity_. Similarly, packets that are received for an incoming payment increase its liquidity. The Account Servicing Entity is notified to withdraw that liquidity via the `incoming_payment.completed` event.
```
BlairCurrey, line 29:
```suggestion
When incoming or outgoing Open Payments are created, a liquidity account is created within the accounting database. Liquidity needs to be added to an outgoing payment before the payment can be processed. The Account Servicing Entity is notified to add liquidity via the `outgoing_payment.created` event, hence the name _Event Liquidity_. Similarly, packets that are received for an incoming payment increase its liquidity. The Account Servicing Entity is notified to withdraw that liquidity via the `incoming_payment.completed` event.
```

packages/documentation/docs/concepts/accounting/liquidity.md
```diff
@@ -0,0 +1,192 @@
...
+```json
+{
+  "input": {
+    "id": "b97fd85a-126e-42ef-b40d-1a50a70ffa6f",
+    "assetId": "7b8b0f65-896d-4403-b7ba-2e24bf20eb35",
+    "amount": "100",
+    "idempotencyKey": "b97fd85a-126e-42ef-b40d-1a50a70ffa6f"
+  }
+}
+```
```
BlairCurrey, line 59: I like the in-context examples because they are clear, but I also wonder about introducing these examples which depend on our current schema/input definitions. I think this is more clear than just linking to the api mutation but now we will need to remember to update these if there are ever any changes. I suppose these are pretty stable but its something to keep the back of our minds.

packages/documentation/docs/concepts/accounting/liquidity.md
sabineschaller: I'm not so sure about this suggestion, tbh. I'd rather not use the term "Open Payments" for payments but only for the standard.

packages/documentation/docs/concepts/accounting/liquidity.md
sabineschaller: They may not be stable because I just realized that `id` and `idempotencyKey` are basically serving the same purpose and we should drop the `id` from the mutation input. I find those examples to be more informing, too, then linking to the schema. Ideally, we could link to a schema with examples with our new docs framework.

packages/documentation/docs/concepts/accounting/liquidity.md
```diff
@@ -0,0 +1,194 @@
+# Liquidity
+
+Rafiki implements a clearing protocol - the [Interledger Protocol](../../reference/glossary.md#interledger-protocol). As such, it does not hold liquidity but keeps track liquidity moving through it.
```
mkurapov, line 3:
```suggestion
Rafiki implements a clearing protocol - the [Interledger Protocol](../../reference/glossary.md#interledger-protocol). As such, it does not hold liquidity but keeps track of liquidity moving through it.
```

packages/documentation/docs/concepts/accounting/liquidity.md
BlairCurrey: Now after out meeting I also wonder, maybe we can link to postman queries? The schema is a nice single source of truth but I do think the usage examples are even clearer.

BlairCurrey (dismissed):

BlairCurrey (approved):

Issue #1502, opened 2023-06-09
Add GraphQL query to fetch list of payments


raducristianpopa:
In order to display all the payments (incoming and outgoing) in the Admin UI, we need a unified query that has the ability to paginate over all of them.

BlairCurrey:
I've been trying to scope this out. This will take some changes to the db. We can't simply combine incoming/outgoing page results together (results would be wrong, we would have 2 cursors, and front end can already do this anyways). I see a few options which I dont have a strong preference on at the moment:

- a new `combinedPaymentsView`. This view is basically a stored query to combine all the payments via a `UNION ALL` of incoming and outgoing (with new field specfying incoming/outgoing). The union will happen whenever the view is used so it's more expensive to read from than a table that just has them all combined, but it is a direct extension of our current schema.
  - slight variation on this is a materialized view which I don't think we want because it trades some payment read-heaviness for write-heaviness and adds some complexity with managing refreshing the view

- refactor `incomingPayments`, `outgoingPayments` tables to inherit from a new parent `payments` table. The parent table would have the columns common to both tables and the child tables would have the non-shared columns. You can read/write to the child tables as we currently do but then we can also read from the parent table to get all of them without doing a `UNION ALL` every time. However, the parent records would not have the non-shared columns (such as `incomingPayment.incomingAmountValue`) which we probably want. In which case we need to `JOIN` which probably mitigates that advantage. 

With either of these we can make a combined payments model using the view/parent table which we can then use our existing `getPage` method on.

Initially I prefered the new parent `payments` table with inheritance until realizing we'd likely need to join against the child tables. Performance-wise I'm not sure there is any meaningful difference, and its probably pre-mature to focus on that anyways. The table inheritance change would be a bit more complicated, although I think it does express a natural relationship between the data.

BlairCurrey:
Some other details shared by Radu in our conversation:

This wil powering a list of payments which should only display the shared fields and they should be filterable by payment pointer and type (incoming or outgoing). From there you can navigate to the individual items which could then use the existing incoming/outcoming resolvers for more details.

Sorting was also discussed but was left out of this as it hasn't been implemented anywhere yet and I wasn't sure what that would take or how complicated that would be in combination with our pagination logic (which controls the sort order in places).

PR #1666, opened 2023-07-27
feat(backend): add payments query

BlairCurrey:

- Adds `Payments` query with `paymenterPointerId` and `type` (incoming or outgoing) filters
- Adds `combinedPaymentsView` which is a view of a union of incoming and outgoing payments
- Adds `CombinedPayment` model and service for getting payments only

## Context

Closes https://github.com/interledger/rafiki/issues/1502

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/open_payments/payment/combined/model.ts
BlairCurrey, line 1: The `CombinedPayment` model is not like our other models. Instead of representing a record in an actual table, it represents a record from a view, which combines the incoming/outgoing payments. It's basically only good for getting a page of combined payments. It can't be written to, updated, or deleted. 

I made a model for it to use our `getPage` function. Initially I tried to use the logic from `getPage` without a model because I wanted to avoid this weirdness. However I found this difficult because the pagination uses a objection query builder which are tied to objection models. All of our pagination tests rely on models as well so I didn't see much way around this.

packages/backend/src/graphql/schema.graphql
```diff
@@ -66,6 +66,20 @@ type Query {
     "Filter webhook events based on specific criteria."
     filter: WebhookEventFilter
   ): WebhookEventsConnection!
+
+  "Fetch a page of combined payments"
+  payments(
```
BlairCurrey, line 71: I named everything `CombinedPayments` in the backend because it more accurately reflects how the data is modeled (it's a composite of two separate things). I used `payments` in the gql schema because I dont think the frontend really cares about that. Still, I'm considering if `combinedPayments` is good here too for explicitness.

packages/backend/migrations/20230718142550_create_combined_payments_view.js
```diff
@@ -0,0 +1,41 @@
...
+          .select([...sharedColumns, knex.raw("'INCOMING' AS type")])
+          .unionAll((union) => {
+            union
+              .select([...sharedColumns, knex.raw("'OUTGOING' AS type")])
+              .from('outgoingPayments')
+          })
+      )
+    }
+  )
+}
```
BlairCurrey, line 34: I debated about the best way to combine the payments in this comment https://github.com/interledger/rafiki/issues/1502#issuecomment-1638438973.

Ultimately I went with the union all view because it's the simplest, cannot accidentally be written to, and didn't have the performance concerns I originally thought it might. I was concerned it would perform a union of all the records but with pagination it does not. Confirmed with some against a lot of dummy data.

packages/backend/src/open_payments/payment/combined/service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const query = CombinedPayment.query(deps.knex)
+
+  if (filter?.paymentPointerId?.in && filter.paymentPointerId.in.length) {
+    query.whereIn('paymentPointerId', filter.paymentPointerId.in)
+  }
+
+  if (filter?.type?.in && filter.type.in.length) {
+    query.whereIn('type', filter.type.in)
+  }
```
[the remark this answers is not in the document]
BlairCurrey: I think we probably could do that but I'm not sure if it's necessary. Initially when we spoke I had concerns about the performance of the union all view but [they ended up being unfounded](https://github.com/interledger/rafiki/pull/1666#discussion_r1276803379). At this point I dont see an advantage to calling the underlying `incomingPaymentService` or `outgoingPaymentService`.

njlie (dismissed):

packages/backend/src/open_payments/payment/combined/service.ts
```diff
@@ -0,0 +1,58 @@
...
+interface CombinedPaymentFilter {
+  paymentPointerId?: FilterString
+  type?: FilterString
+}
+interface GetPageOptions {
+  pagination?: Pagination
+  filter?: CombinedPaymentFilter
+}
+
+export interface CombinedPaymentService {
```
mkurapov, line 18: We should add this to the `AppServices` interface so it's typed when it's used

packages/backend/src/open_payments/payment/combined/model.ts
mkurapov: I agree with you on the decision, I think imagined some sort of "paymentQueryService" that does the raw query as you describe, but the pagination is the tough part here, which I think an objection model is useful for and a good solution.

My only main comment is whether there is a way to hide away the possible functions on the `CombinedPayment` model, so we don't end up exposing functions that we can't use. I'm imagining something like a `Pick<Type, 'query'>`, but on a class instead of an interface, if that is possible.

BlairCurrey: Also correct me if I'm wrong or misunderstanding the specifics of the suggestion but I think we'd need to do the `Pick` (actual or effective) on the query builder, not the model like `Pick<Type, 'query'>`, since the insert, delete, etc. methods are on the query (`CombinedPayment.query().delete()` etc)

packages/backend/src/open_payments/payment/combined/service.ts
BlairCurrey: fixed, thank you

mkurapov (approved):

PR #1677, opened 2023-07-31
fix: use sending asset for base rate when creating quote

### LINKED ISSUE #1676: Fix quote creation rates request
## Context
Currently, whenever we are creating a quote, the rates service is using the receiver's asset code:
https://github.com/interledger/rafiki/blob/8a71d063f4c39d5720ccb2715f89188dbe4d0dbf/packages/backend/src/open_payments/quote/service.ts#L220-L228

This is incorrect, as the base asset should be in relation to the sender's asset code. i.e. if we are sending from USD to EUR, the base asset code should be USD (the sending asset code).

## Todos
- [ ] Fix how rates are called in quote service

mkurapov:

In a discussion with @raducristianpopa, he noticed 
1. we should be using the sending asset code as base rate when creating the quote, instead of (incorrectly) the receiving asset code, and
2. Now that we are using a base rate when asking for rates from the ASE, the rates service caching does not take into account different possible base's. This means when we request rates with base USD, we cache the result, and will continue return USD even if we ask for EUR as base rate. 

This change:
- uses the sending asset code as base rate when creating a quote
- adds in-memory caching as a data store
- updates rate service to use in-memory caching data store & take into account base rate when caching & returning rates

## Context

Fixes #1676

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/quote/service.test.ts
+++ b/packages/backend/src/open_payments/quote/service.test.ts
@@ -39,7 +39,7 @@ describe('QuoteService', (): void => {
   let appContainer: TestContainer
   let quoteService: QuoteService
   let knex: Knex
-  let paymentPointerId: string
+  let sendingPaymentPointer: MockPaymentPointer
   let receivingPaymentPointer: MockPaymentPointer
   let config: IAppConfig
   let quoteUrl: URL
@@ -97,7 +97,7 @@ describe('QuoteService', (): void => {
     const paymentPointer = await createPaymentPointer(deps, {
       assetId: sendAssetId
     })
-    paymentPointerId = paymentPointer.id
+    sendingPaymentPointer = paymentPointer
     const { id: destinationAssetId } = await createAsset(deps, destinationAsset)
     receivingPaymentPointer = await createPaymentPointer(deps, {
       assetId: destinationAssetId,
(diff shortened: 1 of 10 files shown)


packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -97,7 +97,7 @@ describe('QuoteService', (): void => {
     const paymentPointer = await createPaymentPointer(deps, {
       assetId: sendAssetId
     })
-    paymentPointerId = paymentPointer.id
+    sendingPaymentPointer = paymentPointer
```
raducristianpopa, line 100:
```suggestion
    sendingPaymentPointer = await createPaymentPointer(deps, {
      assetId: sendAssetId
    })
```

And we can remove:
https://github.com/interledger/rafiki/blob/a0fce7d005678b26bff3d99353600bbf277a21e5/packages/backend/src/open_payments/quote/service.test.ts#L97-L99

raducristianpopa (approved):
LGTM. A comment:

packages/backend/src/open_payments/quote/service.test.ts
mkurapov: Yup, for sure. Updated.

sabineschaller (commented):
This looks more correct to me then the old version. Why did we include a base asset in the first place?

mkurapov:
> This looks more correct to me then the old version. Why did we include a base asset in the first place?

To support cross-currency payments: for payments between USD > EUR, we _could_ do some math in the rates service to calculate the EUR > USD rate, but often in real world scenarios the rate from USD > EUR does _not_ equal 1 / EUR > USD. Providing the base rate allows the ASE to return their correct exchange rate for each direction.

raducristianpopa (approved):

Issue #1655, opened 2023-07-26
Make AS Interactions more GNAP-like


njlie:
Right now, interaction information is held in a number of fields on the `Grant` model. However, GNAP allows for a grant to require multiple interactions, so it may be best to make a new interaction model that has a relation to the grant it is for. Each interaction would have its own entry in the database, each with its own state management (`pending`, `accepted`, `rejected`, maybe `expired`?)

PR #1710, opened 2023-08-07
feat(auth): use GNAP-ier grant states

### LINKED ISSUE #1646: Update Grant States to new GNAP spec
We discovered that our current grant states are not in line with new states in the GNAP specification in https://github.com/interledger/rafiki/pull/1576

We should update our `GrantState`s to reflect the latest GNAP specificatoin: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-11#section-1.5

More specifically, we need to add the processing and finalized state and likely add a new field to track revoke and rejected as detailed by Nathan's comment here: https://github.com/interledger/rafiki/pull/1576#discussion_r1269956283

njlie:

- Change `GrantStates` to include `Processing` & `Finalized` and remove `Revoked` and `Rejected`
- Rework the transitions to `Pending` and `Approved` states
- Add new `finalizationReason` field to Grant model

## Context

Fixes #1646
GNAP defines a [certain way](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#name-protocol-flow) in which Grants move between states. This PR seeks to update our Grant model to reflect this more closely.

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/auth/src/interaction/routes.test.ts
```diff
@@ -41,7 +47,7 @@ describe('Interaction Routes', (): void => {
   let grant: Grant
 
   const generateBaseGrant = () => ({
-    state: GrantState.Pending,
+    state: GrantState.Processing,
```
BlairCurrey, line 50: Are the "base" grants in the tests supposed to reflect the state of a grant when its created or does it not really matter? That was my impression but not sure. If so, I think the `BASE_GRANT`/`generateBaseGrant()` states could be updated to `Processing` (from `Pending`) in these files:

access/service.test.ts
accessToken/routes.test.ts
accessToken/service.test.ts
grant/routes.test.ts

I changed a few and re-ran and it doesn't look like it causes any failures.

njlie: Before this PR I would say that base grants should reflect its state when it's created. Your question though, did get me thinking about if these base grants ought to reflect the state the grant _should_ be in when the tests are happening (e.g. should test grant states reflect when they are created for `access` tests or reflect when they're approved for `accessToken` tests?).

Fortunately with this PR implementing the `Processing` state, it can be used in tests as a sort of "generic" state. A grant would be processing when it is created and may be processing if it was previously approved, but the client modified the grant after that fact.

BlairCurrey (approved):

packages/auth/src/grant/model.ts
```diff
@@ -16,13 +16,19 @@ export enum FinishMethod {
...
   Revoked = 'REVOKED',
   Rejected = 'REJECTED'
 }
 
+export enum GrantState {
+  Processing = 'PROCESSING',
+  Pending = 'PENDING',
+  Approved = 'APPROVED',
+  Finalized = 'FINALIZED'
+}
```
sabineschaller, line 30: Are both of these defined in the GNAP spec or just the grant states?

packages/auth/src/grant/model.ts
```diff
@@ -147,3 +154,17 @@ export function isPendingGrant(grant: Grant): grant is PendingGrant {
     grant.interactNonce
   )
 }
+
+export function isRejectedGrant(grant: Grant): boolean {
+  return !!(
+    grant.state === GrantState.Finalized &&
+    grant.finalizationReason === GrantFinalization.Rejected
```
sabineschaller, line 161: Can there be a scenario where `finalizationReason` is `Rejected` but `state !== Finalized`?

packages/auth/src/grant/service.ts
```diff
@@ -24,14 +25,15 @@ export interface GrantService {
     interactId: string,
     interactNonce: string,
     options?: GetByInteractionSessionOpts
-  ): Promise<PendingGrant | undefined>
-  issueGrant(grantId: string): Promise<Grant>
+  ): Promise<InteractiveGrant | undefined>
+  markPending(grantId: string): Promise<Grant>
```
sabineschaller, line 29: What would be a reason to manually mark a grant as pending?

njlie: I thought it would make sense to mark a grant as pending only when an interaction is initiated, so it would have been called on the `startInteraction` route. Looking at the GNAP spec more closely though, it implies that a Grant should be marked pending as soon as it knows whether or not interaction is required, irrespective of if it has been initiated or not. I'll remove this in light of that and have the Grant be marked `Pending` on creation if it needs interaction.

BlairCurrey (approved):

packages/auth/src/grant/model.ts
sabineschaller: I'm just wondering why we are checking both but I guess it's a sanity check of our data, too.

sabineschaller (approved):
LGTM

packages/auth/src/accessToken/service.ts
```diff
@@ -78,7 +78,7 @@ async function introspect(
   if (isTokenExpired(token)) {
     return undefined
   } else {
-    if (!token.grant || token.grant.state === GrantState.Revoked) {
+    if (!token.grant || isRevokedGrant(token.grant)) {
```
mkurapov, line 81: At first I was thinking we should check for `isRejectedGrant` also, but I'm assuming we can't issue a token if a grant is finalized /rejected

packages/auth/src/interaction/routes.ts
```diff
@@ -123,7 +123,7 @@ async function startInteraction(
   const { config, grantService } = deps
   const grant = await grantService.getByInteractionSession(interactId, nonce)
 
-  if (!grant) {
+  if (!grant || grant.state !== GrantState.Processing) {
```
mkurapov, line 126:
```suggestion
  if (!grant || grant.state !== GrantState.Pending) {
```
The eCommerce postman example wasn't working for me when starting the interaction, this is what I needed (and I think makes sense from the GNAP diagram).

The difference between Processing and Pending isn't that clear until I saw the diagram 😅 . Now that we don't use `Processing` anywhere, I was hoping we could remove it, but I guess otherwise we wouldn't be able to have a "starting" state...

mkurapov (changes requested):

packages/auth/src/interaction/routes.ts
njlie: A direction I was thinking of taking the AS in the future is to have an explicit `process` function on the Grant service and have it behave a bit more like a state machine. So something like `create` or `finishInteraction` would leave the grant in `Processing` and then call `process` to determine what state to transition the grant next.

I don't think there will be enough information with our current schema to do this until I finish https://github.com/interledger/rafiki/issues/1655

njlie: I'll fix this, though.

packages/auth/src/accessToken/service.ts
njlie: That's correct.

mkurapov (approved):

PR #1716, opened 2023-08-08
refactor: update rates service

### LINKED ISSUE #1711: Update rates to be more intuitive
## Context
Currently, the rates in the MASE are "inversed", and not necessarily intuitive:
https://github.com/interledger/rafiki/blob/bf3bc665c219200baa62cdb79f7bd0118d6caeee/localenv/mock-account-servicing-entity/app/routes/rates.ts#L8-L21

Reading this, IMO, it seems as though the rate is 1 USD = 1.16 EUR, which is not true (although possible of course). What is true is 1 USD = 1 / 1.16 ~ 0.8620 EUR. 

This is the (trickled-up) result of what `rates` are when it comes to the ILP quote. From the ILP pay [documentation](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay#exchange-rates): 

> Exchange rates are represented as the ratio between a destination amount and a source amount, in fractional units.

+ (see [comment](https://github.com/interledgerjs/interledgerjs/blob/fba2dfe8da2e582205632cbbc9433253b7956416/packages/pay/src/index.ts#L343-L346) in the startQuote function)

This means that to ILP pay, these rates
```
{
    USD: 1
    EUR: 1.12
}
```
can be more specifically represented like:
```
{
    USD/USD: 1
    USD/EUR: 1.12
}
```

In my opinion, we should have the `/rates` API be more intuitive for integrators (and in-line with actual FX APIs):
https://www.exchangerate-api.com/docs/standard-requests
https://www.fastforex.io/
https://cdn.jsdelivr.net/gh/fawazahmed0/currency-api@1/latest/currencies/usd.json

to return something like
```
  USD: {
    EUR: 0.89,
    ZAR: 18.7
  },
  EUR: {
    USD: 1.12,
    ZAR: 20.48
  },
```
where it represents 1 USD = 0.89 EUR.

## Todos
- [ ] Return correct rates in the MASE to represent exchange rates for base as shown above
- [ ] Update rates service to handle updated rates
- [ ] Update any ilp pay arguments as necessary

mkurapov:

- Update rates API to _not_ have to expose inversed rates. 
- Create a mapping function from `Rates` to "ILP prices", use when quoting
- Serve rates from config file

See https://github.com/interledger/rafiki/issues/1711 for additional context.

## Context

Fixes #1711 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -49,6 +49,7 @@ services:
       AUTH_SERVER_INTROSPECTION_URL: http://cloud-nine-wallet-auth:3007
       ILP_ADDRESS: test.cloud-nine-wallet
       STREAM_SECRET: BjPXtnd00G2mRQwP/8ZpwyZASOch5sUXT5o0iR5b5wU=
+      SLIPPAGE: 0
       PUBLIC_HOST: http://cloud-nine-wallet-backend
       OPEN_PAYMENTS_URL: http://cloud-nine-wallet-backend
       WEBHOOK_URL: http://cloud-nine-wallet/webhooks
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -49,7 +49,6 @@ services:
       AUTH_SERVER_INTROSPECTION_URL: http://cloud-nine-wallet-auth:3007
       ILP_ADDRESS: test.cloud-nine-wallet
       STREAM_SECRET: BjPXtnd00G2mRQwP/8ZpwyZASOch5sUXT5o0iR5b5wU=
-      SLIPPAGE: 0
       PUBLIC_HOST: http://cloud-nine-wallet-backend
       OPEN_PAYMENTS_URL: http://cloud-nine-wallet-backend
       WEBHOOK_URL: http://cloud-nine-wallet/webhooks
--- a/localenv/cloud-nine-wallet/seed.yml
+++ b/localenv/cloud-nine-wallet/seed.yml
@@ -6,13 +6,13 @@ self:
 assets:
   - code: USD
     scale: 2
-    liquidity: 10000
+    liquidity: 1000000
   - code: EUR
     scale: 2
-    liquidity: 10000
+    liquidity: 1000000
   - code: MXN
     scale: 2
-    liquidity: 10000
+    liquidity: 1000000
 peers:
   - initialLiquidity: '100000'
     peerUrl: http://happy-life-bank-backend:3002
@@ -38,10 +38,10 @@ accounts:
     assetCode: USD
 rates:
   EUR:
-    MXN: 18.68
+    MXN: 18.78
     USD: 1.10
   USD:
-    MXN: 18.68
+    MXN: 17.07
     EUR: 0.91
   MXN:
     USD: 0.059
--- a/localenv/cloud-nine-wallet/seed.yml
+++ b/localenv/cloud-nine-wallet/seed.yml
@@ -10,6 +10,9 @@ assets:
   - code: EUR
     scale: 2
     liquidity: 10000
+  - code: MXN
+    scale: 2
+    liquidity: 10000
 peers:
   - initialLiquidity: '100000'
     peerUrl: http://happy-life-bank-backend:3002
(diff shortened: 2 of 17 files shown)


localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -49,6 +49,7 @@ services:
       AUTH_SERVER_INTROSPECTION_URL: http://cloud-nine-wallet-auth:3007
       ILP_ADDRESS: test.cloud-nine-wallet
       STREAM_SECRET: BjPXtnd00G2mRQwP/8ZpwyZASOch5sUXT5o0iR5b5wU=
+      SLIPPAGE: 0
```
mkurapov, line 52: controversial, but maybe this makes more sense in a single peer environment? this basically takes the network fees down to 0 on the quote (since we assume no FX change variation). Ccould be more straightforward/easier to understand in the localenv setup especially. Would be curious on other's opinions

localenv/mock-account-servicing-entity/app/routes/rates.ts
```diff
@@ -7,16 +7,16 @@ interface Rates {
 
 const exchangeRates: Rates = {
   USD: {
-    EUR: 1.1602,
-    ZAR: 17.3792
+    EUR: 0.89,
```
mkurapov, line 10: now, the rates that the external API needs to return are more intuitive: 1 USD = 0.89 EUR ✅

packages/backend/src/rates/service.ts
```diff
@@ -6,6 +6,11 @@ import { CacheDataStore } from '../middleware/cache/data-stores'
 
 const REQUEST_TIMEOUT = 5_000 // millseconds
 
+export interface Rates {
```
mkurapov, line 9: only use this interface in this service

postman/collections/Interledger.json
```diff
@@ -3501,7 +3502,7 @@
 		{
 			"listen": "prerequest",
 			"script": {
-				"id": "83ff5ca5-119d-4e56-ba79-d7e8dc9aa1f8",
+				"id": "ca1da280-c4d6-43cc-86db-e3cd53bae676",
```
mkurapov, line 3505: not sure why all of these id's have been updated, but don't seem to affect anything 🤷

BlairCurrey: I noticed with environment variables that the id's change if the order changes. It happened when I inserted a new variable in between existing, and they did not change when I added a new variable at the end.

I'm not sure if there is a similar thing happening here but I've noticed the same thing (bunch of id changes) on all of my postman changes.

localenv/cloud-nine-wallet/docker-compose.yml
raducristianpopa: I'm thinking of sticking with the default value of `0.01` for the slippage. This way we can show possible fluctuations and that they can happen.

raducristianpopa (dismissed):

localenv/cloud-nine-wallet/docker-compose.yml
mkurapov: Yep, reverted

localenv/cloud-nine-wallet/seed.yml
```diff
@@ -10,6 +10,9 @@ assets:
   - code: EUR
     scale: 2
     liquidity: 10000
+  - code: MXN
```
mkurapov, line 13: helpful for PCH project

localenv/cloud-nine-wallet/seed.yml
```diff
@@ -33,6 +36,16 @@ accounts:
...
+    USD: 1.10
+  USD:
+    MXN: 18.68
+    EUR: 0.91
+  MXN:
+    USD: 0.059
+    EUR: 0.054
+sendingFee:
+    fixed: 0
+    percentage: 0
```
mkurapov, line 51: IMO I think we should keep this at 0 until we start properly showing the mock account balances

localenv/happy-life-bank/seed.yml
```diff
@@ -39,6 +42,21 @@ accounts:
...
     assetCode: EUR
-fees:
-  - fixed: 1
-    percentage: 0.02
+  - name: 'David'
+    path: accounts/david
+    id: 60257507-3191-4507-9d77-9071fd6b3c30
+    initialBalance: 150000
+    assetCode: MXN
+rates:
```
mkurapov, line 50: rates now from here

raducristianpopa (approved):

raducristianpopa (approved):

Issue #1701, opened 2023-08-04
Spike: look into decoupling Open Payments API and ILP


mkurapov:
## Context
Within the PCH/Mojaloop project, it is looking like we will be going forward with supporting mojaloop transfers instead of ILP transfers between payments. Specifically, payment pointer/incoming payments can now support multiple transfer methods, and this issue is for seeing how we would be able to refactor the code such that we can have "pluggable" payment methods.

PR #1747, opened 2023-08-18
refactor: move ilp related files into /payment-method/ilp

mkurapov:

In order to support adding a support for making payments with Mojaloop (instead of streaming ILP), this PR moves everything ILP related to `/payment-method/ilp`.
This folder will contain all of the possible payment methods Rafiki will support, e.g. `/payment-method/mojaloop`, with a common `PaymentMethod` interface that each method will need to implement from. There will also be a `PaymentMethodManager` class that will know how to select the correct payment methods based on the Open Payment payment.

There are no business logic changes here, purely reorganizing: moving `connector` `peer` `httpToken` (now `peer-http-token`) `spsp` and `ilp_plugin` to `/payment-method/ilp`.

 I recommend pressing `.` to open up the browser VS code editor for this PR to check out the folder:

Screenshot 2023-08-18 at 17 44 27

## Context

Progress towards #1701 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


sabineschaller:
In my mind, we were pulling out the open payments server from the backend but this is a first step in that direction. What do you think?

mkurapov:
> In my mind, we were pulling out the open payments server from the backend but this is a first step in that direction. What do you think?

@sabineschaller that's right, once we decouple ilp & rest of open payments we can see how/if we want to move open payments into it's own "project"

sabineschaller (commented):
I think the rates service should also be moved into `ilp` because it's ilp specific.

mkurapov:
Closing until I pick this up again

Issue #1640, opened 2023-07-25
Add fees to outgoing payment


sabineschaller:
## Context

The `sendAmount` field should be called `debitAmount`, because it includes the total amount that is going to be debited from the user's account, including fees.

## Todo

- ~[ ] add `feeId?: uuid` column to outgoing payment table~ (decided against it, it's already on the quote)
- [ ] rename `sendAmount` -> `debitAmount` in outgoing payment response

  ```
    // Outgoing payment response
      {  ...,
          "receiveAmount": {
                  "value": "10000", 
                  "assetCode": "USD",
                  "assetScale": 2
              },
              "debitAmount": {
                  "value": "10310",
                  "assetCode": "USD",
                  "assetScale": 2
              },
              "sentAmount": {
                  "value": "10302",
                  "assetCode": "USD",
                  "assetScale": 2
              }
      }
  ```
- [ ] update docs

Issue #1639, opened 2023-07-25
Calculate and add fees to quote


sabineschaller:
## Context

The `sendAmount` field should be called `debitAmount`, because it includes the total amount that is going to be debited from the user's account, including fees. The sending fees should be added to the `debitAmount` after Rafiki does the rate probe. 

## Todo
- [ ] add `feeId?: uuid` column to quotes table
- [ ] rename `sendAmount` -> `debitAmount` in quote response
- [ ] calculate sending fee based on `feeId`
- [ ] add sending fee to `debitAmount`
  ```
  // Quote response
    {  ...,
       "receiveAmount": {
            "value": "10000",
            "assetCode": "USD",
            "assetScale": 2
        },
        "debitAmount": { //includes all fees
            "value": "10310",
            "assetCode": "USD",
            "assetScale": 2
        }
    }
  ```
- [ ] update docs

PR #1755, opened 2023-08-21
refactor(backend): move fee calculation to backend

BlairCurrey:

- updates to latest open-payments package and openapi schema
- renames `sendAmount` to `debitAmount` across backend
- moves fee calculation to backend on quote create
- fixes fee calculation for cross-currency (different send/receive assets)

## Context

Fixes: https://github.com/interledger/rafiki/issues/1639 and Fixes: https://github.com/interledger/rafiki/issues/1640

There are lots of file changes, but it's mostly renaming. More interesting code changes are the `packages/backend/src/open_payments/quote/service.ts` ones.

## Checklist

- [X] Tests added/updated
- [X] Documentation added
- [X] Postman collection updated


BlairCurrey:
During the course of this issue, we backtracked on adding receiving fees. All fee's are currently sending fees.

Should we make a new tech debt issue to remove all the receiving fee stuff that was added in https://github.com/interledger/rafiki/pull/1702? Such as the type field on the fee table, requiring a type to get the latest feed, the `receivingFee` graphql resolver on `asset`, etc.

I figured we should remove this but Sabine mentioned that its possible we might want them for the mojaloop integration. Thoughts?

packages/backend/src/index.ts
```diff
@@ -342,17 +342,14 @@ export function initIocContainer(
   container.singleton('quoteService', async (deps) => {
     const config = await deps.use('config')
     return await createQuoteService({
-      slippage: config.slippage,
-      quoteLifespan: config.quoteLifespan,
-      quoteUrl: config.quoteUrl,
-      signatureSecret: config.signatureSecret,
-      signatureVersion: config.signatureVersion,
+      config,
```
BlairCurrey, line 345: passed in the entire config so I could more easily override values for testing.

njlie:
> I noticed this when renaming limits... We are storing the limits as jsonb in our database. See
> 
> https://github.com/interledger/rafiki/blob/ef48146aa3a894e86d7ab658b959a797355110b1/packages/auth/migrations/20220504163024_create_accesses_table.js#L1-L7
> 
> This means there will presumably be some existing ones with the `sendAmount` key in testnet, while our application logic now just looks for `debitAmount`. I suspect this will lead to unhandle exceptions in places but not sure exactly how it will materialize beyond that. I figure its probably safer to just write a migration to rename any such keys but figured I'd mention it here first.

Perhaps in the future we could flatten out the `limits` field into a new model so that it could be migrated more easily.

packages/auth/migrations/20230907201006_rename-limits-jsonb-keys.js
```diff
@@ -0,0 +1,21 @@
+exports.up = function (knex) {
+  // renames sendAmount keys (if any) to debitAmount in limits jsonb column
+  return knex('accesses')
+    .update({
+      limits: knex.raw(
+        "limits - 'sendAmount' || jsonb_build_object('debitAmount', limits->'sendAmount')"
+      )
+    })
+    .whereRaw("limits \\? 'sendAmount'")
+}
```
BlairCurrey, line 10: This is hard to read but I could not find a simpler way.

To test I initialized rafiki with this extra test migration (before `20230907201006_rename-limits-jsonb-keys.js`)  which seeds some dummy data: https://gist.github.com/BlairCurrey/094b5df41f36f94befacbf727c3535ef

Then I verified the accesses records matched expectations after rebuilding rafiki from migrations.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -288,77 +295,58 @@ export async function startQuote(
...
+  if (!maxReceiveAmountValue) {
+    // FixedDelivery
+    const fees = getFees(quote.fee, debitAmountValue)
+    debitAmountValue = BigInt(debitAmountValue) + fees
     if (
-      sendAmount.value !== quote.sendAmount.value ||
-      receiveAmount.value > maxReceiveAmountValue ||
-      receiveAmount.value <= BigInt(0)
+      debitAmountValue < quote.debitAmount.value ||
+      debitAmountValue <= BigInt(0)
```
sabineschaller, line 322: How would that happen?

sabineschaller (commented):
I have to run but here is a first part of a review.

packages/auth/migrations/20230907201006_rename-limits-jsonb-keys.js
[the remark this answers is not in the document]
BlairCurrey: There is some discussion in the main body of the PR. My initial reasoning is here: https://github.com/interledger/rafiki/pull/1755#issuecomment-1710469965

I think it probably is necessary because any existing records will have the old name and the application code has been changed to expect the new one. So it might break something on testnet if we dont do this.

openapi/schemas.yaml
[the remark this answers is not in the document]
BlairCurrey: result of fetching the schemas

packages/backend/migrations/20230904145439_rename-quote-sendamountvalue.js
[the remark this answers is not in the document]
BlairCurrey: For this one I guess we could rename it in the model with a `get` and `set` but I think this is simpler.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -121,13 +121,19 @@ async function createQuote(
       receiver
     })
 
+    const quoteAssetId = paymentPointer.assetId
```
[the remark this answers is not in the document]
BlairCurrey: I guess it's not necessary, especially since they arent multiple assetId's in play here. I assigned it to a variable because the `setFee` `assetId` and `assetId` inserted into the quote should be the same.

packages/backend/src/open_payments/quote/errors.ts
```diff
@@ -2,7 +2,8 @@ export enum QuoteError {
   UnknownPaymentPointer = 'UnknownPaymentPointer',
   InvalidAmount = 'InvalidAmount',
   InvalidReceiver = 'InvalidReceiver',
-  InactivePaymentPointer = 'InactivePaymentPointer'
+  InactivePaymentPointer = 'InactivePaymentPointer',
+  FeeExceedsReceiveAmount = 'FeeExceedsReceiveAmount'
```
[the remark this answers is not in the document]
BlairCurrey: I think it's a problem in the case of fixed send, where this is thrown.

This error is thrown from a check I moved from the MASE's quote endpoint. Thrown when fees are taken out of the receiveAmount for a fixed send (so when the receive amount would be negative): https://github.com/interledger/rafiki/blob/913bc6a91130254fe042242997fd3f3623f87d6d/localenv/mock-account-servicing-entity/app/routes/quotes.ts#L59-L61

I wonder is if this somehow handled elsewhere in the backend though. Or behaves as intended when not handled. Because if not I it means we currently just trust the the ASE to do this check.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -59,7 +57,9 @@ async function getQuote(
   deps: ServiceDependencies,
   options: GetOptions
 ): Promise<Quote | undefined> {
-  return Quote.query(deps.knex).get(options).withGraphFetched('asset')
+  return Quote.query(deps.knex)
+    .get(options)
+    .withGraphFetched('[asset, fee.asset]')
```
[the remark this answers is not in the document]
BlairCurrey: Just fee should work. Originally I was thinking I'd use the asset associated with the fee but I did not (and I suppose we could just use the quote one since we know it's the same).

openapi/schemas.yaml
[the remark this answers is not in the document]
BlairCurrey: actually, let me double check. I misread this comment/change

packages/backend/src/open_payments/quote/errors.ts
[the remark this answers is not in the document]
sabineschaller: Oh, I get it. I want to do a fixed send of $1 but the fees would be $1.1.

> Maybe rename to `FeeExceedsFixedSendReceiveAmount`, `NegativeReceiveAmount` or something else?

I like `NegativeReceiveAmount` because its very obvious.

packages/backend/src/open_payments/quote/service.ts
sabineschaller: I didn't realize it was just a refactor. I also looked more closely now and I think you are right. This can probably be removed since it checks the quote after it is returned from the ASE in the current flow.

packages/backend/src/openapi/webhooks.yaml
```diff
@@ -160,7 +160,7 @@ components:
             - createdAt
             - updatedAt
             - expiresAt
-            - sendAmount
+            - debitAmount
```
sabineschaller, line 163: It seems like we have some validation error here. Check in "files changed" if you can't see it here.

packages/backend/src/config/app.ts
```diff
@@ -88,8 +88,6 @@ export const Config = {
   incomingPaymentWorkers: envInt('INCOMING_PAYMENT_WORKERS', 1),
   incomingPaymentWorkerIdle: envInt('INCOMING_PAYMENT_WORKER_IDLE', 200), // milliseconds
 
-  quoteUrl: envString('QUOTE_URL', 'http://127.0.0.1:4001/quote'),
```
sabineschaller, line 91: Can you remove that from the integration docs, please?

packages/auth/migrations/20230907201006_rename-limits-jsonb-keys.js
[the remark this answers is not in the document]
sabineschaller: Makes sense. I wasn't aware of the discussion and didn't think about testnet. My bad.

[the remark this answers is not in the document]
BlairCurrey: No worries... I fee like higher level discussion on PR's on gh is kind of a mess.

packages/backend/src/open_payments/quote/errors.ts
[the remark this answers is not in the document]
BlairCurrey: updated to `NegativeReceiveAmount`

packages/backend/src/config/app.ts
BlairCurrey: Looks like there were a few places to remove that. I also updated the explanation for setting fees in the integration doc https://github.com/interledger/rafiki/pull/1755/commits/e6f6f340651a97455094a1efd3316487481583c4

packages/backend/src/openapi/webhooks.yaml
BlairCurrey: @sabineschaller Hmm, I see it. More details from the CI job show the warning:

    OpenAPI "servers" must be present and non-empty array
 https://github.com/interledger/rafiki/actions/runs/6174411421/job/16758949762?pr=1755#step:6:12

Looks unrelated to this change and we've been getting it for a while:
- 2 months ago: https://github.com/interledger/rafiki/commit/f5c1f6613de348ad0b266bb811c4b8cf5d729524#diff-90da7214f1ef2d21dca042e424b4079b684ab3017fb6c278e4abab4453855f24
- 3 months ago: https://github.com/interledger/rafiki/commit/22ef1dc9f398ed0a036c8fce9eed42f1d8825192#diff-90da7214f1ef2d21dca042e424b4079b684ab3017fb6c278e4abab4453855f24R1

Seems like it needs the `servers` attribute defined like we do in some of our other specs. Not sure what would be appropriate. What about something like this (similar to exchange rate yaml):

```yaml
servers:
  - url: 'https://account-servicing-entity.com/webhooks'
```

packages/backend/src/open_payments/quote/service.ts
BlairCurrey: I could remove this one OK but I tried the same for fixed send and an existing test failed (which was explicitly looking for this case). Perhaps the inputs _shouldn't_ cause these errors but I think it's probably safer to leave them just in case.

packages/backend/src/openapi/webhooks.yaml
BlairCurrey: Found the webhook url in some config files and concluded that url is right. I added it which seems to have fixed the warning.

sabineschaller (dismissed):
LGTM

localenv/mock-account-servicing-entity/app/routes/quotes.ts
```diff
@@ -19,7 +19,7 @@ export type Quote = {
   paymentType: PaymentType
```
mkurapov, line 19: We could remove this file altogether now, right?

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -288,77 +292,58 @@ export async function startQuote(
   }
 }
 
-type QuoteHeaders = {
-  Accept: string
-  'Content-Type': string
-  'Rafiki-Signature'?: string
+function getFees(fee: Fee | undefined, principal: bigint): bigint {
```
mkurapov, line 295: Maybe this method can live in feeService? just a suggestion

mkurapov (commented):
I would also recommend running `pnpm generate:types` in the token-introspection package to update the types properly.

Also, I'm not sure if we discussed, but for the documentation API schema, we should be seeing `sendAmount` be changed to `debitAmount` as well, correct?

mkurapov (dismissed):
🚀

packages/backend/src/open_payments/quote/service.ts
BlairCurrey: I added the method to the model since we already had the fee model here. It felt wrong to pass the fee into the fee service like `feeService.calculate(fee, principal)` or to re-lookup the fee. Later if we just have the `feeId` or `assetId` we can do `feeService.calculate(id, principal)` which gets the fee then calls the `calculate` method.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -312,7 +303,8 @@ async function finalizeQuote(
 
   if (!maxReceiveAmountValue) {
     // FixedDelivery
-    const fees = getFees(quote.fee, debitAmountValue)
+    // const fees = getFees(quote.fee, debitAmountValue)
```
mkurapov, line 306:
```suggestion
```

mkurapov (approved):

PR #1522, opened 2023-06-23
feat(backend): add metadata to payments

BlairCurrey:
## Changes proposed in this pull request

- Adds `metadata` field to payments
- Removes `description`, `externalRef` from payments (replaced by `metadata`)
- Changed GQL `WebhookEvent.data` type from string to `JSONObject` (to match metadata)
- Updated openapi schema and bumped `@interledger/open-payments` version
- Fix docusaurus config and regenerated GQL docs
- Updated postman. PR: https://interledger.postman.co/workspace/blair-interledger~54a25937-6992-4b7f-829f-91ac2142b7c0/pull-request/f8c94fe6-7695-46fb-8d82-ed161ca373ef

## Context

Progress towards https://github.com/interledger/rafiki/issues/1437
Progress towards https://github.com/interledger/rafiki/issues/1526

After merging, I will remove `externalRef` and `description` from the openapi spec to complete https://github.com/interledger/open-payments/issues/221 and re-fetch the spec in rafiki to complete https://github.com/interledger/rafiki/issues/1437

## Checklist

- [X] Related issues linked using `fixes #number`
- [X] Tests added/updated
- [X] Documentation added
- [X] Make sure that all checks pass

PR #1797, opened 2023-08-28
feat(auth): add interaction model

### LINKED ISSUE #1655: Make AS Interactions more GNAP-like

njlie:

- Adds Interaction model
  - Interaction model has many-to-one relation with Grant
  - Interaction models contain interaction info previously stored on Grant (`grant.interactRef`->`interaction.ref`, etc.)
  - Interaction models contain state of interaction and an expiry amount

## Context

Fixes #1655 
It would be a useful feature to be able to expire interactions. In order to achieve this, a model for individual interactions was added so it is easier to manage them.

It is a big PR, so I am open to breaking it down a bit.

## Checklist

- [X] Tests added/updated
- [ ] Documentation added


packages/auth/migrations/20220504161932_create_grants_table.js
```diff
@@ -15,10 +15,6 @@ exports.up = function (knex) {
     table.string('clientNonce')
     table.string('client').notNullable()
 
-    table.string('interactId').unique()
-    table.string('interactRef').unique()
-    table.string('interactNonce').unique()
-
```
BlairCurrey, line 21: Since testnet has deployed we've moved away from editing the migration files directly and instead making a new migration to update the tables. https://github.com/interledger/rafiki/pull/1522#discussion_r1245523191

I guess in this case it's technically not going to break anything because we are just removing fields here. Testnet would just still have them. Still, I think we should put them in a new migration IMO.

packages/auth/src/signature/middleware.ts
```diff
@@ -69,11 +69,12 @@ export async function grantContinueHttpsigMiddleware(
   )
 
   const grantService = await ctx.container.use('grantService')
+  // TODO: if interactRef exists, check that interaction is found
```
BlairCurrey, line 72: Did you implement this TODO below, on lines 87-97? This seems like it's doing what the TODO describes so maybe this comment can be removed?

```ts
  const interactionService = await ctx.container.use('interactionService')
  const interaction =
    interactRef && (await interactionService.getByRef(interactRef))

  if (!interaction) {
    ctx.status = 401
    ctx.body = {
      error: 'invalid_continuation',
      message: 'invalid interaction'
    }
  }
```

packages/documentation/docs/apis/auth/enums.md
```diff
@@ -0,0 +1,70 @@
+---
+id: enums
+title: Enums
+slug: enums
+sidebar_position: 6
+---
+
+<!-- Do not edit this file, it has been automatically generated by docusaurus-graphql-plugin -->
```
BlairCurrey, line 8: I'm surprised to see this file addition and docusaurus comment since we migrated to astro from docusaurus. Was this generated using the pre-astro migration documentation package? Maybe you need to merge the latest in from main?

I guess im not complete sure what's going on because the auth enums.md file still exists here on main: https://github.com/interledger/rafiki/blob/main/packages/documentation/src/content/docs/apis/auth/enums.md But the gql schemas are generated by graphql-markdown and output here: https://github.com/interledger/rafiki/blob/main/packages/documentation/src/schema-dumps/auth_raw_schema.md

I'm surprised we still have the docusaurus generated stuff (as it will not be updated when we change the schema).

BlairCurrey (commented):
I identified a couple things that might need cleaning up but otherwise looks good to me. Re-using the `generateBaseGrant` everywhere `beforeEach` is a nice change 👍

packages/auth/src/signature/middleware.ts
njlie: Yeah, I just forgot to clean up the TODO. Thanks for pointing that out.

packages/auth/src/interaction/service.ts
```diff
@@ -0,0 +1,142 @@
...
+async function create(
+  deps: ServiceDependencies,
+  grantId: string,
+  trx?: Transaction
+): Promise<Interaction> {
+  const { knex, config } = deps
+  const interactionTrx = trx || (await Interaction.startTransaction(knex))
+
+  try {
+    const interaction = await Interaction.query(interactionTrx).insert({
```
mkurapov, line 80:
```suggestion
  try {
    const interaction = await Interaction.query(trx || deps.knex).insert({
```
would it be sufficient without `startTransaction`, given that it's just a single insert here?

njlie: The way I've implemented the flow of grant creation, interactions are created [in tandem with grants](https://github.com/interledger/rafiki/blob/b2eea04995e1b2b9bf3b9cebb382aa83ff5bc7b2/packages/auth/src/grant/routes.ts#L165-L176) if they require one, so I added a transaction in case the corresponding grant creation fails. 

There might be a case for creating the interaction during the `start interaction` step rather than the `create grant` step.

packages/auth/src/config/app.ts
```diff
@@ -40,6 +40,7 @@ export const Config = {
   ),
   waitTimeSeconds: envInt('WAIT_SECONDS', 5),
   cookieKey: envString('COOKIE_KEY', crypto.randomBytes(32).toString('hex')),
+  interactionExpirySeconds: envInt('INTERACTION_EXPIRY_SECONDS', 10 * 60), // Default 10 minutes
```
sabineschaller, line 43: This needs to be added to the docs, I think.

packages/auth/src/interaction/model.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface InteractionWithGrant extends Interaction {
+  grant: NonNullable<Interaction['grant']>
+}
+
+export function isInteractionWithGrant(
+  interaction: Interaction
+): interaction is InteractionWithGrant {
+  return !!(interaction.grant && isFinishableGrant(interaction.grant))
+}
```
sabineschaller, line 45: Can there be an interaction without a grant?

packages/auth/src/interaction/routes.ts
```diff
@@ -174,14 +203,17 @@ async function handleGrantChoice(
...
+    ) {
       ctx.throw(400, { error: 'request_denied' })
     }
 
-    if (choice === GrantChoices.Accept) {
-      await grantService.approve(grant.id)
-    } else if (choice === GrantChoices.Reject) {
-      await grantService.finalize(grant.id, GrantFinalization.Rejected)
+    if (choice === InteractionChoices.Accept) {
+      await interactionService.approve(interactId)
```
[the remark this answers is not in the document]
BlairCurrey: eCommerce flow is working for me now. Not sure if this was addressed ([here](https://github.com/interledger/rafiki/pull/1797/commits/1f242ba5121323f7a5e2d2d689ee9a192932c629)?) or still an issue.

BlairCurrey (approved):

packages/auth/src/interaction/model.ts
njlie: There shouldn't be able to be one.

PR #1805, opened 2023-08-30
feat: get peering details

### LINKED ISSUE #1763: Add `GET /peer` discoverability route
## Context
In order to provide information that a new peer needs in order to peer with the Rafiki instance, we need to add a new route within the Rafiki backend. We can create a new "ilp peering" server within `backend` that exposes a `GET /peer` route. This route needs to return the
- ilpAddress
- list of supported assets
of the Rafiki instance.

## Todos
- [ ] Create a new ilp server that exposes `GET /peer` route and return the given information
- [ ] Put this new server instantiation under an `ENABLE_AUTO_PEERING` environment flag

mkurapov:

- Adding auto peering server (default port 3005)
- Adding auto peering service & routes that return ilpAddress as well as list of supported assets

## Context

Start of auto-peering story. Adding GET route for returning peering details (for peering discoverability). The new server instantiation itself is under a flag (defaults OFF)

Fixes #1763 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -72,6 +72,8 @@ import { createRedisDataStore } from './middleware/cache/data-stores/redis'
 import { createRedisLock } from './middleware/lock/redis'
 import { CombinedPaymentService } from './open_payments/payment/combined/service'
 import { FeeService } from './fee/service'
+import { AutoPeeringService } from './auto-peering/service'
+import { AutoPeeringRoutes } from './auto-peering/routes'
 
 export interface AppContextData {
   logger: Logger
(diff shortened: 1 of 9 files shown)


packages/backend/src/app.ts
```diff
@@ -477,19 +482,42 @@ export class App {
...
+
+    router.get('/', autoPeeringRoutes.get)
+
+    koa.use(router.routes())
+
+    this.autoPeeringServer = koa.listen(port)
+  }
+
   public async shutdown(): Promise<void> {
     return new Promise((resolve): void => {
```
raducristianpopa, line 501: I'm a little bit confused by this. Why is the promise getting resolved multiple times instead of just once? If there is an OP server, wouldn't the promise be resolved in that specific `if` block, making any further resolves irrelevant?

raducristianpopa (commented):
LGTM. Just a quick question for my understanding:

packages/backend/src/app.ts
mkurapov: I've been trying to update this code, but it's behaving weirdly when I update it to be "correct". I wonder why it was built this way originally, tbh.

I'm going to make a separate PR that hopefully fixes this and makes it more clear

raducristianpopa (approved):

PR #1807, opened 2023-08-31
chore(ci): automating the weekly sync discussion

raducristianpopa:

This PR includes a GitHub action that creates the Weekly Sync discussion every Tuesday at 12AM UTC.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


.github/workflows/create_weekly_sync.yml
```diff
@@ -0,0 +1,58 @@
+name: Create Weekly Sync discussion
+
+# Every Tuesday at 12AM UTC
+on:
+    schedule:
+        - cron: "0 0 * * 2"
```
raducristianpopa, line 6: Let me know if a different time works better.

sabineschaller: Can we do every Wednesday at 12AM? I would like to have it as early as possible but I also don't want to change it if we decide to move the call again. So I think Wednesday for the following week should be enough time.

sabineschaller (approved):
I'm sure you tested this and it works 😉

PR #1814, opened 2023-09-03
feat(docs): add astro-graphql-plugin to the documentation

### LINKED ISSUE #1815: Create plugin that can generate API documentation in markdown from schema
We had migrated from Docusaurus to Starlight for the Rafiki documentation. The Docusaurus version had the [graphql-markdown](https://graphql-markdown.github.io/) plugin installed, which is a Docusaurus-specific package for generating MDX from a GraphQL schema source.

There is no platform agnostic package that could replicate that same feature. The closest thing we could find is [[graphql-markdown](https://github.com/exogen/graphql-markdown)](https://github.com/exogen/graphql-markdown) (same name different package, naming things is hard yo). The generated output from this plugin was slightly unwieldy for the Rafiki schemas because:
1. Our backend schema is really long
2. The tables generated were not the best way to present the large amount of content in certain types (see or maybe not see the extremely tiny screenshot below)

Screenshot 2023-09-04 at 12 44 21 PM

@AlexLakatos proposed pulling out the relevant functionality from the docusaurus plugin for use in our Starlight documentation instead.

AlexLakatos:
Add astro-graphql-plugin plugin that automatically generates GraphQL API docs on every build.

- port old graphql-plugin to astro and add it to the docs
- fix implementedBy link paths
- remove doc generation step from CI

## Context

Closes #1815 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


AlexLakatos:
@huijing: of course I left a bug in there for you 😅 . The sidebar gets expanded and highlighted twice, because both the Backend API and the Auth API have the same menu items. I'm pretty sure that's not a bug with the plugin, and it's an Astro thing though.

Screenshot 2023-09-03 at 21 29 24

AlexLakatos:
@huijing Ignore that last comment, turns out PEBCAK, and it was my fault. I fixed it in [7723574](https://github.com/interledger/rafiki/pull/1814/commits/7723574cc99f281205b050ebc76b3723883065bb)

huijing (commented):
Ran locally and LGTM.
But would like @mkurapov and @sabineschaller to chime in.
Also, how does this work, do the markdown files get generated when the build runs or? How do people update the docs when the schema changes?

sabineschaller:
@mkurapov why is there no preview for this?

mkurapov:
@sabineschaller from the build: 
```
10:36:59 PM: No changes detected in base directory. Returning early from build.
```

I removed the cache and re-built, the preview is there now

packages/documentation/package.json
```diff
@@ -6,18 +6,16 @@
...
   },
   "dependencies": {
     "@astrojs/react": "^2.3.2",
     "@astrojs/starlight": "^0.7.3",
     "@types/react": "^18.2.21",
     "@types/react-dom": "^18.2.7",
     "astro": "2.10.14",
+    "astro-graphql-plugin": "^0.0.6",
     "graphql": "^16.8.0",
-    "graphql-markdown": "^7.0.0",
```
mkurapov, line 20: If we are getting rid of `graphql-markdown` in the same PR we can remove the "Schema Types" files that were generated from the plugin

packages/documentation/package.json
```diff
@@ -6,18 +6,16 @@
     "start": "astro dev",
     "build": "astro build",
     "preview": "astro preview",
-    "astro": "astro",
-    "docs:graphql:generate:backend": "./node_modules/.bin/graphql-markdown ../backend/src/graphql/schema.graphql > src/schema-dumps/backend_raw_schema.md",
-    "docs:graphql:generate:auth": "./node_modules/.bin/graphql-markdown ../auth/src/graphql/schema.graphql > src/schema-dumps/auth_raw_schema.md"
```
mkurapov, line 11: @huijing

> Also, how does this work, do the markdown files get generated when the build runs or? How do people update the docs when the schema changes?

It looks like they are generated on the build. In that case, I propose 
- we don't commit these generated files into the repo and
- we remove the CI step `pnpm --filter documentation docs:graphql:generate` 

Otherwise, we should add a command `docs:graphql:generate`, which will fail a build if the generated files are outdated compared to the graphql schema

mkurapov: If we don't exclude them, we _could_ get weird diffs in PRs as a result of people building the docs locally then pushing up their changes. I was just thinking of treating them as output to a build step, similar to how we ignore `build` folders

AlexLakatos: They need to be put into the "content" folder before the build step, let me see if there's a hook for that in Astro.

AlexLakatos: Removed the files from content folder, added them to gitignore, and now they get generated every time the "build" command is issued.

huijing (approved):

mkurapov:
Just checked on rafiki.dev, all looks 👌  thanks @AlexLakatos & @huijing !

All wanted to mention is we can remove `packages/documentation/src/schema-dumps` folder now, but NABD

AlexLakatos:
@mkurapov missed those, sorry. Here's a PR for it: https://github.com/interledger/rafiki/pull/1888

PR #1837, opened 2023-09-06
feat: fetching retrieves non-revoked keys for a payment pointer

### LINKED ISSUE #1760: Do not return revoked keys when fetching payment pointer keys
## Context
Whenever we revoke a key, we're patching the `PaymentPointerKey` with a `revoked: true` flag. However, when GET the keys, we do not check the flag. 

## Todos
- [ ] Update `getKeysByPaymentPointerId` to not include revoked keys

JoblersTune:

- When fetching payment pointer keys only non-revoked keys are returned

## Context

- fixes #1760 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment_pointer/key/service.test.ts
+++ b/packages/backend/src/open_payments/payment_pointer/key/service.test.ts
@@ -65,6 +65,28 @@ describe('Payment Pointer Key Service', (): void => {
         paymentPointerKeyService.getKeysByPaymentPointerId(paymentPointer.id)
       ).resolves.toEqual([key])
     })
+
+    test('Fetching Only Retrieves Non-Revoked Keys for a Payment Pointer', async (): Promise<void> => {
+      const paymentPointer = await createPaymentPointer(deps)
+
+      const keyOption1 = {
+        paymentPointerId: paymentPointer.id,
+        jwk: TEST_KEY
+      }
+
+      const keyOption2 = {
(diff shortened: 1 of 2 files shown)


JoblersTune:
I have updated the getKeysByPaymentPointerId service function. I wasn't sure if I needed to include any filtering in the routes file where keys are returned based on `deps.config.paymentPointerUrl === ctx.paymentPointerUrl`
```export async function getKeysByPaymentPointerId(
  deps: ServiceDependencies,
  ctx: PaymentPointerKeysContext
): Promise<void> {
  if (ctx.paymentPointer) {
    const keys = await deps.paymentPointerKeyService.getKeysByPaymentPointerId(
      ctx.paymentPointer.id
    )

    ctx.body = {
      keys: keys.map((key) => key.jwk)
    }
  } else if (deps.config.paymentPointerUrl === ctx.paymentPointerUrl) {
    ctx.body = {
      keys: [deps.jwk]
    }
  } else {
    return ctx.throw(404)
  }
}```

mkurapov:
@JoblersTune good q, I would say we don't add any filters here because this is the key that is "issued" for the Rafiki instance itself (used when making inter-Rafiki requests, for example to get grants from another Rafiki instance). The key shouldn't be revoked, or some things might break.

mkurapov (approved):

Issue #1761, opened 2023-08-22
Auto-peering


mkurapov:
## Context
One of the primary goals of the testnet is for to enable other account servicing entities to peer directly with testnet's Rafiki instance. Once an ASE goes online, they should be able to get added to the test network, and make payments between their and the testnet's payment pointers. This story outlines steps to do so.

## Todos
- [ ]

PR #1848, opened 2023-09-11
feat: DuplicatePeer error

mkurapov:

- Adds DuplicatePeer error during peering creation via checking for new unique constraint/index in the DB

## Context

While building out the auto-peering service, I noticed that it didn't really make sense to allow creating a Peer with the same combination of `staticIlpAddress` and `assetId` (this would mean a duplicate peering relationship). This adds a unique constraint for these two columns in the DB.

Related to #1761 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/peer/service.test.ts
```diff
@@ -173,6 +173,21 @@ describe('Peer Service', (): void => {
...
+      const peer = await peerService.create(options)
+      assert.ok(!isPeerError(peer))
+
+      await expect(
+        peerService.create({
+          ...options,
+          staticIlpAddress: options.staticIlpAddress,
+          assetId: options.assetId
+        })
+      ).resolves.toEqual(PeerError.DuplicatePeer)
```
BlairCurrey, line 189:
```suggestion
      await expect(
        peerService.create(options)
      ).resolves.toEqual(PeerError.DuplicatePeer)
```

BlairCurrey (approved):
Left a suggestion but looks good. Tested on postman as well and see the 409 duplicate peer error.

packages/backend/src/peer/errors.ts
```diff
@@ -24,6 +26,7 @@ export const errorToMessage: {
   [key in PeerError]: string
 } = {
   [PeerError.DuplicateIncomingToken]: 'duplicate incoming token',
+  [PeerError.DuplicatePeer]: 'duplicate peer',
```
raducristianpopa, line 29: Would it be better to provide a more specific error message rather than simply "Duplicate peer"?

mkurapov: Updated to be more specific

sabineschaller (approved):
Thanks for making this so digestible.

raducristianpopa (approved):

PR #1849, opened 2023-09-11
feat: accept peering request

### LINKED ISSUE #1802: Add `POST /peer` route handler
## Context
By adding a `POST /peer` route handler, we can supply the peer information that it needs to make the peering connection on its own side, as well as return the necessary information for the requesting Rafiki to create the peer locally. This route will be the main connection point between two peers trying to connect, and exchange peering information.

## Todos
- [ ] Create a POST `/peer` handler to the existing "peering" service built in #1763. The request will take in `{ staticIlpAddress, ilpConnectorAddress, asset: { code, scale }, httpToken }` as the arguments.

Within the handler:
- [ ] Call the peerService.createPeer method internally, to create the (requesting) peer locally in the DB
- [ ] If peering creation was successful, return the configured `staticIlpAddress` `ilpConnectorAddress` so the requesting peer can connect on it's own end

mkurapov:

- Add POST / handler for auto-peering, which:
    - validates the incoming request
    - creates the requested peer locally
    - returns its _own_ peering information if successful
- Removes GET peering details handler (to simplify service)

## Context

Steps 3 & 4 of the given diagram: 
image

Fixes #1802 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/app.ts
```diff
@@ -177,6 +179,7 @@ const PAYMENT_POINTER_PATH = '/:paymentPointerPath+'
 export interface AppServices {
   logger: Promise<Logger>
   knex: Promise<Knex>
+  axios: Promise<AxiosInstance>
```
BlairCurrey, line 182: do we need this? I'm not seeing this referenced, nor axios added in the ioc container

packages/backend/src/auto-peering/errors.ts
```diff
@@ -0,0 +1,38 @@
...
+export const errorToCode: {
+  [key in AutoPeeringError]: number
+} = {
+  [AutoPeeringError.InvalidIlpConfiguration]: 400,
+  [AutoPeeringError.InvalidPeerIlpConfiguration]: 400,
+  [AutoPeeringError.UnknownAsset]: 404,
+  [AutoPeeringError.PeerUnsupportedAsset]: 400,
+  [AutoPeeringError.InvalidPeerUrl]: 400,
+  [AutoPeeringError.InvalidPeeringRequest]: 400,
+  [AutoPeeringError.DuplicatePeer]: 400
```
raducristianpopa, line 24:
```suggestion
  [AutoPeeringError.DuplicatePeer]: 409
```

packages/backend/src/auto-peering/errors.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const errorToMessage: {
+  [key in AutoPeeringError]: string
+} = {
+  [AutoPeeringError.InvalidIlpConfiguration]:
+    'The ILP configuration is misconfigured',
+  [AutoPeeringError.InvalidPeerIlpConfiguration]: `Requested peer's ILP configuration is misconfigured`,
+  [AutoPeeringError.UnknownAsset]: 'Unknown asset',
+  [AutoPeeringError.PeerUnsupportedAsset]: 'Peer does not support asset',
```
raducristianpopa, line 34: How does the error related to unsupported assets work?
- when one of the provided assets is not supported it rollbacks everything
- when one of the provided assets is not supported it keeps creating the others but will return the message `AutoPeeringError.PeerUnsupportedAsset`

And what is the difference between `AutoPeeringError.PeerUnsupportedAsset` and `AutoPeeringError.UnknownAsset`

packages/backend/src/auto-peering/routes.ts
```diff
@@ -1,13 +1,27 @@
 import { AppContext } from '../app'
 import { BaseService } from '../shared/baseService'
+import { errorToCode, errorToMessage, isAutoPeeringError } from './errors'
 import { AutoPeeringService } from './service'
 
+interface PeeringRequestArgs {
+  staticIlpAddress: string
+  ilpConnectorAddress: string
+  asset: { code: string; scale: number }
```
raducristianpopa, line 9: Would it be easier if we ask for an array of assets?

packages/backend/src/auto-peering/errors.ts
mkurapov: Originally, I was thinking about changing how we define the "Peer" in Rafiki to be relationship to be 1 Peer many assets, but after consideration w Sabine we decided it's better to keep what we have now 1 "Peer" defines an ILP relationship over a single asset: it keeps liquidity management a little bit easier to understand, and doesn't require a big overhaul in the service/DB. 

As a result, 
> when one of the provided assets is not supported
this isn't a scenario that would happen, given we only try to attempt to auto-peer using one asset

mkurapov: > And what is the difference between AutoPeeringError.PeerUnsupportedAsset and AutoPeeringError.UnknownAsset

from the perspective of peer 1 trying to peer with peer 2, AutoPeeringError.UnknownAsset
 is if your own (peer 1) instance does not support the asset and AutoPeeringError.PeerUnsupportedAsset is when trying to peer with peer 2, peer 2 does not support the asset

packages/backend/src/app.ts
mkurapov: Ah, it's used in the follow up PR: https://github.com/interledger/rafiki/pull/1857/files#diff-2ef3cf9bf12d21de5f65f0c473654ec6d62ac76a214fbf53a4afa2d3a92e7bf3

This is the result of my splitting of changes into PRs...

packages/backend/src/auto-peering/routes.ts
mkurapov: replied: https://github.com/interledger/rafiki/pull/1849#discussion_r1324279743

packages/backend/src/app.ts
mkurapov: Updated

raducristianpopa (approved):

njlie (approved):

raducristianpopa:
It looks like the `httpbis-digest-headers` issue still exists for some reason. 😅 I'll look into it.

raducristianpopa:
> It looks like the `httpbis-digest-headers` issue still exists for some reason. 😅 I'll look into it.

It looks like that `pnpm` is attempting to download the packages again instead of copying them from the global store in some scenarios (still haven't figured them out) and it still tries to run `npm build`. Can you try adding the `--ignore-scripts` option to `pnpm i` when building the production deps so we can avoid running the `build` script from `httpbis-digest-headers`. This _should_ fix the issue for now (in both scenarios - downloading or reusing the dependencies), until I have a closer look at what is actually happening.

Changes: https://github.com/interledger/rafiki/pull/1868/commits/2f3ef2e90e8e7fe76bac823babc5a36b63656d3d

mkurapov:
I merged in Blair's changes before it was merged into main, hence the many commits here

raducristianpopa (approved):

dragosp1011 (approved):

packages/backend/src/auto-peering/errors.ts
mkurapov: Yes, it's just the case for the wrong input

PR #1857, opened 2023-09-12
feat: createOrUpdatePeerByUrl mutation

### LINKED ISSUE #1765: Auto-peering: update `createPeerByUrl` mutation
## Context
Now that we have the `POST /peer` route that returns ILP connection details for the peers, we don't need to manually enter the `staticIlpAddress` for the peer: the rafiki instance can just fetch that information via `POST /peer`. As a result, we can create a new `createPeerByUrl` mutation:

## Todos
- [ ] Takes in `peerUrl` `assetId` `maxPacketAmount` `name`
- [ ] Call `POST <peerUrl>` with `{ staticIlpAddress, ilpConnectorAddress, asset: { code, scale }, httpToken }` to get back the request peer's `staticIlpAddress` `ilpConnectorAddress`, as well as generate an httpToken for the peering relationship. 
- [ ] Create the peer in the DB as normal

mkurapov:

- Adds `createPeerOrUpdateByUrl` mutation
- creates auto peering service method that makes a POST request to the other peer (via `peerUrl`) to get peering details (#1802), and creates the peer locally

## Context

Steps 1, 2, 5 of the diagram:
image

Fixes #1765  

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


mkurapov:
> Just from looking at your diagram: Does the response (step 4) not also have to include httpToken?

A single token is generated for the (auto) peering relationship, and is the incoming _and_ outgoing token for both of the parties. I could change it to be different for each counterparty, but it isn't too different.

> Also, why are there no docs? 🙃
I still have the "issue a default number of liquidity during auto-peering" issue so I want to finish everything and then add a  page

packages/backend/src/auto-peering/routes.ts
```diff
@@ -52,7 +52,8 @@ async function acceptPeerRequest(
     ctx.body = {
       error: {
         code: errorCode,
-        message: errorToMessage[peeringDetailsOrError]
+        message: errorToMessage[peeringDetailsOrError],
+        type: peeringDetailsOrError
```
sabineschaller, line 56: Is `type` the right property name for this?

packages/backend/src/auto-peering/service.ts
```diff
@@ -39,8 +56,110 @@ export async function createAutoPeeringService(
...
+    maxPacketAmount: args.maxPacketAmount,
+    name: args.name,
+    assetId: asset.id,
+    staticIlpAddress: peeringDetailsOrError.staticIlpAddress,
+    http: {
+      incoming: {
+        authTokens: [httpToken]
+      },
+      outgoing: {
+        authToken: httpToken,
```
sabineschaller, line 100: I don't think these should be the same token

mkurapov: At a high level, it doesn't make a difference, but now that I think about it, this matters in supporting the case where if the receiving peer (peer 2) has already registered the peer (peer 1) that is making the request: https://github.com/interledger/rafiki/blob/d5089e2a94c94c9c0815c1fd5b7a351bf833b309/packages/backend/src/auto-peering/service.ts#L194-L209

to support this functionality, we'd need to send back the outgoing token that peer 2 would use during outgoing packets. Going to update this :+1:

packages/backend/src/auto-peering/routes.ts
mkurapov: What would you propose as an alternative?

I think 
```js
{
    error: {
        code: 400
        message: ...
        type: "UnknownAsset"
    }
}
```
works fine in this case. Similar to stripe docs: https://stripe.com/docs/api/errors#errors-type

sabineschaller: Yeah, I can't come up with anything better. Let's keep it.

packages/backend/src/auto-peering/service.ts
```diff
@@ -56,8 +68,107 @@ export async function createAutoPeeringService(
...
+    return AutoPeeringError.UnknownAsset
+  }
+
+  const outgoingHttpToken = uuid()
+
+  const peeringDetailsOrError = await sendPeeringRequest(deps, peerUrl, {
+    ilpConnectorAddress: deps.config.ilpConnectorAddress,
+    staticIlpAddress: deps.config.ilpAddress,
+    asset: { code: asset.code, scale: asset.scale },
+    httpToken: outgoingHttpToken,
```
sabineschaller, line 94: Why not let the Rafiki instance receiving the auto-peering request generate that token?

sabineschaller: Oh, ok. Maybe I didn't read the code correctly.

packages/backend/src/auto-peering/service.ts
```diff
@@ -56,8 +68,107 @@ export async function createAutoPeeringService(
...
+    staticIlpAddress: peeringDetailsOrError.staticIlpAddress,
+    http: {
+      incoming: {
+        authTokens: [peeringDetailsOrError.httpToken]
+      },
+      outgoing: {
+        authToken: outgoingHttpToken,
+        endpoint: peeringDetailsOrError.ilpConnectorAddress
+      }
+    }
```
sabineschaller, line 116: This is where I thought it would be 
```js
   http: {
      incoming: {
        authTokens: [incomingHttpToken] //generated here and communicated to the peer
      },
      outgoing: {
        authToken: peeringDetailsOrError.httpToken, //generated by the peer for me to use when I send a request to the peer
        endpoint: peeringDetailsOrError.ilpConnectorAddress
      }
    }
```

mkurapov: I see what you mean, but is there a big difference in this case? I was thinking about conceptually, the peer just presents _its_ connection details, including the token it will be sending for payments

sabineschaller: I thought for the use case, where I'm requesting to auto-peer but the other Rafiki for some reason already has me as a peer, that other Rafiki can just respond with the token it already has on file for me. Otherwise, it would have to update its peering relationship with the token I'm now presenting. But I guess that is also an option.

mkurapov: I'm actually already doing the updating: https://github.com/interledger/rafiki/pull/1849/files#diff-7b03b8ad222a5bb774fbc5e8afa3e8588dd1a28da1acabd65f33222d263e325dR98-R115

This allows for the receiving peer to update any of the arguments it receives from the first peer

sabineschaller: Since this should only be used with fake money and in sandbox environments, I guess that is fine.

njlie (dismissed):
LGTM

packages/backend/src/auto-peering/routes.ts
dragosp1011: type or category make sense
another option is to rename code to status, and use the code yo define the "type" instead

dragosp1011 (dismissed):

packages/backend/src/auto-peering/service.ts
mkurapov: Talked in person, each peer will present its own outgoing token

sabineschaller (approved):
Thank you!

PR #1860, opened 2023-09-12
feat: add test coverage commands

BlairCurrey:

- adds command to run tests and show coverage for files that jest thinks have changed or were related to changes since main

## Context

Closes: https://github.com/interledger/rafiki/discussions/1821

Allows us to get current coverage of tests, but does not show any difference between current branch coverage and main.

If you're really motivated to show the difference from main, you can generate a summary on main and the base branch and run this command: 

    pnpx cov-diff coverage-summary1.json coverage-summary2.json

Not sure if that will prove useful but thought I'd share. https://www.npmjs.com/package/cov-diff

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -12,6 +12,7 @@
     "build": "pnpm build:deps && pnpm clean && tsc --build tsconfig.json && pnpm copy-files",
     "clean": "rm -fr dist/",
     "test": "jest --passWithNoTests --maxWorkers=50%",
+    "test:cov": "pnpm test -- --coverage --changedSince=main",
     "prepack": "pnpm build",
     "copy-files": "cp src/graphql/schema.graphql dist/graphql/ && cp -r ./src/openapi ./dist/"
   },
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -12,7 +12,9 @@
     "build": "pnpm build:deps && pnpm clean && tsc --build tsconfig.json && pnpm copy-files",
     "clean": "rm -fr dist/",
     "test": "jest --passWithNoTests --maxWorkers=50%",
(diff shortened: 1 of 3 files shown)


packages/auth/package.json
```diff
@@ -12,6 +12,7 @@
     "build": "pnpm build:deps && pnpm clean && tsc --build tsconfig.json && pnpm copy-files",
     "clean": "rm -fr dist/",
     "test": "jest --passWithNoTests --maxWorkers=50%",
+    "test:cov": "pnpm test -- --coverage --changedSince=main",
```
BlairCurrey, line 15: Keep in mind that this will only test things that jest thinks changed. So it does not just run all the tests like `test` does. https://jestjs.io/docs/cli#--changedsince

In limited testing this seems to include. a lot (1 line change in auth grant service tested 10/14 tests)

I also considered adding a `test:sincemain` that runs all tests with `--changedSince=main`. Then adding `test:sincemain:cov` and `test:cov`.

sabineschaller, line 15: Are the two `--` after `test` intentional since nothing is following them immediately?

sabineschaller: > I also considered adding a `test:sincemain` that runs all tests with `--changedSince=main`. Then adding `test:sincemain:cov` and `test:cov`.

I think that's a good idea.

BlairCurrey: yeah its weird, but that will pass the `--coverage` and `--changedSince` args to the `test` command. Without it it gets interpreted as arguments for the `pnpm` command.

sabineschaller: The more you know... Thank you!

BlairCurrey: added

mkurapov (approved):

PR #1937, opened 2023-09-20
feat(backend): rename "payment pointer" tables & columns to use "wallet address"

### LINKED ISSUE #1934: Backend: rename 'payment pointer' to 'wallet address' in database
Rename all database table and column names that refer to 'payment pointers' to refer to 'wallet addresses' instead.

njlie:

- Rename all tables & columns that refer to 'payment pointers' to refer to 'wallet addresses' instead.

## Context

Fixes #1934.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
+++ b/packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
@@ -5,11 +5,10 @@
 exports.up = function(knex) {
   return Promise.all([
     knex.schema.renameTable('paymentPointers', 'walletAddresses'),
-  	knex.schema.renameTable('paymentPointerKeys', 'walletAddressKeys').then(() => {
-      knex.schema.alterTable('walletAddressKeys', function (table) {
-        table.renameColumn('paymentPointerId', 'walletAddressId')
-        table.foreign('walletAddressId').references('walletAddresses.id')
-      })
+  	knex.schema.renameTable('paymentPointerKeys', 'walletAddressKeys'),
+    knex.schema.alterTable('walletAddressKeys', function (table) {
+      table.renameColumn('paymentPointerId', 'walletAddressId')
+      table.foreign('walletAddressId').references('walletAddresses.id')
     }),
   	knex.schema.alterTable('quotes', function (table) {
   	  table.renameColumn('paymentPointerId', 'walletAddressId')
@@ -36,7 +35,8 @@ exports.up = function(knex) {
 exports.down = function(knex) {
   return Promise.all([
   	knex.schema.renameTable('walletAddresses', 'paymentPointers'),
-  	knex.schema.renameTable('walletAddressKeys', 'paymentPointerKeys').then(() => {
+  	knex.schema.renameTable('walletAddressKeys', 'paymentPointerKeys'),
+    knex.schema.alterTable('paymentPointerKeys', function (table) {
       table.renameColumn('walletAddressId', 'paymentPointerId')
       table.foreign('paymentPointerId').references('paymentPointers.id')
     }),
--- a/packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
+++ b/packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
@@ -2,19 +2,19 @@
  * @param { import("knex").Knex } knex
  * @returns { Promise<void> }
  */
-exports.up = function(knex) {
+exports.up = function (knex) {
   return Promise.all([
     knex.schema.renameTable('paymentPointers', 'walletAddresses'),
-  	knex.schema.renameTable('paymentPointerKeys', 'walletAddressKeys'),
+    knex.schema.renameTable('paymentPointerKeys', 'walletAddressKeys'),
     knex.schema.alterTable('walletAddressKeys', function (table) {
       table.renameColumn('paymentPointerId', 'walletAddressId')
       table.foreign('walletAddressId').references('walletAddresses.id')
     }),
-  	knex.schema.alterTable('quotes', function (table) {
-  	  table.renameColumn('paymentPointerId', 'walletAddressId')
-  	  table.foreign('walletAddressId').references('walletAddresses.id')
+    knex.schema.alterTable('quotes', function (table) {
+      table.renameColumn('paymentPointerId', 'walletAddressId')
+      table.foreign('walletAddressId').references('walletAddresses.id')
       table.index(['walletAddressId', 'createdAt', 'id'])
-  	}),
+    }),
     knex.schema.alterTable('incomingPayments', function (table) {
       table.renameColumn('paymentPointerId', 'walletAddressId')
       table.foreign('walletAddressId').references('walletAddresses.id')
@@ -26,25 +26,25 @@ exports.up = function(knex) {
       table.index(['walletAddressId', 'createdAt', 'id'])
     })
   ])
-};
+}
 
 /**
  * @param { import("knex").Knex } knex
  * @returns { Promise<void> }
  */
-exports.down = function(knex) {
+exports.down = function (knex) {
   return Promise.all([
-  	knex.schema.renameTable('walletAddresses', 'paymentPointers'),
-  	knex.schema.renameTable('walletAddressKeys', 'paymentPointerKeys'),
+    knex.schema.renameTable('walletAddresses', 'paymentPointers'),
+    knex.schema.renameTable('walletAddressKeys', 'paymentPointerKeys'),
     knex.schema.alterTable('paymentPointerKeys', function (table) {
       table.renameColumn('walletAddressId', 'paymentPointerId')
       table.foreign('paymentPointerId').references('paymentPointers.id')
     }),
-  	knex.schema.alterTable('quotes', function (table) {
-  	  table.renameColumn('walletAddressId', 'paymentPointerId')
-  	  table.foreign('paymentPointerId').references('paymentPointers.id')
+    knex.schema.alterTable('quotes', function (table) {
+      table.renameColumn('walletAddressId', 'paymentPointerId')
+      table.foreign('paymentPointerId').references('paymentPointers.id')
       table.index(['paymentPointerId', 'createdAt', 'id'])
-  	}),
+    }),
     knex.schema.alterTable('incomingPayments', function (table) {
       table.renameColumn('walletAddressId', 'paymentPointerId')
       table.foreign('paymentPointerId').references('paymentPointers.id')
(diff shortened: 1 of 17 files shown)


njlie:
The backend service isn't going to build in this branch (yet), but the migration should run successfully.

BlairCurrey (commented):
This concern spans this PR as well as the code changes. There are payment pointer webhook events which are being renamed to wallet address. I think there will be webhook events with JSON payloads with a `paymentPointer` key in existing deployments (e.g. testnet). Can we rule out that this will cause any problems? Are we never referencing the `walletAddress` key in the event payload read from the db? Otherwise I think we'd need to update the keys to `walletAddress`. 

This is the same potential issue I ran into when renaming `sendAmount` -> `debitAmount` here: https://github.com/interledger/rafiki/pull/1755#issuecomment-1710469965

This is what the migration to rename the keys looked like: https://github.com/interledger/rafiki/blob/main/packages/auth/migrations/20230907201006_rename-limits-jsonb-keys.js

packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
```diff
@@ -24,7 +24,15 @@ exports.up = function (knex) {
...
-    })
+    }),
+    knex('webhookEvents')
+      .update({
+        // renames paymentPointer keys (if any) to walletAddress in data json column
+        data: knex.raw(
+          "data - 'paymentPointer' || json_build_object('walletAddress', data->'paymentPointer')"
+        )
+      })
+      .whereRaw("data \\? 'paymentPonter'")
```
BlairCurrey, line 35: I'm getting an error when this migration runs: `operator does not exist: json ? unknown` because this column is `json` (not `jsonb`) which doesnt have the `?` operator. Similar errors are lurking behind this one. This is working for me, although perhaps not the most efficient because its casting the `json` to `jsonb` twice:

```js
    knex('webhookEvents')
      .update({
        data: knex.raw(
          "data::jsonb -'paymentPointer' || jsonb_build_object('walletAddress', data::jsonb->'paymentPointer')"
        )
      })
      .whereRaw("data->'paymentPointer' is not null")
```
This is working for me as well and should be better because it casts once in a CTE and then references that in the update. Harder to read though. 🤷 

```js
    knex.raw(`
      WITH cte_data AS (
        SELECT
          id,
          data::jsonb AS json_data
        FROM
          "webhookEvents"
        WHERE
          data->'paymentPointer' IS NOT NULL
      )
      UPDATE "webhookEvents" w
      SET
        "data" = (cte_data.json_data - 'paymentPointer' || jsonb_build_object('walletAddress', cte_data.json_data->'paymentPointer'))
      FROM cte_data
      WHERE
        w.id = cte_data.id;
    `)
```
Not sure how much we should care about the performance difference with testnet as the only deployment, but casting to jsonb is relatively expensive compared to the other operations we are doing here.

njlie: Perhaps we can handwave the performance differences since this would only run once per deployment. I'll see if there are other approaches that may work as well.

packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
```diff
@@ -0,0 +1,115 @@
...
+  return Promise.all([
+    knex.schema.renameTable('paymentPointers', 'walletAddresses'),
+    knex.schema.renameTable('paymentPointerKeys', 'walletAddressKeys'),
+    knex.schema.alterTable('walletAddressKeys', function (table) {
+      table.renameColumn('paymentPointerId', 'walletAddressId')
+      table.foreign('walletAddressId').references('walletAddresses.id')
+    }),
+    knex.schema.alterTable('quotes', function (table) {
+      table.renameColumn('paymentPointerId', 'walletAddressId')
+      table.foreign('walletAddressId').references('walletAddresses.id')
```
mkurapov, line 15: I'm not actually sure, what happens to the existing `paymentPointer.id` foreign key here? Does it get properly handled in the rename, or dropped?

BlairCurrey: This also applies the indexes (not pictured above) in  line 16: `table.index(['walletAddressId', 'createdAt', 'id'])`.

Looks like the migration is duplicating these. Two fkey constraints with different fkey names but the same relationship. I think we can just remove the `table.foreign('walletAddressId').references('walletAddresses.id')` line (and the indexes). This is what I see in `Indexes` and `Foreign-key constraints:` from doing a `\d quotes` after this migration. 

quote:

```
Indexes:
    "quotes_pkey" PRIMARY KEY, btree (id)
    "quotes_paymentpointerid_createdat_id_index" btree ("walletAddressId", "createdAt", id)
    "quotes_walletaddressid_createdat_id_index" btree ("walletAddressId", "createdAt", id)
Foreign-key constraints:
    "quotes_assetid_foreign" FOREIGN KEY ("assetId") REFERENCES assets(id)
    "quotes_feeid_foreign" FOREIGN KEY ("feeId") REFERENCES fees(id)
    "quotes_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    "quotes_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
 ```
 
 and here is some of \d "walletAddresses" for good measure (see the duplicate quote fkey):
```
Referenced by:
    TABLE ""incomingPayments"" CONSTRAINT "incomingpayments_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE ""incomingPayments"" CONSTRAINT "incomingpayments_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE ""outgoingPayments"" CONSTRAINT "outgoingpayments_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE ""outgoingPayments"" CONSTRAINT "outgoingpayments_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE ""walletAddressKeys"" CONSTRAINT "paymentpointerkeys_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE "quotes" CONSTRAINT "quotes_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE "quotes" CONSTRAINT "quotes_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
    TABLE ""walletAddressKeys"" CONSTRAINT "walletaddresskeys_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)
```

File: packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+        )
+      })
+      .whereRaw("data->'paymentPointerUrl' is not null"),
+    knex('webhookEvents')
+      .update({
+        // renames payment_pointer.not_found values (if any) to wallet_address.not_found for type key in data json column
+        type: knex.raw("REPLACE(type, 'payment_pointer.', 'wallet_address.')")
+      })
+      .whereLike('type', 'payment_pointer%')
+  ])
```
BlairCurrey, line 58: i inserted some dummy data before this migration and this and similar all seem to be working correctly

packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js
BlairCurrey: Actually instead of removing these lines we might want to just drop the old one. This way we end up with `"quotes_walletaddressid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id)` instead of "quotes_paymentpointerid_foreign" FOREIGN KEY ("walletAddressId") REFERENCES "walletAddresses"(id).

So doing like:

```
    knex.schema.alterTable('quotes', function (table) {
      table.dropForeign(['paymentPointerId'])
      table.dropIndex(['paymentPointerId', 'createdAt', 'id'])
      table.renameColumn('paymentPointerId', 'walletAddressId')
      table.foreign('walletAddressId').references('walletAddresses.id')
      table.index(['walletAddressId', 'createdAt', 'id'])
    }),
 ```

File: packages/backend/migrations/20230918113102_rename_payment_pointer_tables.js [hunk 4]
```diff
@@ -0,0 +1,159 @@
...
+      'ALTER TABLE "walletAddressKeys" DROP CONSTRAINT "paymentpointerkeys_paymentpointerid_foreign"'
+    ),
+    knex.schema.alterTable('quotes', function (table) {
+      table.dropForeign(['paymentPointerId'])
+      table.dropIndex(['paymentPointerId', 'createdAt', 'id'])
+      table.renameColumn('paymentPointerId', 'walletAddressId')
+      table.foreign('walletAddressId').references('walletAddresses.id')
+      table.index(['walletAddressId'])
+      table.index(['createdAt'])
+      table.index(['id'])
```
njlie, line 45: The indexes are not aggregated because the original indexes weren't either. Aggregating them produces, for example:
```
Indexes:
"outgoingpayments_walletaddressid_createdat_id_index" btree ("walletAddressId", "createdAt", id)
```

Whereas originally the indexes were not aggregated:
```
"outgoingpayments_paymentpointerid__index" btree ("paymentPointerId")
"outgoingpayments_createdat_index" btree ("createdAt")
"outgoingpayments_id_index" btree (id)
```

So `table.index` is called separately each time to preserve this.

BlairCurrey: > The indexes are not aggregated because the original indexes weren't either.

It seems like they are originally 1 composite index to me but maybe I'm missing something... without running this migration I see `"outgoingpayments_paymentpointerid_createdat_id_index" btree ("paymentPointerId", "createdAt", id)` (same for `quotes`, did not check the rest). If they were seperate indexes I also don't think we'd need to do `table.dropIndex(['paymentPointerId', 'createdAt', 'id'])`, but instead would just drop the payment pointer one (and thus not have to re-add the others).

njlie: Oh yeah, they're composite. A misreading on my part earlier. So we'll keep the composite dropping & adding of these indexes.

Issue #1910, opened 2023-09-19
Update Rafiki backend to allow for an unauthenticated `GET /incoming-payments/:id` request


njlie:
Allow a request to `GET /incoming-payments/:id` to be unauthenticated, and consequently return a reduced subset of data (`receivedAmount`).

BlairCurrey:
fixed in https://github.com/interledger/rafiki/pull/1952

Issue #1896, opened 2023-09-18
Non-authenticated GET Incoming payment request


njlie:
Introduce unauthenticated GET incoming payment request that returns a subset of the details normally returned from an authenticated request to this same route.

This subset should only include the `receiveAmount` field.

- [x] #1909
- [x] #1910

BlairCurrey:
fixed in https://github.com/interledger/rafiki/pull/1952

PR #1952, opened 2023-09-21
feat(backend): unauthenticated get incoming payment

BlairCurrey:

- adds unauthenticated version of the incoming-payments route
- makes new middleware to set authentication status and switch between authed/unauthed gets
- modifies token introspection middleware to not throw on failure (for this route only) via new `bypassErrors`option
  - token introspection still needs to happen (otherwise I just wouldn't use it on this route) because it sets info on the context which is required in case valid auth details are provided

## Context

fixes: #1896
fixes: #1910

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/app.ts
```diff
@@ -129,6 +130,9 @@ export type HttpSigContext = AppContext & {
   client: string
 }
 
+export type HttpSigWithAuthenticatedStausContext = HttpSigContext &
```
njlie, line 133: nit: "HttpSigWithAuthenticated**Staus**Context" -> "HttpSigWithAuthenticatedStatusContext"

packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -69,6 +75,42 @@ describe('Auth Middleware', (): void => {
...
+        requestType: type,
+        requestAction: action,
+        bypassError: true
+      })
+      ctx.request.headers.authorization = ''
+      const error = new Error('Unknown')
+      ctx.throw = jest
+        .fn()
+        .mockImplementation(
+          (message: string, code?: number, properties?: {}) => {
```
njlie, line 104: Does `mockImplementation` require these parameters? I don't think the Github Actions warnings are a big deal but if it can be called with an empty parameter list we may as well do that so Github Actions won't complain.

BlairCurrey: didnt need. removed

openapi/resource-server.yaml
BlairCurrey, line 1: I plan on pulling in the latest open-payments spec when we do the open-payments release. Leaving these for now as it is required to run/tests.

njlie (approved):

PR #1960, opened 2023-09-25
fix: grant related fixes

### LINKED ISSUE #1958: Unable to reject a grant
When declining a grant in Testnet we receive an `Internal Server Error`. The cause of this issue is that a new migration was not generated for the `finalizationReason` column (https://github.com/interledger/rafiki/commit/bf17908a0dc99beab0cd066ab59a065696faaebc). Hence, any action aimed at modifying the `finalizationReason` value of a grant will result in a failure.

```
err= DBError: update "grants" set "state" = $1, "finalizationReason" = $2, "updatedAt" = $3 where "grants"."id" = $4 - column "finalizationReason" of relation "grants" does not exist
```

### LINKED ISSUE #1959: `getByContinue` retrieves an incorrect grant
Logs from Testnet:
```
getByContinue param continueId 70523ebc-de90...
getbyContinue param continueToken C2A5286D1...

getByContinue result Grant {
  id: 'a78a1fa1-62fb-...',
  state: 'FINALIZED',
  startMethod: [ 'redirect' ],
  continueToken: '3D2B0DE37...',
  continueId: '3a444877-cfcc-...',
  wait: null,
  // ..
}
```

The retrieved grant has completely different values for `continueToken` and `continueId` compared to the grant that should be returned. I _think_ that the issue might be related to Objection's `QueryBuilder` component here:

https://github.com/interledger/rafiki/blob/6a2ae152c4724630e42f9611e05da86605c8dfff/packages/auth/src/grant/service.ts#L256-L266

mkurapov:

Two changes:

**1. Fixing get grant by continuation id** 

In testnet, Radu discovered that when calling grant continue, it would always return the same grant, regardless of the `continueToken` and `continueId`. This was because of how we were building the query to fetch those grants:
https://github.com/interledger/rafiki/blob/6a2ae152c4724630e42f9611e05da86605c8dfff/packages/auth/src/grant/service.ts#L256-L266

This results in a postgres query:
```psql
 'select "grants".* from "grants" where "continueId" = ? and "continueToken" = ? and "finalizationReason" is null or not "finalizationReason" = REVOKED'
```
which ends up returning the first grant that is _not_ revoked. 

This PR fixes this behaviour by adding the `finalizationReason` check within a separate "where" clause.

**2. Fixing finalization reason migration** 
From the issue:

> When declining a grant in Testnet we receive an `Internal Server Error`. The cause of this issue is that a new migration was not generated for the `finalizationReason` column (https://github.com/interledger/rafiki/commit/bf17908a0dc99beab0cd066ab59a065696faaebc). Hence, any action aimed at modifying the `finalizationReason` value of a grant will result in a failure.

This change makes adding `finalizationReason` as a separate migration

## Context

Fixes #1959
Fixes #1958

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/migrations/20220504161932_create_grants_table.js
```diff
@@ -3,7 +3,6 @@ exports.up = function (knex) {
     table.uuid('id').notNullable().primary()
 
     table.string('state')
-    table.string('finalizationReason')
```
raducristianpopa, line 6: I think we can keep this here and ... (👇)

mkurapov:
I redeployed the Netlify build and it works, but the checks aren't being updated in here

packages/auth/migrations/20230925183132_add_finalization_reason_grants.js
```diff
@@ -0,0 +1,31 @@
...
+exports.up = function (knex) {
+  return knex.schema
+    .hasColumn('grants', 'finalizationReason')
+    .then((exists) => {
+      if (!exists) {
+        return knex.schema.alterTable('grants', (table) => {
+          table.enum('finalizationReason', ['ISSUED', 'REVOKED', 'REJECTED'])
+        })
+      }
+    })
```
raducristianpopa, line 14: Let me know if this makes sense:
```suggestion
    .then((exists) => {
      if (!exists) {
        return knex.schema.alterTable('grants', (table) => {
          table.enum('finalizationReason', ['ISSUED', 'REVOKED', 'REJECTED'])
        })
      }
      return knex.schema.alterTable('grants', (table) => {
        table.enum('finalizationReason', ['ISSUED', 'REVOKED', 'REJECTED']).alter()
      })
    })
```

If the column does not exist, create it. 
If the column already exists, alter it to use check constraints.

mkurapov: Unfourtunately wasn't as easy as adding `alter()` to change from string to enum type. I've been looking into several ways to do so, I've come up with a way, just going to be testing it

packages/auth/migrations/20230925183132_add_finalization_reason_grants.js
```diff
@@ -0,0 +1,54 @@
...
+          })
+        })
+      }
+
+      return knex.schema.raw(
+        `CREATE TYPE ${finalzationReasonType} AS ENUM ${toPsqlList(
+          finalizationReasonList
+        )};
+        ALTER TABLE "${grantsTable}" ALTER COLUMN "${finalizationReasonColumn}" TYPE ${finalzationReasonType} USING "${finalizationReasonColumn}"::${finalzationReasonType}`
+      )
```
mkurapov, line 30: https://www.postgresql.org/docs/current/datatype-enum.html#DATATYPE-ENUM-DECLARATION

mkurapov: Screenshot 2023-09-26 at 17 05 09

raducristianpopa (approved):

njlie (approved):

PR #1966, opened 2023-09-28
feat: connect local playground to rafiki.money

### LINKED ISSUE #1957: Autopeer local playground to rafiki.money
rafiki.money supports autopeering.
To peer local playground to rafiki.money it is required to expose the local endpoints.
There are few tasks that should be completed for the autopeering to work:
- [x] enable autopeering on rafiki.money
- [x] create the tunnels for local playground ports
- [x] write .env file with generated urls
- [x] add request for autopeering in playground ase
- [ ] setup autopeering service to a rafiki.money subdomain
- [ ] start the script to generate the urls on running `pnpm localenv:compose up`
- [ ] test payments from rafiki.money to local playground payment pointer

dragosp1011:

- Add a script to create tunnels for local playground services
- Peer local playground with rafiki.money
- Add an env variable for trusting the proxy protocol. This will be used by Koa application to set the protocol from`X-Forwarded-Proto` header if present.

## Context

fixes #1957 
## Checklist

- [ ] Tests added/updated
- [x] Documentation added


sabineschaller:
Is this contradicting?
- https://github.com/interledger/rafiki/pull/1993

dragosp1011:
> Is this contradicting?
> 
> * [fix: Remove hostname from local playground #1993](https://github.com/interledger/rafiki/pull/1993)

does not look so, but there will be conflicts to solve

sabineschaller:
Are you autopeering both of the local instances? I think just autopeering one (cloud-nine-wallet) would be enough.

dragosp1011:
> Are you autopeering both of the local instances? I think just autopeering one (cloud-nine-wallet) would be enough.

yes, only cloud-nine-wallet is autopeering to rafiki.money

sabineschaller:
@dragosp1011 can you add some docs on what is happening? Does the local playground auto-auto-peer (automatically auto-peers with rafiki.money 😉) or do I have to set a variable? How do I turn it off? Do I need an ngrok account? All these kind of things? 

Also, how would I test this working? Can you add a new environment to postman (`Autopeer Environment` or something) that let's us do the Open Payments example with a local sender and a remote receiver?

localenv/README.md
```diff
@@ -109,6 +109,35 @@ The secondary Happy Life Bank docker compose file (`./happy-life-bank/docker-com
 data stores created by the primary Rafiki instance so it can't be run by itself.
 The `pnpm localenv:compose up` command starts both the primary instance and the secondary.
 
+#### Autopeering
+
+If you want to start the local env and peer it automatically to rafiki.money, you can run the following commands:
+
+```
+pnpm localenv:autopeer:start
```
sabineschaller, line 117:
```suggestion
pnpm localenv:autopeer:start -d
```

localenv/README.md
```diff
@@ -109,6 +109,35 @@ The secondary Happy Life Bank docker compose file (`./happy-life-bank/docker-com
...
+// OR to start with Postgres db
+pnpm localenv:autopeer:start:psql
+```
+
+Your local cloud nine rafiki instance will be peered automatically in this case with https://rafiki.money instance.
+The required services will be exposed externally using [ngrok](https://www.npmjs.com/package/ngrok) and [tunnelmole](https://www.npmjs.com/package/tunnelmole) packages.
+Ngrok is used to expose open-payments api and auth api. These apis require a https protocol for correctly validating the requests signatures and ngrok will add X-Forwarded-Proto to secure requests that helps koa to identify requests as secure.
+Tunnelmole package is used for exposing externally ilp connector.
+
+There is no need for ngrok token in case you are planning to try out only the API requests.
```
sabineschaller, line 128: Can you clarify which API requests you mean?

localenv/README.md
```diff
@@ -109,6 +109,35 @@ The secondary Happy Life Bank docker compose file (`./happy-life-bank/docker-com
...
+In case you are looking to test the open payments example with the interaction flow, then the ngrok token should be provided in `~/cloud-nine-wallet/.env` file
+as `NGROK_TOKEN={YOUR TOKEN}`. Ngrok does not serve html without providing the auth token.
```
sabineschaller, line 130: @AlexLakatos can we get an ngrok account and use that token globally? Otherwise the developer experience isn't great.

localenv/README.md
sabineschaller: I guess the logs are required for finding out the wallet address, though 🤔. Can we log that some other way?

localenv/README.md
```diff
@@ -109,6 +109,35 @@ The secondary Happy Life Bank docker compose file (`./happy-life-bank/docker-com
...
+
+To use the postman collection examples - copy the created payment pointer and set it into `senderPaymentPointer` postman variable in `Remote Environment`.
+
+After stopping the script it is necessary to clear the environment using the command described in setup. This is necessary as on a new run of the scripts (with autopeeing or not) the payment pointer url will differ.
```
sabineschaller, line 134:
```suggestion
After stopping the script it is necessary to clear the environment using the command described in [Shutting down](#Shutting-down). This is necessary as on a new run of the scripts (with autopeeing or not) the payment pointer url will differ.
```

localenv/README.md
```diff
@@ -109,6 +109,35 @@ The secondary Happy Life Bank docker compose file (`./happy-life-bank/docker-com
...
+
+```
+// tear down and remove volumes
+pnpm localenv:compose down --volumes
```
sabineschaller, line 138: And then maybe remove this if we reference the next section in the previous paragraph?

sabineschaller (commented):
Should we move the docs to the docs package?

localenv/README.md
AlexLakatos: I thought we weren't using ngrok. What happened to the other options?

dragosp1011: I was using tunnelmole, but the issue was that the requests are redirected as `http` and the signature is not validated as it is expecting `https`.
Koa application has an option to trust the proxy and then it will check the presence of `X-Forwarded-Proto` header. This header is added by ngrok, but not by tunnelmole - this is why ngrok was used for the services that validate signature requests.

The issue with ngrok is that it works fine without a token on api request, but it does not serve `html`. So for testing the interaction part on open payments it requires a token (a free one works) to be provided.

dragosp1011: @sabineschaller @mkurapov does these changes make sense to you?
https://github.com/interledger/rafiki/commit/ae604cc691323ea4d89f69e06b755f2c02655c39

localenv/README.md
dragosp1011: added it in the command and aditional instructions how to find the payment pointer url from docker logs (it will be much easier to find it in ase logs)

localenv/README.md
sabineschaller: Would that mean we get away without using ngrok?

dragosp1011: yes, it will allow to use only tunnelmole

sabineschaller: Let's do that but check `Config.env === 'development'` rather than `Config.env === 'testing'`, or just both.

package.json
```diff
@@ -23,20 +23,30 @@
     "localenv:compose:psql": "docker compose -f ./localenv/cloud-nine-wallet/docker-compose.yml -f ./localenv/happy-life-bank/docker-compose.yml ${BEFORE_MERGED:-} -f ./localenv/merged/docker-compose.yml",
     "localenv:compose": "BEFORE_MERGED='-f ./localenv/tigerbeetle/docker-compose.yml' pnpm localenv:compose:psql",
     "localenv:seed:auth": "pnpm -C ./packages/auth knex seed:run --env=development && pnpm -C ./packages/auth knex seed:run --env=peerdevelopment",
-    "sanity": "pnpm -r build && pnpm -r test"
+    "sanity": "pnpm -r build && pnpm -r test",
+    "localenv:autopeer:start": "run-p tunnel:start wait-tunnel:localenv:compose",
+    "localenv:autopeer:start:psql": "run-p tunnel:start wait-tunnel:localenv:compose:psql",
```
sabineschaller, line 28: 2 comments:
* should we call it `localenv:compose:autopeer` and `localenv:compose:psql:autopeer`?
* I think we would ideally only start cloud-nine-wallet. But that can be a follow up issue/PR

sabineschaller (commented):
Not finished with my review but I'm summoned for lunch ;-)

localenv/mock-account-servicing-entity/app/lib/requesters.ts
```diff
@@ -116,6 +118,52 @@ export async function createPeer(
...
+      }
+    }
+  `
+
+  const addedLiquidity = '10000' as unknown as bigint
+  const createPeerInput: { input: CreateOrUpdatePeerByUrlInput } = {
+    input: {
+      peerUrl,
+      assetId,
+      addedLiquidity
```
sabineschaller, line 150: Isn't that missing the http token?

packages/auth/src/app.ts
```diff
@@ -367,7 +367,9 @@ export class App {
   }
 
   private async createKoaServer(): Promise<Koa<Koa.DefaultState, AppContext>> {
-    const koa = new Koa<DefaultState, AppContext>()
+    const koa = new Koa<DefaultState, AppContext>({
+      proxy: this.config.trustProxy
+    })
```
sabineschaller, line 372: Should we make sure that also only happens in the `autopeer` environment?

packages/backend/src/app.ts
```diff
@@ -696,7 +696,9 @@ export class App {
   }
 
   private async createKoaServer(): Promise<Koa<Koa.DefaultState, AppContext>> {
-    const koa = new Koa<DefaultState, AppContext>()
+    const koa = new Koa<DefaultState, AppContext>({
+      proxy: this.config.trustProxy
+    })
```
sabineschaller, line 701: Same comment as above

packages/auth/src/app.ts
dragosp1011: this is an environment variable with default value to `false`
enabling it will allow koa application to check `X-Forwarded-Proto` header when determining the request protocol.
We are going to use it on rafiki.money to fix the issue with requests that are redirected as `http`.
It should be available in all environments IMO

localenv/mock-account-servicing-entity/app/lib/requesters.ts
dragosp1011: no, there is not http token in this request.
I believe this is taken care of in rafiki

sabineschaller: Indeed: https://github.com/interledger/rafiki/blob/111a5040d158fe345172ca875979ae978b9bf7d3/packages/backend/src/payment-method/ilp/auto-peering/service.ts#L96

sabineschaller (approved):
Let's get it done! Thank you 🙂

Issue #1967, opened 2023-09-28
Create "payment method handler" service, that abstracts away quoting & paying


mkurapov:
In order to be able to support multiple payment methods, we need a service that knows how to do quoting and transfers **without** having the callers of the service (QuoteService, OutgoingPaymentService) know exactly about the details of _how_ quoting and paying are done for different payment methods.

## Tasks
- [x] Introduce "payment method handler" service that abstracts away quoting and paying
- [x] Have `quoteService.createQuote` call the "payment manager" service "getQuote" method
- [ ] Move ILP quoting & transfers into own service
- [ ] Have `outgoingPaymentService` call the "payment manager" service "pay" method

PR #1974, opened 2023-10-03
feat: payment method handler service (quoting)

mkurapov:

- Adds `paymentManagerHandlerService` which will know how to call different payment method services for each payment method (for now, just `ilpPaymentService`)
- Adds `ilpPaymentService` that has a `getQuote` method which handles ILP payments only (essentially copied over from `QuoteService.create`
- Adds `additionalFields` property to quotes to store any payment method-specific quote information

_Note: these new services are only added in this PR_, not used (to break up into smaller PRs). Next up:
- Using `paymentMethodHandlerService.getQuote` in `QuoteService.create` to abstract away ILP quoting

## Context

Progress towards #1967:
> In order to be able to support multiple payment methods, we need a service that knows how to do quoting and transfers without having the callers of the service (QuoteService, OutgoingPaymentService) know exactly about the details of how quoting and paying are done for different payment methods.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/payment-method/ilp/service.test.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    describe('successfully gets ilp quote', (): void => {
+      describe('with incomingAmount', () => {
+        test.each`
+          incomingAssetCode | incomingAmountValue | debitAssetCode | expectedDebitAmount | exchangeRate | slippage | description
+          ${'EUR'}          | ${100n}             | ${'USD'}       | ${101n}             | ${1.0}       | ${0}     | ${'same currency, no slippage'}
+          ${'USD'}          | ${100n}             | ${'USD'}       | ${102n}             | ${1.0}       | ${0.01}  | ${'same currency, some slippage'}
+          ${'EUR'}          | ${100n}             | ${'USD'}       | ${113n}             | ${0.9}       | ${0.01}  | ${'cross currency, exchange rate < 1'}
+          ${'EUR'}          | ${100n}             | ${'USD'}       | ${51n}              | ${2.0}       | ${0.01}  | ${'cross currency, exchange rate > 1'}
+        `(
```
mkurapov, line 256: This is a good demonstration of the interledger `Pay` library behaviour. This may look odd, but some of these are programmed into the library, for example, the off-by-one "issue", where an extra 1 unit is added to the debitAmount to account for possible packet loss. 

Some behavior is _weird_ for example, slippage being taken into account for a payment within the same asset. 
We have taken note of this behaviour, and will continue investigating.

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -0,0 +1,130 @@
...
+    ...deps_,
+    logger: deps_.logger.child({ service: 'IlpPaymentService' })
+  }
+
+  return {
+    getQuote: (quoteOptions) => getQuote(deps, quoteOptions)
+  }
+}
+
+async function getQuote(
```
mkurapov, line 34: basically copy of the ILP-related quoteService.create method

packages/backend/src/tests/quote.ts
```diff
@@ -6,12 +6,56 @@ import { AppServices } from '../app'
...
   withFee?: boolean
 }
 
+type MockQuoteArgs = {
+  receiver: Receiver
+  paymentPointer: PaymentPointer
+  exchangeRate?: number
+} & ({ debitAmountValue: bigint } | { receiveAmountValue: bigint })
+
+export function mockQuote(
```
mkurapov, line 24: not used in this PR, but will make sense in the next

packages/backend/src/index.ts
```diff
@@ -432,6 +434,24 @@ export function initIocContainer(
...
+  })
+
+  container.singleton('paymentMethodHandlerService', async (deps) => {
+    return createPaymentMethodHandlerService({
+      logger: await deps.use('logger'),
+      knex: await deps.use('knex'),
+      ilpPaymentService: await deps.use('ilpPaymentService')
+    })
+  })
+
```
mkurapov, line 454: Not used yet, but will be in the next PR

packages/backend/src/payment-method/handler/service.ts
```diff
@@ -0,0 +1,68 @@
...
+  paymentPointer: PaymentPointer
+  receiver: Receiver
+  debitAmount: Amount
+  receiveAmount: Amount
+  additionalFields: Record<string, unknown>
+}
+
+export interface PaymentMethodService {
+  getQuote(quoteOptions: StartQuoteOptions): Promise<PaymentQuote>
+}
```
mkurapov, line 24: Interfaces that will need to be satisfied for all payment method handlers

packages/backend/src/index.ts
```diff
@@ -432,6 +434,24 @@ export function initIocContainer(
...
+  container.singleton('paymentMethodHandlerService', async (deps) => {
+    return createPaymentMethodHandlerService({
+      logger: await deps.use('logger'),
+      knex: await deps.use('knex'),
+      ilpPaymentService: await deps.use('ilpPaymentService')
+    })
+  })
```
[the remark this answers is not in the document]
njlie: To try and build on this idea, maybe it could take in an object whose keys are the names of supported payment methods, and the values are its services. So `getQuote` would then look like:
```
/*
  PaymentMethods: {
    ilp: IlpPaymentService
    ...other future payment method services
  }

  paymentMethods: PaymentMethods
*/

getQuote: (method, quoteOptions) => paymentMethods[method].getQuote(quoteOptions)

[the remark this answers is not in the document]
mkurapov: @raducristianpopa the `PaymentMethodHandlerService` is our "factory" in this case already, but instead of a new object instance we're just returning the service/singleton itself

@njlie yeah I think I'll change it to the suggestion since the types make it (more) "safe" and prevent passing in an invalid payment method  :+1:

njlie (approved):
LGTM

PR #1976, opened 2023-10-04
feat: add automatic version bumping workflow #1946

### LINKED ISSUE #1946: Automatically bump `package.json` Rafiki version when releasing new package
## Context
When we release a new version of Rafiki, we should automatically bump `package.json` in the root folder, similar to how 
- https://github.com/interledger/rafiki/pull/1945 

is done manually.

### EDIT:
A bump in package.json is a new commit that should not happen after we tag. Hence, what we really need is a workflow that detects that package.json has seen a version bump and then automatically tags on push to main.

error9098x:
fixes #1946 

- Added a new workflow file `.github/workflows/version_bump.yml`.
- The workflow uses `pnpm version prerelease` to automatically bump the version.

Please do let me know if there is any problem with this solution and any other possible improvements.  @mkurapov 

## Context

- The workflow runs on every push to the `main` branch, bumps the prerelease version in `package.json` using the `pnpm version prerelease` command, and pushes the changes back to the `main` branch.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


sabineschaller (commented):
Hi! Thank you for the contribution! 
We would only like to bump the version when we tag. So let's say we tag main with 'v1.0.0-alpha.4", we would also want that to be reflected in the package.json. Also, when we finally get out of alpha into beta or even into release and let's say we tag with "v1.0.0", that should be the version in the package.json as well.

error9098x:
> We would only like to bump the version when we tag. So let's say we tag main with 'v1.0.0-alpha.4", we would also want that to be reflected in the package.json. Also, when we finally get out of alpha into beta or even into release and let's say we tag with "v1.0.0", that should be the version in the package.json as well.

Sure thing, I will add the workflow change the version according to the tag :)

error9098x:
> Hi! Thank you for the contribution! We would only like to bump the version when we tag. So let's say we tag main with 'v1.0.0-alpha.4", we would also want that to be reflected in the package.json. Also, when we finally get out of alpha into beta or even into release and let's say we tag with "v1.0.0", that should be the version in the package.json as well.

Hi, I have updated my code accordingly. Now, it checks for new tag and updated the version in package.json, please do tell me if anything else has to be updated.

sabineschaller:
Looks good to me but I asked @mkurapov to also have a look since he did most of our GH workflows. Thank you!

mkurapov:
@sabineschaller 
I think we should rethink this, since we need the new tag + package version bump _before_ we actually draft & publish a release. What do you think of having a workflow that is just manually ran to create a PR that bumps the package version?

.github/workflows/version_bump.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Bump version
+        run: |
+          git config --global user.name 'github-actions'
+          git config --global user.email 'github-actions@github.com'
+          pnpm version ${GITHUB_REF#refs/tags/v} --no-git-tag-version 
+          git commit -am "Bump version"
+          git push origin main
```
mkurapov, line 29: Since `main` is a protected branch, we would need to instead create a PR before merging into `main`.
We could use the https://github.com/marketplace/actions/create-pull-request action here to do the creating/updating of a PR. 

The `pnpm version ${GITHUB_REF#refs/tags/v} --no-git-tag-version` looks good though :+1:

.github/workflows/version_bump.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18'
+
+      - name: Install dependencies
+        run: pnpm install
```
mkurapov, line 21: I think we can get away without installing all of the packages, just need to install `pnpm` using `- uses: pnpm/action-setup@v2`

.github/workflows/version_bump.yml
```diff
@@ -0,0 +1,31 @@
+name: Version Bump
+
+on:
+  create: # It will trigger whenever a new tag or branch is created
```
mkurapov, line 4: This will probably be triggered by a manual workflow instead, pending https://github.com/interledger/rafiki/pull/1976#issuecomment-1757881005

sabineschaller:
> I think we should rethink this, since we need the new tag + package version bump _before_ we actually draft & publish a release.

Yes, you are right. I was somehow hoping we could do that as a pre-commit hook or something but tagging is not committing...

sabineschaller:
@error9098x @mkurapov 
What if we did it the other way around. What if we had a workflow that tagged the repo when we push a change to main that includes a version bump in the package.json?

PR #1994, opened 2023-10-06
feat: use payment manager handler in quote service

mkurapov:

- Abstracts away ILP-specific details during `quoteService.create` by using `paymentMethodHandlerService.getQuote`
- Minor refactoring of quote service
-Adding `additionalFields` to the quote model for storing method-specific information that may need to be stored in between quoting and transfers (like the ILP maxPacketAmount and exchangeRate info)
    - There is a separate issue for removing the ILP specific columns in the quote table here: https://github.com/interledger/rafiki/issues/2000

## Context

Progress towards #1967 

In the previous PR, #1974, the `PaymentMethodHandlerService` was introduced, but not used by anything. In this PR, `quoteService` will now use this service to get a quote, without needing to know exactly how an "ILP" quote happens.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/quote/service.ts
+++ b/packages/backend/src/open_payments/quote/service.ts
@@ -140,6 +140,7 @@ async function createQuote(
           expiresAt: new Date(0),
           client: options.client,
           feeId: sendingFee?.id,
+          estimatedExchangeRate: quote.estimatedExchangeRate,
           additionalFields: quote.additionalFields
         })
         .withGraphFetched('[asset, fee]')
@@ -218,7 +219,7 @@ function calculateQuoteAmountsAfterFees(
     // FixedSend
     const fees = quote.fee?.calculate(receiveAmountValue) ?? 0n
     const exchangeAdjustedFees = BigInt(
-      Number(fees) * quote.lowEstimatedExchangeRate.valueOf()
+      Number(fees) * quote.estimatedExchangeRate.valueOf()
     )
     receiveAmountValue = BigInt(receiveAmountValue) - exchangeAdjustedFees
 
--- a/packages/backend/migrations/20230927223235_make_quotes_generic.js
+++ b/packages/backend/migrations/20230927223235_make_quotes_generic.js
@@ -5,6 +5,7 @@
 exports.up = function (knex) {
   return knex.schema.alterTable('quotes', (table) => {
     table.json('additionalFields').nullable()
+    table.decimal('estimatedExchangeRate', 20, 10).notNullable()
   })
 }
 
@@ -15,5 +16,6 @@ exports.up = function (knex) {
 exports.down = function (knex) {
   return knex.schema.alterTable('quotes', (table) => {
     table.dropColumn('additionalFields')
+    table.dropColumn('estimatedExchangeRate')
   })
 }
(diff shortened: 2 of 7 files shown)


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -116,69 +106,52 @@ async function createQuote(
...
-          },
-          receiveAmount: {
-            value: ilpQuote.minDeliveryAmount,
-            assetCode: receiver.assetCode,
-            assetScale: receiver.assetScale
-          },
+          debitAmount: quote.debitAmount,
+          receiveAmount: quote.receiveAmount,
           // Cap at MAX_INT64 because of postgres type limits.
           maxPacketAmount:
```
njlie, line 132: Thoughts on refactoring Quote so that ILP-related fields are more delineated from "base" quote stuff? I was thinking it might help keep the amount of columns in quote reasonable when other payment methods become supported.

mkurapov: 💯 
I have it in a separate issue:
- https://github.com/interledger/rafiki/issues/2000

packages/backend/migrations/20230927223235_add_additionalFields_quotes.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('quotes', (table) => {
+    table.json('additionalFields').nullable()
```
BlairCurrey, line 7: Is this meant to be temporary until ilp related stuff is migrated in https://github.com/interledger/rafiki/issues/2000?

otherwise I'm concerned about the schemaless-ness of jsonb

packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -506,49 +473,55 @@ describe('QuoteService', (): void => {
...
-            })
-            assert.ok(!isQuoteError(quote))
-
-            expect(quote.debitAmount).toEqual({
+        async ({
+          incomingAmountValue,
+          fixedFee,
+          basisPointFee,
+          expectedQuoteDebitAmountValue
+        }): Promise<void> => {
```
BlairCurrey, line 487: Can you elaborate on how there isnt an off by one issue here? Did we gain any insight into that issue?

As I recall the default slippage introduced some variance in this test and manual tests (via postman), which is why I tried to test without slippage originally. Although it appears that using the default slippage doesnt modify the values at all here?

packages/backend/migrations/20230927223235_add_additionalFields_quotes.js
mkurapov: Long term, temporary. I need to determine what non generic items are needed for the mojaloop quote, and decide how to store those as well. I imagine we end up having a separate table for method-specific quoting/transfer items

packages/backend/src/open_payments/quote/service.test.ts
BlairCurrey: aha, makes sense. Good, glad there is something preserving that issue.

BlairCurrey (dismissed):

BlairCurrey (approved):

PR #2003, opened 2023-10-06
fix(auth): fix noninteractive grant creation

### LINKED ISSUE #1968: Fix non-interactive grant creation
## Context
When we are creating a grant in `grantService.create`, we are setting the status of the grant to be `Approved` or `Pending` depending on whether some `interact` property is passed in:
https://github.com/interledger/rafiki/blob/7eedb1e57a0c5b5b346b7fb4ee6bc6fc8e1fca5f/packages/auth/src/grant/service.ts#L216

However, in the route: https://github.com/interledger/rafiki/blob/7eedb1e57a0c5b5b346b7fb4ee6bc6fc8e1fca5f/packages/auth/src/grant/routes.ts#L131
we take in the whole request body into the service, meaning, that even if we've set up the grant to be non-interactive (like for incoming payments or quotes), and the requestor still passes in an `interact` property, the grant will incorrectly be Pending, not  Approved.

We should fix this behaviour to make sure that even if `interact` property is passed into the request, we create an "Approved" grant for resources that do not require interaction.

## Todos
- [ ] Make sure non-interactive grants cause grants to be "Approved" regardless of whether the `interact` field was passed into the request

njlie:

- Refactored `canSkipInteraction` as a utility function.
- Pulls in `canSkipInteraction` in the Grant service so its `create` function can properly create a noninteractive grant.

## Context

Fixes #1968.

Grant creation was handling noninteractive grant creation by checking if there is an `interact` property in the grant request. But that property may be present in a request whose requested access does not necessitate interaction.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/service.test.ts
raducristianpopa, line 1: Would it be worth adding a test scenario where we provide all three accesses?

raducristianpopa (commented):
LGTM. You need to run `pnpm format`. One question I have:

packages/auth/src/grant/service.test.ts
njlie: That sounds like a good idea, will add.

packages/auth/src/grant/service.test.ts
```diff
@@ -135,38 +135,46 @@ describe('Grant Service', (): void => {
...
-            {
-              ...BASE_GRANT_ACCESS,
-              type: AccessType.IncomingPayment
-            }
-          ]
-        },
-        interact: undefined
-      }
+    test.each`
+      type                          | state                  | interact                       | not       | description
```
mkurapov, line 139:
```suggestion
      type                          | expectedState                  | interact                       | not       | description
```

packages/auth/src/grant/service.test.ts
```diff
@@ -135,38 +135,46 @@ describe('Grant Service', (): void => {
...
+      ${AccessType.IncomingPayment} | ${GrantState.Approved} | ${undefined}                   | ${''}     | ${'incoming payment'}
+      ${AccessType.Quote}           | ${GrantState.Approved} | ${undefined}                   | ${''}     | ${'quote'}
+      ${AccessType.OutgoingPayment} | ${GrantState.Pending}  | ${BASE_GRANT_REQUEST.interact} | ${' not'} | ${'outgoing payment'}
+    `(
+      'Does$not approve a grant without interaction - $description',
```
mkurapov, line 144: Nit-picky, but instead of using `$not` we can just put everything in the description

mkurapov (approved):

packages/auth/src/grant/service.test.ts
sabineschaller: I tried but I couldn't come up with something that made sense for all scenarios

raducristianpopa (approved):

mkurapov (approved):

PR #2028, opened 2023-10-12
feat: add pay method to ilpPaymentService

mkurapov:

This PR adds `pay` method support to the `ilpPaymentService`. This PR is basically prepping the service to be used during the lifecycle of an Outgoing Payment (stay tuned shortly), to abstract out all of the ILP pay stuff, so the outgoingPayment lifecycle is payment method agnostic:
https://github.com/interledger/rafiki/blob/c16a60fdfa10563998ffc26ea3b669109a98fe15/packages/backend/src/open_payments/payment/outgoing/lifecycle.ts#L91-L131

It also introduces `PaymentMethodHandlerError` error class that will be used to signify details about a thrown error during the `paymentMethodHandler.pay` and  `paymentMethodHandler.startQuote` methods, so the caller can react according, for example, whether it is safe to retry the payment/quote or not.

## Context

Progress towards #1967 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/payment-method/handler/service.ts
+++ b/packages/backend/src/payment-method/handler/service.ts
@@ -1,4 +1,5 @@
 import { Amount } from '../../open_payments/amount'
+import { OutgoingPayment } from '../../open_payments/payment/outgoing/model'
 import { PaymentPointer } from '../../open_payments/payment_pointer/model'
 import { Receiver } from '../../open_payments/receiver/model'
 import { BaseService } from '../../shared/baseService'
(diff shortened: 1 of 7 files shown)


packages/backend/src/payment-method/handler/service.ts
```diff
@@ -20,8 +21,16 @@ export interface PaymentQuote {
   additionalFields: Record<string, unknown>
 }
 
+export interface PayOptions {
+  receiver: Receiver
+  outgoingPayment: OutgoingPayment
+  finalDebitAmount: bigint
+  finalReceiveAmount: bigint
```
mkurapov, line 28: these are called 'finalX' because the amount is adjusted in the outgoing payment service before calling ILP pay (due to retries, for example, where in the case of ILP pay, we've already sent some amount of currency)

packages/backend/src/payment-method/ilp/service.test.ts
```diff
@@ -305,65 +342,368 @@ describe('IlpPaymentService', (): void => {
...
-              const sendingPaymentPointer = paymentPointerMap[debitAssetCode]
-
-              const options: StartQuoteOptions = {
-                paymentPointer: sendingPaymentPointer,
-                receiver: await createReceiver(deps, receivingPaymentPointer),
-                debitAmount: {
-                  assetCode: sendingPaymentPointer.asset.code,
-                  assetScale: sendingPaymentPointer.asset.scale,
-                  value: debitAmountValue
+      describe('with debitAmount', () => {
```
mkurapov, line 346: just moved into the parent describe block, no changes

njlie (approved):

PR #2031, opened 2023-10-13
feat: use paymentMethodHandler in outgoing payment lifecycle

### LINKED ISSUE #1967: Create "payment method handler" service, that abstracts away quoting & paying

mkurapov:

- Building up from https://github.com/interledger/rafiki/pull/2028, this PR allows the Outgoing Payment lifecycle to use `paymentMethodHandlerService.pay` method. This allows abstracting away the ILP pay stuff from the outgoing payment lifecycle. The original retry logic for ILP payments is kept: it has been abstracted away to the ilpPaymentService instead, which will appropriately throw a "retryable" `PaymentMethodHandlerError`.

## Context

Officially closes #1967 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -257,7 +256,8 @@ async function validateGrant(
   deps: ServiceDependencies,
   payment: OutgoingPayment,
   grant: Grant,
-  callback?: (f: unknown) => NodeJS.Timeout
+  callback?: (f: unknown) => NodeJS.Timeout,
+  grantLockTimeoutMs: number = 5000
```
mkurapov, line 260: allow setting this locking value so we can set it to something much lower in tests

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -42,95 +38,76 @@ export async function handleSending(
...
     )
     await handleCompleted(deps, payment)
     return
-  } else if (
-    newMaxSourceAmount <= BigInt(0) ||
-    newAmountToDeliver <= BigInt(0)
-  ) {
+  }
+
+  if (maxDebitAmount <= BigInt(0)) {
```
sabineschaller, line 64: Why is the second condition dropped?

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -42,95 +38,76 @@ export async function handleSending(
...
-      // Ignore connection close failures, all of the money was delivered.
-      deps.logger.warn(
-        {
-          destination: destination.destinationAddress,
-          error: err.message
-        },
-        'close pay connection failed'
-      )
-    })
+  await deps.paymentMethodHandlerService.pay('ILP', {
```
sabineschaller, line 77: Instead of hardcoding `ILP`, should we not make that configurable right away?

packages/backend/src/open_payments/payment/outgoing/worker.ts
```diff
@@ -35,8 +35,7 @@ export async function processPendingPayment(
 }
 
 // Fetch (and lock) a payment for work.
-// Exported for testing.
```
sabineschaller, line 38: Did we not test it or are we not testing it anymore or do we test somewhere else?

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -226,22 +227,18 @@ describe('OutgoingPaymentService', (): void => {
...
-          XRP: 0.5
-        }
-      }))
-      .persist()
-    deps = await initIocContainer(Config)
+    const exchangeRatesUrl = 'https://test.rates'
+
+    mockRatesApi(exchangeRatesUrl, () => ({
+      XRP: exchangeRate
+    }))
```
sabineschaller, line 234: This is not the format that we use right now, is it?

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
mkurapov: We need several changes to the quote model for this to not be hardcoded so I think this should be changed during the PR that adds support for open payments payment methods

packages/backend/src/open_payments/payment/outgoing/worker.ts
mkurapov: This was never tested directly, the tests for `processNext` tested this

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
mkurapov: it wasn't necessary, since it was the same as the first `if` statement

packages/backend/src/open_payments/payment/outgoing/service.test.ts
mkurapov: Those are just for the rates themselves, the actual mocking returns the usual `{ base, rates: { ... }}` object
```ts
export function mockRatesApi(
  exchangeRatesUrl: string,
  getRates: (baseAssetCode: unknown) => Rates['rates']
) {
  return nock(exchangeRatesUrl)
    .get('/')
    .query(true)
    .reply(200, (url) => {
      const base = url.split('=')[1]
      const rates = getRates(base)

      return {
        base,
        rates
      }
    })
    .persist()
}

```

sabineschaller: Gotcha. Must have missed this with all the changes there were...

sabineschaller (approved):
LGTM, but maybe get one more review?

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: We could address this in our mojaloop payment method effort when we actually begin utilizing a second payment method. It's not strictly necessary for updating rafiki to use the latest open-payments so I'm concerned it would unnecessarily slow that effort down. I can include it if we want its just not clear to me what this would take.

PR #1940, opened 2023-09-20
feat(backend): remove connections route

omertoast:
- remove connections route from open payments spec
- remove connections route and any associated service functions from backend

## Changes proposed in this pull request

- 

## Context

- fixes #1895 
## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass
- [ ] Postman collection updated

PR #2051, opened 2023-10-18
feat(backend): remove connections route (#1940)

njlie:
- remove connections route from open payments spec
- remove connections route and any associated service functions from backend

- 

## Context

This is the same as #1940, but with some rebasing work.

Somehow, during the maintenance of `nl-open-payments` to be kept up to date with main, this commit was lost. This PR will re-introduce the changes.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


openapi/resource-server.yaml
```diff
@@ -103,42 +103,6 @@ paths:
       operationId: get-wallet-addres-keys
       description: Retrieve the public keys of the Wallet Address.
       security: []
-  '/connections/{id}':
```
mkurapov, line 106: is this via the `fetch-schemas.sh`?

packages/backend/src/open_payments/connection/routes.ts
```diff
@@ -1,35 +0,0 @@
-import { Logger } from 'pino'
```
mkurapov, line 1: I think we have tests for this file

openapi/resource-server.yaml
njlie: I think when this commit was originally written Omer modified this file by hand. Could remove this change and commit `fetch-schemas.sh` changes separately.

Issue #1962, opened 2023-09-26
Implement payment methods and finish removing connections


BlairCurrey:
- use payment method from incoming-payments instead of connection in lifecycle/receiver
- remove unused connection request in postman

PR #2054, opened 2023-10-18
refactor(backend): swap connection for methods

BlairCurrey:

- updates to latest open-payments openapi spec  (5.2.2)
- refactors to remove connection and replace with methods
- connection service was reduced to just get the stream credentials and has been renamed to `streamCredentialService`

## Context

This branch is does not build yet, and not all tests pass, which is expected. This is due to other unimplemented changes. I fixed all build errors related to these change and the test failures I see locally appear related to pagination (possibly due to bug in our down migration in our dev branch at the moment?).

- fixes https://github.com/interledger/rafiki/issues/1962

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/quote/model.ts
```diff
@@ -152,6 +153,10 @@ export class Quote extends WalletAddressSubresource {
     this.highEstimatedExchangeRateDenominator = value.b.value
   }
 
+  public get method(): 'ilp' {
+    return 'ilp'
+  }
+
```
BlairCurrey, line 159: @mkurapov wasn't quite sure what our intention is for modeling payment methods.

The new openapi spec requires the quote type in `POST /quote` body but I didn't actually add it to the create path (`quoteRoutes.create`, `quoteService.create`, `createQuote` test util etc) since its not stored anywhere and I'm not sure how we want to do that (method column on Quote? fkey column on Quote point to Methods table?)

BlairCurrey:
the `ilp` payment method is hardcoded a number of places. I'm not against abstracting it a bit to handle multiple methods where it's clear and simple but I wanted to keep the scope to what is necessary to maintain functionality of `ilp` payment methods. In additional to maximizing the amount of time we have to update testnet, I think the right way to handle multiple methods will probably be more clear when we actually add the next method.

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -74,39 +52,31 @@ export class Receiver extends ConnectionBase {
...
-      connection.ilpAddress,
-      base64url.toBuffer(connection.sharedSecret),
-      connection.assetCode,
-      connection.assetScale
-    )
+    this.ilpAddress = ilpMethod.ilpAddress
+    this.sharedSecret = base64url.toBuffer(ilpMethod.sharedSecret)
+    // TODO: ensure these are the right assetCode/Scales
+    this.assetCode = incomingPayment.receivedAmount.assetCode
+    this.assetScale = incomingPayment.receivedAmount.assetScale
```
BlairCurrey, line 70: I will address this TODO before merging. Previously we were getting these off the connection and I wasn't sure where we should get it now - Max suggested this and I believe it is correct. Tests seem to back it up but any more common-sense checks would be welcome https://github.com/interledger/rafiki/blob/04051d144787eb0bee75283416ccd6002f07d3f2/packages/backend/src/open_payments/receiver/model.test.ts#L49-L51

packages/backend/src/open_payments/connection/service.ts
```diff
@@ -44,19 +46,8 @@ function getConnection(
...
-  })
-}
-
-function getConnectionUrl(
-  deps: ServiceDependencies,
-  payment: IncomingPayment
-): string | undefined {
-  if (!payment.connectionId) {
-    return undefined
+  return {
```
mkurapov, line 49: we can probably get away with the `if (!payment.connectionId)` check, and remove `connectionid` altogether from the incoming payment model

packages/backend/src/open_payments/quote/model.ts
mkurapov: I like having the idea of a `PaymentMethods` table that determines what types of payment methods are active or inactive, but for the sake of time this (hard-coding ilp) works well

packages/backend/src/open_payments/receiver/model.ts
mkurapov: Other than fetching the walletAddress asset, I don't think there is really any other option that we have

mkurapov (commented):
Looks good

sabineschaller:
> We still have a concept of connections internally because we still need to get the stream credentials for the ilp payment method.
Should we call it `streamCredentials` then? Or something else that is not as generic as `connection`? Since I'm assuming we'll have something similar for MJL?

mkurapov:
Stream credentials is good

mkurapov:
Almost makes me think we can just pass in the `streamServer` into `incomingPaymentService` and remove `connectionService` altogether, thoughts? @sabineschaller @BlairCurrey

packages/backend/src/open_payments/connection/service.ts
BlairCurrey: I removed the `connectionId` from `IncomingPayment`. I kept this check to return undefined when payment is completed or expired to maintain the behavior tested here: 

https://github.com/interledger/rafiki/blob/4f1c1787d2efc14108a56344b42de055590aeafb/packages/backend/src/open_payments/connection/service.test.ts#L93-L107

A `beforeInsert` hook ensured the `connectionId` was always set, and then it would be set to `null` in a `beforeUpdate` hook when the state was completed or expired. So it seems this was basically a roundabout way to check if the payment was expired or complete.

mkurapov: Wouldn't this mean that the Open Payments GET /incoming-payment/ fails if the payment is expired or completed since we won't be returning the methods array?

mkurapov: In this case, if there is no `ilpStreamConnection` passed into the `incomingPayment.toOpenPaymentsType`, we can just return an empty methods array

packages/backend/src/open_payments/quote/model.ts
```diff
@@ -20,7 +20,8 @@ export class Quote extends WalletAddressSubresource {
       'receiveAmount',
       'minExchangeRate',
       'lowEstimatedExchangeRate',
-      'highEstimatedExchangeRate'
+      'highEstimatedExchangeRate',
+      'method'
```
mkurapov, line 24: we should also be able to expect the `method` field as an argument to quote creation

BlairCurrey: I'm not sure if I fully understand so please clarify if I have something wrong. Given that this is an objection model, these quotes objects are created from inserting and querying the database. Rather than doing something like `new Quote({..., method: 'ilp'})`. So I cant pass it in to the quote creation because it doesnt exist in the database and [I think we agreed that the database changes didn't need to be in this PR](https://github.com/interledger/rafiki/pull/2054#discussion_r1364740532). And if its defined as a real class attribute, it will try to write `ilp` to a non-existant `method` column where ever we insert a quote.

At face value I think it would be appropriate to include all of that here (add to `quoteService.create`, adjust everywhere that's called, add migration). However I think we're sacrificing debugging of `nl-open-payments-updates` at this point to do so, and it's functional as-is. I can capture this improvement in a task if we like, and work on that if we have extra time, otherwise do it after the code-freeze.

packages/backend/src/open_payments/connection/service.ts
BlairCurrey: > Wouldn't this mean that the Open Payments GET /incoming-payment/ fails if the payment is expired or completed since we won't be returning the methods array?

I think you're right but I dont think we can just add an empty methods array to `toOpenPaymentsType` when no connections are passed in. Sometimes it should return the payment without methods to satisfy the base `incoming-payment` openapi type. I guess we could make a seperate `toOpenPaymentsTypeWithMethods` (not overloaded) and call as appropriate. But should the method really be empty when the payment is expired/complete? Do we not want the actual payment method with credentials?

I'm thinking your original suggestion makes the most sense now. If we dont return `undefined` for completed or expired payments then we can handle completed/expired where its called if we want, and do nothing otherwise (such as `GET /incoming-payment/{id}`). I implemented that in this new commit https://github.com/interledger/rafiki/pull/2054/commits/8396970b59a2483728d210d8623d1f2e977b34bb

packages/backend/src/open_payments/quote/model.ts
sabineschaller: I think what Max meant was adding `method` to the arguments of `quoteService.create`.

BlairCurrey: We can, but it will be unused until we make the db changes

packages/backend/src/open_payments/payment/incoming/service.test.ts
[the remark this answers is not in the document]
BlairCurrey: credentials were never being generated in the incoming payment service. so I think `streamCredentials` generation is a non-factor here unless Im missing something. I dont see anywhere I could add such a check.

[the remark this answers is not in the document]
BlairCurrey: im also going to restore the original behavior to `streamCredentials.get` where it returned undefined when the incoming payment was complete or expired.

packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -194,47 +198,24 @@ export class IncomingPayment
...
-  ):
-    | OpenPaymentsIncomingPaymentWithConnection
-    | OpenPaymentsIncomingPaymentWithConnectionUrl
+    ilpStreamCredentials?: IlpStreamCredentials
+  ): OpenPaymentsIncomingPayment | OpenPaymentsIncomingPaymentWithPaymentMethod
 
   public toOpenPaymentsType(
     walletAddress: WalletAddress,
-    ilpStreamConnection?: Connection | string
+    ilpStreamCredentials?: IlpStreamCredentials
```
njlie: To spitball a future idea for this, maybe this parameter could be genericized into `method` which has a type that all payment methods implement and can call a `.toOpenPayments()` function.

packages/backend/src/open_payments/connection/service.ts
BlairCurrey: updated. The caller of `toOpenPaymentsType` has to distinguish between which type (with methods or without) so that we ensure we're returning the right object with respect to the methods field. We can no longer use the presence of the `ilpStreamCredentials` to determine it alone. An undefined `ilpStreamCredentials` should sometimes return no methods, and sometimes `[]` depending on where its called.

As such I removed the overloading (rather than perhaps adding an additional `withMethods` boolean) since it makes the internal function logic much simpler and the caller already has to differentiate between which type it wants anyways.

packages/backend/src/open_payments/payment/incoming/model.ts
BlairCurrey: Im not sure if I'm following but wanted to note there have been some recent changes that might effect this suggestion: https://github.com/interledger/rafiki/pull/2054#discussion_r1367369897

packages/backend/src/open_payments/payment/incoming/routes.test.ts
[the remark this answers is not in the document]
BlairCurrey: done

postman/collections/Interledger.json
```diff
@@ -1654,12 +1620,10 @@
 									}
 								],
 								"url": {
-									"raw": "http://localhost:4000/incoming-payments/{{incomingPaymentId}}",
-									"protocol": "http",
+									"raw": "{{pfryWalletAddress}}/incoming-payments/{{incomingPaymentId}}",
```
njlie, line 1623: I think the postman collection needs to be updated again - looks like it swallowed the "remove wallet address from url" changes.

njlie (approved):

njlie:
@mkurapov @sabineschaller If there's any other feedback you'd like to give in this PR feel free to make comment on pertinent lines as usual, but in the interest of keeping things moving I'd also like for you to create an issue so we can address it in a subsequent PR on the `nl-open-payments` feature branch.

PR #2090, opened 2023-10-23
feat: add option to paginate in desc order

### LINKED ISSUE #2088: Add a sorting option to the getPage function
For the Admin UI we want to be able to have payments, webhooks and asset fees display in descending order. Currently the getPage function used for pagination displays results in ascending order only.

JoblersTune:

- Add a sortOrder paramater to the getPage function used for pagination so that results can be retrieved for ascending / descending order.

## Context

- fixes #2088 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/fee/service.ts
+++ b/packages/backend/src/fee/service.ts
@@ -2,7 +2,7 @@ import { ForeignKeyViolationError } from 'objection'
 import { BaseService } from '../shared/baseService'
 import { FeeError } from './errors'
 import { Fee, FeeType } from './model'
-import { Pagination } from '../shared/baseModel'
+import { Pagination, SortOrder } from '../shared/baseModel'
 
 export interface CreateOptions {
   assetId: string
@@ -15,7 +15,11 @@ export interface CreateOptions {
 
 export interface FeeService {
   create(CreateOptions: CreateOptions): Promise<Fee | FeeError>
-  getPage(assetId: string, pagination?: Pagination): Promise<Fee[]>
+  getPage(
+    assetId: string,
+    pagination?: Pagination,
+    sortOrder?: SortOrder
+  ): Promise<Fee[]>
   getLatestFee(assetId: string, type: FeeType): Promise<Fee | undefined>
 }
 
@@ -33,8 +37,11 @@ export async function createFeeService({
   }
   return {
     create: (options: CreateOptions) => createFee(deps, options),
-    getPage: (assetId: string, pagination: Pagination) =>
-      getFeesPage(deps, assetId, pagination),
+    getPage: (
+      assetId: string,
+      pagination: Pagination,
+      sortOrder = SortOrder.Desc
+    ) => getFeesPage(deps, assetId, pagination, sortOrder),
     getLatestFee: (assetId: string, type: FeeType) =>
       getLatestFee(deps, assetId, type)
   }
@@ -43,9 +50,12 @@ export async function createFeeService({
 async function getFeesPage(
   deps: ServiceDependencies,
   assetId: string,
-  pagination?: Pagination
+  pagination?: Pagination,
+  sortOrder?: SortOrder
 ): Promise<Fee[]> {
-  const query = Fee.query(deps.knex).where({ assetId }).getPage(pagination)
+  const query = Fee.query(deps.knex)
+    .where({ assetId })
+    .getPage(pagination, sortOrder)
 
   return await query
 }
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -12,8 +12,8 @@ type Query {
     first: Int
     "Paginating backwards: The last **n** elements from the page."
     last: Int
-    "Choose ascending or descending order for results."
-    sortOrder: String
+    "Ascending or descending order of creation."
+    sortOrder: SortOrder
   ): AssetsConnection!
 
   "Fetch a peer"
@@ -29,8 +29,8 @@ type Query {
     first: Int
     "Paginating backwards: The last **n** elements from the page."
     last: Int
-    "Choose ascending or descending order for results."
-    sortOrder: String
+    "Ascending or descending order of creation."
+    sortOrder: SortOrder
   ): PeersConnection!
 
   "Fetch a wallet address"
@@ -46,8 +46,8 @@ type Query {
     first: Int
     "Paginating backwards: The last **n** elements from the page."
     last: Int
-    "Choose ascending or descending order for results."
-    sortOrder: String
+    "Ascending or descending order of creation."
+    sortOrder: SortOrder
   ): WalletAddressesConnection!
 
   "Fetch an Open Payments quote"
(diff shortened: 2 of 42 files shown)


packages/backend/src/fee/service.ts
```diff
@@ -45,7 +45,9 @@ async function getFeesPage(
   assetId: string,
   pagination?: Pagination
 ): Promise<Fee[]> {
-  const query = Fee.query(deps.knex).where({ assetId }).getPage(pagination)
+  const query = Fee.query(deps.knex)
+    .where({ assetId })
+    .getPage(pagination, 'desc')
```
sabineschaller, line 50: I think we should rather pass the sorting order as a parameter. That way, we can have a toggle on the Admin UI in the future allowing us to sort both ways.

packages/backend/src/webhook/service.ts
```diff
@@ -186,5 +186,5 @@ async function getWebhookEventsPage(
     query.whereIn('type', filter.type.in)
   }
 
-  return await query.getPage(pagination)
+  return await query.getPage(pagination, 'desc')
```
sabineschaller, line 189: Same comment as for fees. Let's pass the sorting order as a parameter.

packages/backend/src/shared/baseModel.ts
```diff
@@ -57,33 +57,34 @@ class PaginationQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<
...
       return this.whereRaw(
-        `("${tableName}"."createdAt", "${tableName}"."id") < (select "${tableName}"."createdAt" :: TIMESTAMP, "${tableName}"."id" from ?? where "${tableName}"."id" = ?)`,
+        `("${tableName}"."createdAt", "${tableName}"."id") ${comparisonOperator} (select "${tableName}"."createdAt" :: TIMESTAMP, "${tableName}"."id" from ?? where "${tableName}"."id" = ?)`,
         [this.modelClass().tableName, pagination.before]
       )
         .orderBy([
-          { column: 'createdAt', order: 'desc' },
-          { column: 'id', order: 'desc' }
+          { column: 'createdAt', order: order },
+          { column: 'id', order: order }
```
sabineschaller, line 87:
```suggestion
          { column: 'createdAt', order },
          { column: 'id', order }
```

packages/backend/src/shared/baseModel.ts
```diff
@@ -45,7 +45,7 @@ class PaginationQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<
    * @param pagination Pagination - cursors and limits.
    * @returns Model[] An array of Models that form a page.
    */
-  getPage(pagination?: Pagination): this {
+  getPage(pagination?: Pagination, sortOrder: 'asc' | 'desc' = 'asc'): this {
```
raducristianpopa, line 48: I would probably go for `desc` as a default.

JoblersTune:
@raducristianpopa and @sabineschaller, how is this looking now? I've added a sortOrder parameter to the GraphQL query and we are now sorting in desc order by default.

packages/backend/src/shared/baseModel.ts
```diff
@@ -57,33 +65,34 @@ class PaginationQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<
...
         ])
         .limit(first)
     }
-
     /**
      * Backward pagination
      */
     if (typeof pagination?.before === 'string') {
+      const comparisonOperator = sortOrder === 'asc' ? '<' : '>'
+      const order = sortOrder === 'asc' ? 'desc' : 'asc'
```
mkurapov, line 88:
```suggestion
      const comparisonOperator = sortOrder === SortOrder.Asc ? '<' : '>'
      const order = sortOrder === SortOrder.Asc ? SortOrder.Desc : SortOrder.Asc 
```

packages/backend/src/graphql/schema.graphql
```diff
@@ -12,6 +12,8 @@ type Query {
     first: Int
     "Paginating backwards: The last **n** elements from the page."
     last: Int
+    "Choose ascending or descending order for results."
+    sortOrder: String
```
mkurapov, line 16: we should make this an enum in the schema, maybe all upper case similar to how we do most enums

packages/backend/src/payment-method/ilp/peer/service.test.ts
```diff
@@ -427,8 +427,9 @@ describe('Peer Service', (): void => {
   describe('Peer pagination', (): void => {
     getPageTests({
       createModel: () => createPeer(deps, { assetId: asset.id }),
-      getPage: (pagination: Pagination | undefined) =>
-        peerService.getPage(pagination)
+      getPage: (pagination?: Pagination, sortOrder?: SortOrder) =>
+        peerService.getPage(pagination, sortOrder),
+      sortOrder: Math.random() < 0.5 ? SortOrder.Asc : SortOrder.Desc
```
mkurapov, line 432: up to you, but we could actually put this line in `baseModel.test.ts` so we don't have to add this for every `getPageTests` separately

packages/backend/src/graphql/schema.graphql
mkurapov: Ideally, in the future `sortOrder` becomes something more customizable such that you can give it the key you want to sort on e.g. `createdAt` `updatedAt` `activatedAt` but this is good for the time being since we are hardcoding `createdAt` in the pagination code

mkurapov: But maybe we can be more specific in the comment "Choose ascending or descending order for results." -> "Ascending or descending order of creation" or something

packages/backend/src/shared/baseModel.test.ts
```diff
@@ -1,13 +1,15 @@
...
-  getPage: (pagination?: Pagination) => Promise<Type[]>
+  getPage: (pagination?: Pagination, sortOrder?: SortOrder) => Promise<Type[]>
+  sortOrder: SortOrder
 }
 
 export const getPageTests = <Type extends BaseModel>({
   createModel,
-  getPage
+  getPage,
+  sortOrder
```
mkurapov, line 12: maybe there is a nice way to explicitly test sortOrder for all of the 4 cases in here (forwards/backwards x asc/desc), so that we wouldn't even need the `Math.random() < 0.5 ? SortOrder.Asc : SortOrder.Desc` line

mkurapov (commented):
looks good, a few comments!

packages/backend/src/shared/baseModel.test.ts
JoblersTune: I am happy to do implement this. The reason that I did it randomly was just to try and keep test time in mind. But it's probably better to be thorough.

sabineschaller:
Let's merge this and upgrade the tests in a follow up PR, @JoblersTune  @mkurapov

sabineschaller (approved):
LGTM, it works 😅

mkurapov (approved):

packages/backend/src/shared/baseModel.test.ts
JoblersTune: Note that we're moving this into a separate issue.

PR #2123, opened 2023-10-25
Change timeout to ms in balance.ts and transfers.ts

### LINKED ISSUE #1112: Fix tigerbeetle transfer timeouts
In the connector middleware, we pass in 5 seconds into the `createTransfer` method, expecting the argument takes in nanoseconds:

https://github.com/interledger/rafiki/blob/40091b649120fdcd327706a424c5c672c4c3674a/packages/backend/src/connector/core/middleware/balance.ts#L56

However, the Tigerbeetle service method was actually expecting `ms`, since it converts the argument to `ns`:

https://github.com/interledger/rafiki/blob/40091b649120fdcd327706a424c5c672c4c3674a/packages/backend/src/accounting/tigerbeetle/transfers.ts#L88

We should standardize the `timeout` argument, maybe using `timeoutMs` for the method signature.

AniqJaved:

- Timeout parameters should be named with the units associated with their particular context because we will be coverting them. In general timeouts are typically milliseconds but TigerBeetle uses nanoseconds. As we are using both of them, so the same name (e.g. timeout) it is creating an ambiguity.

## Context

- This fix allows us to convert the seconds into milliseconds instead of nanoseconds as the Tigerbeetle service method was actually expecting milliseconds to be converted into nanoseconds.

Fixes #1112 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/accounting/tigerbeetle/transfers.ts
+++ b/packages/backend/src/accounting/tigerbeetle/transfers.ts
@@ -23,7 +23,7 @@ export interface NewTransferOptions {
   destinationAccountId: AccountId
   amount: bigint
   ledger: number
-  timeout?: bigint
+  timeoutMs?: bigint
   postId?: never
   voidId?: never
 }
@@ -65,7 +65,7 @@ export async function createTransfers(
       user_data: 0n,
       reserved: 0n,
       pending_id: 0n,
-      timeout: 0n,
+      timeoutMs: 0n,
       ledger: 0,
       code: ACCOUNT_TYPE,
       flags: 0,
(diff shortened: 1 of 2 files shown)


sabineschaller:
Hi @AniqJaved ! Thank you for your contribution. Could you please add some context and the issue (see first item on the checklist) to your description?

AniqJaved:
Hello @sabineschaller , so I have made necessary changes in the description. And regarding the merge conflicts, so they are happening as I am changing the name of timeout to timeoutMs.

mkurapov (commented):
On the right path, there are a few spots that need to be updated as a result. 

It would be also to update the `TransferOptions` interface in the `packages/backend/src/accounting/service.ts` file, to use `timeoutMs` instead of `timeout` as well.

Make sure to merge in main as well, it has some changes

devcer:
Thank you for Contributing to Interledger during Hacktoberfest. If you have any accepted PRs, You can claim your Swag [here](https://community.interledger.org/devcer/hacktoberfest-swag-redemption-claim-your-cool-gear-2152). Last date to fill up the form is December 10th, 2023. @AniqJaved

PR #2135, opened 2023-10-26
fix(backend): prevent palindromic asset codes from causing test failure

### LINKED ISSUE #2112: Fix incoming payment test so that it doesn't fail when generated asset code is a palindrome
In the following test:
```
Incoming Payment Service 
› Create IncomingPayment 
› Cannot create incoming payment with different asset details than underlying wallet address
```

Where or not it passes is dependent on if the asset is a palindrome or not, as it generates "different asset details" by reversing the original asset code.

njlie:

- Updates the `Cannot create incoming payment with different asset details than underlying wallet address` test to shift the character codes of an asset code by one rather than reversing it.

## Context

Fixes #2112.

This test is intended to test a case where the asset code in an incoming payment creation call is discrepant with what is attached to the associated wallet address. However, because reversing the order was the strategy to create discrepancy, palindromic asset codes would produce the same asset code as the original.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/service.test.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.test.ts
@@ -125,14 +125,9 @@ describe('Incoming Payment Service', (): void => {
           incomingAmount: {
             value: BigInt(123),
             assetCode: String.fromCharCode(
-              ...asset.code.split('').map((letter) => {
-                let charCode = letter.charCodeAt(0)
-                if (charCode < 91) {
-                  return --charCode
-                } else {
-                  return ++charCode
-                }
-              })
+              ...asset.code
+                .split('')
+                .map((letter) => ((letter.charCodeAt(0) + 1 - 65) % 26) + 65)
             ),
             assetScale: asset.scale
           },
--- a/packages/backend/src/open_payments/payment/incoming/service.test.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.test.ts
@@ -127,7 +127,7 @@ describe('Incoming Payment Service', (): void => {
             assetCode: String.fromCharCode(
               ...asset.code.split('').map((letter) => {
                 let charCode = letter.charCodeAt(0)
-                if (charCode === 91) {
+                if (charCode < 91) {
                   return --charCode
                 } else {
                   return ++charCode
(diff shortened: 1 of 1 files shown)


packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -124,7 +124,16 @@ describe('Incoming Payment Service', (): void => {
           walletAddressId,
           incomingAmount: {
             value: BigInt(123),
-            assetCode: asset.code.split('').reverse().join(''),
+            assetCode: String.fromCharCode(
+              ...asset.code.split('').map((letter) => {
+                let charCode = letter.charCodeAt(0)
+                if (charCode === 91) {
```
BlairCurrey, line 130: I think this should be 90. I'm seeing 91 is `[` when I do `String.fromCharCode(91)` (so `ZZZ` becomes `[[[`). For whatever it's worth, it still works in that case but I think the intention is to keep it alphabetic.

Alternatively it looks like `assetCode: asset.code + 'x'` also works but I suppose its probably prudent to keep it a realistic assetCode

BlairCurrey (approved):
left some suggestions but they are non-blocking

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -124,7 +124,16 @@ describe('Incoming Payment Service', (): void => {
           walletAddressId,
           incomingAmount: {
             value: BigInt(123),
-            assetCode: asset.code.split('').reverse().join(''),
+            assetCode: String.fromCharCode(
+              ...asset.code.split('').map((letter) => {
+                let charCode = letter.charCodeAt(0)
+                if (charCode < 91) {
```
mkurapov, line 130: nice solution. I think we need to check the lower bound though (65), so we don't go under 'A'. (maybe we always decrease and then wrap around to 'Z' (90)?)

BlairCurrey: Feel free to completely ignore this, but I had fun thinking about Max's suggestion and came up with this.
 
```ts
assetCode: String.fromCharCode(
  ...asset.code
    .split('')
    .map((letter) => ((letter.charCodeAt(0) + 1 - 65) % 26) + 65)
),
```

njlie: I like it. I'll just go ahead and steal that...

BlairCurrey (approved):

PR #2137, opened 2023-10-26
feat(auth): add finalizationReason grant filter

### LINKED ISSUE #2125: Add `finalizationReason` to grants filter on Open Payments Auth Admin API
Add `finalizationReason` as a filter option in the `grants` query for the Auth Server Admin API.

BlairCurrey:
- adds `finalizationReason` to grant filter
- fixed bug where `finalizationReason` wasnt actually being returned in gql
- refactors the grant filter tests a bit to make the filter results more predictable (doesn't have to account for records inserted for other tests).

## Context
Tested manually in postman

fixes #2125 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -220,6 +221,96 @@ describe('Grant Resolvers', (): void => {
...
+              return query.data.grants
+            } else {
+              throw new Error('Data was empty')
+            }
+          })
+        expect(query.edges).toHaveLength(2)
+        expect(grants).toHaveLength(2)
+      })
+
+      test('finalizationReason: not in', async (): Promise<void> => {
```
mkurapov, line 270: we're expecting either or, correct? not both `in` and `not`?

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -220,6 +221,96 @@ describe('Grant Resolvers', (): void => {
         expect(query.edges).toHaveLength(0)
         expect(grants).toHaveLength(2)
       })
+
+      test('finalizationReason: in', async (): Promise<void> => {
+        const grants: GrantModel[] = []
+        for (let i = 0; i < 2; i++) {
```
mkurapov, line 227: maybe we do 3 so that we can have an odd one out with a different status, if that makes sense?

packages/auth/src/graphql/schema.graphql
```diff
@@ -46,6 +46,7 @@ type GrantEdge {
 input GrantFilter {
   identifier: FilterString
   state: FilterGrantState
+  finalizationReason: FilterFinalizationReason
```
mkurapov, line 49: I think we should also add this to the postman example to show that its there

BlairCurrey: I considered that but figured the default get request should be pretty inclusive and not only return some narrow subset of grants. IE, if me make a few grants and run Get Grants it should have results. 

I guess in that spirit I could do something like finalizationReason: notIn: [REVOKED]. A little non-sensical because state isnt filtered to FINALIZED but demonstrates the existence of the filter.

packages/auth/src/graphql/resolvers/grant.test.ts
BlairCurrey: that's correct. `finalizationReason: notIn` would be probably be better.

packages/auth/src/graphql/schema.graphql
BlairCurrey: I added added a finalizationReason filter so it's now:

```ts

    "filter": {
        "state": {
            "in": ["PROCESSING", "PENDING", "APPROVED", "FINALIZED"]
        },
        "finalizationReason": {
            "notIn": ["REVOKED"]
        }
    }
```

I think postman should document a realistic usage since this is an actual request. We do have docs which outlines all the filter options. A realistic request wouldn't Include all filter options because some conflict or are dependent in some way (`state`, `finalizationReason`). Exhausting all filter options would look something like this and seems more misleading than helpful. It also requires more work to undo the conflicting bits to get a working query when you want to use it


```ts
"filter": {
        "identifier": {
            "in": ["50c0b796-59fc-4878-a3d9-de6cf5485195"] 
        },
        "state": {
            "in": ["PROCESSING", "PENDING", "APPROVED", "FINALIZED"],
            "notIn": ["PROCESSING", "PENDING", "APPROVED", "FINALIZED"]
        },
        "finalizationReason": {
            "in": ["REVOKED", "ISSUED", "REJECTED"],
            "notIn": ["REVOKED", "ISSUED", "REJECTED"]
        }
    }
```

packages/auth/src/graphql/resolvers/grant.test.ts
BlairCurrey: I did that here as well as the other tests. Also verified the returned records' fields filtered by match the filter (or don't for `notIn`). I think these assertions are in a better place now. I also wonder about not testing each filter here since they are tested in services and just making sure the service `getPage` function is called correctly...

packages/auth/src/graphql/resolvers/grant.test.ts
BlairCurrey: changed the name here

mkurapov (approved):

packages/auth/src/graphql/schema.graphql
mkurapov: I think what you have is reasonable, I agree that listing out everything ends up being a lot

PR #2269, opened 2023-12-08
fix(auth): enable interaction on list-all grant requests by default

### LINKED ISSUE #2265: Enable more granular auth interaction
## Context

Right now, we have flags for grant interaction for incoming payments and quotes, which per default, are set to false. However, this means that clients can request `read-all` and `list-all` access for incoming payments without the user having to interact. Even worse, they could request that without an `identifier`, meaning they request to see all incoming payment data an ASE has for all of their users. While quotes cannot be listed, an adversary could still request a grant to `read-all` quotes and then guess quote ids to read someone else's quote.

## Todos
- [ ] require an `identifier` when requesting a grant including an `-all` action, such that a client can only see a specific user's transaction history
- [ ] turn on interaction for `-all` grants by default but include a flag to allow ASEs to turn it off

sabineschaller:

- add environment variable to specify whether interaction is required when requesting a grant with a `list-all` interaction (default: true)
- check whether `list-all` action is in a grant request and if so, require interaction
- check whether when a grant that requires interaction is requested, the `identifier` is present

## Context

- fixes #2265 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/auth/src/grant/routes.ts
```diff
@@ -137,7 +138,18 @@ async function createPendingGrant(
...
+  }
+
+  if (
+    body.access_token.access.every(
+      (item) =>
+        (item.actions.includes(AccessAction.ReadAll) ||
+          item.actions.includes(AccessAction.ReadAll)) &&
+        item.identifier !== ''
+    )
+  ) {
```
mkurapov, line 151: Can we simplify this to:
```suggestion
  if (body.access_token.access.find((item) => !item.identifier)) {
```

If this:
> check whether when a grant that requires interaction is requested, the identifier is present

is what we want?

mkurapov (commented):
I believe this would break the Rafiki to Rafiki grant requests (to get the remote incoming payments), since we are expecting a non-interactive grant in the response

packages/auth/src/grant/routes.test.ts
```diff
@@ -119,24 +119,35 @@ describe('Grant Routes', (): void => {
...
-          interactionFlagsEnabled | description
-          ${true}                 | ${'enabled'}
-          ${false}                | ${'disabled'}
+          resourceInteractionFlagsEnabled | allGrantInteractionFlagEnabled | allGrant | description
+          ${true}                         | ${false}                       | ${false} | ${'no -all grant'}
+          ${false}                        | ${false}                       | ${false} | ${'no -all grant'}
+          ${true}                         | ${false}                       | ${true}  | ${'including an -all grant'}
+          ${false}                        | ${false}                       | ${true}  | ${'including an -all grant'}
+          ${true}                         | ${true}                        | ${true}  | ${'including an -all grant'}
+          ${false}                        | ${true}                        | ${true}  | ${'including an -all grant'}
```
mkurapov, line 128: I _believe_ we are missing 2 cases here to capture all of the possibilities (8)?

packages/auth/src/grant/routes.ts
sabineschaller: I don't think that is what we want. This condition would also fire on accesses that don't need an identifier. But I want to make sure that the identifier is within the access that triggered the interaction. (This is so hard to explain...)

packages/auth/src/grant/routes.test.ts
sabineschaller: There are but having the flag enabled but no list-all grant isn't really worth testing, I think.

packages/auth/src/grant/routes.ts
```diff
@@ -96,7 +96,13 @@ async function createGrant(
   deps: ServiceDependencies,
   ctx: CreateContext
 ): Promise<void> {
-  if (canSkipInteraction(deps.config, ctx.request.body)) {
+  let noInteractionRequired: boolean
+  try {
+    noInteractionRequired = canSkipInteraction(deps.config, ctx.request.body)
+  } catch (err) {
+    ctx.throw(400, 'identifier_required', { error: 'identifier_required' })
```
njlie, line 103: Note to self, or some other enterprising person: we should make it a point to change all the calls to `ctx.throw` so that they also use this pattern.

njlie (approved):

PR #2285, opened 2023-12-11
feat(auth): add continuation polling

### LINKED ISSUE #2193: Add polling to Rafiki Auth Server continuation endpoint
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#name-continuing-during-pending-i

The Rafiki Auth Server should support polling the grant continuation endpoint to determine if interaction is finished, to better support clients without browsers. We should also add and implement the `wait` field from the grant initiation response's `continue` field: https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#response-continue

njlie:

- Adds polling to authorization server
- Enforces the `wait` value in grants during grant continuation

## Context

Closes #2193.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/routes.ts
```diff
@@ -206,13 +266,20 @@ async function continueGrant(
     interactionService
   } = deps
 
-  const interaction = await interactionService.getByRef(interactRef)
+  if (!ctx.request.body || Object.keys(ctx.request.body).length === 0) {
+    await pollGrantContinuation(deps, ctx, continueId, continueToken)
+    return
+  }
+
+  const interaction = await interactionService.getByRef(interactRef as string)
```
BlairCurrey, line 274: Looks like `interactRef` could be `undefined` since its now optional. Could check it before line 274 instead of the string type cast to preserve the same error response as before for a required and missing `interactRef`, as well as making it more typesafe.

```ts
  if (!interactRef) {
    ctx.throw(401, { error: 'invalid_request' })
  }
 ```
 
Might also be able to keep [this 'Cannot issue access token without continue token' test](https://github.com/interledger/rafiki/pull/2285/files#diff-a9f1841aeba665aeaf6b6a3e2065cc618245ab8c7a5124641ecd655b24a4cd0dR508) and adapt to this case.  `interactRef required if no continueId`, or similar.

**Comment by njlie** [comment]
Editor's note: I need to update how `wait` is handled during polling so that it is treated like an interval between attempts, rather than just being anchored to the creation time.

File: packages/auth/src/grant/routes.ts [hunk 2]
```diff
@@ -179,6 +180,65 @@ function isContinuableGrant(grant: Grant): boolean {
...
+): Promise<void> {
+  const { config, grantService, accessService, accessTokenService } = deps
+
+  const grant = await grantService.getByContinue(continueId, continueToken)
+  if (!grant) {
+    ctx.throw(404, { error: 'unknown_request' })
+  }
+
+  if (isGrantStillWaiting(grant)) {
+    ctx.throw(401, { error: 'too_fast' })
```
sabineschaller, line 205: Can we be more descriptive in those error messages?

packages/auth/src/grant/model.ts
```diff
@@ -120,6 +121,20 @@ interface ToOpenPaymentsGrantArgs {
...
+  args: ToOpenPaymentsGrantArgs
+): OpenPaymentsGrantContinuation {
+  return {
+    continue: {
+      access_token: {
+        value: grant.continueToken
+      },
+      uri: `${args.authServerUrl}/continue/${grant.continueId}`
+    }
+  }
```
mkurapov, line 135: do we need to return `wait` here as well?

packages/auth/src/grant/routes.ts
mkurapov: I think @njlie has been following https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#response-error

I think we should keep these errors by putting them in the `code` field (as defined in the spec), while adding the `description` field

packages/auth/src/grant/routes.ts
```diff
@@ -179,6 +180,65 @@ function isContinuableGrant(grant: Grant): boolean {
...
+    grant.state === GrantState.Processing
+  ) {
+    ctx.body = toOpenPaymentsGrantContinuation(grant, {
+      authServerUrl: config.authServerDomain
+    })
+    return
+  } else if (
+    grant.state !== GrantState.Approved ||
+    !isContinuableGrant(grant)
+  ) {
```
mkurapov, line 225: If it's already a finalized grant, we should error always during continuation requests, correct? Just checking

packages/auth/src/grant/routes.ts
```diff
@@ -206,13 +266,20 @@ async function continueGrant(
...
+
+  const interaction = await interactionService.getByRef(interactRef as string)
   if (
     !interaction ||
     !isContinuableGrant(interaction.grant) ||
     !isMatchingContinueRequest(continueId, continueToken, interaction.grant)
   ) {
     ctx.throw(404, { error: 'unknown_request' })
+  } else if (isGrantStillWaiting(interaction.grant)) {
+    ctx.throw(401, { error: 'too_fast' })
```
mkurapov, line 282: I see this [here](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#name-continuing-a-grant-request):
> If the client instance does not respect the given wait period, the AS MUST return the too_fast error ([Section 3.6](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#response-error)).

More of a question about the spec, but is it an actual issue if the continuation request (somehow) is made before the `wait` time with the correct `interact_ref`?

packages/auth/src/client/service.test.ts
```diff
@@ -38,7 +38,8 @@ describe('Client Service', (): void => {
           publicName: TEST_CLIENT_DISPLAY.name,
           assetCode: 'USD',
           assetScale: 9,
-          authServer: faker.internet.url({ appendSlash: false })
+          authServer: faker.internet.url({ appendSlash: false }),
+          resourceServer: faker.internet.url({ appendSlash: false })
```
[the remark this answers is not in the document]
sabineschaller: We should but in a follow up PR, I think.

packages/auth/src/grant/routes.ts
njlie: I interpreted this as meaning that the wait time should be respected irregardless if the request is made with the correct `interact_ref`.

packages/auth/src/client/service.test.ts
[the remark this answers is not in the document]
njlie: In today's team call @sabineschaller mentioned taking time to update Rafiki to match the latest OP spec, so I'll leave this comment out of scope for this PR.

packages/auth/src/grant/routes.ts
mkurapov: Makes sense, I doubt that is something that would realistically happen anyway

packages/auth/src/grant/routes.ts
njlie: Yeah, that's correct. Finalized grants cannot be continued. 

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-15#section-1.5
> Once in this state, no new access tokens can be issued, no subject information can be returned, and no interactions can take place. Once in this state, the grant request is dead and cannot be revived. If future access is desired by the client instance, a new grant request can be created, unrelated to this grant request.

packages/auth/package.json
```diff
@@ -26,7 +26,7 @@
     "@graphql-tools/load": "^8.0.1",
     "@graphql-tools/schema": "^10.0.2",
     "@interledger/http-signature-utils": "2.0.0",
-    "@interledger/open-payments": "6.0.0",
+    "@interledger/open-payments": "6.5.0",
```
mkurapov, line 29: Now that we are bumping to 6.5.0, can you please add `useHttp: process.env.NODE_ENV === 'development'` when we instantiate the Open Payments client please? 

change in question: https://github.com/interledger/open-payments/releases/tag/%40interledger%2Fopen-payments%406.1.0

mkurapov: This will also make the Examples > Open Payments Postman folder work

packages/auth/src/grant/middleware.ts
```diff
@@ -0,0 +1,21 @@
+import { ContinueContext } from './routes'
+
+export async function grantLastContinueAttemptMiddleware(
```
mkurapov, line 3: Should this always be called, regardless of result of `grantRoutes.continue` function? Can you get stuck in a situation where you keep trying to continue, extending `lastContinuedAt` and never being able to successfully continue it?

packages/auth/package.json
mkurapov: After updating this and trying locally, the Grant Requests for Incoming Payment and Quotes work, but the Outgoing payment grant request does not, will need to keep digging

packages/auth/src/grant/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+  ctx: ContinueContext,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  next: () => Promise<any>
+): Promise<void> {
+  await next()
```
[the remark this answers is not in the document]
njlie: Yeah, that's the intent here.

packages/auth/src/grant/routes.ts
```diff
@@ -185,6 +186,80 @@ function isContinuableGrant(grant: Grant): boolean {
...
+    ctx.throw(404, {
+      error: {
+        code: 'unknown_request',
+        description: 'grant not found'
+      }
+    })
+  }
+
+  if (isGrantStillWaiting(grant)) {
+    ctx.throw(401, {
```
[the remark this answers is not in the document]
njlie: In hindsight, 400 does seem more appropriate.

packages/auth/package.json
njlie: That last part should work after https://github.com/interledger/rafiki/pull/2337 is in

packages/auth/src/grant/middleware.ts
njlie: I took a closer look at the spec, and I think I now understand that a `wait` time should be anchored to the last time the "continue" response is given, so I'll change it to reflect that instead.

packages/auth/src/grant/routes.ts
```diff
@@ -212,17 +295,46 @@ async function continueGrant(
...
+
   const interaction = await interactionService.getByRef(interactRef)
   if (
     !interaction ||
     !isContinuableGrant(interaction.grant) ||
     !isMatchingContinueRequest(continueId, continueToken, interaction.grant)
   ) {
-    ctx.throw(404, { error: 'unknown_request' })
+    ctx.throw(404, {
+      error: { code: 'unknown_request', description: 'grant not found' }
```
mkurapov, line 320: Should this also be `invalid_continuation`? or maybe `invalid_request` (to pick something from the list of codes?

packages/auth/src/grant/routes.ts
```diff
@@ -198,10 +277,14 @@ async function continueGrant(
...
     'GNAP '
   )[1]
-  const { interact_ref: interactRef } = ctx.request.body
 
-  if (!continueId || !continueToken || !interactRef) {
-    ctx.throw(401, { error: 'invalid_request' })
+  if (!continueId || !continueToken) {
+    ctx.throw(401, {
+      error: {
+        code: 'invalid_request',
```
mkurapov, line 284: should this be `invalid_continuation`? 


> The AS MUST be able to tell from the client instance's request which specific ongoing request is being accessed, using a combination of the continuation URI and the continuation access token. If the AS cannot determine a single active grant request to map the continuation request to, the AS MUST return an invalid_continuation error ([Section 3.6](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#response-error)).


[Link](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-1.5)

mkurapov (commented):
Just comments about the code responses, but logic LGTM :+1:

packages/auth/src/grant/routes.ts
njlie: Will use `invalid_continuation`, I feel it fits best given the text quoted in https://github.com/interledger/rafiki/pull/2285#discussion_r1452857289

mkurapov (approved):
looks good, should we wait for https://github.com/interledger/rafiki/pull/2337?

sabineschaller:
> looks good, should we wait for https://github.com/interledger/rafiki/pull/2337?

#2337 should be good to go now. Awaiting final approval.

sabineschaller (approved):

packages/auth/src/grant/routes.ts
```diff
@@ -198,10 +277,14 @@ async function continueGrant(
...
     'GNAP '
   )[1]
-  const { interact_ref: interactRef } = ctx.request.body
 
-  if (!continueId || !continueToken || !interactRef) {
-    ctx.throw(401, { error: 'invalid_request' })
+  if (!continueId || !continueToken) {
+    ctx.throw(401, {
+      error: {
+        code: 'invalid_continuation',
```
mkurapov, line 284: maybe we can provide these error codes as the second argument to `ctx.throw` such that we get them served up to us in the response: https://github.com/koajs/koa/blob/master/docs/api/context.md#ctxthrowstatus-msg-properties

njlie:
> Were you able to test this via postman requests after merging in local (and making [this](https://github.com/interledger/rafiki/pull/2285#discussion_r1463919477) change)? I'm unable to call Continue grant (after not providing finish method in the original grant request + not passing in `interact_ref`)

I'll look into it before merging in.

njlie:
For some reason, adding [this snippet](https://github.com/interledger/rafiki/pull/2285/files#diff-a9f1841aeba665aeaf6b6a3e2065cc618245ab8c7a5124641ecd655b24a4cd0dR738-R742) causes the test to fail with this error:
```
  ● Test suite failed to run

    DataCloneError: Symbol(jest.asymmetricMatcher) could not be cloned.
```

Nobody seems to have this problem on Google/Stack Overflow as far as I've searched, and it seems pretty arbitrary since the other uses of `jest.useFakeTimers` don't cause this to happen...

njlie:
Re-requesting reviews as I made some pretty big changes to how finishing interaction is handled while addressing some issues that were highlighted in https://github.com/interledger/rafiki/pull/2285#issuecomment-1906852181.

packages/auth/src/interaction/routes.test.ts
```diff
@@ -256,142 +256,266 @@ describe('Interaction Routes', (): void => {
         })
       })
 
-      test('Can finish accepted interaction', async (): Promise<void> => {
-        const grant = await Grant.query().insert({
-          ...generateBaseGrant(),
-          state: GrantState.Approved
-        })
+      describe('Interactions for grant with finish method', (): void => {
```
njlie, line 259: This whole describe block contains tests that existed previously, and were just moved inside this. There are no other changes.

mkurapov (commented):
Happy flow worked for me in Postman ✅ 

I noticed though that in the MASE I believe we need to call the `finish` interaction for Rejected cases at some point, to have the grant marked as Finalized -> Rejected in the DB

mkurapov (approved):
Tested the rejected flow ✅

Issue #1956, opened 2023-09-22
[Telemetry] Set up AWS Prometheus & Grafana


beniaminmunteanu:
## Context
Rafiki's Telemetry Feature now exposes metrics on the OpenTelemetry Standard, through an OpenTelemetry Collector Instance.
We should be able to gather metrics from all Rafiki instances, analyse and visualize them.
## Todos

- [x] Set up Managed Prometheus
- [x] Set up Managed Grafana with Prometheus as a Data source

- [x] Create an IAM user for OpenTelemetryCollector
- [x] Get access keys for the Opentelemetry collector user 
- [x] Use Access Keys to authenticate opentelemetry collectors
- [x] Create Role with write permissions to prometheus
- [x] Have collector IAM users use "assume role" in order to get write permissions to prometheus

Issue #1923, opened 2023-09-19
[Telemetry] - Set up Grafana to view metrics


JoblersTune:
Visualise the collected Rafiki metrics using Grafana.
Be able to view number of transactions, total transaction amount, and average transaction amount per day / week / month. 
Possibly embed this on a frontend (location TBD)

Issue #1921, opened 2023-09-19
Telemetry


JoblersTune:
Aim

- Provide telemetry for the network
- Enabled by default
- Note this enabled by default

Desired metrics

- Total money [per time frame] taken from packet data
- Average money [per time frame] taken from packet data
- Amount of transactions taken from outgoing payments that have been at least partially successful (i.e. some value in the send amount)

Basic setup

- Rafiki will include the OpenTelemetry SDK to generate metrics, and contextual information [[rafiki repo]](https://github.com/interledger/rafiki)
- Using the sidecar pattern with an OpenTelemetry Collector to gather metrics
- Will need the option to enable/disable metric collection
- This functionality will be included within the Rafiki repo
- Using Prometheus for storing, and querying metrics (using the counter primitive)
- Using Grafana for creating dashboards and visualizing metrics
- Running Prometheus and Grafana in containers managed on AWS

Integration

- [Rafiki](https://github.com/interledger/rafiki) code base needs to expose the desired metrics as HTTP endpoints
- Configure the OpenTelemetry Collector to gather the metrics from Rafiki
- Configure Prometheus to scrape metrics from the OpenTelemetry Collector (uses a pull-based approach)
- In Grafana, you'll configure Prometheus as a data source (can use PromQL to write queries against the data collected by Prometheus.)

JoblersTune:
AWS access key and secret for the open collector user are stored in 1Password

PR #2299, opened 2023-12-23
feat(telemetry): instrument rafiki

### LINKED ISSUE #1913: [Telemetry] - Instrument Rafiki
## Context
In order to add telemetry features for Rafiki, We need to instrument rafiki to collect and expose metrics according to the Opentelemetry standard.

## Tech:
We should creating a pipeline for collecting, analysing, and visualizing metrics in regards to **transactions** by using the respective set of technologies : Opentelemetry, Prometheus, Grafana.

- We will be opting for a sidecar pattern, and have rafiki push metrics to a new (configurable) dependency:
[The Opentelemetry Collector.](https://hub.docker.com/r/otel/opentelemetry-collector)

- The Otel collector will push metrics to an AWS Prometheus

- Grafana should have Prometheus as a datasource and provide dashboards

(uses instances created in) #1906 

As such, we will first attempt to visualize the following data (aggregated in daily, weekly & monthly):

- How many transactions were in the ILP network 
- What is the average amount in transactions
- What is the total transactions amount

## Todos
- [x] Add the Opentelemetry Collector as a configurable dependency
- [x] Add the Opentelemetry SDK
- [x] Collect transaction metrics
- [x] Periodically send metrics to Opentelemetry Collector
- [x] Authenticate Opentelemetry Collector in AWS
- [x] Send metrics from opentelemetry collector to AWS Prometheus
- [x] Send the transaction amount in USD to to Opentelemetry Collector and AWS Prometheus

## Linked Issues:

- #1925

beniaminmunteanu:

This PR introduces telemetry into Rafiki to observe and measure the growth and activities of the ILP Network. This will provide valuable insights for future improvements and developments. The implementation is built from scratch and includes the following key features:

- Rafiki instances are instrumented with Opentelemetry to adhere to a standard format and allow integrating account servicing entities to build their own telemetry.
- Instrumented Rafiki communicates with an Opentelemetry collector through the otel protocol (in grpc).
- Rafiki can send metrics to multiple Opentelemetry collectors, defined in the .env file.
- Telemetry is entirely optional and controlled through an env variable.
- Two counter metrics are introduced: one for counting the number of transactions and another for counting amounts being sent. These metrics are used to answer three key questions about the network: What is the average amount that a transaction holds? What is the total number of transactions? How much money is being sent through the network? The solution provides answers for these questions for the past day, week, and month.
- A centralized Telemetry service is introduced to handle telemetry. The telemetry rates service uses the public S3 URL that the AWS Lambda function dumps data into.
- Currency conversions are performed on transactions to standardize data in telemetry.
- An AWS Lambda function is deployed to retrieve daily exchange rates from the Coingate API, transform the data into the Rafiki Rates service format, and save it to a public S3 in AWS.
- The telemetry data is sent to Amazon Managed Service for Prometheus (AMP), and then used in Grafana Cloud to build our "Network insights" dashboards, which are already set up and ready to use.
- Privacy considerations have been a major focus in this PR, with a full documentation on our privacy solution included as an .md file in this PR, which describes how we implement Local Differential Privacy (LDP).

## Context

The goal of this PR is to provide a foundation for observing the ILP Network at the money transfer level. By instrumenting Rafiki with Opentelemetry, we can collect data that will help us understand the network's growth and make informed decisions about future developments.

The telemetry data is sent to an Opentelemetry collector in AWS, but the system is designed to allow integrating ASEs to have their own collectors and build their own telemetry solutions. The data is then used in Grafana Cloud to build our "Network insights" dashboards.

This PR also introduces a new Telemetry service in Rafiki and a new RatesService within it. The RatesService reuses the existing Rafiki Rates service but queries a different URL for rates and has a different cache TTL setup.

Infrastructure was a major point of this feature as it was non-existent beforehand. We are using AWS ECS Fargate, with a cluster of custom ADOT (AWS Distro for OpenTelemetry) collectors. There is also a Network Load Balancer (NLB) in AWS that sits in front of the cluster and load balances the multiple ECS otel collector tasks. All these otel collectors then send data further downstream to Prometheus. The infrastructure details will be documented separately to keep this PR focused on the Rafiki changes.

closes #1913 #1923 #1921 #1956

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/index.ts
```diff
@@ -347,10 +355,26 @@ export function initIocContainer(
       peerService: await deps.use('peerService'),
       ratesService: await deps.use('ratesService'),
       streamServer: await deps.use('streamServer'),
-      ilpAddress: config.ilpAddress
+      ilpAddress: config.ilpAddress,
+      telemetry: config.enableTelemetry
+        ? await deps.use('telemetry')
+        : undefined
```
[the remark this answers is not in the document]
beniaminmunteanu: This was indeed in the accounting service, using 'onDebit', but it moved to connector so that we can collect at packet level. 
I see your scalability concern. Is there another way to collect at packet level?

packages/backend/src/payment-method/ilp/connector/core/middleware/telemetry.ts
```diff
@@ -0,0 +1,69 @@
...
+      Number(request.prepare.amount) &&
+      !state.unfulfillable
+    ) {
+      const senderAsset = accounts.outgoing.asset
+      const convertOptions = {
+        sourceAmount: BigInt(request.prepare.amount),
+        sourceAsset: { code: senderAsset.code, scale: senderAsset.scale },
+        destinationAsset: {
+          code: services.telemetry.getBaseAssetCode(),
+          scale: 4
```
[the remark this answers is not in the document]
beniaminmunteanu: Looking at the metrics we collect, and at the fact that accuracy is not really conserved in telemetry due to privacy concerns, i would argue that scale 4 is more precision than what is actually needed.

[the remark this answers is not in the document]
beniaminmunteanu: I will however also make the asset scale configurable

localenv/collector/otel-collector-config.yaml
```diff
@@ -0,0 +1,42 @@
...
+    protocols:
+      grpc:
+      http:
+        cors:
+          allowed_origins:
+            - http://*
+            - https://*
+
+processors:
+  batch:
```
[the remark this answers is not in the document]
beniaminmunteanu: yes, it is using the defaults.

packages/backend/src/index.ts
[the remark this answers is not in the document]
beniaminmunteanu: Having another look into this.
We need to:

- collect at packet level,
-  keep both the count and the amount metrics on the same side of a transaction, to avoid discrepancies.

By the current implementation, they are both on the Sending side. I belive the `onDebit` method in the accounting service is on the receiving side of a transaction.

I will explore moving both back to the receiving side. I remember having some trouble with that implementation though, in the local playground both happy-life and cloud-nine fired the metric instead of just one, and the amount ended up doubling if i recall correctly. Does it make sense for that to happen?

Will give it another shot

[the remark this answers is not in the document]
JoblersTune: I believe we can use the accounting service and still achieve gathering metrics on the sending side. I'm discussing this with Beni in Slack.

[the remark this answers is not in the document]
JoblersTune: In `createAccountToAccountTransfer` in the `packages/backend/src/accounting/service.ts` file there is a `post` method and inside that there we can add a `if (destinationAccount.onDebit)` conditional. If we set our metric to only be collected on when the destination account is being debited then then I believe we will be collecting the accounting, packet level, sending side only, meeting all our objectives.
Like so:
```
post: async (): Promise<void | TransferError> => {
      const error = await postTransfers(pendingTransferIdsOrError)

      if (error) {
        return error
      }

      await Promise.all([
        onDebit(sourceAccount),
        onDebit(destinationAccount.asset)
      ])

      if (destinationAccount.onCredit) {
        const totalReceived = await getAccountReceived(destinationAccount.id)

        if (totalReceived === undefined) {
          throw new Error()
        }

        await destinationAccount.onCredit({
          totalReceived,
          withdrawalThrottleDelay
        })
      }
      
      if (destinationAccount.onDebit) {
        // collect metrics here on destinationAmount
      }
    },
```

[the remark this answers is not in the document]
mkurapov: @JoblersTune is correct here. Some context: this works because in cases where we are doing sending over Interledger, the destination account will be a `Peer`, that _does_ have an `onDebit` method. See [here](https://rafiki.dev/concepts/accounting/accounts-and-transfers/#interledger).

If you want to be more explicit, you could try using `if (destinationAccount instanceof Peer)` to get the same effect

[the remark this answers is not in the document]
beniaminmunteanu: I already started moving the codebase back to the accounting service, and using the suggested conditionals to filter sending transactions.

@mkurapov  just to clarify, so if the payment is done within the same rafiki instance, there would be no onDebit method on the destinationAccount? 

In a broader view, I do not think local payments are necessarily the best indicator of ILP network growth. 
It could however be a factor in having a feel about the activity within the network. How many users there are, or at least how active the served users are from a transaction standpoint. That would be the argument I see in supporting those local payments.

That being said, not having that condition means that we need another way to differentiate between sending and receiving transactions, so we react just on one side (suggesting sending side).  

If we don't, we could have another approach:
Having an additional buffer for metrics before they are sent, that is also able to de-duplicate amount metrics that come up both from the sending and from the receiving side of a transaction.
This is however might have diminishing returns in terms of complexity vs achievement.

Thanks for your insights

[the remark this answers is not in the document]
beniaminmunteanu: update: latest commit moved everything back to the accounting service.
Please have a look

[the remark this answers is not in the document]
mkurapov: @beniaminmunteanu 
> @mkurapov just to clarify, so if the payment is done within the same rafiki instance, there would be no onDebit method on the destinationAccount and the telemetry instrumentation code would not execute?

That's right, since the source/debit account is the outgoing payment and the destination/credit account is the incoming payment that only has the onCredit method, and not the onDebit (since it doesn't make too much sense to ever debit an incoming payment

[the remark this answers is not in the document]
mkurapov: @beniaminmunteanu 

> I do not think local payments are necessarily the best indicator of ILP network growth.
I agree with you, if we wanted to track that, I think that should belong in its own metric.

packages/backend/src/telemetry/privacy.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  let originalModule: typeof privacy
+
+  beforeEach(() => {
+    originalModule = { ...privacy }
+
+    jest.mock('./privacy', () => ({
+      ...originalModule,
+      applyPrivacy: jest.fn()
+    }))
```
mkurapov, line 16: why do we need this mock here?

packages/backend/src/accounting/service.ts
```diff
@@ -183,6 +195,21 @@ export async function createAccountToAccountTransfer(
...
         })
       }
+      if (
+        destinationAccount.onDebit &&
+        telemetry &&
+        aseRatesService &&
+        sourceAccount.asset.code &&
+        sourceAccount.asset.scale
+      ) {
+        collectTelemetryAmount(telemetry, aseRatesService, {
```
mkurapov, line 205: should this be awaited?

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (converted === ConvertError.InvalidDestinationPrice) {
+    return
+  }
+
+  telemetryService
+    .getOrCreate('transactions_amount', {
+      description: 'Amount sent through the network',
+      valueType: ValueType.DOUBLE
+    })
+    .add(privacy.applyPrivacy(Number(converted)))
```
mkurapov, line 38: can `.add` throw? we should make sure to handle this to not break the main process. It's ok if telemetry doesn't end up sending a metric, but wouldn't be nice obviously if it prevents a trx from completing

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export async function convertAmount(
+  aseRates: RatesService,
+  telemetryRates: RatesService,
+  convertOptions: Omit<ConvertOptions, 'exchangeRate'>
+) {
+  try {
+    const aseConvert = await aseRates.convert(convertOptions)
+    return aseConvert
+  } catch (error) {
```
mkurapov, line 49: we should log in here

packages/backend/src/telemetry/mocks.ts
```diff
@@ -0,0 +1,56 @@
+import { Attributes, Counter, MetricOptions } from '@opentelemetry/api'
```
mkurapov, line 1: we should put this file in the `tests` folder IMO

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -81,6 +81,14 @@ export async function handleSending(
     finalReceiveAmount: maxReceiveAmount
   })
 
+  deps.telemetry
+    ?.getOrCreate('transactions_total', {
+      description: 'Count of funded transactions'
+    })
+    .add(1, {
+      source: deps.telemetry.getServiceName()
```
mkurapov, line 89: The setting of `source` (and probably calling of the `add`) should be encapsulated in the telemetryService, the services that use this as a dependency shouldn't need to worry about setting `source: deps.telemetry.getServiceName()` every time

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  const convertOptions = {
+    sourceAmount: amount,
+    sourceAsset: { code: asset.code, scale: asset.scale },
+    destinationAsset: {
+      code: telemetryService.getBaseAssetCode(),
+      scale: telemetryService.getBaseScale()
+    }
+  }
+  const converted = await convertAmount(
```
mkurapov, line 24: My first thought seeing this was that the conversion should possibly live in the telemetry service itself so that it doesn't need to expose its own rates service outside of the telemetry class. Thoughts?

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,53 @@
...
+    aseRates,
+    telemetryService.getRatesService(),
+    convertOptions
+  )
+  if (converted === ConvertError.InvalidDestinationPrice) {
```
mkurapov, line 29: log here maybe as well

mkurapov (commented):
Looking solid, thank you for adding those docs!

packages/backend/src/telemetry/transaction-amount.ts
beniaminmunteanu: I think that was my initial solution as well. I think i moved it out in the previous implementation on the connector side for proper unit testing (which i had trouble with there)

Giving this some more thought, we could also have another approach:
How about the existing ratesService has an optional fallback retrieval url?
since telemetry tries the ASE rates first as well,and only then tries out the external rates from coingate.

so then telemetry would not have it's own rates service.
When we need just ASE conversion, we use the rates service as usual.
When we need fallback to external rates, we just use the rates service with the flag set.

This way, telemetry itself would not have the concern of rates conversion at all, it would just use it.

Which solution would you opt for?

packages/backend/src/accounting/psql/service.ts
```diff
@@ -33,8 +33,12 @@ import {
   voidTransfers
 } from './ledger-transfer'
 import { LedgerTransfer, LedgerTransferType } from './ledger-transfer/model'
+import { TelemetryService } from '../../telemetry/service'
+import { RatesService } from '../../rates/service'
 
 export interface ServiceDependencies extends BaseService {
+  telemetry?: TelemetryService
+  aseRatesService?: RatesService
```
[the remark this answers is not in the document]
beniaminmunteanu: That makes sense.

I'll have another look at the optionality of the services, but i'm pretty sure they ended up that way because telemetry is optional.
I did the optionality at the first possible injection layer, such that if enableTelemetry is not true, then the telemetry singleton service is not ever instantiated. (and as such would be undefined => ? on most uses.)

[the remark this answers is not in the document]
mkurapov: > but i'm pretty sure they ended up that way because telemetry is optional.
Right, that makes sense. This makes me think whether we should support optional dependencies though. Might lead to some unexpected things. An approach could be just creating a noop telemetry service (with the same interface, but maybe just does console logs instead) and injecting that instead, when telemetry is disabled.

packages/backend/src/telemetry/transaction-amount.ts
mkurapov: @sabineschaller do you see ASEs being ok with falling back to our (external) rates in case something goes wrong on their rates API?

packages/backend/src/telemetry/transaction-amount.ts
beniaminmunteanu: I'm waiting on the previous comment resolution, so i know where to inject the app context logger for it

packages/backend/src/accounting/service.ts
mkurapov: Or I'm guessing you don't need to await since we don't use the response of this function? 
It makes sense, I'm just always wary of not `await`ing an async fn mostly because of error handling - I don't think an error can be caught from this method - (I think) it would always throw an unhandled promise rejection.

packages/backend/src/telemetry/transaction-amount.ts
beniaminmunteanu: > but I'm not entirely sure if an ASE would be ok with using some external rates
We would not be doing that in ASE direct use cases though. I see it as optional, and It would just be used in telemetry now, i would not enforce the fallback. It's just used when calling the `convert` method of the `ratesService` with an optional 'fallback' param

sabineschaller: > @sabineschaller do you see ASEs being ok with falling back to our (external) rates in case something goes wrong on their rates API?

I don't think so. That could put a big amount of risk on them. Technically, the fallback should only be used for telemetry but I don't think we can 100% guarantee that. Hence, they wouldn't want to add a fallback that could put them out of business.

mkurapov: @beniaminmunteanu I understand this now:
>  I see it as optional, and It would just be used in telemetry now, i would not enforce the fallback.

@sabineschaller, why do you say 
> I don't think we can 100% guarantee that?

mkurapov: I think for now let's just have two separate rates services like we have now and use them only in telemetry service (ie. do the conversion in telemetry service itself)

sabineschaller: > why do you say "I don't think we can 100% guarantee that"?

Because if it's a fallback, the connector could also fallback to it if the ASE's rates service is not available, couldn't it?

mkurapov: > the connector could also fallback to it if the ASE's rates service is not available, couldn't it?

Yes, if we just have one rates service. I think Beni was referring to having an optional fallback URL as a parameter for the convert method, and we'd only have it set when doing the conversion the telemetry service

sabineschaller: Gotcha. I thought it would be a general fallback. But I'm sure I didn't read carefully before 😉

Thanks for clarifying!

mkurapov: You still have a good point though about being careful. I think this is why I still prefer separate rate [services](https://github.com/interledger/rafiki/pull/2299#discussion_r1463798330)

beniaminmunteanu: Separate it is then! 👌

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
beniaminmunteanu: >Might lead to some unexpected things

anything in particular?

I'm leaning a bit more into the optionality camp for now,
but I'm happy to change it and use a noop service if you think it's better.

packages/documentation/src/content/docs/telemetry/privacy.md
```diff
@@ -0,0 +1,57 @@
...
+
+The sensitivity of a function in differential privacy is the maximum amount that any single observation can change the output of the function. In this case, the sensitivity is considered to be the maximum of the rounded value and the bucket size.
+
+The privacy parameter is computed as one-tenth of the sensitivity. This parameter controls the trade-off between privacy and utility: a smaller privacy parameter means more privacy but less utility, and a larger privacy parameter means less privacy but more utility.
+
+The noise, selected from the Laplacian distribution, is then generated using this privacy parameter and added to the rounded value. If the resulting value is zero, it is set to half the bucket size to ensure that the noise does not completely obscure the transaction value.
+
+## Currency Conversion
+
+Another factor that obscures sensitive data is currency conversion. In cross-currency transactions, we use exchange rates that are not traced back. This introduces an additional layer of noise and further protects the privacy of the transactions.
```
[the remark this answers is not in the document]
beniaminmunteanu: Please have a look at the additional context provided there.

packages/backend/src/accounting/service.ts
beniaminmunteanu: not awaiting is a choice, i'd rather have the accounting execution keep flowing without it bothering about telemetry. 
It gets executed in the background, as telemetry should be.

in terms of error handling, you are correct.
The function itself will just catch whatever goes wrong inside it, and just log.
Does that sound ok to you?

packages/backend/src/telemetry/privacy.test.ts
beniaminmunteanu: that is mocking the module, but keeping it's implementation for all other methods except for applyPrivacy.
That is so that i can properly mock return values.

beniaminmunteanu: leftover. thx

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
beniaminmunteanu: On another look at this, the source is not something all metrics send.
The object that gets sent as the second parameter when calling `add` represents the additional properties that will be attached to the metric.
I think this should be controlled externally, as it is now, by the place that creates the metric through the `getOrCreate`method.

It is just the case on this particular metric that it sends the 'source' which happens to be a property of the telemetry service itself through DI, but the additional properties object could hold anything else.

packages/backend/src/telemetry/transaction-amount.ts
beniaminmunteanu: moved the covnersion inside of telemetry service.
It now logs using telemetry's logger in case both ase rates service and fallback rates service conversions fail.
This just remains an early return

packages/backend/src/telemetry/transaction-amount.ts
beniaminmunteanu: Caught and logged all bussines code here to make sure that does not happen. Have another 👀

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -118,6 +124,22 @@ services:
     depends_on:
       - cloud-nine-backend
 
+  #Serves as example for optional local collector
```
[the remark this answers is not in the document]
beniaminmunteanu: Created a new documentation section for integration.
Please do give feedback on it 🙏

beniaminmunteanu:
new docs section added for integrating your own otel collector (the information was previously just commented configs)
Please have a look.

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
mkurapov: > anything in particular?

Mostly to avoid cases where you are trying to do `container.use(...)` expecting to get an instance of a service, but getting nothing/error instead. Edge casey - probably not a big deal for us. optional is ok

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
mkurapov: >  the source is not something all metrics send.

In which cases would it be better not setting `source`? (If `source` would always just be telemetry service, maybe we can even leave it out?)

mkurapov (commented):
Will get back to this!

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
beniaminmunteanu: the source is not always telemetryService. the serviceName in telemetry Service is the config.instanceName of rafiki.
This is so that we have the counts of transactions with distinction between instances in the network in the final dashboards.
I dropped the source attribute in the transaction_amount metric for privacy concerns, but i'd rather keep it here, on the count metric

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
beniaminmunteanu: Ok then, will leave it as optional for now 👍 🙇

njlie:
How does one get access to the Grafana dashboard? I tried signing in with an account created using my Github account and I wasn't able to see it.

beniaminmunteanu:
> How does one get access to the Grafana dashboard? I tried signing in with an account created using my Github account and I wasn't able to see it.

Me | Sarah need to add your emails to the grafana cloud instance. Unfortunately since AMP wasn't a good fit for us, we can't use the already existing AWS accounts.
Will add your emails, you should receive an invite

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -0,0 +1,88 @@
...
+    - ../collector/otel-collector-config.yaml:/etc/otel-collector-config.yaml
+  networks:
+    - rafiki
+  expose:
+    - 4317
+  ports:
+    - '13132:13133' # health_check extension
+```
+
+#### OpenTelemetry OTEL collector config:
```
mkurapov, line 42: maybe worth linking to https://opentelemetry.io/docs/collector/configuration/ in this section?

packages/documentation/src/content/docs/telemetry/overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+The telemetry feature is optional for ASEs.
+
+## Architecture
+
+The architecture of the telemetry feature is illustrated below:
+
+![Telemetry architecture](/img/telemetry-architecture.png)
+
+## Opentelemetry
```
mkurapov, line 27:
```suggestion
## OpenTelemetry
```

packages/backend/src/tests/telemetry.ts
```diff
@@ -0,0 +1,64 @@
...
+    return 'USD'
+  }
+
+  public getBaseScale(): number {
+    return 4
+  }
+
+  // public applyPrivacy(rawValue: number): number {
+  //   return rawValue + Math.random() * 100
+  // }
```
mkurapov, line 63: we can just remove this commented code

packages/backend/src/telemetry/privacy.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  })
+
+  test('generateLaplaceNoise should return a different number each time', () => {
+    const scale = 0.5
+    const noise1 = privacy.generateLaplaceNoise(scale)
+    const noise2 = privacy.generateLaplaceNoise(scale)
+    expect(noise1).not.toBe(noise2)
+  })
+
+  test('computePrivacyParameter should return 0 when sensitivity is 0', () => {
```
mkurapov, line 19: we should add a non-0 test here

packages/backend/src/telemetry/privacy.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  const clipParams = {
+    minBucketSize: 1000,
+    maxBucketSize: 10000
+  }
+
+  afterEach(() => {
+    jest.unmock('./privacy')
+  })
+
+  test('generateLaplaceNoise should return a different number each time', () => {
```
mkurapov, line 12: I think in this test we should just mock `Math.random` and verify the deterministic parts

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -0,0 +1,88 @@
+---
+title: Building on Top of Rafiki Metrics
```
mkurapov, line 2: Maybe something like: Deploying Custom Telemetry Service? Since "Building on Top of Rafiki Metrics" sounds more like adding additional metrics

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    telemetryService
+      .getOrCreate('transactions_amount', {
+        description: 'Amount sent through the network',
+        valueType: ValueType.DOUBLE
+      })
+      .add(privacy.applyPrivacy(Number(converted)))
+  } catch (e) {
+    console.error(`Unable to collect telemetry`, e)
```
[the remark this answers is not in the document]
beniaminmunteanu: I will. I was still looking for a way to get the usual logger in there without passing it as a function param.
But will just do it that way for simplicity's sake

packages/backend/src/telemetry/transaction-amount.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!amount) {
+    return
+  }
+
+  const convertOptions = {
+    sourceAmount: amount,
+    sourceAsset: { code: asset.code, scale: asset.scale },
+    destinationAsset: {
+      code: telemetryService.getBaseAssetCode(),
+      scale: telemetryService.getBaseScale()
```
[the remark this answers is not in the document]
beniaminmunteanu: yes that makes total sense. Didn't notice it, thanks!

packages/backend/src/app.ts
```diff
@@ -204,6 +204,7 @@ const WALLET_ADDRESS_PATH = '/:walletAddressPath+'
 
 export interface AppServices {
   logger: Promise<Logger>
+  telemetry: Promise<TelemetryService>
```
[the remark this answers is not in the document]
beniaminmunteanu: will look into it 👍

packages/backend/src/telemetry/transaction-amount.ts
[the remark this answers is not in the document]
mkurapov: I would probably just have `createAccountToAccountTransfer` take in `deps: ServiceDependencies` (from the AccountingServices) as an argument, and then just do `deps.logger`

packages/backend/src/telemetry/privacy.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const roundedValue = privacy.roundValue(rawValue, bucketSize, clipParams)
+    expect(roundedValue).toBe(clipParams.maxBucketSize)
+  })
+
+  test('roundValue should return a number within the specified range', () => {
+    const rawValue = 5000
+    const bucketSize = 1000
+    const roundedValue = privacy.roundValue(rawValue, bucketSize, clipParams)
+    expect(roundedValue).toBeGreaterThanOrEqual(clipParams.minBucketSize)
+    console.log(roundedValue)
```
njlie, line 44: nit: clean up this log?

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -0,0 +1,88 @@
...
+---
+
+Rafiki allows for integrating [Account Servicing Entities](/reference/glossary#account-servicing-entity) (ASE) to build their own telemetry solution based on the [OpenTelemetry](https://opentelemetry.io/) standardized metrics format that Rafiki exposes.
+
+In order to do so, the integrating ASE must deploy its own OpenTelemetry collector that should act as a sidecar container to Rafiki. It needs to provide the OpenTelemetry collector's ingest endpoint so that Rafiki can start sending metrics to it.
+
+## Rafiki Telemetry Environment Variables
```
njlie, line 9: Nitpicky, I know, but I think this section would benefit from more explicit descriptions. I was able to discern what they do based on the name of the variable, but I think some readers would benefit from less ambiguity. For example:

```
- `ENABLE_TELEMETRY`: boolean, defaults to `true`. -> - `ENABLE_TELEMETRY`: boolean, defaults to `true`. Enables the telemetry service on Rafiki.
- `OPEN_TELEMETRY_COLLECTOR_URL`: CSV of URLs (e.g., `http://otel-collector-NLB-e3172ff9d2f4bc8a.elb.eu-west-2.amazonaws.com:4317,http://happy-life-otel-collector:4317`). -> - `OPEN_TELEMETRY_COLLECTOR_URL`: CSV of URLs for Open Telemetry collectors (e.g., `http://otel-collector-NLB-e3172ff9d2f4bc8a.elb.eu-west-2.amazonaws.com:4317,http://happy-life-otel-collector:4317`).

File: packages/documentation/src/content/docs/telemetry/overview.md [hunk 26]
```diff
@@ -0,0 +1,76 @@
...
+## Instrumentation
+
+Rafiki currently has two counter metrics. All data points (counter increases) are exported to collection endpoints at a configurable interval (default recommended to 15s).
+
+Currently collected metrics:
+
+- `transactions_total` (Open-payments) - Counter metric
+  - Description: “Count of funded transactions”
+  - This counter metric increases by 1 for each successfully sent transaction.
+  - It is collected in the open-payments module, outgoing payment service.
```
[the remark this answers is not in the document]
beniaminmunteanu: ok so it still didn't update. I made these changes a while ago but git somehow misses the diff. Will investigate why

packages/backend/src/telemetry/privacy.test.ts
beniaminmunteanu: i added another test for your suggestion, have a look.

Also:
would you want me to remove this one entirely? I'm presuming you are thinking that the test might fail every now and then since theoretically Math.random is pseudo-random.
I'd argue that we won't hit it :D

packages/backend/src/telemetry/transaction-amount.ts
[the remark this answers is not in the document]
beniaminmunteanu: done. Please have a look, as DI changes might also affect accounting service

packages/backend/src/telemetry/privacy.test.ts
beniaminmunteanu: done. I hope that's what you meant by it

packages/backend/src/telemetry/service.ts
```diff
@@ -0,0 +1,147 @@
...
+    diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG)
+    this.serviceName = deps.serviceName
+    this.fallbackRatesService = deps.fallbackRatesService
+    this.aseRatesService = deps.aseRatesService
+
+    if (
+      deps.collectorUrls &&
+      Array.isArray(deps.collectorUrls) &&
+      deps.collectorUrls.length === 0
+    ) {
```
BlairCurrey, line 63: small thing but maybe we can just do `if(!deps.collectorUrls.length){`? we know from TS at this point that it's `string[]`.

packages/backend/src/telemetry/service.ts
```diff
@@ -0,0 +1,147 @@
...
+      deps.logger.info(
+        'No collector URLs specified, metrics will not be exported'
+      )
+      return
+    }
+
+    this.meterProvider = new MeterProvider({
+      resource: new Resource({ 'service.name': 'RAFIKI_NETWORK' })
```
BlairCurrey, line 71: What's the service name here? Just something internal to the meter provider (I don't see it anywhere else)? Something that should match some configuration on somewhere AWS? I'm just wondering if this is the best way to set it - looks like a potential 'magic value'.

packages/backend/src/telemetry/service.ts
```diff
@@ -0,0 +1,147 @@
...
+    const destinationAsset = {
+      code: this.deps.baseAssetCode,
+      scale: this.deps.baseScale
+    }
+
+    let converted = await this.aseRatesService.convert({
+      ...convertOptions,
+      destinationAsset
+    })
+    if (typeof converted !== 'bigint' && converted in ConvertError) {
```
BlairCurrey, line 119: Kinda just a style thing but if you wanted you could do this:
```ts
    if (isConvertError(converted)) {
      converted = await this.fallbackRatesService.convert(convertOptions)
      if (isConvertError(converted)) {
        this.deps.logger.error(`Unable to convert amount: ${converted}`)
      }
    }
```

After adding this to `rafiki/packages/backend/src/rates/service.ts`
```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
export const isConvertError = (o: any): o is ConvertError =>
  Object.values(ConvertError).includes(o)
```

This is a pattern we follow a lot of other places but looks like we just didnt add such a util for `ConvertError`. I like it but I guess its kinda personal preference since both narrow the types. Seems like the `&& converted in ConvertError` is not strictly necessary in any case.

packages/backend/src/telemetry/service.ts
```diff
@@ -0,0 +1,147 @@
...
+    })
+
+    metrics.setGlobalMeterProvider(this.meterProvider)
+  }
+
+  private createCounter(
+    name: string,
+    options: MetricOptions | undefined
+  ): Counter {
+    const counter = metrics.getMeter('Rafiki').createCounter(name, options)
```
BlairCurrey, line 94: same quest as here basically: https://github.com/interledger/rafiki/pull/2299/files#r1474772665

Is it OK to hardcode `'Rafiki'` like this? I was not able to work out where exactly this is coming from.

packages/backend/src/telemetry/service.test.ts
```diff
@@ -0,0 +1,47 @@
+import { ConvertError } from '../rates/service'
+import { MockTelemetryService, mockCounter } from '../tests/telemetry'
+
+const telemetryService = new MockTelemetryService()
+describe('TelemetryServiceImpl', () => {
+  it('should create a counter when getOrCreate is called for a new metric', () => {
+    const counter = telemetryService.getOrCreateMetric('testMetric')
+    expect(counter).toBe(mockCounter)
+  })
```
BlairCurrey, line 9: I get mocking the telemetry service when testing `collectTelemetryAmount` for example, but doesn't this just test our `MockTelemetryService`, which may or may not be entirely different from our actual `TelemetryServiceImpl`? We could make `TelemetryServiceImpl.getOrCreateMetric` not return a counter and this would still pass. I would expect to test this by creating an actual `TelemetryServiceImpl` and mock the dependencies as needed.

BlairCurrey:
I see it working locally when runnig through the postman flow and viewing the public dashboard - very cool!

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
mkurapov: I think `aseRatesService` is unused here?

packages/backend/src/telemetry/service.ts
beniaminmunteanu: The service name here is intended to be the same on all rafiki instrumented instances. it will serve as an aggregator property for metrics that are related to the "rafiki network / ILP Network" as a whole. By having it set on the sdk, all metrics will have their "job" property set as "RAFIKI_NETWORK".
You can think of it as the one property that identifies all the metrics with the same intent: Observability of the Rafiki/ILP Network.

packages/backend/src/telemetry/service.ts
beniaminmunteanu: sure thing, i'm pro typescript guards. Will add, thanks

packages/backend/src/telemetry/service.ts
beniaminmunteanu: right. thanks

packages/backend/src/telemetry/service.ts
beniaminmunteanu: `metrics` is imported directly from the '`@opentelemetry/api`' package.
By calling `getMeter`, a new Meter named "Rafiki" in this case will be created and returned from the global meter provider.
It's just the name of the "Meter" that will be responsible of creating all our counter metrics.
The name is a required parameter, and I don't think it should be externaly configurable or something.

Moved it out of inline string definition into a const in the same file.

packages/backend/src/telemetry/service.test.ts
beniaminmunteanu: valid. Will come back to this ❤️

packages/backend/src/tests/telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+        GBP: 1.25,
+        RON: 0.22,
+        USD: 1,
+        XRP: 0.5994
+      }
+    }
+  }
+}
+
+export class MockTelemetryService implements TelemetryService {
```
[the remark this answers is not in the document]
beniaminmunteanu: will look into it, thanks!

packages/backend/src/index.ts
```diff
@@ -127,6 +128,36 @@ export function initIocContainer(
...
+      const config = await deps.use('config')
+      return createTelemetryService({
+        logger: await deps.use('logger'),
+        aseRatesService: await deps.use('ratesService'),
+        fallbackRatesService: createRatesService({
+          logger: await deps.use('logger'),
+          exchangeRatesUrl: config.telemetryExchangeRatesUrl,
+          exchangeRatesLifetime: config.telemetryExchangeRatesLifetime
+        }),
+        serviceName: config.instanceName,
```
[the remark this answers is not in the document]
beniaminmunteanu: renamed to instanceName

packages/backend/src/accounting/service.ts
```diff
@@ -9,11 +12,16 @@ export enum LiquidityAccountType {
...
+  ledger: number
+}
+
 export interface LiquidityAccount {
   id: string
-  asset: {
-    id: string
-    ledger: number
+  asset: LiquidityAccountAsset & {
     onDebit?: (options: OnDebitOptions) => Promise<LiquidityAccount>
```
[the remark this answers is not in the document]
beniaminmunteanu: done

packages/backend/src/telemetry/service.test.ts
beniaminmunteanu: done

packages/backend/src/tests/telemetry.ts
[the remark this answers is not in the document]
mkurapov: If you want to avoid using this class altogether (in `transaction-amount` tests), you can either just do `jest.mock` like we were chatting about earlier to define `telemetryService`, or initialize the dependencies, and just mockImplementation whenever you need to spy on a method. Up to you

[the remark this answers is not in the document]
beniaminmunteanu: Yop, thanks. 
I decided I would keep it for now and see on further telemetry development if i should replace it with jest mocks altogheter.

packages/backend/src/index.ts
```diff
@@ -127,6 +128,40 @@ export function initIocContainer(
...
+    const config = await deps.use('config')
+    return createRatesService({
+      logger: await deps.use('logger'),
+      exchangeRatesUrl: config.exchangeRatesUrl,
+      exchangeRatesLifetime: config.exchangeRatesLifetime
+    })
+  })
+
+  if (config.enableTelemetry) {
+    container.singleton('fallbackRatesService', async (deps) => {
```
mkurapov, line 142: I think the name should maybe be `externalRatesService` now that it's a separate thing, and doesn't just live in the telemetry service, thoughts?

packages/backend/src/telemetry/service.ts
```diff
@@ -0,0 +1,145 @@
...
+class TelemetryServiceImpl implements TelemetryService {
+  private instanceName: string
+  private meterProvider?: MeterProvider
+  private fallbackRatesService: RatesService
+  private aseRatesService: RatesService
+
+  private counters = new Map()
+  constructor(private deps: TelemetryServiceDependencies) {
+    // debug logger:
+    // diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG)
```
mkurapov, line 52: Do we need to uncomment these?

localenv/happy-life-bank/seed.yml
```diff
@@ -5,55 +5,55 @@ self:
...
     path: accounts/pfry
     id: 97a3a431-8ee1-48fc-ac85-70e2f5eba8e5
-    initialBalance: 50000
+    initialBalance: 1
     postmanEnvVar: pfryWalletAddress
     assetCode: USD
   - name: 'PlanEx Corp'
     id: a455cc54-b583-455b-836a-e5275c5c05b7
-    initialBalance: 2000
+    initialBalance: 2000000
```
mkurapov, line 37: making everyone rich hey

packages/documentation/src/content/docs/telemetry/privacy.md
```diff
@@ -0,0 +1,57 @@
...
+
+The privacy parameter is computed as one-tenth of the sensitivity. This parameter controls the trade-off between privacy and utility: a smaller privacy parameter means more privacy but less utility, and a larger privacy parameter means less privacy but more utility.
+
+The noise, selected from the Laplacian distribution, is then generated using this privacy parameter and added to the rounded value. If the resulting value is zero, it is set to half the bucket size to ensure that the noise does not completely obscure the transaction value.
+
+## Currency Conversion
+
+Another factor that obscures sensitive data is currency conversion. In cross-currency transactions, we use exchange rates that are not traced back. The rates are provided by the integrating [ASE](/reference/glossary#account-servicing-entity) at runtime and are not persisted. If the needed rates are not provided, external API's for exchange rates are used. The obtained exchange rates are overwritten frequently in this case, with no versioning or history access. This introduces an additional layer of noise and further protects the privacy of the transactions.
```
mkurapov, line 33: maybe we can re-word "traced back" better? something like, "since exchange rates are provided by the ASE internally, they cannot be correlated to an individual transaction"

packages/documentation/src/content/docs/telemetry/privacy.md
```diff
@@ -0,0 +1,57 @@
...
+
+- [Local Differential Privacy for Human-Centered Computing](https://proceedings.neurips.cc/paper_files/paper/2017/file/253614bbac999b38b5b60cae531c4969-Paper.pdf)
+- [Collecting Telemetry Data Privately](https://www.microsoft.com/en-us/research/blog/collecting-telemetry-data-privately/)
+- [Collecting Telemetry Data Privately - NeurIPS Publication](https://proceedings.neurips.cc/paper_files/paper/2017/file/253614bbac999b38b5b60cae531c4969-Paper.pdf) by Bolin Ding, Janardhan Kulkarni, Sergey Yekhanin from Microsoft Research.
+- [RAPPOR: Randomized Aggregatable Privacy-Preserving Ordinal Response](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/42852.pdf)
+
+# Experimental Transaction Values when using the Algorithm
+
+The following table shows the values in the algorithm when running transactions for different amounts. The raw value increases as you go down the rows of the table.
+(all values are in scale 4)
```
mkurapov, line 47: is " (all values are in scale 4)" necessary? if so, maybe it's worth expanding on this just a bit

packages/documentation/src/content/docs/telemetry/privacy.md
mkurapov, line 33:
```suggestion
Another factor that obscures sensitive data is currency conversion. In cross-currency transactions, we use exchange rates that are not traced back. The rates are provided by the integrating [ASE](/reference/glossary#account-servicing-entity) at runtime and are not persisted. If the necessary rates are not provided or not available from the ASE, an external API for exchange rates are used. The obtained exchange rates are overwritten frequently in this case, with no versioning or history access. This introduces an additional layer of noise and further protects the privacy of the transactions.
```

packages/documentation/src/content/docs/telemetry/privacy.md
```diff
@@ -0,0 +1,57 @@
+---
+title: Privacy
+---
+
+Rafiki telemetry is designed with a strong emphasis on privacy. The system anonymizes user data and refrains from collecting identifiable information. Since transactions can originate from any user to a Rafiki instance, the privacy measures are implemented at the Rafiki instance level in the network. This means that at the individual level, the data is already anonymous as single Rafiki instances service transactions for multiple users.
```
mkurapov, line 5:
```suggestion
Rafiki telemetry is designed with a strong emphasis on privacy. The system anonymizes user data and refrains from collecting identifiable information. Since transactions can originate from any user to a Rafiki instance, the privacy measures are implemented directly at the source (each Rafiki instance). This means that at the individual level, the data is already anonymous as single Rafiki instances service transactions for multiple users.
```

packages/documentation/src/content/docs/telemetry/overview.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Purpose
+
+The objective of the telemetry feature is to gather metrics and establish an infrastructure for visualizing valuable network insights. The metrics we collect include:
+
+- The total amount of money transferred via packet data within a specified time frame (daily, weekly, monthly).
+- The number of transactions from outgoing payments that have been at least partially successful.
+- The average amount of money held within the network per transaction.
+
+Our goal is to use these data for our own insights and to enable [Account Servicing Entities](/reference/glossary#account-servicing-entity) (ASEs) to gain their own insights. We aim to track the growth of the network in terms of transaction sizes and the number of transactions processed.
```
mkurapov, line 13: Maybe the "we aim to track the growth of the network..." sentence should go first?

packages/backend/src/index.ts
beniaminmunteanu: technically speaking, the rates service that hits the ase provided rates endpoint is also external in rafiki's perspective, right?
That's why i went for fallback

packages/backend/src/telemetry/service.ts
beniaminmunteanu: no, I'm leaving the comment there just for when debugging is needed on local development setup.
keeping the debug logger on at all times would also log on every 15 seconds, even if new metrics are not there.
So i'd rather just keep them off.

packages/documentation/src/content/docs/telemetry/privacy.md
beniaminmunteanu: As i mentioned before, I would argue that scale 2 is better for telemetry than scale 4.
But scale 4 came down as a requirement for some reason (which I don't know)

beniaminmunteanu: by:
1. enforcing privacy (rounding + noise addition)
2. using promql operators that change monotonic increasing counters into rate-of-change graphs

We are already disregarding complete accuracy.

mkurapov: I just meant more so in the context of the example table. We don't need to provide the scale necessarily, since it doesn't affect the values in the example table is all.

packages/backend/src/index.ts
beniaminmunteanu: and i belive that other occurances of using this external rates service will also be as a fallback to ASE

packages/documentation/src/content/docs/telemetry/integrating.md
hajjimo, line 1: The link at the end of the following sentence throws a 404:  "It defines the endpoint that Rafiki will query for exchange rates, as a fallback when ASE does not [provide them]"

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -0,0 +1,91 @@
...
+
+- `ENABLE_TELEMETRY`: boolean, defaults to `true`. Enables the telemetry service on Rafiki.
+- `OPEN_TELEMETRY_COLLECTOR_URL`: CSV of URLs for Open Telemetry collectors (e.g., `http://otel-collector-NLB-e3172ff9d2f4bc8a.elb.eu-west-2.amazonaws.com:4317,http://happy-life-otel-collector:4317`).
+- `OPEN_TELEMETRY_EXPORT_INTERVAL`: number in milliseconds, defaults to `15000`. Defines how often the instrumented Rafiki instance should send metrics.
+- `TELEMETRY_EXCHANGE_RATES_URL`: string URL, defaults to `https://telemetry-exchange-rates.s3.amazonaws.com/exchange-rates-usd.json`. It defines the endpoint that Rafiki will query for exchange rates, as a fallback when ASE does not [provide them](../integration/getting-started.md#exchange-rates). If set, the response format of the external exchange rates API should be of type Rates, as the rates service expects.
```
melissahenderson, line 14: The getting-started.md#exchange-rates link breaks in preview mode

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -0,0 +1,91 @@
...
+
+In order to do so, the integrating ASE must deploy its own OpenTelemetry collector that should act as a sidecar container to Rafiki. It needs to provide the OpenTelemetry collector's ingest endpoint so that Rafiki can start sending metrics to it.
+
+## Rafiki Telemetry Environment Variables
+
+- `ENABLE_TELEMETRY`: boolean, defaults to `true`. Enables the telemetry service on Rafiki.
+- `OPEN_TELEMETRY_COLLECTOR_URL`: CSV of URLs for Open Telemetry collectors (e.g., `http://otel-collector-NLB-e3172ff9d2f4bc8a.elb.eu-west-2.amazonaws.com:4317,http://happy-life-otel-collector:4317`).
+- `OPEN_TELEMETRY_EXPORT_INTERVAL`: number in milliseconds, defaults to `15000`. Defines how often the instrumented Rafiki instance should send metrics.
+- `TELEMETRY_EXCHANGE_RATES_URL`: string URL, defaults to `https://telemetry-exchange-rates.s3.amazonaws.com/exchange-rates-usd.json`. It defines the endpoint that Rafiki will query for exchange rates, as a fallback when ASE does not [provide them](../integration/getting-started/#exchange-rates). If set, the response format of the external exchange rates API should be of type Rates, as the rates service expects.
```
mkurapov, line 14:
```suggestion
- `TELEMETRY_EXCHANGE_RATES_URL`: string URL, defaults to `https://telemetry-exchange-rates.s3.amazonaws.com/exchange-rates-usd.json`. It defines the endpoint that Rafiki will query for exchange rates, as a fallback when ASE does not [provide them](/integration/getting-started/#exchange-rates). If set, the response format of the external exchange rates API should be of type Rates, as the rates service expects.
```
to fix

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
beniaminmunteanu: it does, I just missed it. Thanks!
Regarding the noop, Is it ok if I do it after this gets in?

packages/backend/src/accounting/service.ts
```diff
@@ -183,6 +196,21 @@ export async function createAccountToAccountTransfer(
           withdrawalThrottleDelay
         })
       }
+
+      if (
+        destinationAccount.onDebit &&
```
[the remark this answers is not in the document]
beniaminmunteanu: this was suggested in:
https://github.com/interledger/rafiki/pull/2299#discussion_r1451048121

Basically this gets executed both on the sender and the receiver of a transaction.
By doing the onDebit check on the destinationAccount, we are making sure that we are on the sender side

packages/backend/src/telemetry/privacy.ts
```diff
@@ -0,0 +1,67 @@
...
+  minBucketSize: number
+  maxBucketSize: number
+}
+
+export const privacy = {
+  getBucketSize: function (rawValue: number, clip: ClipParams): number {
+    const { minBucketSize, maxBucketSize } = clip
+    const base = 2
+    const scale = 5000
+    const threshold = 20000
```
sabineschaller, line 11: Can you add a comment to explain those hardcoded values?

packages/backend/src/telemetry/service.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  it('should return the instance name when calling getInstanceName', () => {
+    const serviceName = telemetryService.getInstanceName()
+
+    expect(serviceName).toBe('Rafiki')
+  })
+
+  describe('conversion', () => {
+    it('should try to convert using aseRatesService and fallback to internalRatesService', async () => {
```
sabineschaller, line 74: This only tests the fallback. Should we add a test to test that using the aseRatesService works?

packages/backend/src/accounting/psql/service.ts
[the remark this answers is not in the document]
mkurapov: > However, I wouldn't console.log but rather add proper logging to the noop service. 

Yup, agree.

> Is it ok if I do it after this gets in?

We can a follow up one to close this.

packages/backend/src/tests/telemetry.ts
```diff
@@ -0,0 +1,63 @@
...
+    _name: string,
+    _options?: MetricOptions | undefined
+  ): Counter<Attributes> {
+    return mockCounter
+  }
+  public getInstanceName(): string | undefined {
+    return 'serviceName'
+  }
+  public shutdown(): void {
+    console.log('telemetry service shutdown')
```
mkurapov, line 43:
```suggestion
```

mkurapov (approved):
I think anything after can be added to separate PR

mkurapov (approved):

sabineschaller:
Tested it and works smoothly. Final question: Should this work on localenv?
@beniaminmunteanu @JoblersTune @AlexLakatos @mkurapov

mkurapov:
@sabineschaller you mean enabled by default on localenv, correct? 

I think it'll be good to have it on at rafiki.money (of course) but I don't think that would be particularly useful on localenv

beniaminmunteanu:
@sabineschaller  @mkurapov 
Updated such that telemetry is still enabled by default, but localenv docker-compose files turn it off.

mkurapov (dismissed):

packages/backend/src/config/app.ts
```diff
@@ -38,9 +38,15 @@ dotenv.config({
 export const Config = {
   logLevel: envString('LOG_LEVEL', 'info'),
   enableTelemetry: envBool('ENABLE_TELEMETRY', true),
-  openTelemetryCollectors: envStringArray('OPEN_TELEMETRY_COLLECTOR_URL', [
-    'http://otel-collector-NLB-e3172ff9d2f4bc8a.elb.eu-west-2.amazonaws.com:4317'
-  ]),
+  livenet: envBool('LIVENET', false),
+  openTelemetryCollectors: envStringArray(
+    'OPEN_TELEMETRY_COLLECTOR_URL',
```
mkurapov, line 43:
```suggestion
    'OPEN_TELEMETRY_COLLECTOR_URLS',
```
should it be plural?

mkurapov (dismissed):

packages/documentation/src/content/docs/telemetry/integrating.md
```diff
@@ -9,6 +9,7 @@ In order to do so, the integrating ASE must deploy its own OpenTelemetry collect
 ## Rafiki Telemetry Environment Variables
 
 - `ENABLE_TELEMETRY`: boolean, defaults to `true`. Enables the telemetry service on Rafiki.
+- `LIVENET`: boolean. Should be set to `true` on production environments. If it is not set, it will default to `false`, and metrics will get sent to the testnet otel-collector
```
sabineschaller, line 12:
```suggestion
- `LIVENET`: boolean. Should be set to `true` on production environments dealing with real money. If it is not set, it will default to `false`, and metrics will get sent to the testnet otel-collector
```

mkurapov (approved):

sabineschaller (approved):

PR #2306, opened 2024-01-05
feat(grants): add sorting option

### LINKED ISSUE #2305: Support sort order for grant list
## Context
Grant query is ordered by default ascending. Usually an user will be interested to see the most recent grants instead. 
Default sort order should be changed to descending and add a parameter to indicate desired sort order.
## Todos
- [x] Change default grants order to descending
- [x] Add parameter for sort order on grant list

dragosp1011:

- Add sort order for grants graphql query

## Context

fixes #2305

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/shared/baseModel.test.ts
+++ b/packages/auth/src/shared/baseModel.test.ts
@@ -1,8 +1,9 @@
-import { BaseModel, Pagination } from './baseModel'
+import { BaseModel, Pagination, SortOrder } from './baseModel'
+import { getPageInfo } from './pagination'
 
 interface PageTestsOptions<Type> {
   createModel: () => Promise<Type>
-  getPage: (pagination?: Pagination) => Promise<Type[]>
+  getPage: (pagination?: Pagination, sortOrder?: SortOrder) => Promise<Type[]>
 }
 
 export const getPageTests = <Type extends BaseModel>({
@@ -17,6 +18,7 @@ export const getPageTests = <Type extends BaseModel>({
       for (let i = 0; i < 22; i++) {
         modelsCreated.push(await createModel())
       }
+      modelsCreated.reverse() // default sort order is DESC
     })
 
     test.each`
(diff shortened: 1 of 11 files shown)


packages/auth/src/shared/pagination.ts
```diff
@@ -1,8 +1,24 @@
...
+  last,
+  cursor
+}: PaginationArgs): Pagination {
+  return {
+    first,
+    last,
+    before: last ? cursor : undefined,
+    after: cursor && !last ? cursor : undefined
+  }
+}
```
BlairCurrey, line 16: Looks like this function is not used anywhere?

packages/auth/src/shared/baseModel.test.ts
```diff
@@ -17,6 +18,7 @@ export const getPageTests = <Type extends BaseModel>({
       for (let i = 0; i < 22; i++) {
         modelsCreated.push(await createModel())
       }
+      modelsCreated.reverse() // default sort order is DESC
```
BlairCurrey, line 21: Small thing, but I think we can optimize the test execution time a bit here by avoiding this `reverse` and the subsequent `reverse` calls in tests where we want ascending.

Like:

```ts
let modelsCreatedAsc: Type[] = []
let modelsCreatedDesc: Type[] = []

  for (let i = 0; i < 22; i++) {
    modelsCreatedAsc.push(await createModel())
    modelsCreatedDesc[21 - i] = await createModel()
  }
})
```

Then use `modelsCreatedAsc` where we use `modelsCreated.reverse()` and `modelsCreatedDesc` where we use `modelsCreated`.

Maybe not a huge factor time-wise since there are 22 only items but it could only help. By my count the `beforeEach` is executing 22 times plus the 7 times where we re-reverse inside tests multiplied by every time `getPage` is called (0 currently? but should probably be used for grants). So we're cutting out 29 o(n) operations (assuming we add it to grants service test).

Side note, I was surprised this works: `modelsCreatedDesc[21 - i] = await createModel()` since the index is out of bounds. I figured it needed an array of length 22. Was a fun little rabbit hole: https://stackoverflow.com/a/22850666/12893676

packages/auth/src/shared/baseModel.test.ts
BlairCurrey, line 1: I see that this test was already here, but it looks like it's not used anywhere? I guess we should probably be doing this in the grant service test?

```ts
  describe('getPage', (): void => {
    getPageTests({
      createModel: () => createGrant(deps),
      getPage: (pagination?: Pagination, sortOrder?: SortOrder) =>
        grantService.getPage(pagination, undefined, sortOrder)
    })
  })
```

JoblersTune:
LGTM with Blair's feedback.

Issue #917, opened 2022-12-21
Audit use of @ts-ignore


wilsonianb:
Particularly
https://github.com/interledger/rafiki/blob/e773fb7cd533a1a8495e056987a772154651130b/packages/backend/src/app.ts#L369-L370

kurayami07734:
Please assign me to this issue

PR #2340, opened 2024-01-12
fix: improve type safety

BlairCurrey:

- removes uses of `ts-ignore` in favor of type assertions or fixing underlying types

## Context

These changes fall under a few categories:
  - there was no ts error (code or tsconfig probably changed after comment was committed?) so I just removed the `ts-ignore` comment
  - (most common) could not make it typesafe in which case I used type assertions instead of the `ts-ignore` comment. This is only a marginal improvement but I preferred this because it narrows the errors typescript will ignore and codifies the assumptions we are making.  For example, one of our `ts-ignore` comments was unintentionally hiding a `ReferenceError`: https://github.com/interledger/rafiki/pull/2340/commits/e6a643f294b71c57f5d622e0a0ce4e0aeb9e5f7e#r1450753470
  - fixed the underlying typescript errors 🥳 
  
If I couldn't find a reasonable way to do type assertions (using `any` or 
 too complicated I guess) then I would have left the ts-ignore comment, but that did not end up being the case.

fixes: https://github.com/interledger/rafiki/issues/917

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/graphql/resolvers/page.test.ts
+++ b/packages/auth/src/graphql/resolvers/page.test.ts
@@ -52,19 +52,17 @@ export const getPageTests = <T extends Model, M extends BaseModel>({
   const toConnection = (query: ApolloQueryResult<T>): Connection<T> => {
     if (query.data) {
       if (parent) {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        if (query.data[parent.query]) {
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
-          return query.data[parent.query][pagedQuery]
+        const parentData = query.data[parent.query as keyof typeof query.data]
+
+        if (parentData) {
+          return (parentData as Record<string, Connection<T>>)[pagedQuery]
         } else {
           throw new Error(`Parent ${parent.query} was empty`)
         }
       } else {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        return query.data[pagedQuery]
+        return query.data[
+          pagedQuery as keyof typeof query.data
+        ] as Connection<T>
       }
     } else {
       throw new Error('Data was empty')
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -230,9 +230,7 @@ export const start = async (
       logger.info('completed graceful shutdown.')
       process.exit(0)
     } catch (err) {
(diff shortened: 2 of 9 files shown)


packages/auth/src/index.ts
```diff
@@ -230,9 +230,7 @@ export const start = async (
       logger.info('completed graceful shutdown.')
       process.exit(0)
     } catch (err) {
-      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-      // @ts-ignore
-      const errInfo = err instanceof Error && error.stack ? err.stack : err
```
BlairCurrey, line 235: This case illustrates why we should prefer type assertions over `ts-ignore` comments. This `ts-ignore` was inadvertently hiding a `ReferenceError`. We call `error.stack` but error is not defined anywhere.

packages/backend/src/app.ts
```diff
@@ -553,8 +553,6 @@ export class App {
       quoteRoutes.get
     )
 
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
BlairCurrey, line 556: This was the one specifically called out in the original issue https://github.com/interledger/rafiki/issues/917 and I could use a sanity check on the changes. Here is the full commit: https://github.com/interledger/rafiki/pull/2340/commits/7237c7090cf46970b897a6fb1bb079b5d058c83f

Mostly want to double check that the type and tests are safe to remove. I believe it should have been removed alongside the other connection stuff when we replaced it with incoming payment methods. I only see one place the spsp middleware is used (here) so the removed context type and tests should no longer be needed.

mkurapov (approved):

packages/auth/src/graphql/resolvers/page.test.ts
```diff
@@ -52,19 +52,17 @@ export const getPageTests = <T extends Model, M extends BaseModel>({
...
+          return (parentData as Record<string, Connection<T>>)[pagedQuery]
         } else {
           throw new Error(`Parent ${parent.query} was empty`)
         }
       } else {
-        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-        // @ts-ignore
-        return query.data[pagedQuery]
+        return query.data[
+          pagedQuery as keyof typeof query.data
```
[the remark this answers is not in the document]
BlairCurrey: Yeah I was hoping to do something like that but moved on after some initial attempts did not work. More sophisticated types here would be nice. Not worried about it ATM but I would be curious to see what that would look like.

PR #1459, opened 2023-05-18
refactor(frontend): update message storage for notifications

raducristianpopa:
## Changes proposed in this pull request

Toast notifications are now appearing on physical mobile devices too. The primary issue was that in the development environment, `secure` was mistakenly set to `true`. Additionally, the message storage was refactored and a close button was included in the toast notification.

## Context

- fixes #1451 

## Checklist

- [x] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Documentation added
- [x] Make sure that all checks pass

PR #2349, opened 2024-01-18
chore(frontend): remix v2 upgrade

### LINKED ISSUE #1464: Prepare for Remix future changes

BlairCurrey:

- upgrades remix to v2 which includes:
  - refactoring `ErrorBoundary` (now handles what was previously seperate `ErrorBoundary` and `CaughtBoundary`)
  - updates `MetaFunction` args
  - specifies `cjs` as module type (default switched to `esm`)
  - renames some types whose names were changed
  - renames routes in MASE to conform to new route nesting pattern (denoted by `.` rather than sub directory)
- adds `typecheck` command to `package.json` and run in ci (as well as fixing a pre-existing type error)

## Context

fixes #1464
renovate pr https://github.com/interledger/rafiki/pull/1886

## Checklist

- [x] Tests added/updated
- [x] Documentation added


BlairCurrey:
I noticed form the renovate pr that we're also using remix in the MASE. going to update that as well.

sabineschaller:
This needs frontend-y reviewers. Not me 🙃

localenv/mock-account-servicing-entity/Dockerfile
```diff
@@ -59,4 +59,5 @@ COPY --from=builder /home/rafiki/localenv/mock-account-servicing-entity/build ./
 COPY --from=builder /home/rafiki/localenv/mock-account-servicing-entity/public ./localenv/mock-account-servicing-entity/public
 
 WORKDIR /home/rafiki/localenv/mock-account-servicing-entity
-CMD ["sh", "./node_modules/.bin/remix-serve", "build"]
\ No newline at end of file
+
+CMD ["sh", "-c", "NODE_ENV=production ./node_modules/.bin/remix-serve ./build/index.js"]
```
BlairCurrey, line 63: (this comment also applies to [this same change here](https://github.com/interledger/rafiki/pull/2349/files#diff-3e0c81413453a46c12208866c9fb0686af8eaee0c49cfd701c1fdabd7aa3a660R60))

Couple changes here.

Target was changed to `./build/index.js` to resolve [this error](https://github.com/remix-run/remix/discussions/7586) which is pretty straightforward.

`NODE_ENV` changed to production to resolve [this error](https://github.com/remix-run/remix/discussions/7338). I don't fully understand what caused this to break nor why this was originally running with `NODE_ENV=development` (coming from docker-compose). I would think it should be production given it's running code built in production mode (confirmed here and in main by observing logs from `pnpm --filter mock-account-servicing-entity build`). Docker container's start fine and was able to use MASE UI's and frontends without issue.

localenv/mock-account-servicing-entity/remix.config.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
   // assetsBuildDirectory: "public/build",
   // serverBuildPath: "build/index.js",
   // publicPath: "/build/",
-  serverDependenciesToBundle: ['axios']
+  serverModuleFormat: 'cjs',
+  browserNodeBuiltinsPolyfill: { modules: { querystring: true } }
```
BlairCurrey, line 9: To fix an error from trying to use a node lib in the browser:

> Node builtin "querystring" (imported by "app/lib/utils.ts") must be polyfilled for the browser. You can enable this polyfill in your Remix config, e.g. `browserNodeBuiltinsPolyfill: { modules: { querystring: true } }` [plugin browser-node-builtins-polyfill-plugin]

Going to try and re-implement the util with a URLSearchParams

BlairCurrey: even better, I deleted `app/lib/utils.ts` entirely. Apparently it wasn't doing anything that `URLSearchParams` couldn't. https://github.com/interledger/rafiki/pull/2349/commits/29c4a24a291cab900017c697b18a04182cf45f24

localenv/mock-account-servicing-entity/Dockerfile
JoblersTune: NODE_ENV=development is needed for things like the message server (`/packages/frontend/app/lib/message.server.ts`) to work over HTTP instead of HTTPS
```
export const messageStorage = createCookieSessionStorage<
  SessionData,
  MessageStorageFlashData
>({
  cookie: {
    name: '__message',
    path: '/',
    httpOnly: true,
    sameSite: 'lax',
    maxAge: ONE_MINUTE_IN_S,
    secrets: ['MY_SUPER_SECRET_TOKEN'],
    secure: process.env.NODE_ENV === 'production'
  }
})
```

If the environment is set to production and our local setup doesn't use https then this is gonna break.

packages/frontend/app/entry.server.tsx
```diff
@@ -1,6 +1,6 @@
 import { PassThrough } from 'stream'
 import type { EntryContext } from '@remix-run/node'
-import { Response } from '@remix-run/node'
+import { createReadableStreamFromReadable } from '@remix-run/node'
```
raducristianpopa, line 3: It looks like there is a small change in `handleBotRequest` and `handleBrowserRequest` for the `renderToPipeableStream` method, in the `onError` handler.

This is the default `entry.server.tsx` when creating a new Remix project:

```diff
import { PassThrough } from "node:stream";
import type { AppLoadContext, EntryContext } from "@remix-run/node";
import { createReadableStreamFromReadable } from "@remix-run/node";
import { RemixServer } from "@remix-run/react";
import { isbot } from "isbot";
import { renderToPipeableStream } from "react-dom/server";

const ABORT_DELAY = 5_000;

export default function handleRequest(
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext,
  // This is ignored so we can keep it in the template for visibility.  Feel
  // free to delete this parameter in your app if you're not using it!
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  loadContext: AppLoadContext
) {
  return isbot(request.headers.get("user-agent") || "")
    ? handleBotRequest(
        request,
        responseStatusCode,
        responseHeaders,
        remixContext
      )
    : handleBrowserRequest(
        request,
        responseStatusCode,
        responseHeaders,
        remixContext
      );
}

function handleBotRequest(
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext
) {
  return new Promise((resolve, reject) => {
+    let shellRendered = false;
    const { pipe, abort } = renderToPipeableStream(
      <RemixServer
        context={remixContext}
        url={request.url}
        abortDelay={ABORT_DELAY}
      />,
      {
        onAllReady() {
 +         shellRendered = true;
          const body = new PassThrough();
          const stream = createReadableStreamFromReadable(body);

          responseHeaders.set("Content-Type", "text/html");

          resolve(
            new Response(stream, {
              headers: responseHeaders,
              status: responseStatusCode,
            })
          );

          pipe(body);
        },
        onShellError(error: unknown) {
          reject(error);
        },
        onError(error: unknown) {
+          responseStatusCode = 500;
+          // Log streaming rendering errors from inside the shell.  Don't log
+          // errors encountered during initial shell rendering since they'll
+          // reject and get logged in handleDocumentRequest.
+          if (shellRendered) {
+           console.error(error);
+          }
        },
      }
    );

    setTimeout(abort, ABORT_DELAY);
  });
}

function handleBrowserRequest(
  request: Request,
  responseStatusCode: number,
  responseHeaders: Headers,
  remixContext: EntryContext
) {
  return new Promise((resolve, reject) => {
+    let shellRendered = false;
    const { pipe, abort } = renderToPipeableStream(
      <RemixServer
        context={remixContext}
        url={request.url}
        abortDelay={ABORT_DELAY}
      />,
      {
        onShellReady() {
+          shellRendered = true;
          const body = new PassThrough();
          const stream = createReadableStreamFromReadable(body);

          responseHeaders.set("Content-Type", "text/html");

          resolve(
            new Response(stream, {
              headers: responseHeaders,
              status: responseStatusCode,
            })
          );

          pipe(body);
        },
        onShellError(error: unknown) {
          reject(error);
        },
        onError(error: unknown) {
+          responseStatusCode = 500;
+          // Log streaming rendering errors from inside the shell.  Don't log
+          // errors encountered during initial shell rendering since they'll
+          // reject and get logged in handleDocumentRequest.
+          if (shellRendered) {
+            console.error(error);
+          }
        },
      }
    );

    setTimeout(abort, ABORT_DELAY);
  });
}

```

BlairCurrey: I updated the frontend and MASE to follow the pattern @raducristianpopa shared: [0c660ae](https://github.com/interledger/rafiki/pull/2349/commits/0c660ae9d1e61a79f0988cde3e6376dba9fe6684)

localenv/mock-account-servicing-entity/app/routes/quotes.ts
```diff
@@ -1,4 +1,4 @@
-import type { ActionArgs } from '@remix-run/node'
+import type { ActionFunctionArgs } from '@remix-run/node'
```
mkurapov, line 1: I think we can get rid of this file altogether: we don't call the ASE during quotes anymore since we introduced fees via Admin API

localenv/mock-account-servicing-entity/Dockerfile
BlairCurrey: Looking into this... we haven't been able to reproduce such an issue yet. Seems like the check was added here (https://github.com/interledger/rafiki/pull/1459) so maybe only a problem with physical mobile device?

Also trying to get this working with `NODE_ENV=development`. I get this error which suggests we need to set a `REMIX_DEV_HTTP_ORIGIN` env var but cant find a value that its happy with:

```
/Users/blaircurrey/code/interledger/rafiki/node_modules/.pnpm/@remix-run+server-runtime@2.5.1_typescript@4.9.5/node_modules/@remix-run/server-runtime/dist/dev.js:17
  if (!origin) throw Error("Dev server origin not set");
                     ^
Error: Dev server origin not set
    at Object.broadcastDevReady (/Users/blaircurrey/code/interledger/rafiki/node_modules/.pnpm/@remix-run+server-runtime@2.5.1_typescript@4.9.5/node_modules/@remix-run/server-runtime/dist/dev.js:17:22)
    at Server.onListen (/Users/blaircurrey/code/interledger/rafiki/node_modules/.pnpm/@remix-run+serve@2.5.1_typescript@4.9.5/node_modules/@remix-run/serve/dist/cli.js:123:17)
    at Object.onceWrapper (node:events:627:28)
    at Server.emit (node:events:513:28)
    at emitListeningNT (node:net:1774:10)
    at processTicksAndRejections (node:internal/process/task_queues:81:21)
```

packages/frontend/app/root.tsx
```diff
@@ -95,81 +98,74 @@ export default function App() {
...
-                </Button>
-              </div>
-            </main>
-          </div>
+    <ErrorPage>
+      <div className='flex items-center justify-center flex-col bg-offwhite p-10 rounded-md shadow-md space-y-5'>
+        <div className='grid place-items-center'>
+          <XCircle className='w-10 h-10 text-red-500' />
+          <p className='text-lg font-semibold'>
+            There was an issue with your request.
```
mkurapov, line 158: how does this get triggered?

BlairCurrey: throw some error in a page:

```ts
// rafiki/packages/frontend/app/routes/assets.$assetId.tsx
export default function ViewAssetPage() {
  throw new Error('test')
```

localenv/mock-account-servicing-entity/app/routes/quotes.ts
BlairCurrey: good cath, removed

localenv/mock-account-servicing-entity/Dockerfile
BlairCurrey: After resolving a mobile bug I was finally able to test the toast notifications on mobile devices and they are indeed broken because `secure` is being set to `true`.

After a few renewed attempts to figure out how to configure remix-serve for development mode I found this comment by the remix staff engineer that suggests remix-serve isnt really intended to run in development mode (which makes sense, thats what `remix dev` is for) https://github.com/remix-run/remix/issues/8456#issuecomment-1885166316

I don't see a reason to use development mode here, given that we're not using HMR and our dockerfile is setup to install remix with prod dependencies. It is controlling this `secure` value but that's incidental - I think instead of trying to do `remix-serve` in dev mode we can just pass in a different environment variable to control the secure cookie value.

packages/frontend/app/lib/message.server.ts
```diff
@@ -20,10 +21,9 @@ export const messageStorage = createCookieSessionStorage<
     sameSite: 'lax',
     maxAge: ONE_MINUTE_IN_S,
     secrets: ['MY_SUPER_SECRET_TOKEN'],
-    secure:
-      process.env.COOKIE_SECURE === undefined
-        ? true
-        : ['true', 't', '1'].includes(process.env.COOKIE_SECURE.toLowerCase())
+    secure: process.env.ENABLE_INSECURE_MESSAGE_COOKIE
+      ? !parseBool(process.env.ENABLE_INSECURE_MESSAGE_COOKIE)
+      : process.env.NODE_ENV === 'production'
```
BlairCurrey, line 26: basically, you can pass in `ENABLE_INSECURE_MESSAGE_COOKIE=true` to set this to false otherwise it uses `NODE_ENV` like we currently do.

this means it's secure by default and when running in dev mode we dont need to set `ENABLE_INSECURE_MESSAGE_COOKIE`

for good measure here's a state table:

| ENABLE_INSECURE_MESSAGE_COOKIE | NODE_ENV      | Secure |
| ------------------------------ | ------------- | ------ |
| true                           | any           | false  |
| false                          | any           | true   |
| undefined                      | 'production'  | true   |
| undefined                      | 'development' | false  |
| undefined                      | undefined     | true   |

localenv/mock-account-servicing-entity/Dockerfile
BlairCurrey: added an insecure cookie flag here instead of running in dev mode: https://github.com/interledger/rafiki/pull/2349/commits/2db091cc06f3773a457a0709a7ea533b69cdd3d3

packages/frontend/Dockerfile
```diff
@@ -56,4 +56,5 @@ COPY --from=builder /home/rafiki/packages/frontend/build ./packages/frontend/bui
 COPY --from=builder /home/rafiki/packages/frontend/public ./packages/frontend/public
 
 WORKDIR /home/rafiki/packages/frontend
-CMD ["sh", "./node_modules/.bin/remix-serve", "build"]
\ No newline at end of file
+
+CMD ["sh", "-c", "NODE_ENV=production ENABLE_INSECURE_MESSAGE_COOKIE=true ./node_modules/.bin/remix-serve ./build/index.js"]
```
raducristianpopa, line 60: I think the `ENABLE_INSECURE_MESSAGE_COOKIE` should be passed as a build argument in the compose file. `ENABLE_INSECURE_MESSAGE_COOKIE` will be `true` when starting the frontend application every time (for everyone that is going to use the Rafiki's frontend image), since we use this specific Dockerfile to build the image.

https://docs.docker.com/compose/compose-file/compose-file-v3/#args

```suggestion
ARG ENABLE_INSECURE_MESSAGE_COOKIE=false # default value if the argument is not present at build time

ENV NODE_ENV=production
ENV ENABLE_INSECURE_MESSAGE_COOKIE=$ENABLE_INSECURE_MESSAGE_COOKIE

CMD ["sh", "-c", "./node_modules/.bin/remix-serve ./build/index.js"]
```

I experimented with this method and it appears to be effective for me. I would greatly appreciate any additional feedback on it to see if it works for everyone since I'm not entirely sure about this approach.

raducristianpopa: The above approach might not even be necessary, we can just get rid of the inline environment variables:

```suggestion
CMD ["sh", "-c", "./node_modules/.bin/remix-serve ./build/index.js"]
```

openapi/resource-server.yaml
```diff
@@ -31,7 +31,7 @@ servers:
   - url: 'https://ilp.rafiki.money'
     description: 'Server for wallet address subresources (ie https://ilp.rafiki.money/alice)'
   - url: 'https://ilp.rafiki.money/.well-known/pay'
-    description: 'Server for when the wallet address has no pathname (ie https://ilp.rafiki.money)'
+    description: 'Server for when Payment Pointer has no pathname (ie https://ilp.rafiki.money)'
```
mkurapov, line 34: is this an old version maybe?

packages/frontend/package.json
```diff
@@ -32,7 +32,7 @@
     "@types/react": "^18.2.48",
     "@types/react-dom": "^18.2.18",
     "@types/uuid": "^9.0.7",
-    "autoprefixer": "^10.4.17",
+    "autoprefixer": "^10.4.16",
```
mkurapov, line 35: package went down in version?

openapi/resource-server.yaml
mkurapov: I think just need to remerge main in?

BlairCurrey: good catch... lost it on a reverted merge. will fix

BlairCurrey: fixed

mkurapov (approved):

raducristianpopa (approved):

PR #2359, opened 2024-01-24
chore(backend): upgrade ilp-protocol-stream

BlairCurrey:

- upgrades backend devDependency `ilp-protocol-stream` from `^2.7.1` to `^2.7.2-alpha.2`

## Context

Renovate didn't pick it up because of the `alpha` and I noticed it because it has a very old and deprecated uuid sub dependancy that `pnpm install` warns about. It is only used in one test, hence the `devDependency`. The other ilp packages that use it under the hood are using this latest alpha version.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -31,7 +31,7 @@
     "@types/uuid": "^9.0.7",
     "cross-fetch": "^4.0.0",
     "ilp-protocol-ildcp": "^2.2.3",
-    "ilp-protocol-stream": "^2.7.1",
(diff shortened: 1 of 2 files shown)


sabineschaller (approved):
Good catch! Can you check whether we use any other packages from that monorepo that have not been updated?

BlairCurrey:
> Good catch! Can you check whether we use any other packages from that monorepo that have not been updated?

Will do. looks like there are a few.

Issue #1438, opened 2023-05-16
Use tigerbeetle production release


sabineschaller:
The first production release for TigerBeetle is `v0.15.3`. 
See https://npm.io/package/tigerbeetle-node

<img width="737" alt="image" src="https://github.com/interledger/rafiki/assets/1992108/b46eb509-e80b-4ceb-a493-9bb8be27cdb7">

<img width="740" alt="image" src="https://github.com/interledger/rafiki/assets/1992108/d89ecfb1-719a-491c-9c9c-bd0f9daa8821">

koekiebox:
# PR for PROD release update:
https://github.com/interledger/rafiki/pull/2653

sabineschaller:
Accomplished with #2653

PR #2364, opened 2024-01-25
chore: upgrade to newest tigerbeetle version

### LINKED ISSUE #1112: Fix tigerbeetle transfer timeouts

sabineschaller:

- upgrade to newest tigerbeetle version such that the changes aren't big as soon as we get the TB production release

## Context

- progress towards #1438
- fixes #1112 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/accounting/service.ts
+++ b/packages/backend/src/accounting/service.ts
@@ -36,7 +36,7 @@ export interface Deposit {
 }
 
 export interface Withdrawal extends Deposit {
-  timeout?: bigint
+  timeout?: number
 }
 
 export interface TransferOptions {
@@ -44,7 +44,7 @@ export interface TransferOptions {
   destinationAccount: LiquidityAccount
   sourceAmount: bigint
   destinationAmount?: bigint
-  timeout: bigint
+  timeout: number
 }
 
 export interface Transaction {
--- a/packages/backend/src/accounting/tigerbeetle/accounts.ts
+++ b/packages/backend/src/accounting/tigerbeetle/accounts.ts
@@ -8,8 +8,6 @@ import { ServiceDependencies, TigerbeetleAccountCode } from './service'
 import { TigerbeetleCreateAccountError } from './errors'
 import { AccountId, toTigerbeetleId } from './utils'
 
-const ACCOUNT_RESERVED = Buffer.alloc(48)
-
 export interface CreateAccountOptions {
   id: AccountId
   ledger: number
@@ -23,8 +21,10 @@ export async function createAccounts(
   const errors = await deps.tigerbeetle.createAccounts(
     accounts.map((account) => ({
       id: toTigerbeetleId(account.id),
-      user_data: 0n,
-      reserved: ACCOUNT_RESERVED,
+      user_data_32: 0,
+      user_data_64: 0n,
+      user_data_128: 0n,
+      reserved: 0,
       ledger: account.ledger,
       code: account.code,
       // Credit and debit accounts can both send and receive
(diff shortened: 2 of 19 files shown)


packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -23,8 +21,10 @@ export async function createAccounts(
   const errors = await deps.tigerbeetle.createAccounts(
     accounts.map((account) => ({
       id: toTigerbeetleId(account.id),
-      user_data: 0n,
-      reserved: ACCOUNT_RESERVED,
+      user_data_32: 0,
+      user_data_64: 0n,
+      user_data_128: 0n,
```
sabineschaller, line 26: These are now all required.

sabineschaller:
Talking to TB team regarding failing tests. I've tried a couple of things including `setInterval` but TB never throws the `TransferExpired` error for me.

packages/backend/src/accounting/service.ts
```diff
@@ -36,15 +36,15 @@ export interface Deposit {
...
+  timeout?: number
 }
 
 export interface TransferOptions {
   sourceAccount: LiquidityAccount
   destinationAccount: LiquidityAccount
   sourceAmount: bigint
   destinationAmount?: bigint
-  timeout: bigint
+  timeout: number
```
sabineschaller, line 47: Tigerbeetle now uses seconds (instead of nanoseconds). Hence, I changed everything to seconds and number. Within the psql accounting, seconds are transformed into milliseconds and bigints. We can discuss whether we should also change that to seconds.

BlairCurrey: From a kind of detached view, I don't think we need to update psql accounting to be seconds. Since we've abstracted the accounting service to work with tigerbeetle or postgres I think we should judge the postgres implementation based on the inherent merits of the time format rather than relative to what tigerbeetle is doing. I can see why we'd be consistent to start but at this point I wouldn't bother personally.

And looking more closely at the code the `timeoutMs` detail does seem to be fully contained to the psql service. Ultimately drives the transfer `state` and `expiresAt` - so its not like there is some environment variable or display field that means one thing if using tigerbeetle and another if using postgres.

BlairCurrey (approved):

packages/backend/src/accounting/tigerbeetle/transfers.ts
```diff
@@ -62,10 +62,11 @@ export async function createTransfers(
       id: 0n,
       debit_account_id: 0n,
       credit_account_id: 0n,
-      user_data: 0n,
-      reserved: 0n,
```
njlie, line 66: Is `reserved` no longer required here?

njlie (approved):

packages/backend/src/accounting/service.ts
sabineschaller: I made the base time unit seconds because you can easily convert that to milliseconds but you can't go from milliseconds to seconds without rounding.

packages/backend/src/accounting/tigerbeetle/transfers.ts
sabineschaller: No, it was dropped, see [here](https://docs.tigerbeetle.com/reference/operations/create_transfers). There is only a `reserved_flag` now, but if I read that correctly, that defaults to 0.

Issue #2229, opened 2023-11-28
Enable multi-hop ILP payments


mkurapov:
## Context

We want to see if we are able to send ILP packets across different peers running Rafiki.

Say we have Rafiki instances A, B, C, where A is peered with B, and B is peered with C. Since they are all running ILP connectors, we should theoretically be able to send a payment from A to C.

It looks like currently this isn't supported. Whenever we try a payment from A to C, we end up not being able to a) resolve `state.streamDestination` in `createStreamAddressMiddleware` and by extension, fetch the outgoing account in `createAccountMiddleware` (which should end up being the peer of the next hop).

 However, we already have files that seem to do the functionality that we want in `packages/backend/src/payment-method/ilp/connector/ilp-routing`. There is a README in the folder that might be helpful.

Additionally, we have some commented out code in the connector that *should* be able to enable the routing:

https://github.com/interledger/rafiki/blob/48fb07a7e587aa6d54bb88c2db5cec773a9cf0b7/packages/backend/src/payment-method/ilp/connector/index.ts#L42-L66

This issue will be about making this code work to support multi-hop ILP payments. We should test this by spinning up another Rafiki instance in the local playground, peering with one of the default instances only (ie. happy-life-bank), and sending a payment from cloud-nine to the new instance. A branch with a 3-Rafiki instance localplayground is available here: https://github.com/interledger/rafiki/compare/main...rp--multi-hop-setup

Resources

https://github.com/interledger/rafiki/blob/7f332c725a9050fcd70566ea0e8b310a86578f18/packages/connector/src/connector.ts

https://interledger.org/developers/rfcs/ilp-addresses/
https://github.dev/interledger/rafiki/tree/7f332c725a9050fcd70566ea0e8b310a86578f18

## Todos
- [x] #2742

PR #2369, opened 2024-01-25
chore: update ilp dependencies

BlairCurrey:

- removes unused `ilp-protocol-ccp` dependency
- updates `ilp-protocol-ildcp` dependency 

## Context

https://github.com/interledger/rafiki/pull/2359#pullrequestreview-1843209797

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/package.json
```diff
@@ -30,7 +30,6 @@
     "@types/tmp": "^0.2.6",
     "@types/uuid": "^9.0.7",
     "cross-fetch": "^4.0.0",
-    "ilp-protocol-ildcp": "^2.2.3",
```
BlairCurrey, line 33: We had this in both `dependencies` and `devDependencies`. Not sure why but we shouldn't need it as both so I removed the dev one.

packages/backend/src/payment-method/ilp/connector/core/factories/ccp-packets.ts
BlairCurrey, line 1: I removed this (and the `ilp-protocol-ccp` dependency) because its not used anywhere. Unless there is some reason to keep it ... ?

mkurapov: I think this commented out code was kept in all this time because of
- https://github.com/interledger/rafiki/issues/2229

since `ilp-protocol-ccp` deals with routing. Since we have more commented out code that shows how routing would work, we should either: a) not delete this package/code for now b) delete all of the routing code in a single PR so it can be easily referenced when we _do_ start #2229

BlairCurrey: thanks for the context. I will keep it for now.

mkurapov (approved):

Issue #418, opened 2022-08-16
Declarative Ledgers and Accounts for TB


matdehaast:
Explore adding the ability to declaretivly add ledgers and liquidity accounts to tigerbeetle.

sabineschaller:
@matdehaast Can you elaborate on this? What exactly do you want to do?

matdehaast:
So this was allowing the ability to declare the default ledgers and accounts for TB via a file and then Rafiki would auto setup it up for you. This meant you wouldn't have to use the API on startup to configure it.

Example
```
ledgers:
  - id: 123
    name: "LocalUSD"
    asset: "USD"
    scale: 2
  - id: 124
    name: "LocalZAR"
    asset: "ZAR"
    scale: 2

accounts:
  - id: "46d4a2bd-e29b-4a63-9aa8-7990776c714e"
    ledger_id: 124
    code: 2
    linked: false
    debits_must_not_exceed_credits: false
    credits_must_not_exceed_debits: false
  - id: "c54aa8a9-b303-4b75-9bf4-203a9cf15f68"
    ledger_id: 123
    code: 2
    linked: true
    debits_must_not_exceed_credits: false
    credits_must_not_exceed_debits: false
  - id: "29e5aa54-0dc8-4e92-a9dd-b99a373525f0"
    ledger_id: 123
    code: 2
    linked: false
    debits_must_not_exceed_credits: false
    credits_must_not_exceed_debits: false

```

mkurapov:
@sabineschaller ^ this makes sense, but I'm wondering if it would be applicable for us now because we also support PSQL accounting, since (IMO) the seeding should be DB-agnostic.
Also, I don't know whether it would make sense to create TB accounts that aren't tied to `backend` accounts like Assets, Peers, Incoming Payments and Outgoing Payments

PR #2380, opened 2024-01-30
feat: integration tests

### LINKED ISSUE #2470: Test P2p flow
Use the apollo client to implement the more basic Peer-to-Peer Payment flow from Bruno. Should prove to be the simpler flow to implement given some complications in the docker environment relating to the open payments client, running tests from host machine, and services in docker using proper domain names.

### LINKED ISSUE #2471: Test Open payments flow
Jest tests using the test mock account servicing entities and open payments client to recreate the Bruno Open Payments flow.

TODO:
- [ ] #2474
- [ ] #2475

### LINKED ISSUE #2474: Tests through Grant Request Outgoing Payment
Recreate the requests from the Bruno Open Payments collection up through the grant request for outgoing payment. There are a fair number of configuration tweaks that will probably need to be ironed out in doing so, as well as potentially adding more functionality to the integration server for webhook handling (and similar).

### LINKED ISSUE #2475: Test through Create Outgoing Payment (including mocking consent screen)
Should implement tests for the Bruno Open Payment flow from Continuation through the end. Will including mocking the consent screen. May also want to modify the Grant Request Outgoing Payment to do continuation polling (rather than redirect for the `interact_ref`) if not doing so already.

BlairCurrey:

- extracts common MASE functionality (`AccountProvider`, `setupFromSeed`, etc.) into new `mock-account-servicing-lib`
- new test package in `rafiki/test/integration` containing test environment configuration and integration test akin to Bruno Open Payments and P2P flows.
- new `./test/integration/scripts/run-tests.sh` command to launch test environment and run tests (via `run-tests` `package.json` script)

## Context

fixes #2470
fixes #2471
fixes #2474
fixes #2475

## Test Running Instructions:
Currently there are port conflicts with localenv, so you will need to spin that down first. Fixing this is captured here: https://github.com/interledger/rafiki/issues/2529

To run for the first time:

- add `127.0.0.1 host.docker.internal` to `/etc/hosts` file
- `pnpm --filter mock-account-servicing-lib build`
- `pnpm install`
- `pnpm --filter integration run-tests -b`

Then normally just run:

`pnpm --filter integration run-tests -b` (or without -b if you dont need to rebuild images) 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


pnpm-workspace.yaml
```diff
@@ -2,4 +2,4 @@ packages:
   - 'packages/*'
   - 'localenv/mock-account-servicing-entity'
   # exclude packages that are inside test directories
-  - '!**/test/**'
```
BlairCurrey, line 5: Was not excluding anything from what I can tell (using `pnpm ls --depth -1 -r` to get packages). Removed to accommodate the `test/integration` workspace pkg but can rework this if it's actually doing something.

test/integration/lib/MockASE.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  private async initAsync() {
+    await setupFromSeed(this.config, this.apolloClient, this.accounts)
+
+    this.opClient = await createAuthenticatedClient({
+      privateKey: this.config.key,
+      keyId: this.config.keyId,
+      walletAddressUrl: this.config.walletAddressUrl
+    })
```
BlairCurrey, line 43: Not quite right. Getting an invalid signature error from auth's grant initialization middleware. 

The `keyId` and `walletAddressUrl` are coming from uncommitted environment variables. They are `rafiki` and `https://cloud-nine-wallet-test-backend/.well-known/pay` (for the coud nine client) respectively. The key is loaded from this file (same key as from localenv): https://github.com/interledger/rafiki/pull/2380/files#diff-059e91a06656fac95dd3ee765c22dc8b3ae16d4e41074ce510a111bce7bd3f59

I'm not sure the `keyId` is right but using one of the `keyId`'s formed in `setupFromSeed` (`keyid-{account.number}`) gave me `invalid client` which happens earlier on in the auth middleware where it cant find the key by the `client` and `keyId`.

BlairCurrey: This ended up being fine, but I needed to use the same private key file everywhere (here in the test and in each backend service).

packages/auth/src/app.ts
```diff
@@ -322,9 +322,7 @@ export class App {
           maxAge: 60 * 1000,
           signed: true
         },
-        // Only accepts Middleware<DefaultState, DefaultContext> for some reason, this.koa is Middleware<DefaultState, AppContext>
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        koa as any
```
BlairCurrey, line 327: Just didnt need to do this anymore... this might have been where it was fixed https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65100

localenv/mock-account-servicing-entity/app/entry.server.tsx
```diff
@@ -27,7 +30,10 @@ async function callWithRetry(fn: () => any, depth = 0): Promise<void> {
 }
 
 if (!global.__seeded) {
-  callWithRetry(runSeed)
+  callWithRetry(async () => {
+    console.log('setting up from seed...')
+    return setupFromSeed(CONFIG, apolloClient, mockAccounts, { debug: true })
+  })
```
BlairCurrey, line 36: same as it was before, but inlined `runSeed`. I did this because everything else from `'./lib/run_seed.server'` was moved to the new shared `mock-account-servicing-lib` and `runSeed` was the only thing left.

packages/mock-account-servicing-lib/src/setupFromSeed.ts
```diff
@@ -1,25 +1,33 @@
...
+  const {
+    createAsset,
+    depositAssetLiquidity,
+    setFee,
+    createPeer,
+    depositPeerLiquidity,
+    createAutoPeer,
+    createWalletAddress,
+    createWalletAddressKey
+  } = createRequesters(apolloClient, debug)
```
BlairCurrey, line 29: This was kind of a tough line to draw between "what should be in mock-account-servicing-lib" and "what should be in the mock account servicing entites (in localenv, testenv)".

I added these requests to the lib because `setupFromSeed` depends on them and `setupFromSeed` is needed by both testenv and localenv. However, I dont think defining all requests in general from the library makes much sense (each MASE might want to call them differently for example). So I just added the ones `setupFromSeed` needed and kept them internal to this package. I did `createRequesters` instead of exporting each request individually because this is the only thing that needs them (nothing that needs to import a single request, nor are they exported from the package) and so that we don't have to pass in `apolloClient` to each one.

packages/mock-account-servicing-lib/src/setupFromSeed.ts
```diff
@@ -35,7 +43,9 @@ export async function setupFromSeed(config: Config): Promise<void> {
     )
 
     assets[code] = asset
-    console.log(JSON.stringify({ asset, initialLiquidity }, null, 2))
+    if (debug) {
+      console.log(JSON.stringify({ asset, initialLiquidity }, null, 2))
+    }
```
BlairCurrey, line 48: the setup from seed and requests just log everything in our local environment but this made less sense for testing. This debug option was a kinda quick solution - perhaps a pino debug log would make more sense?

scripts/run-integration-tests.sh
BlairCurrey, line 1: This could change or go away depending on how [this issue](https://github.com/interledger/rafiki/issues/2472) is resolved so I think the important thing at this point is simply that it works to launch the tests

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+# This script runs the integration tests. It starts the test environment, runs the tests,
+# saves the container logs to a file, and stops the containers.
+# Usage:
+#   ./script.sh            # Run the script with default options
+#   ./script.sh --build    # Re-build the docker images (-b or --build)
+
+timeout=30
+log_file="/tmp/rafiki_integration_logs.txt"
```
BlairCurrey, line 10: Container logs get saved here for each run. Useful for developing tests but should not normally need.

I could change to local tmp folder instead but didn't have much preference.

test/integration/testenv/cloud-nine-wallet/.env.example
BlairCurrey, line 1: Some general notes on configuration/env vars: I wanted to scope env variables to a mock ase instead of having everything in 1 global configuration file (`CLOUD_NINE_WALLET_ADDRESS`, `HAPPY_LIFE_BANK_WALLET_ADDRESS`). Not sure if we'll ever need to, but it should make adding an additional MASE easier than if we bake the assumption that there are only two with specific prefixes.

Unfortunately There is some duplication with the env vars here (used in tests) and docker-compose. I tried to share env vars by loading from this env file in the compose but it did not work as expected when merging on the docker-compose files. 

This part is kinda in flux depending on this issue https://github.com/interledger/rafiki/issues/2472 so I would like to focus on if this works. For example, this might all be moved into a docker-compose if tests are moved into a docker container (which is a pro for that option).

BlairCurrey:
We have used both "integration" and "end-to-end" tests to refer to these tests and I would like to align on this. What do you all think? 

I have come to the opinion these are integration tests so I've used that terms in these changes but I've went back and forth. These are some of the most loaded terms though, so here are the basic definitions I'm working from.

- integration: tests how parts work together, but not the entire system
- end-to-end: tests from the user’s perspective against the full system (maybe some mocking where required, but very minimal)

Here is a very simplified graph representing layers of Rafiki Proper:

```mermaid
graph TD;
    A[Database] --> C[Backend];
    C --> D[Frontend];
```

And one for these tests:

```mermaid
graph TD;
    A[Database] --> C[Backend];
    C --> E[OP Client];
```

The frontend will never be part of these tests without a major refactor to support some sort of browser agent. They just aren't in scope.

Despite that, I think its tempting to call these an end-to-end test because the open payments client is arguable as much or more of an end user than frontend. It's as much or more because it's a primary way of interacting with Rafiki and because it has more functionality. In fact, frontend could not conduct the required test flows. However, and maybe this is too much of a technicality, the open payments client isnt actually part of rafiki, so we're really just testing rafiki's backend (and auth) and database.

The hydra authentication is another wrinkle though. In theory I think using hydra it out of scope as well, but it might be required to run the tests. I suppose even in the case that it is, it's not going to be a full test - it would use the hydra api, which will test the usage of hydra itself, but not our actual login flow.

test/integration/integration.test.ts
```diff
@@ -0,0 +1,533 @@
...
+                type: 'incoming-payment',
+                actions: ['create', 'read', 'list', 'complete']
+              }
+            ]
+          }
+        }
+      )
+
+      assert(!isPendingGrant(grant))
+      accessToken = grant.access_token.value
```
BlairCurrey, line 116: Note that these tests dont have too much in the way of assertions. I am open to feedback on more meaningful assertions if you see an opportunity, but the flow completing is really the whole test. 

Most of the assertions are based around making the code typesafe, providing some minimal level of validation that the response isn't bad (which the subsequent test would catch anyways), or verifying we received webhooks. Many of these really just make it easier to see where things went wrong rather than ensuring everything works - whether or not it works really boils down to if they all pass and they wont if each step doesn't complete without error. Our service unit tests are really the place to make more detailed assertions about what we expect to see. We can validate some additional things here like receiving 

In this way its actually kind of misleading to have separate tests per request but that's something I elaborate on and will tackle in another issue (https://github.com/interledger/rafiki/issues/2529).

test/integration/integration.test.ts
```diff
@@ -0,0 +1,533 @@
...
+          walletAddress: senderWalletAddressUrl.replace('http', 'https'),
+          receiver: incomingPayment.id.replace('https', 'http'),
+          method: 'ilp'
+        }
+      )
+    })
+
+    // --- GRANT CONTINUATION WITH FINISH METHOD ---
+    // TODO: Grant Continuation w/ finish in another Open Payments Flow test
+    test.skip('Grant Request Outgoing Payment', async (): Promise<void> => {
```
BlairCurrey, line 193: Currently the "Open Payments Flow" uses continuation polling but we can test both. The only thing stopping me ATM is that we'd have to duplicate all the other test steps, which can be improved and is captured here: https://github.com/interledger/rafiki/issues/2529 So I'm just leaving these alternate steps for now

test/integration/testenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
...
+    image: rafiki-backend
+    build:
+      context: ../../../..
+      dockerfile: ./packages/backend/Dockerfile
+    restart: always
+    privileged: true
+    ports:
+      - '3000:3000'
+      - '3001:3001'
+      - '3002:3002'
```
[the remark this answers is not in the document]
mkurapov: Ah, don't worry about it then in this one, just as long as you have it for the next one :+1:

test/integration/testenv/cloud-nine-wallet/.env.example
```diff
@@ -0,0 +1,7 @@
+OPEN_PAYMENTS_URL=https://host.docker.internal:3000
```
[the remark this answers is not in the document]
BlairCurrey: yeah I agree. Updated the .gitignore to not exclude the env's in `rafiki/test`.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,58 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+pnpm --filter integration testenv:compose down --volumes
+pnpm --filter integration testenv:compose up -d $build_flag
+pnpm --filter integration testenv:compose logs -f --tail=0 > "$log_file" 2>&1 &
+
+# perform health check until OK, or errors if timeout.
```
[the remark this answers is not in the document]
BlairCurrey: I'm not sure how well we can utilize it here. I see how we can add a health check to one of the services but it wont block our shell script from continuing on to run the tests which are launched here. We would need to poll that status and the only way I'm seeing how to do it doesn't seem simpler than simply polling the url here.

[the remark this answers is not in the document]
BlairCurrey: Aha... I will try that.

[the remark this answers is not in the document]
mkurapov: Also, I think we'll need to use `wget` instead of curl for the healthcheck since it's an alpine image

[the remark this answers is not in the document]
BlairCurrey: This isn't running in docker.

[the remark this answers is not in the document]
BlairCurrey: added https://github.com/interledger/rafiki/pull/2380/commits/366d02f7108f3fda419d4341de3dfd6bc3f3d678

sabineschaller:
I followed your instructions and I got 
```
> integration@1.0.0 test /Users/sabine/git/coil/rafiki/test/integration
> jest

 FAIL   integration  ./integration.test.ts
  ● Test suite failed to run

    Cannot find module 'mock-account-servicing-lib' from 'test/integration/lib/MockASE.ts'

    Require stack:
      test/integration/lib/MockASE.ts
      test/integration/integration.test.ts

      10 | import { TestConfig } from './config'
      11 |
    > 12 | /** Mock Account Servicing Entity */
         |                                  ^
      13 | export class MockASE {
      14 |   private config: TestConfig
      15 |   private apolloClient: ApolloClient<NormalizedCacheObject>

      at Resolver._throwModNotFoundError (node_modules/.pnpm/jest-resolve@29.7.0/node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (test/integration/lib/MockASE.ts:12:34)
      at Object.<anonymous> (test/integration/integration.test.ts:9:18)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        1.207 s
Ran all test suites.
/Users/sabine/git/coil/rafiki/test/integration:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  integration@1.0.0 test: `jest`
Exit status 1

> integration@1.0.0 testenv:compose /Users/sabine/git/coil/rafiki/test/integration
> docker-compose -f ./testenv/cloud-nine-wallet/docker-compose.yml -f ./testenv/happy-life-bank/docker-compose.yml -f ./testenv/docker-compose.yml "down" "--volumes"

[+] Running 7/6
 ✔ Container rafiki-test-happy-life-bank-test-backend-1    Removed                                                                                                                                    0.2s 
 ✔ Container rafiki-test-happy-life-auth-1                 Removed                                                                                                                                    0.2s 
 ✔ Container rafiki-test-cloud-nine-wallet-test-auth-1     Removed                                                                                                                                    0.2s 
 ✔ Container rafiki-test-cloud-nine-wallet-test-backend-1  Removed                                                                                                                                    0.3s 
 ✔ Container rafiki-test-shared-redis-1                    Removed                                                                                                                                    0.2s 
 ✔ Container rafiki-test-shared-database-1                 Removed                                                                                                                                    0.2s 
 ✔ Network rafiki-test_rafiki-test                         Removed      
```
😞 what am I doing wrong?

mkurapov:
@sabineschaller that's odd, even after `pnpm i` ? What happens if you `pnpm clean && pnpm i`?

BlairCurrey:
> > What happens if you `pnpm clean && pnpm i`
> 
> Let me try that.
> 
> EDIT: @mkurapov same

@sabineschaller do `pnpm --filter mock-account-servicing-lib build` first. It's a new lib which you wont have the compiled js for. Was able to reproduce by deleting the dist folder. Let me know if that works for you and I'll update the instructions. Not sure why it might have worked for Max though.

mkurapov:
@BlairCurrey @sabineschaller I forgot I also did that at one point...

sabineschaller:
Yes, works now 🙂

packages/mock-account-servicing-lib/jest.config.js
sabineschaller, line 1: Just a general question: Should we rename this to `mock-account-service-lib` or even `account-service-lib`, since actual ASEs may for instance use the seeding functionality?

packages/mock-account-servicing-lib/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "mock-account-servicing-lib",
+  "description": "Common library for mock account servicing",
```
sabineschaller, line 3:
```suggestion
  "description": "Common library for mock account service",
```

packages/mock-account-servicing-lib/src/setupFromSeed.ts
sabineschaller: I think these ones make sense to be in the library.

packages/mock-account-servicing-lib/src/setupFromSeed.ts
sabineschaller: We want to more towards proper logging so yes, please let's do pino debug logs.

pnpm-workspace.yaml
sabineschaller: this was probably a typo and should have been `tests`

sabineschaller (commented):
I think this is a great start and we should just merge it. I would like to add tests, though, that check that the webhook events are fired correctly and also that the Admin APIs behave correctly.

scripts/run-integration-tests.sh
mkurapov: I think local `tmp` folder is better IMO, then I can just see it right there.

Also, I think we can move this script inside `test/integration`?

packages/mock-account-servicing-lib/src/WebhookEventType.ts
```diff
@@ -0,0 +1,11 @@
+export enum WebhookEventType {
```
mkurapov, line 1: we can also just throw this in `types.ts`

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,30 @@
...
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
+
+pnpm --filter integration testenv:compose down --volumes
+pnpm --filter integration testenv:compose up -d --wait $build_flag
+pnpm --filter integration testenv:compose logs -f --tail=0 > "$log_file" 2>&1 &
+pnpm --filter integration test
```
mkurapov, line 29: I think it would be nice to prevent `pnpm --filter integration test` from running at all if we had errors starting up in `pnpm --filter integration testenv:compose up -d --wait $build_flag`

mkurapov (commented):
~For some reason I'm getting errors running now:~ 
resolved now by having a later version of node (I was on an old version, 18.14)

```

  ● Integration tests › Open Payments Flow › Continuation Request

    TypeError: response.headers.getSetCookie is not a function

      31 | export function parseCookies(response: Response) {
      32 |   return response.headers
    > 33 |     .getSetCookie()
         |      ^
      34 |     .map((header) => {
      35 |       const parts = header.split(';')
      36 |       const cookiePart = parts[0]

      at getSetCookie (test/integration/lib/utils.ts:33:6)
      at Object.<anonymous> (test/integration/integration.test.ts:343:34)
```

BlairCurrey:
> I think this is a great start and we should just merge it. I would like to add tests, though, that check that the webhook events are fired correctly and also that the Admin APIs behave correctly.

Can you elaborate a bit on "fired correctly"? For most (all?) of the webhooks in these flows we are checking that we received them and identify them by type... were you thinking about checking the data on them more specifically?

mkurapov:
To add to the webhook discussion, we can also check the `sentAmount` on outgoing payment or check the `receivedAmount` on the incoming one to check that the payment did in fact go through

BlairCurrey:
> To add to the webhook discussion, we can also check the `sentAmount` on outgoing payment or check the `receivedAmount` on the incoming one to check that the payment did in fact go through

Yeah that seems like a good one to add and I'm open to suggestions on more checks. I don't think the assertions I've added cover all the assertions we should make.

I just wanted to keep these tests focused on answering the question, "do the payments flows work?" and not "do the *(webhook | open payments) response meet their full specifications?" because that question is better answered by unit tests. For example, adding a new field (ie metadata on outgoing payment), or changing the type of a field, or removing a field, etc. should not fail these tests unless it actually causes the flow to fail or to completely incorrectly. I mean we can have a wider discussion on that but that's the spirit that's guiding me here.

packages/mock-account-servicing-lib/jest.config.js
BlairCurrey: I did not think about potential use by real ASEs but I agree `setupFromSeed` could be useful.  However, it also includes the `AccountProvider` so I feel like we would be implicitly suggesting it's for real ASE's (which it probably shouldnt be?). So maybe `mock-account-service-lib`?

But anything to make it shorter would be great 🤣 . I'm not married to the name, mostly just wanted to draw a clear relationship to `mock-account-servicing-entity`.

.github/workflows/lint_test_build.yml
```diff
@@ -84,6 +84,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/workflows/rafiki/env-setup
+      - run: pnpm --filter mock-account-servicing-entity build:deps
```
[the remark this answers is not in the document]
BlairCurrey: Yes - a typescript error in `mock-account-servicing-lib` surfaces in this command. We are already doing `pnpm --filter mock-account-servicing-entity build` 2 steps down. I guess we can just switch the `typecheck` and `build` command instead of adding this new step.

For context, I did this after I saw CI fail the typecheck command because the `mock-account-servicing-lib` wasnt built. But bulding `mock-account-servicing-entity` first covers that.

scripts/run-integration-tests.sh
BlairCurrey: yeah makes sense... easier to have it locally i guess. changed to a test/integration/tmp and moved inside test/integration with a package.json script to start it

packages/mock-account-servicing-lib/src/WebhookEventType.ts
BlairCurrey: moved

scripts/run-integration-tests.sh
BlairCurrey: makes sense. added https://github.com/interledger/rafiki/pull/2380/commits/bb98b70f7402fee1d098d3cf9f0ae85cde5c0294

packages/mock-account-servicing-lib/src/setupFromSeed.ts
BlairCurrey: I changed to pino but these logs were basically unreadable (lots of fairly large objects) so I added [pino-pretty](https://www.npmjs.com/package/pino-pretty) which helps (looks similar to how it was). It's just really tedious to digest any of it without `pino-pretty` IMO.

Without pino-pretty:

Screenshot 2024-03-16 at 9 49 28 PM

With pino-pretty:
Screenshot 2024-03-16 at 10 08 23 PM

For whatever it's worth, this is something we can do elsewhere (backend, auth, etc), if we wanted. It just probably shouldn't be used in production. If this wont be used in prod we can hardcode it on, but currently it's controlled by the `pinoPretty` option (false by default) because `mock-account-servicing-entity` actually runs with a `NODE_ENV` in production due to some expectations by remix, if I recall correctly, and I wasn't sure if reading the env in the library made total sense anyways.

packages/mock-account-servicing-lib/jest.config.js
sabineschaller: Then let's call it `mcok-account-service-lib` and re-evaluate in the future if we want to pull some code into a lib that actual ASEs can use. I think Matt had asked about seeding from file before.

sabineschaller:
> @sabineschaller Can you elaborate a bit on "fired correctly"? For most (all?) of the webhooks in these flows we are checking that we received them and identify them by type... were you thinking about checking the data on them more specifically?

Sorry, for some reason, I didn't see your webhook spy in the code. Must have been tired or distracted or both. I think the webhook tests are fine.

>To add to the webhook discussion, we can also check the `sentAmount` on outgoing payment or check the `receivedAmount` on the incoming one to check that the payment did in fact go through

Is that necessary? Should we not trust our unit tests that these things have happened before the webhook event was sent?

sabineschaller:
Once we have renamed the lib (because I think `mock-account-servicing-lib` doesn't make as much sense as `mock-account-service-lib`) this is a go from my side.

packages/mock-account-servicing-lib/jest.config.js
sabineschaller: Found it:
- https://github.com/interledger/rafiki/issues/418

packages/mock-account-service-lib/src/AccountProvider.ts
```diff
@@ -0,0 +1,271 @@
+interface Account {
```
mkurapov, line 1: nitpick, but should we lower pascal/kebab case these file names?

localenv/mock-account-servicing-entity/app/lib/requesters.ts
```diff
@@ -1,169 +1,12 @@
 import { gql } from '@apollo/client'
```
mkurapov, line 1: is this file necessary at all now given we have `packages/mock-account-service-lib/src/requesters.ts`?

test/integration/integration.test.ts
```diff
@@ -0,0 +1,535 @@
...
+            value: '100',
+            assetCode: receiverWalletAddress.assetCode,
+            assetScale: receiverWalletAddress.assetScale
+          },
+          metadata: { description: 'Free Money!' },
+          expiresAt: tomorrow.toISOString()
+        }
+      )
+
+      // Delay gives time for webhook to be received
```
mkurapov, line 152: could we also poll here instead?

mkurapov:
> Is that necessary? Should we not trust our unit tests that these things have happened before the webhook event was sent?

I think checking the OutgoingPayment sentAmount is a good extra check to have alongside making sure the ASEs received the corresponding webhooks

mkurapov (commented):
Non-blocking comments, other than https://github.com/interledger/rafiki/pull/2380#pullrequestreview-1935414130

packages/mock-account-service-lib/src/AccountProvider.ts
BlairCurrey: No thats totally fare. Such as `accountProvider` or `account-provider`? Was accustomed to this way for classes but I see its not our pattern. Looks like we are kinda inconsistent about `accountProvider` vs `account-provider`... renamed this to `accountProvider` (and similar in the `integration` package).

mkurapov: we are pretty inconsistent, so `accountProvider` is fine for now

localenv/mock-account-servicing-entity/app/lib/requesters.ts
BlairCurrey: Yes, it has requests not in the lib and the ones in the lib are only used internally (in setupFromSeed). Commented more on this here: https://github.com/interledger/rafiki/pull/2380/files/4cd2646220d3024331afd6b933f3ee1b52dab6d2#r1521562424

If we want, we can move all the requests to the library. Could perhaps move the entire apollo client and admin client (from the integreation package) in and just expose the admin client. My initial assumption was the caller may want to form their own requests, such as different fields, or return the whole response vs. just the resource data. But none of those are necessarily blockers.

Thinking about it more, having an adminClient could make sense. It could even be useful outside the context of the tests and `mock-account-servicing-entity`. If we want to do that I would like to create a new task for it though.

packages/mock-account-service-lib/src/AccountProvider.ts
BlairCurrey: i dont really have a strong preference for whatever its worth. This is kinda more what i am used to https://github.com/airbnb/javascript#naming--filename-matches-export but see snake/kebab is recommended by google's style guide https://google.github.io/styleguide/jsguide.html#file-name. I wouldn't be against some filename cleanup.

test/integration/lib/integration-server.ts
```diff
@@ -0,0 +1,182 @@
...
+          base,
+          rates: {
+            ...this.config.seed.rates[base as string]
+          }
+        }
+      }
+    })
+
+    this.server = this.app.listen(port, () => {
+      console.log(`Integration server listening on port ${port}`)
```
mkurapov, line 58: I think for the next PR we should just make this a pino logger and set it to silent during tests

mkurapov (approved):
🚀 LGTM 🚀 
Good to merge, can do any follow ups later

test/integration/lib/integration-server.ts
BlairCurrey: yeah we probably dont even need these tbh. will add it to the cleanup to address one way or another.

test/integration/integration.test.ts
mkurapov: Makes sense, we could even decrease the polling frequency (although its very much not a big deal)

localenv/mock-account-servicing-entity/app/lib/requesters.ts
mkurapov: I agree with the fact that different implementations might require different fields/requests, its not difficult to add or change these requests at all anyway. Was just trying to see how much can be taken out of the MASE.

In any case, I think the MASE would benefit from a nice overhaul/update anyway, so we shouldn't spend too much time on it until we have a proper plan for where we want to take it.

PR #2394, opened 2024-02-05
chore: update test memory leaks

### LINKED ISSUE #1094: Gracefully stop individual TB testcontainers
When upgrading Tigerbeetle to the LSM tree version (plus node client to 0.11.11), github action backend tests would fail when stopping the individual tigerbeetle testcontainers for accounting and asset tests within the `afterAll`. A quick fix was to just remove that line and let the action itself stop all running containers once finished. 

We should find a way to reenable gracefully stopping these containers, potentially using jest [testEnvironment](https://jestjs.io/docs/configuration#testenvironment-string).

### LINKED ISSUE #2393: Fix memory leaks in tests
## Context
When trying to debug remaining open handles in Jest tests reviewing [this PR](https://github.com/interledger/rafiki/pull/2373), I was unable to run jest tests serially via `--runInBand`/`--detectOpenHandles` because they were running out of memory due to a number of possible memory leaks. 

There is an experimental jest argument: `--detectLeaks` that will fail a test if it detects that it is leaking memory: 
```
max ~/interledger/rafiki/packages/backend [main] $ node ../../node_modules/jest/bin/jest.js --runInBand --detectLeaks --detectOpenHandles
{"level":30,"time":1707131063083,"pid":12851,"hostname":"max.local","msg":"creating knex"}
 FAIL   backend  src/open_payments/payment/outgoing/service.test.ts
  ● Test suite failed to run

    EXPERIMENTAL FEATURE!
    Your test suite is leaking memory. Please ensure all references are cleaned.

    There is a number of things that can leak memory:
      - Async operations that have not finished (e.g. fs.readFile).
      - Timers not properly mocked (e.g. setInterval, setTimeout).
      - Keeping references to the global scope.

      at onResult (node_modules/.pnpm/@jest+core@29.7.0/node_modules/@jest/core/build/TestScheduler.js:150:18)
      at node_modules/.pnpm/@jest+core@29.7.0/node_modules/@jest/core/build/TestScheduler.js:254:19
      at node_modules/.pnpm/emittery@0.13.1/node_modules/emittery/index.js:363:13
          at Array.map (<anonymous>)
      at Emittery.emit (node_modules/.pnpm/emittery@0.13.1/node_modules/emittery/index.js:361:23)

 FAIL   backend  src/open_payments/payment/combined/service.test.ts
  ● Test suite failed to run

    EXPERIMENTAL FEATURE!
    Your test suite is leaking memory. Please ensure all references are cleaned.

    There is a number of things that can leak memory:
      - Async operations that have not finished (e.g. fs.readFile).
      - Timers not properly mocked (e.g. setInterval, setTimeout).
      - Keeping references to the global scope.

      at onResult (node_modules/.pnpm/@jest+core@29.7.0/node_modules/@jest/core/build/TestScheduler.js:150:18)
      at node_modules/.pnpm/@jest+core@29.7.0/node_modules/@jest/core/build/TestScheduler.js:254:19
      at node_modules/.pnpm/emittery@0.13.1/node_modules/emittery/index.js:363:13
          at Array.map (<anonymous>)
      at Emittery.emit (node_modules/.pnpm/emittery@0.13.1/node_modules/emittery/index.js:361:23)
      
      
.... and so on      
```

This fails most of our tests due to several reasons.

It looks like the main culprit is the `nock` library: 
- https://github.com/nock/nock?tab=readme-ov-file#memory-issues-with-jest
- https://github.com/nock/nock/issues/2358

Some other culprits are the 
- tigerbeetle testcontainer not being torn down properly after the test has been ran (related: #1094)

mkurapov:

- Fixes memory leak related to `nock`, *backend* tests can now be run serially (via `--runInBand`) without running out of memory. This was verified using the `--detectLeaks` flag for jest tests, which would throw an error if detected a memory leak during a test.
- Properly spins down tigerbeetle container during tests using a custom environment
- Sets log level to silent for tests
- Improves commands related to app shutdown & tests teardown (should help with the amount of open handles the tests have). I fixed this error message:

> "A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them." 

when running the test:ci command, but at 100% workers, it still appears on my machine 🤷 . But at least some progress there.

## Context

Fixes #2393 
Fixes #1094 (even though it seemed like the thread panic wasn't an issue with newer versions of tigerbeetle)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


package.json
```diff
@@ -43,7 +43,6 @@
     "jest": "^29.7.0",
     "npm-run-all2": "^5.0.2",
     "prettier": "^3.2.4",
-    "ts-jest": "^29.1.2",
```
mkurapov, line 46: we use `@swc/jest` instead, not needed

packages/backend/src/app.ts
```diff
@@ -611,8 +611,8 @@ export class App {
     if (this.openPaymentsServer) {
       await this.stopServer(this.openPaymentsServer)
     }
-    if (this.adminServer) {
-      await this.stopServer(this.adminServer)
+    if (this.apolloServer) {
+      await this.apolloServer.stop()
     }
```
mkurapov, line 616: since we do `plugins: [ApolloServerPluginDrainHttpServer({ httpServer })]` when creating the apollo server, the httpServer will be automatically stopped/connections will be drained. However, before we weren't closing the actual apollo server that does other operations as well: https://www.apollographql.com/docs/apollo-server/api/apollo-server/#stop

packages/backend/src/middleware/cache/data-stores/redis.test.ts
```diff
@@ -13,12 +13,12 @@ describe('Redis Data Store', (): void => {
...
 
   afterEach(async () => {
+    jest.useRealTimers()
     await redis.flushall()
   })
 
   afterAll(async () => {
-    redis.disconnect()
-    jest.useRealTimers()
+    await redis.quit()
```
mkurapov, line 21: this seems to be the preferred way to stop redis connections: https://github.com/redis/ioredis/issues/1088

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -33,10 +36,14 @@ describe('Accounting Service', (): void => {
   }
 
   beforeAll(async (): Promise<void> => {
-    const { port } = await startTigerbeetleContainer()
-    Config.tigerbeetleReplicaAddresses = [port.toString()]
+    const tigerbeetlePort = (global as unknown as { tigerbeetlePort: number })
```
mkurapov, line 39: gets injected via the tigerbeetle test environment

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -106,7 +112,7 @@ describe('Accounting Service', (): void => {
           },
           LiquidityAccountType.ASSET
         )
-      ).rejects.toThrowError(
```
mkurapov, line 109: using `toThrowError` gave me a deprecated warning

packages/backend/src/payment-method/ilp/connector/core/factories/ilp-packet.ts
```diff
@@ -5,7 +5,7 @@ import { faker } from '@faker-js/faker'
 
 export const IlpPrepareFactory = Factory.define<IlpPrepare>('IlpPrepare').attrs(
   {
-    amount: faker.finance.amount(1, 100, 0),
```
mkurapov, line 8: this type signature is deprecated (was giving me warnings)

package.json
BlairCurrey: looks like its used here? https://github.com/interledger/rafiki/blob/main/jest.config.js#L4

BlairCurrey:
I was not familiar with `--runInBand` so I looked it up. In doing so it seems like [it might be faster to use `--runInBand` in CI](https://stackoverflow.com/questions/43864793/why-does-jest-runinband-speed-up-tests)? This is tangential, but perhaps its worth testing if we haven't already. `--runInBand` took about 50% longer for me locally FWIW but thats on an m2 Max and the ubuntu runner is [4 cores/16gb memory](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) so maybe it could be faster? 🤷‍♂️

packages/backend/src/graphql/resolvers/auto-peering.test.ts
```diff
@@ -20,6 +19,8 @@ import { CreateOrUpdatePeerByUrlInput } from '../generated/graphql'
 import { AutoPeeringService } from '../../payment-method/ilp/auto-peering/service'
 import { v4 as uuid } from 'uuid'
 
+const nock = (global as unknown as { nock: typeof import('nock') }).nock
```
BlairCurrey, line 22: What about this?

add this to  a new file at `rafiki/packages/backend/src/global.d.ts`:

```ts
declare module globalThis {
  var nock: typeof import('nock')
}
```

which lets us just do:


```ts
const nock = global.nock
```

packages/backend/jest.custom-environment.ts
```diff
@@ -0,0 +1,9 @@
+import { TestEnvironment } from 'jest-environment-node'
```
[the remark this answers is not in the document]
BlairCurrey: I guess these nock changes are needed even after this fix? https://github.com/nock/nock/pull/2572 It looks like we are on the latest and have this fix. I suppose there were two issues:
- nock leaking from simply being imported (fixed above)
- us not cleaning it up properly which still require these changes

BlairCurrey:
Runs fine for me locally (including `maxWorkers=100%` for whatever that is worth)

packages/backend/src/graphql/resolvers/auto-peering.test.ts
BlairCurrey: Although I dont love that this suggests nock is on global *anywhere* (including outside tests, or outside the that specific test environment) ... was just thinking maybe we could be easier to reference than doing all this casting everywhere. alternatively maybe we can initialize it like this once somewhere and export it?

packages/backend/src/middleware/cache/data-stores/redis.test.ts
sabineschaller: What about resetting the timers?

mkurapov: Moved it to the `afterEach`

sabineschaller:
>I was not familiar with --runInBand so I looked it up. In doing so it seems like [it might be faster to use --runInBand in CI](https://stackoverflow.com/questions/43864793/why-does-jest-runinband-speed-up-tests)? This is tangential, but perhaps its worth testing if we haven't already. --runInBand took about 50% longer for me locally FWIW but thats on an m2 Max and the ubuntu runner is [4 cores/16gb memory](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) so maybe it could be faster? 🤷‍♂️

Very interesting. I think we should give it a try. I also didn't know that flag.

sabineschaller:
@mkurapov ♥️♥️♥️ Thank you for tackling this ♥️♥️♥️

packages/backend/src/graphql/resolvers/auto-peering.test.ts
mkurapov: I'm not too fond of this long casting, but I think I prefer not mixing the shipped code vs the testing code IMO

mkurapov:
@BlairCurrey `--runInBand` instead of `--maxWorkers=2` in the CI took about 50% longer

https://github.com/interledger/rafiki/actions/runs/7919005784/job/21618820101 (~6 mins)
https://github.com/interledger/rafiki/actions/runs/7788547057 (~ 4 mins)

sabineschaller:
Are we ready to get this merged as is then?

mkurapov:
@sabineschaller going to try one final test to move some imports around to see if the "fix" from the nock issue might work for us, otherwise, I'll merge

mkurapov:
@sabineschaller yeah, the order of imports don't seem to make a difference in the repo

BlairCurrey:
> @BlairCurrey `--runInBand` instead of `--maxWorkers=2` in the CI took about 50% longer
> 
> https://github.com/interledger/rafiki/actions/runs/7919005784/job/21618820101 (6 mins) https://github.com/interledger/rafiki/actions/runs/7788547057 (4 mins)

Good to know, thanks for testing it out.

BlairCurrey (approved):

package.json
mkurapov: @BlairCurrey was about to merge until I saw this, fixed now :+1:

PR #2400, opened 2024-02-07
feat(auth): improve GNAP error responses

### LINKED ISSUE #2344: Improve Auth Server Errors
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#name-error-response

> For example, if the RO denied the request while interacting with the AS, the AS would return the following error when the client instance tries to continue the grant request:
> {
>    "error": {
>        "code": "user_denied",
>        "description": "The RO denied the request"
>    }
>}

The Rafiki Auth server currently returns the minimal amount of information that's compliant with the spec - just the error code. But with https://github.com/interledger/rafiki/pull/2285 the more informative error response is being returned during grant continuation - this pattern should be applied to the rest of the errors in the auth server.

njlie:

- Expands error responses to include a `code` field that refers to a particular GNAP error and a `description` field that explains what went wrong.

## Context

Fixes #2344.

Just a small PR to update all the GNAP error responses to be more detailed.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/grant/routes.ts
```diff
@@ -364,16 +364,16 @@ async function revokeGrant(
     'GNAP '
   )[1]
   if (!continueId || !continueToken) {
-    ctx.throw(401, { error: 'invalid_request' })
+    ctx.throw(401, { error: { code: 'invalid_request', description: 'invalid continuation information' } })
```
mkurapov, line 367: Similar to how you did above, should we add the description also as the second parameter to have it also exposed to the caller/client (like postman)? Or not for these 401 errors?

packages/auth/src/interaction/routes.ts
```diff
@@ -202,28 +199,48 @@ async function handleInteractionChoice(
       Buffer.from(config.identityServerSecret)
```
mkurapov, line 199: should there be also similar changes for L120 and L125?

BlairCurrey (commented):
Should error responses in middleware such as this be update as well?

https://github.com/interledger/rafiki/blob/main/packages/auth/src/signature/middleware.ts#L44

njlie:
> Should error responses in middleware such as this be update as well?
> 
> [`main`/packages/auth/src/signature/middleware.ts#L44](https://github.com/interledger/rafiki/blob/main/packages/auth/src/signature/middleware.ts?rgh-link-date=2024-02-12T17%3A52%3A06Z#L44)

Good catch, I forgot about those.

packages/auth/src/grant/routes.ts
```diff
@@ -144,12 +154,22 @@ async function createPendingGrant(
   const { body } = ctx.request
   const { grantService, interactionService, config } = deps
   if (!body.interact) {
-    ctx.throw(400, 'interaction_required', { error: 'interaction_required' })
+    ctx.throw(400, 'interaction_required', {
+      error: {
+        code: 'interaction_required',
+        description: 'missing required request field'
```
sabineschaller, line 160: Should this be different from the `invalid_client` description? Maybe 
```
missing required request field 'interact'
```
and 
```
missing required request field 'client'
```
respectively.

packages/auth/src/interaction/routes.ts
njlie: I thought at first not to change those, as they're not being served by a GNAP endpoint. But now I think I'll make them uniform across the board.

packages/auth/src/grant/routes.ts
```diff
@@ -364,16 +389,25 @@ async function revokeGrant(
...
+      error: {
+        code: 'invalid_request',
+        description: 'invalid continuation information'
+      }
+    })
   }
   const grant = await deps.grantService.getByContinue(continueId, continueToken)
   if (!grant) {
-    ctx.throw(404, { error: 'unknown_request' })
+    ctx.throw(404, 'unknown_request', {
```
mkurapov, line 401: should this also be `invalid_request`?

mkurapov (commented):
What do you think about making GNAP error responses into an enum and then having a function to handle all of these throws?

packages/auth/src/grant/routes.ts
```diff
@@ -101,7 +102,12 @@ async function createGrant(
   try {
     noInteractionRequired = canSkipInteraction(deps.config, ctx.request.body)
   } catch (err) {
-    ctx.throw(400, 'identifier_required', { error: 'identifier_required' })
```
njlie, line 104: I have no idea where I got stuff like `identifier_required` or `internal_server_error`. I've restricted error responses to the ones that are listed in https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-18#section-3.6

packages/auth/src/grant/routes.ts
```diff
@@ -144,12 +157,26 @@ async function createPendingGrant(
   const { body } = ctx.request
   const { grantService, interactionService, config } = deps
   if (!body.interact) {
-    ctx.throw(400, 'interaction_required', { error: 'interaction_required' })
```
njlie, line 147: I changed this from `interaction_required` because that seems to be more for if a resource (likely only a grant) requires interaction in order to be accessed.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -294,9 +297,11 @@ describe('Signature Service', (): void => {
         resource_server: 'test'
       }
       await expect(tokenHttpsigMiddleware(ctx, next)).rejects.toMatchObject({
-        status: 400,
-        error: 'invalid_request',
-        message: 'invalid signature headers'
```
njlie, line 299: I've change this and other error responses to `401` as it is a failure to verify the signature. I also chose `invalid_client` as the code because of the following description:
> "invalid_client":
The request was made from a client that was not recognized or allowed by the AS, **or the client's signature validation failed.**

(emphasis mine)

sabineschaller:
> Just a small PR to update all the GNAP error responses to be more detailed.

Not so small anymore...

packages/auth/src/accessToken/routes.ts
```diff
@@ -119,20 +120,25 @@ async function rotateToken(
     newToken = await deps.accessTokenService.rotate(ctx.accessToken.id, trx)
 
     if (!newToken) {
-      ctx.throw()
+      throw new Error('invalid access token')
```
sabineschaller, line 123: Why not throw an internal server error with this error message? I know that this is kind of what we are doing here, but should we do a `ctx.throw`?

packages/auth/src/grant/routes.ts
```diff
@@ -101,7 +102,12 @@ async function createGrant(
   try {
     noInteractionRequired = canSkipInteraction(deps.config, ctx.request.body)
   } catch (err) {
-    ctx.throw(400, 'identifier_required', { error: 'identifier_required' })
+    ctx.throw(400, GNAPErrorCode.InvalidRequest, {
+      error: {
+        code: GNAPErrorCode.InvalidRequest,
+        description: 'access identifier required'
+      }
```
mkurapov, line 109: Should we call `generateGNAPErrorResponse` here?

packages/auth/src/accessToken/routes.ts
mkurapov: I think because we end up catching the error and then call `ctx.throw` below

packages/auth/src/accessToken/routes.ts
```diff
@@ -119,20 +120,25 @@ async function rotateToken(
...
     deps.logger.error(
       { err: error instanceof Error && error.message },
       errorMessage
     )
-    ctx.throw(400, { message: errorMessage })
+    ctx.throw(
+      400,
+      GNAPErrorCode.InvalidRotation,
+      generateGNAPErrorResponse(GNAPErrorCode.InvalidRotation, errorMessage)
+    )
```
mkurapov, line 141:
```suggestion
    throwGNAPError(ctx, 400, GNAPErrorCode.InvalidRotation, message)
```
what do you think of something like this also? then we wouldn't need to define the `GNAPErrorCode.X` value twice

packages/auth/src/grant/routes.ts
njlie: Good catch, yes we should

packages/auth/src/accessToken/routes.ts
njlie: I like it, looks cleaner that way too

mkurapov (approved):
LGTM ✅ 
🚢 it

sabineschaller (approved):

PR #2537, opened 2024-03-14
chore: enable graphql protection - maxDepth, blockFieldSuggestions, maxCost

### LINKED ISSUE #2417: Enforce maximum depth limit on Graphql queries
## Description

In the context of GraphQL, a circular relationship vulnerability arises when the schema allows for recursive relationships without set depth restrictions. This means that a type can reference another type, which in turn either directly or indirectly references the original type, leading to a cyclical loop.
If not managed with appropriate checks or constraints, this vulnerability can be exploited by malicious actors. They can design queries that leverage these recursive relationships, compelling the server into endless processing cycles or pulling an overwhelming amount of data. Such actions have the potential to deteriorate performance, drain server resources, or even trigger Denial of Service (DoS) incidents.

## Observation
IntroSpection Circular Relationship query’s are possible on the Rafiki Admin API endpoints.

## Recommendation
The implementation of a maximum depth limit for GraphQL queries is crucial. Such a limit curtails the ability of malicious entities to exploit circular relationships, preventing them from crafting endlessly recursive queries. Many middleware solutions tailored for GraphQL servers, such as Apollo Server, offer mechanisms to enforce these depth constraints. 

## References
* https://www.apollographql.com/docs/technotes/TN0021-graph-security/ 
* https://www.apollographql.com/blog/graphql/security/securing-your-graphql-api-from-malicious-queries/#depth-limiting

sabineschaller:

- adds graphql armor to enable
  - maxDepth
  - blockFieldSuggestion
  - maxCost

To test `maxDepth`, run the following query
```
{
  __schema {
    types {
      name 
      fields {
        name
        type {
          name
          fields {
            name
            type {
              name
              fields {
                name
                type {
                  name
                  fields {
                    name
                    ...
                }
              } 
            }
          }
        }
      }
    }
  }
}
```

To test `blockFieldSuggestion` just misspell something slightly in an graphql query.

For the `maxCost` protection, I enabled it with default values. 

## Context

- fixes #2417 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


mkurapov (approved):
Depth test works for me.

To confirm, for the `blockFieldSuggestion`, should I be seeing

```
{
  "errors": [
    {
      "message": "Cannot query field \"receivdeAmount\" on type \"OutgoingPayment\". [Suggestion hidden]?",
      "locations": [
        {
          "line": 11,
          "column": 5
        }
      ],
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED"
      }
    }
  ]
}
```

sabineschaller:
> To confirm, for the blockFieldSuggestion, should I be seeing

That's it. the suggestion:hidden part. Otherwise it would say something like "did you mean...?"

packages/auth/src/app.ts
```diff
@@ -153,10 +154,35 @@ export class App {
...
+        ignoreIntrospection: false
+      },
+      costLimit: {
+        enabled: true,
+        maxCost: 5000,
+        objectCost: 2,
+        scalarCost: 1,
+        depthCostFactor: 1.5,
+        ignoreIntrospection: true
+      }
```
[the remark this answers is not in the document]
sabineschaller: https://escape.tech/graphql-armor/docs/plugins/cost-limit

packages/auth/src/app.ts
```diff
@@ -153,10 +154,35 @@ export class App {
...
+
     // Setup Apollo
     const apolloServer = new ApolloServer({
       schema: schemaWithResolvers,
-      plugins: [ApolloServerPluginDrainHttpServer({ httpServer })],
+      ...protection,
+      plugins: [
+        ...protection.plugins,
+        ApolloServerPluginDrainHttpServer({ httpServer })
+      ],
```
BlairCurrey, line 185: I realize this is merged but wanted to comment here. We can add new issues for this. 

The addition of `...protection` breaks introspection. It's working on main (before this was merged) but not here. Commenting out `...protection` fixes it.

And I know this is how the docs show it, but destructuring protection is probably not the best way to add it IMO because it has a plugins array. If we were to change the order of `...protection` and `plugins` on this object then the plugin array we are forming ourslves would get overridden:

```ts
const apolloServer = new ApolloServer({
      schema: schemaWithResolvers,
      plugins: [
        ...protection.plugins,
        ApolloServerPluginDrainHttpServer({ httpServer })
      ],
      ...protection, // overrides the above plugins
      introspection: this.config.env !== 'production'
    })
```

Again, it's techniaclly fine as-is but it's not obvious how close it is to not fine. I think something like this would be safer:

```ts
    const { plugins: protectionPlugins, ...protectionRest } = protection

    this.apolloServer = new ApolloServer({
      schema: schemaWithMiddleware,
      ...protectionRest,
      plugins: [
        ...protectionPlugins,
        ApolloServerPluginDrainHttpServer({ httpServer })
      ],
      introspection: this.config.env !== 'production'
    })
```

BlairCurrey: it's the `maxDepth.ignoreIntrospection` breaking it. Although I see that's intentionally set that way to resolve an issue with max depth being ignored. Maybe it can also be controlled by NODE_ENV (not disabled in not prod)?

PR #2538, opened 2024-03-14
chore(frontend): add X-Frame-Options header

### LINKED ISSUE #2418: Configure X-Frame-Options Header on Admin UI
## Description
Clickjacking, also known as a UI redress attack, is when an attacker uses multiple transparent or opaque layers to trick a user into clicking on a button or link on another page when they were intending to click on the top-level page. Thus, the attacker is hijacking clicks meant for their page and routing them to another page, most likely owned by another application, domain, or both. Using a similar technique, keystrokes can also be hijacked. With a carefully crafted combination of stylesheets, iframes, and text boxes, a user can be led to believe they are typing in the password to their account, but are instead typing into an invisible frame controlled by the attacker.

## Observation
Frameable response is possible on the Rafiki Admin Service http://localhost:3010/

## Recommendation
To effectively prevent framing attacks, the application should return a response header with the name X- Frame-Options and the value DENY to prevent framing altogether. If framing is necessary for application functionality, use the same header with the value SAMEORIGIN to allow framing only by pages from the same origin as the response itself. The preferred method to control framing is setting the Content-Security-Policy (CSP) header with the frame-ancestors directive set to 'none', 'self', or an allowed source. Example: Content-Security-Policy: frame-ancestors 'none' Example: Content-Security- Policy: frame-ancestors 'self' https://www.example.org;

## References
OWASP Clickjacking: https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html
Testing for Clickjacking: https://www.owasp.org/index.php/Testing_for_Clickjacking_(OWASP-CS-004)

sabineschaller:

I think this is my first frontend PR ever 😂
Adds X-Frame-Options header to Admin UI. 

## Context

- fixes #2418 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/frontend/app/entry.server.tsx
+++ b/packages/frontend/app/entry.server.tsx
@@ -46,6 +46,8 @@ function handleBotRequest(
           const stream = createReadableStreamFromReadable(body)
 
           responseHeaders.set('Content-Type', 'text/html')
+          // Set X-Frame-Options Header
+          responseHeaders.set('X-Frame-Options', 'SAMEORIGIN')
 
           resolve(
             new Response(stream, {
@@ -93,6 +95,8 @@ function handleBrowserRequest(
           const stream = createReadableStreamFromReadable(body)
 
           responseHeaders.set('Content-Type', 'text/html')
+          // Set X-Frame-Options Header
+          responseHeaders.set('X-Frame-Options', 'SAMEORIGIN')
 
           resolve(
             new Response(stream, {


JoblersTune:
LGTM

mkurapov:
Should we add the `Content-Security-Policy: frame-ancestors 'self';` header as well?
https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html#defending-with-content-security-policy-csp-frame-ancestors-directive

sabineschaller:
> Should we add the Content-Security-Policy: frame-ancestors 'self'; header as well?
https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html#defending-with-content-security-policy-csp-frame-ancestors-directive

The [limitations of that mentioned in the cheat sheet](https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html#limitations) are that X-Frame-Options takes priority.

Issue #2473, opened 2024-03-04
Run tests in CI


BlairCurrey:
Add github action to run the end to end tests in ci

Issue #2568, opened 2024-03-18
Add Admin API authorization capability to Bruno collection


njlie:
## Context
With the Admin API secured in https://github.com/interledger/rafiki/pull/2466, the Bruno collection should ideally be able to support getting the Hydra auth token and include it in requests to the Admin API.

## Todos
- [ ] Add api token variable to Bruno collection that can store auth token and gets used in each request
- [ ] Add Bruno request to initiate authorization flow for Admin API

Issue #2557, opened 2024-03-15
frontend: Add exception in Admin API for seed file


njlie:
## Context
Sub-issue of #2218.

The frontend is unable to seed the database using the Admin API as it currently expects either an active token from Hydra or a development/test-only string that bypasses the check on the backend. While the Admin API supports this, the frontend has not yet been updated to support it.

## Todos
- [ ] Augment API requests from frontend seed file with "excepted" auth token

njlie:
Closing since the seed files use the same Apollo client as the actual app - which all contain an `ApolloLink`  that [generates the signature and adds it to the headers before sending the request.](https://github.com/interledger/rafiki/pull/2632/files#diff-65709ed1fdeb0560954f00ab9030ac2fc39561669983981e7fc96b050574beb7)

Issue #2555, opened 2024-03-15
Fix Hostnames in Integration Tests


BlairCurrey:
## Feature Request

These are the conclusions from https://github.com/interledger/rafiki/issues/2472

**Describe the feature you'd like to request**

Update the integration tests with "real" urls. For example, from `http://host.docker.internal:3000` to `http://cloud-nine-wallet-backend:3000`.

Todo:
- [x] Use ~[sudo-prompt](https://www.npmjs.com/package/sudo-prompt)~ and [hostile](https://www.npmjs.com/package/hostile) to automate hostfile edits in `run-integration` script
  - this should work in ci (consider how sudo-prompt behaves) and locally
  - locally, this shouldnt prompt for password if the hostfile already has the required entries
- [x] Update all the urls in the tests, `.env`s, and `docker-compose.yml`

I'm unsure exactly how sudo-prompt will work in CI but something like this may be simpler than getting/passing credentials in in that context:

in ci 
- add step to add hostnames to run before run-integration script

in run-integration script
- define required entries
- look for required entries (non-root access)
- if not found, write them (root access)

This way the host editing is a no-op in CI and locally if the hosts have already been added. That is, it wont unnecessarily prompt for credentials. At least this is how I'm envisioning it.

PR #2569, opened 2024-03-19
ci: add integration test to lint_test_build

BlairCurrey:

- adds integration tests to ci

## Context

fixes https://github.com/interledger/rafiki/issues/2473

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


test/integration/package.json
```diff
@@ -4,7 +4,8 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "testenv:compose": "docker-compose -f ./testenv/cloud-nine-wallet/docker-compose.yml -f ./testenv/happy-life-bank/docker-compose.yml -f ./testenv/docker-compose.yml",
+    "build:deps": "pnpm --filter mock-account-service-lib build",
+    "testenv:compose": "docker compose -f ./testenv/cloud-nine-wallet/docker-compose.yml -f ./testenv/happy-life-bank/docker-compose.yml -f ./testenv/docker-compose.yml",
```
BlairCurrey, line 8: from `docker-compose` -> `docker compose`

This is the same for me locally, but in github actions docker-compose is v1 and finds our compose files invalid (because of the top-level name attribute).

.github/workflows/integration_test.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    branches:
+      - '!main'
+  pull_request:
+    types:
+      - opened
+      - synchronize
```
BlairCurrey, line 8: same event trigger as lint_test_build

.github/workflows/lint_test_build.yml
```diff
@@ -102,6 +102,27 @@ jobs:
         run: |
           npx @stoplight/spectral-cli lint ./packages/token-introspection/openapi/*.yaml
 
+  integration-test:
+    runs-on: ubuntu-22.04
+    needs: checkout
+    timeout-minutes: 5
```
BlairCurrey, line 108: Originally this was a seperate workflow file but I figured I'd include it here since it's the same event trigger.

I considered making this depend on auth, backend, token-introspection but decided not to because I think it gives us a useful signal if integration tests pass and auth/backend do not. This could happen if the failures just dont overlap with the tests cases, or if the backend/auth tests are flakey. Additionally, not making it depend on our longest running jobs ensures that we aren't increasing the time it takes to run CI. The integration tests take ~2 minutes, so they should finish before backend.

.github/workflows/lint_test_build.yml
```diff
@@ -102,6 +102,27 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup environment
+        uses: ./.github/workflows/rafiki/env-setup
+
+      - name: Setup hosts
+        run: |
+          echo "127.0.0.1 host.docker.internal" | sudo tee -a /etc/hosts
```
BlairCurrey, line 118: We're eventually going to use sudo and hostile to do this in the `run-tests` script but I think this will end up being permanent for the reason outlined in the end of the body in this issue: https://github.com/interledger/rafiki/issues/2555

BlairCurrey:
Presumably backend (and auth) gql APIs will soon require a hydra service which is not included in these tests. So the integration tests will fail at that point and we will need to add hydra to the docker environment and handle authentication in the tests.

We should decide which we want to merge first. Either we merge this first and the authentication changes will be blocked until the testenv and tests handle it, or merge the authentication changes and this will be blocked.

BlairCurrey:
> Presumably backend (and auth) gql APIs will soon require a hydra service which is not included in these tests. So the integration tests will fail at that point and we will need to add hydra to the docker environment and handle authentication in the tests.
> 
> We should decide which we want to merge first. Either we merge this first and the authentication changes will be blocked until the testenv and tests handle it, or merge the authentication changes and this will be blocked. It seems like we can probably piggy-back off either of these tasks to authenticate with the tests: #2568 #2557
> 
> I think being a much larger change, maybe we merge the authentication first (one less thing to implement there). But I don't know how close it is.

We talked about this in sprint planning and concluded the tests should be runnable w/out spinning up hydra. should work with a NODE_ENV that is test/development and using a special header. So the tests will break with those changes but we can simply update the header in the test requests. Given the changes required are small and dont look imminent I dont think we need to hold of on merging this IMO.

.github/workflows/lint_test_build.yml
sabineschaller: I'm not sure I'm following. Why do we need it twice? Because that way, the script itself doesn't prompt?

sabineschaller (approved):
Just one clarifying question but otherwise, LGTM

mkurapov (dismissed):
Approving for now, just one comment

.github/workflows/lint_test_build.yml
BlairCurrey: Good questions - short answer is that I wonder if hostile will work fine in CI after all, in which case we can get rid of this step. I will test that.

The thinking was hostile will not work in github actions due to sudo prompt (needs confirming). Thus the extra step. And that appending to `/etc/hosts` and otherwise manually performing a "append if not exist" was better managed by something like hostile, thus not using  `echo "127.0.0.1 host.docker.internal" | sudo tee -a /etc/hosts`. And to answer Max's question, that does actually prompt from the command line - that's what makes me think my original assumptions about github actions and sudo might just be bad.

Also questioning if need hostile at this point. We are already checking for the existence of some line before setting a new one, so we shouldn't be adding duplicates. Beyond that and being cross-platform, which I don't think matters because we're already assuming a unix-like environment (including WSL) with this shell script and elswhere, it basically just does this:
- erroring if you try to overwrite local or localhost or the cli args are missing https://github.com/feross/hostile/blob/master/bin/cmd.js#L62-L70
- not setting if it exists (we're already checking that) https://github.com/feross/hostile/blob/master/index.js#L93
- adding a newline such that it preserves any existing whitespace at the end of the file.  https://github.com/feross/hostile/blob/master/index.js#L94-L101

So I'm not really seeing a value-add over checking if a line already exists and if not adding it.

BlairCurrey: I guess I forgot where I was... editing `/etc/hosts` via hostile in `run-tests` is not in this branch and adding host step is required regardless here.

I will test using hostile in CI on that other branch when this is merged.

mkurapov (approved):

PR #2586, opened 2024-03-19
feat: use real hostnames & auto edit host file

BlairCurrey:

- updates all urls to use service-based hostname (e.g. `http://cloud-nine-wallet-test-backend:3000`)
- adds hosts (if not already set) automatically in `test/integration/scripts/run-tests.sh`

## Context

fixes https://github.com/interledger/rafiki/issues/2555
## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


test/integration/scripts/run-tests.sh
```diff
@@ -23,16 +23,32 @@ while [[ $# -gt 0 ]]; do
...
+addHost() {
+  local hostname="$1"
+  
+  # check first to avoid sudo prompt if host is already set
+  if hostile list | grep -q "127.0.0.1 $hostname"; then
+    echo "$hostname already set"
+  else
+    sudo hostile set 127.0.0.1 "$hostname"
+  fi
+}
```
BlairCurrey, line 36: If hosts aren't present you will be prompted for your password so it can write to /etc/hosts. If hosts are present, it will not prompt you. In CI, there is a step before running tests that add them so we dont need to navigate sudo here in ci.

Note that we didnt need sudo-prompt here. Unless I misunderstood the original suggestion, its just not necessary for this use-case.

BlairCurrey:
With this change you can remove the `127.0.0.1 host.docker.internal` entry from `/etc/hosts`. Doesn't hurt anything if you don't though.

sabineschaller:
there are still some `host.docker.internal` in the docker-compose files. I'm assuming that is intentional because those are not referring to the webhook and the exchange rate urls?

BlairCurrey:
> there are still some `host.docker.internal` in the docker-compose files. I'm assuming that is intentional because those are not referring to the webhook and the exchange rate urls?

Good catch, should have pointed that out. Yes, that's exactly right. That's so our backend services can reach the test's server.

On that note, I removed the `extra_hosts: host.docker.interal` but maybe I should return it. It works without for me but @mkurapov suggested it may be necessary for linux, if I recalled correctly. I will read up on that a bit and probably just return it to be safe.

test/integration/scripts/run-tests.sh
sabineschaller: > Note that we didnt need sudo-prompt here. 

So why do we include it?

test/integration/lib/integration-server.ts
```diff
@@ -54,9 +54,7 @@ export class IntegrationServer {
       }
     })
 
-    this.server = this.app.listen(port, () => {
-      console.log(`Integration server listening on port ${port}`)
-    })
+    this.server = this.app.listen(port)
```
sabineschaller, line 57: Is removing thing console.log part of us not wanting to console.log things anymore but wanting to do proper logging in the future?

test/integration/scripts/run-tests.sh
sabineschaller: > Note that we didnt need sudo-prompt here.

So why do we include it?

test/integration/scripts/run-tests.sh
```diff
@@ -23,16 +23,32 @@ while [[ $# -gt 0 ]]; do
...
 
+# setup hosts
+addHost() {
+  local hostname="$1"
+  
+  # check first to avoid sudo prompt if host is already set
+  if hostile list | grep -q "127.0.0.1 $hostname"; then
+    echo "$hostname already set"
+  else
+    sudo hostile set 127.0.0.1 "$hostname"
```
mkurapov, line 34:
```suggestion
  if pnpm hostile list | grep -q "127.0.0.1 $hostname"; then
    echo "$hostname already set"
  else
    sudo pnpm hostile set 127.0.0.1 "$hostname"
```
right? or did I not install it correctly?

mkurapov:
@BlairCurrey I think we can get rid of 

```
    extra_hosts:
      - 'host.docker.internal:host-gateway'  
```
from the testenv docker compose files, since we are now using the "proper" hosts to resolve the addresses (the docker containers already can resolve those hosts for containers on the same network)

test/integration/scripts/run-tests.sh
BlairCurrey: I'm referring to this npm package https://www.npmjs.com/package/sudo-prompt, not the use of `sudo` here.

test/integration/lib/integration-server.ts
BlairCurrey: Somewhat I guess - I added it initially when I was debugging service to this test server requests.

BlairCurrey: On that note I don think using pino logger here or anywhere else that is dev only would have any benefit. They aren't run with our other services and aren't going to be consumed by some sort of log aggregator or cloud logging viewer. Just development.

I'm mainly concerned about readability which `pino-pretty` would resolve. Jest has it's own "prettyfied" logs though so if using pino we should not [pipe it](https://github.com/pinojs/pino-pretty?tab=readme-ov-file#usage) and instead add it [in code](https://github.com/pinojs/pino-pretty?tab=readme-ov-file#usage) to preserve that.

BlairCurrey:
For adding hosts in CI, we can't just remove the `Setup hosts` step and I see these options:

1. keeping the setup host step
2. removing the setup host step and passing in a pnpm binary path to the script for us in CI? like `pnpm --filer integration run-tests -p /home/runner/setup-pnpm/node_modules/.bin/pnpm` or maybe just `-p $(which pnpm)`
3. removing the setup host step and adding a step to add pnpm to the binary path

Just removing the setup host step fails because root doesn't find pnpm:

> sudo: pnpm: command not found

https://github.com/interledger/rafiki/actions/runs/8392609267/job/22985668877?pr=2586

At this point I think sticking with 1 makes sense. No matter what we do this isn't going to "just work" without either passing in the path or doing some extra setup. I made some attempts at adding pnpm to the path (the more obvious move over passing in the path I think) but did not get it working. Apparently its not like just appending to the $PATH variable and if someone knows a simple way to do it I'm all ears, but it's still going to be an extra setup step.

mkurapov (approved):

mkurapov:
Keeping the setup host step in CI is the most straightforward solution.

Issue #2580, opened 2024-03-19
Remove all `console.log()`s (if any)


sabineschaller:

golobitch:
@sabineschaller I just did a quick search through rafiki repository and this is the result

```bash
❯ grep "console.log" -R .
./test/integration/lib/integration-server.ts:        console.log(`unknown event type: ${webhookEvent.type}`)
./bruno/collections/Rafiki/Open Payments APIs/Outgoing Payments/List Outgoing Payments.bru:  // console.log(request.url)
./bruno/collections/Rafiki/Open Payments APIs/Incoming Payments/List Incoming Payments.bru:  // console.log(request.url)
./localenv/mock-account-servicing-entity/app/entry.server.tsx:    console.log('setting up from seed...')
./localenv/mock-account-servicing-entity/app/entry.server.tsx:      console.log(
./localenv/mock-account-servicing-entity/app/lib/requesters.ts:      console.log(data)
./localenv/mock-account-servicing-entity/app/lib/requesters.ts:      console.log(data)
./localenv/mock-account-servicing-entity/app/lib/requesters.ts:      console.log(data)
./localenv/mock-account-servicing-entity/app/routes/webhooks.ts:  console.log('received webhook: ', JSON.stringify(wh))
./localenv/mock-account-servicing-entity/app/routes/webhooks.ts:        console.log(`unknown event type: ${wh.type}`)
./localenv/mock-account-servicing-entity/app/routes/webhooks.ts:    console.log(errorInfo)
./scripts/localenv-tunnel-setup.js:  console.log(`Created tunnel for port ${port}: ${tunnel}`)
./scripts/localenv-tunnel-setup.js:  console.log('Starting the tunnels and preparing .env file...')
./scripts/localenv-tunnel-setup.js:  console.log('Tunnels and .env file are ready!')
./scripts/localenv-tunnel-setup.js:  console.log('Tunnels are closing...')
./packages/backend/src/tests/tigerbeetle.ts:      .on('data', (line) => console.log(line))
./packages/backend/src/tests/tigerbeetle.ts:      .on('end', () => console.log('Stream closed for [tb-format]'))
./packages/backend/src/tests/tigerbeetle.ts:      .on('data', (line) => console.log(line))
./packages/backend/src/tests/tigerbeetle.ts:      .on('end', () => console.log('Stream closed for [tb-start]'))
```

as you can see, `console.log` is not really used in any of the packages.

sabineschaller:
[...]

PR #2588, opened 2024-03-20
feat!: securing the Admin UI

### LINKED ISSUE #2200: design auth mechanism for admin UI

### LINKED ISSUE #2491: Session and Token Handling
## Context
Auth Token Handling
- Develop mechanisms for secure token handling, including session validation and invalidation and token refresh to maintain user sessions. This requires handling sensitive session data securely on the server.
- Investigate token lifespan and refresh strategies. 
Session handling
- Implement session management with Ory Kratos to validate user sessions for every request. This involves setting up checks in the loaders to check session validity through Kratos' endpoints for all loader functions, ensuring users are authenticated before accessing any resources.
- Handle logout, expiry, invalidation etc.
- Consider what should happen to access tokens when a user signs out
- Understand storage for invalid or expired tokens
- Do we want to invalidate tokens upon logout or keep them and reuse them, how to manage 'dangling' tokens then
- Handle refresh tokens

## Time Estimate
2 days

### LINKED ISSUE #2492: Complete flows
## Context
Currently there is a simple registration, email verification, login flow.
Still need to implement all authentication and account management flows provided by Kratos, including handling errors, user settings, account recovery, user logout, and login processes. Additionally, set up a logout flow that effectively terminates the session in Kratos.
The Hydra consent flow is still a dummy flow
- Need to investigate if we can actually skip it and if not, then integrate it properly
User profile management
- Allow users to view and edit their profile information through a UI connected to Kratos' endpoints. Implement backend support for fetching and updating user data, including password resets and email updates.

## Time Estimate
3 days

### LINKED ISSUE #2493: Create a Production-Ready Setup
## Context
The Docker environment and config files are still set to dev appropriate values. These need to be checked and configured appropriately.
This also requires configuring all settings to be dynamically manageable through environment variables.
Implement Role-Based Access Control
- Currently anyone who logs in simply has full_access
- Define and implement role-based access controls within your application to manage user permissions effectively. This involves creating roles, assigning roles to users, and enforcing permission checks before executing sensitive operations.
- Ory Kratos session lifespans should be considered
- Ory Kratos access token lifespans should be considered
- Ensure proper error handling
- Ensure proper handling of all secrets
- Ensure proper handling of all environment variables

## Time Estimate
3 days

### LINKED ISSUE #2494: Documentation
## Context
Document the integration process, setup instructions, and operational guidelines for using the secured Admin UI with Ory Hydra and Kratos. Include details on managing auth flows, session management, user profile management, etc.
Write docs on how to get access tokens without Admin UI.
Explain version requirements.

## Time Estimate
1/2 day

### LINKED ISSUE #2495: Fronted implementation
## Context
Ensure a smooth user experience on the fronted throughout all flows

## Time Estimate
1 day

JoblersTune:

- Securing the Admin UI using Ory Kratos

## Context

fixes #2200
fixes #2492
fixes #2495
fixes #2494 
fixes #2493

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


JoblersTune:
This is a work in progress. Note all the TODO comments in the code as well as all outstanding issues for this PR which also list the remaining items that need attention.

To try out this PR:
1. `pnpm localenv:compose up`
2. `docker exec -it admin-container-name npm run invite-user -- example@mailer.com` where `admin-container-name` should be the actual docker container name of whichever ASE you are acting as (either cloud-nine, or happy-life).
3. This script will log a recovery link to the console. Copy that link into the browser and it will direct you to a user settings page where you can enter a password, save it and you're good to go.

localenv/happy-life-bank/kratos.yml
```diff
@@ -0,0 +1,96 @@
...
+    login:
+      ui_url: http://127.0.0.1:4010/auth/login
+      lifespan: 10m
+
+    registration:
+      enabled: false
+
+log:
+  level: debug
+  format: text
```
BlairCurrey, line 62: we can make this json to match the structure of our other json logs from pino

localenv/happy-life-bank/docker-compose.yml
```diff
@@ -86,6 +86,47 @@ services:
...
+      dockerfile: ./packages/frontend/kratos/Dockerfile
+      args:
+        PATH_TO_KRATOS_CONFIG: ./localenv/happy-life-bank/kratos.yml
+    depends_on:
+      - shared-database
+      - mailslurper
+    environment:
+      DSN: postgres://happy_life_kratos:kratos_password@shared-database:5432/happy_life_kratos?sslmode=disable&max_conns=20&max_idle_conns=4
+    ports:
+      - "5433:4433"
```
BlairCurrey, line 114: not a big deal, but can we use a different port than 5433? it's a common alternative port for postgres (with default being 5432).

packages/frontend/README.md
```diff
@@ -4,24 +4,6 @@
...
-To start the application run:
-
-```sh
-pnpm --filter frontend dev
-```
-
-Now you can access the application on [http://localhost:3005](http://localhost:3005).
-
-> NOTE: When running Rafiki Admin in the development environment, it will connect to Cloud Nine Wallet Admin GraphQL - [http://localhost:3001/graphql](http://localhost:3001/graphql).
-
```
BlairCurrey, line 24: I take it frontend cannot be run without kratos anymore and this is why this is removed? I do see an error when doing `pnpm --filter frontend dev`. 

```
Error: Environment variable kratosContainerPublicUrl is missing
    at /Users/blaircurrey/code/interledger/rafiki/packages/frontend/app/lib/envConfig.server.ts:8:11
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/Users/blaircurrey/code/interledger/rafiki/packages/frontend/app/lib/envConfig.server.ts:6:27)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Function.Module._load (node:internal/modules/cjs/loader:960:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:169:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)
```

I do think it's important to keep that working for development purposes so that we dont have to rebuild to see changes which not only takes ~20 seconds to rebuild but also wipes state.

packages/frontend/app/root.tsx
```diff
@@ -30,19 +36,37 @@ export const meta: MetaFunction = () => [
...
+      `${variables.kratosContainerPublicUrl}/self-service/logout/browser`,
+      {
+        headers: {
+          cookie: cookies
+        },
+        withCredentials: true
+      }
+    )
+    //TODO: In the scenario where a user is logged in but we hgave an issue retrieving the logoutUrl is it okay to simply not have a logoutUrl and not display the button? Given that the session will expire? Not sure how else to handle the error.
+    logoutUrl = response.data.logout_url
```
BlairCurrey, line 55: > Given that the session will expire

When would it expire? Like on the next page visit which requires authorization action? I guess I don't fully understand the scenario, but could we make the logout button click just _look_ like they visited the logout link in this scenario (deleting a browser cookie or something)? Or should we force-logout if we know they are now unauthorized but are still seeing a logged in view until the next action?

I was actually wondering how the logout link works. I was thinking it would be nice if we didnt have to make this request in root (so on every page?) and either store it (pass it to the client to store somewhere?) or get the link on logout click from the client, depending on just how public the`kratosContainerPublicUrl` is (can we share it with the browser)?

BlairCurrey: Looking at it further, the token/session are storedd in http only cookies so we cant delete them in the browser. We'd have to do it using Set-Cookie header and the cookie name (like we are doing just below here for the http only message cookie on line 72).

BlairCurrey:
> 3. Is there a cleaner way to check the Kratos session without needing so much repeated code in every loader?
(Note it is not sufficient to only do the check in the root file because of how Remix works that check doesn’t necessarily run when navigating on the same page only if you open a new tab)

I take your word for it, and that's consistent with what I'm reading, but I actually can't reproduce. I removed all redirects except for the root one and still cannot access any of the routes (gets redirected to `/auth`).

Apparently remix suggests just reusing some function everywhere, so not really anything different than what we have. https://remix.run/docs/en/main/guides/faq#how-can-i-have-a-parent-route-loader-validate-the-user-and-protect-all-child-routes

It appears this is a common problem for people and there has been lots of talk about remix middleware but nothing yet:
- https://github.com/remix-run/remix/discussions/1432
- https://github.com/remix-run/remix/discussions/6437
- https://github.com/remix-run/react-router/issues/9566
- https://github.com/remix-run/remix/discussions/7642

packages/frontend/app/root.tsx
```diff
@@ -101,6 +128,7 @@ export default function App() {
 export function ErrorBoundary() {
   const error = useRouteError()
 
+  // TODO: is there a better approach for handling the logout Url on the ErrorPage given that the user may / may not eb logged in, and we don'tr have access to a loader.
```
BlairCurrey, line 131: I realize this doesn't solve the problem but I don't think we should show the sidebar at all when there is an error and they're not authenticated. It's the only way they can see the sidebar without authenticating. Could probably argue the information leaking doesn't matter (tab names) but it's also just inconsistent/unexpected from a UI perspective IMO.  

I think what you have is reasonable but I also think it would be reasonable to never show the sidebar (even when authenticated) in service of hiding it when unauthenticated. 

It does make you wonder how we can get logged in state client side. Since they're http only cookies we cant check for them, so I guess we would need to track it and save on the context or something. Or perhaps we could make a request here in the error page, but I don't think that's worth it.

packages/frontend/README.md
BlairCurrey: I suppose that would require not throwing an error when in dev and trying to read the kratos ENV vars and not finding them. And then when in dev and the kratos env var(s) are undefined we don't do any of the login, redirectIfUnauthorized, etc. Kinda sucks....

packages/frontend/app/root.tsx
JoblersTune: I don't think the error boundary supports async calls. The better approach seems to be removing the sidebar since all error pages have a `Go to homepage` button. So this will resolve appropriately whether the user is logged in or not and there is a cl;ear option to click for someone stuck on an error page without the sidebar to use for navigation. Is that okay?

JoblersTune:
> > 3. Is there a cleaner way to check the Kratos session without needing so much repeated code in every loader?
> >    (Note it is not sufficient to only do the check in the root file because of how Remix works that check doesn’t necessarily run when navigating on the same page only if you open a new tab)
> 
> I take your word for it, and that's consistent with what I'm reading, but I actually can't reproduce. I removed all redirects except for the root one and still cannot access any of the routes (gets redirected to `/auth`).
> 
> Apparently remix suggests just reusing some function everywhere, so not really anything different than what we have. https://remix.run/docs/en/main/guides/faq#how-can-i-have-a-parent-route-loader-validate-the-user-and-protect-all-child-routes
> 
> It appears this is a common problem for people and there has been lots of talk about remix middleware but nothing yet:
> 
> * [How to Implement Middleware in Remix? remix-run/remix#1432](https://github.com/remix-run/remix/discussions/1432)
> * [What is the idiomatic way to protect your whole Remix app? remix-run/remix#6437](https://github.com/remix-run/remix/discussions/6437)
> * [🗺 Loader+Action Context (Middleware) remix-run/react-router#9566](https://github.com/remix-run/react-router/issues/9566)
> * [🗺️ Route Middleware remix-run/remix#7642](https://github.com/remix-run/remix/discussions/7642)

There's no further intervention required here then, right?

packages/frontend/app/root.tsx
JoblersTune: The scenario I'm sketching here is just if we encounter an error in extracting the logout URL from the response data
```
const response = await axios.get(
      `${variables.kratosContainerPublicUrl}/self-service/logout/browser`,
      {
        headers: {
          cookie: cookies
        },
        withCredentials: true
      }
    )
    logoutUrl = response.data.logout_url
    ```
    
    Maybe we got an error back or there were network issues or whatever.

File: packages/frontend/app/root.tsx (hunk 5 above)
Comment (BlairCurrey) [reply packages/frontend/app/root.tsx]: seems good to me

**Comment by BlairCurrey** [comment]
> > > 3. Is there a cleaner way to check the Kratos session without needing so much repeated code in every loader?
> > >    (Note it is not sufficient to only do the check in the root file because of how Remix works that check doesn’t necessarily run when navigating on the same page only if you open a new tab)
> > 
> > 
> > I take your word for it, and that's consistent with what I'm reading, but I actually can't reproduce. I removed all redirects except for the root one and still cannot access any of the routes (gets redirected to `/auth`).
> > Apparently remix suggests just reusing some function everywhere, so not really anything different than what we have. https://remix.run/docs/en/main/guides/faq#how-can-i-have-a-parent-route-loader-validate-the-user-and-protect-all-child-routes
> > It appears this is a common problem for people and there has been lots of talk about remix middleware but nothing yet:
> > 
> > * [How to Implement Middleware in Remix? remix-run/remix#1432](https://github.com/remix-run/remix/discussions/1432)
> > * [What is the idiomatic way to protect your whole Remix app? remix-run/remix#6437](https://github.com/remix-run/remix/discussions/6437)
> > * [🗺 Loader+Action Context (Middleware) remix-run/react-router#9566](https://github.com/remix-run/react-router/issues/9566)
> > * [🗺️ Route Middleware remix-run/remix#7642](https://github.com/remix-run/remix/discussions/7642)
> 
> There's no further intervention required here then, right?

Correct - just chiming in that I see the same thing. Seems like a common feature request.

File: packages/frontend/app/root.tsx [hunk 6]
```diff
@@ -30,19 +36,39 @@ export const meta: MetaFunction = () => [
...
+        headers: {
+          cookie: cookies
+        },
+        withCredentials: true
+      }
+    )
+    logoutUrl = response.data.logout_url
+    if (!logoutUrl) {
+      logoutUrl = '/auth/manual-logout'
+    }
```
BlairCurrey, line 57: I think we should probably log that there was some issue here. 

Also, related, I think we should probably use pino to log here instead (I see some console.logs elsewhere). judging by this recent audit I think we don't want to introduce console.logs here thinkhttps://github.com/interledger/rafiki/issues/2580

JoblersTune:
Also we're now 400 words over the word limit again so I'm gonna have to remove stuff again.

mkurapov:
@JoblersTune what do you think about providing a default user/pass for the localenv? So that we don't need to call `docker exec -it admin-container-name npm run invite-user -- example@mailer.com` everytime we want to access the admin UI locally?

packages/frontend/app/routes/auth.recovery.tsx
```diff
@@ -0,0 +1,127 @@
...
+      statusText: 'No Kratos account recovery flow ID found.'
+    })
+  }
+}
+
+export default function Recovery() {
+  const { responseData } = useLoaderData<typeof loader>()
+  const uiContainer: UiContainer = responseData.ui
+  const uiNodes = uiContainer.nodes
+  const actionUrl = uiContainer.action
```
mkurapov, line 49: out of curiosity, how do these work exactly? the `/self-service/recovery/flows` return a list of UI elements? for a given flow?

packages/frontend/kratos/config/identity.schema.json
```diff
@@ -0,0 +1,34 @@
+{
```
mkurapov, line 1: is this a default preset of how the identity schema is stored in kratos?

packages/frontend/README.md
```diff
@@ -49,4 +36,13 @@ To add a new typed apollo request, you will need to add an untyped request and r
   - `routes`: outer layer of the application
   - `shared`: utilility functions or types
   - `styles`: CSS files
+- `kratos`: Dockerfile and setup files
+  - `config`: contains the Kratos identity schema
+  - `scripts`: scripts to start Kratos, as well as, add and delete users
```
mkurapov, line 41:
```suggestion
  - `scripts`: scripts to start Kratos, as well as to add and delete users
```

packages/frontend/app/root.tsx
```diff
@@ -30,19 +36,39 @@ export const meta: MetaFunction = () => [
 ]
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
-  const session = await messageStorage.getSession(request.headers.get('cookie'))
+  const cookies = request.headers.get('cookie')
+  await redirectIfUnauthorizedAccess(request.url, cookies)
+
+  let logoutUrl = ''
```
mkurapov, line 42:
```suggestion
  let logoutUrl
```

packages/frontend/README.md
```diff
@@ -2,25 +2,9 @@
...
-To start the application run:
-
-```sh
-pnpm --filter frontend dev
-```
-
-Now you can access the application on [http://localhost:3005](http://localhost:3005).
-
-> NOTE: When running Rafiki Admin in the development environment, it will connect to Cloud Nine Wallet Admin GraphQL - [http://localhost:3001/graphql](http://localhost:3001/graphql).
+The Admin UI runs in development mode in the local playground, so when you have it running, you can see live code changes by simply refreshing your page.
```
mkurapov, line 7: should we keep the links pointing to the admin pages here?

packages/frontend/Dockerfile
```diff
@@ -8,48 +8,16 @@ ENV PATH="$PNPM_HOME:$PATH"
...
 RUN corepack prepare pnpm@8.7.4 --activate
 
-COPY pnpm-lock.yaml ./
+COPY pnpm-lock.yaml package.json pnpm-workspace.yaml .npmrc tsconfig.json tsconfig.build.json ./
 
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     pnpm fetch \
     | grep -v "cross-device link not permitted\|Falling back to copying packages from store"
 
-FROM base AS prod-deps
```
mkurapov, line 17: why do we need to remove the multi-step builds here?

packages/frontend/app/root.tsx
JoblersTune: Does this resolve the logging issue: https://github.com/interledger/rafiki/commit/f1a25c4b56f6a502b976a9de7cd124e0f4293376

localenv/cloud-nine-wallet/kratos.yml
```diff
@@ -0,0 +1,91 @@
+version: v0.13.0
+
+dsn: postgres://cloud_nine_kratos:kratos_password@shared-database:5432/cloud_nine_kratos?sslmode=disable&max_conns=20&max_idle_conns=4
+
+serve:
+  public:
+    base_url: http://127.0.0.1:4433/
```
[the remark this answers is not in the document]
JoblersTune: It can be either so long as it's consistent. I'll change it.

packages/frontend/app/root.tsx
```diff
@@ -30,19 +36,39 @@ export const meta: MetaFunction = () => [
...
+  const loggedIn = await isLoggedIn(cookies)
```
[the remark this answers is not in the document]
JoblersTune: If the user is logged in then we display the sidebar with the navigation options as well as a logoutURL which are not displayed if a user is not logged in. This file still renders pages like /auth/login so it must have configurations to know when to render the sidebar or not. 

`<html
      lang='en'
      className='h-full bg-polkadot bg-cover bg-no-repeat bg-center bg-fixed'
    >
      <head>
        <Meta />
        <Links />
      </head>
      <body className='h-full text-tealish'>
        <div className='min-h-full'>
          {loggedIn && <Sidebar logoutUrl={logoutUrl} />}
          <div
            className={`pt-20 md:pt-0 flex ${loggedIn ? 'md:pl-60' : ''} flex-1 flex-col`}
          >
            <main className='pb-8 px-4'>
              <Outlet />
            </main>
          </div>
        </div>
        <Snackbar
          id='snackbar'
          onClose={() => setSnackbarOpen(false)}
          show={snackbarOpen}
          message={message}
          dismissAfter={2000}
        />
        <ScrollRestoration />
        <PublicEnv env={publicEnv} />
        <Scripts />
        <LiveReload />
      </body>
    </html>`

packages/frontend/README.md
JoblersTune: The http://localhost:3005 link is no longer in use and doesn't work anymore. 
I can include the http://localhost:3001/graphql link but then why just that one, why not the rest of them - which is why I just advise people to to the localenv readme for that info.

packages/frontend/Dockerfile
JoblersTune: We are running the admin UI container in dev mode now so we need the dev dependencies not the prod dependencies. 

This Dockerfile now uses only a builder stage basically, because we're mounting code in, which we still need to build and run inside the container. This is so that we can see live changes with a  simple refresh instead of rebuilding the container container each time we want to see our changes reflect.

JoblersTune:
> @JoblersTune what do you think about providing a default user/pass for the localenv? So that we don't need to call `docker exec -it admin-container-name npm run invite-user -- example@mailer.com` every time we want to access the admin UI locally?

We can do this. Kratos will not allow you to use esay passwords like 'password' though. So unless you think it's easier for people to use a secure password that's preset rather than clicking a link we can do that.

Another option would be to run the invite-user script on startup so that a link is already there and can just be clicked directly without running a script. The link may expire if people leave it too long but that's okay because they can just run invite-user again then.

BlairCurrey:
> Another option would be to run the invite-user script on startup so that a link is already there and can just be clicked directly without running a script.

I was more on board with Max's original suggestion at first but I wonder if this might actually be more convenient. Clicking a link in the logs (might be buried though?) vs. navigating to the right port (which I often forget) and knowing the right password to put in.

packages/frontend/app/root.tsx
BlairCurrey: yes looks good

packages/frontend/Dockerfile
BlairCurrey: and this is required because we cant do hot module reloading when launching with `remix-serve` (even with the bind mount), is that correct?

BlairCurrey: is it a bad practice to have separate docker files for dev/prod? I see some mixed conclusions on the subject ([1](https://stackoverflow.com/questions/46440909/how-to-configure-different-dockerfile-for-development-and-production), [2](https://www.reddit.com/r/kubernetes/comments/ukhi3t/different_dockerfile_for_dev_vs_stagingprod/), [3](https://stackoverflow.com/questions/48116279/dockerfile-setup-for-both-production-and-development)).

Just thinking we'll have no reason to run it locally so errors might go unnoticed. Thinking in particular with remix upgrades. I suppose at least the image build would be verified by CI.

I guess HMR necessitates different run commands and dev deps but I wonder if we could unify it a bit more up to that point. Perhaps using multi-stage builds and targets like suggested [here](https://stackoverflow.com/a/56114663/12893676) or perhaps controlling with 
[ARG](https://docs.docker.com/reference/dockerfile/#arg) (as some of the above links have alluded to).

BlairCurrey: in any case, can we install the dev dependencies in the dockerfile instead of in the container after start? And we can move the dev command to the dockerfile as well I think, instead of this in the compose: `command: sh -c "pnpm install && pnpm dev"`

JoblersTune: Thanks for that catch. I had it that way for a previous configuration. That additional `pnpm install` is no longer needed in the docker-compose. 

As for your concern about having a production use Dockerfile that we're not forced to interact with, or even worse, can't easily interact with... That is a valid problem...

localenv/README.md
```diff
@@ -8,6 +8,8 @@ These packages include:
 - `auth` (GNAP auth server)
```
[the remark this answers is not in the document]
JoblersTune: How do I edit the localenv-architecture image?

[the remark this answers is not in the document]
mkurapov: (resolved in slack).
We should also commit the file into the repo so we always have it.

mkurapov (commented):
A few comments, but basically there :+1:

This should be a breaking change, correct? If so we should add "!" in the title to signify this:

`feat!: securing the Admin UI`

(This makes it easier to generate release notes with the different breaking changes)

packages/frontend/Dockerfile
BlairCurrey: would need to install dev dependencies too so we can use the `remix dev` cmd

packages/frontend/app/routes/auth.recovery.tsx
JoblersTune: Sorry, I missed this comment. Kratos allows you to render all your own UIs so their API calls return elements needed on your page and then it's up to you to group them, place them and style them as you see fit. 
It uses the identity schema you defined so it knows what fields are required when and also what email to use for recovery etc. 

Do you think it would be better to include a comment about this somewhere? Or add a point on this in the docs?

packages/frontend/README.md
```diff
@@ -49,4 +36,16 @@ To add a new typed apollo request, you will need to add an untyped request and r
...
+  - `config`: contains the Kratos identity schema
+  - `scripts`: scripts to start Kratos, as well as to add and delete users
 - `public`: static files and assets that are served to the browser
+
+## Login
+
+We have secured access to the Admin UI using [Ory Kratos](https://www.ory.sh/docs/kratos/ory-kratos-intro), a secure and fully open-source identity and user management solution. Check it out on [GitHub](https://github.com/ory/kratos). Since access to the UI is on an invitation-only basis the registration flow is not publicly available. As such, in order to access the UI you'll need to add a new user with the invite-user script. Run `docker exec -it <admin-container-name> npm run invite-user -- example@mail.com` and it will output recovery link to the terminal. The recovery link doubles as the invitation method. Copy and paste this link in your browser and you will automatically be logged in and directed to the account settings page. The next step is changing your password. We're using a simple email and password authentication method.
+
+We've also included a script to remove users: `docker exec -it <admin-container-name> npm run delete-user -- example@mail.com`.
```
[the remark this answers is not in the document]
JoblersTune: | If you only changed the CMD in the original Dockerfile to either run remix-serve or remix-build, would it still work?

It takes me down a rabbit hole of issues, I have to change pnpm to npm for the command, then I must ensure the build steps are remove because tsc is no longer available, fair enough, since we build a few lines above, but then the next is "Error: unable to determine transport target for "pino-pretty"" And I don't think we'd want to move that out of dev dependencies. So we can keep trying to jump over these hurdles but ultimately we are running a dev build process without our dev dependencies which is not ideal.

So in short, I'd argue, no, it doesn't work.

packages/frontend/tsconfig.json
```diff
@@ -1,13 +1,14 @@
 {
   "include": ["remix.env.d.ts", "**/*.ts", "**/*.tsx"],
   "compilerOptions": {
-    "lib": ["DOM", "DOM.Iterable", "ES2019"],
+    "lib": ["DOM", "DOM.Iterable", "ES2019", "ESNext", "ESNext.Promise"],
```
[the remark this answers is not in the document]
JoblersTune: I don't remember anymore at what point I included these. They don't seem to be necessary. So I removed the ESNext stuff and we can keep with ES2019.

mkurapov:
Is there a way we can change this message to not mention social-sign in? (and maybe format the minutes nicely?)

Screenshot 2024-04-18 at 12 35 28

JoblersTune:
> Is there a way we can change this message to not mention social-sign in? (and maybe format the minutes nicely?)
> 
> Screenshot 2024-04-18 at 12 35 28

There isn't a good way to change this. It's a generic message returned from Kratos. They have had other people complain about this before too. There is no way to customise it and you can't easily choose not to display it without also not displaying other messages you may receive, like, "Your changes have been saved!"

The only way I could see how to change this is check if this is the message we receive and in that case rather display our own custom message but that feels very messy.

Issue #2529, opened 2024-03-12
General Test Cleanup


BlairCurrey:
Several miscellaneous things are needed to polish up the integration tests
- [x] switch ports in configuration to resolve conflicts with localenvironment. should be able to run local env and tests.
- [x] possibly rework test structure (perhaps some type of [pipe pattern](https://blog.logrocket.com/how-to-create-compose-function-typescript/)?) Idea is to make the requests functions and make them more composable. Prevents lots of duplication if we want to do something like seperate `Open Payments with Continuation via Polling` and `Open Payments with Continue via "interact_ref"` and makes it clear what each request requires and returns (instead of a mess in global scope). Also enforces the dependant relationship between tests better (ie, you cant skip one or run in parallel). Some ideas with regards to failing after first test: https://stackoverflow.com/questions/51250006/jest-stop-test-suite-after-first-fail
- [x] if using `run_integration.sh`, probably set to always build instead of an option, or at least to build by default because its the main use case and easy to forget to include it.
- [x] Add continuation via `interact_ref` `Open Payments Flow` variant (have the steps, but currently only test the continuation polling way)
- [x] remove console logs and/or convert to pino logs
- [x] some more useful assertions
  - https://github.com/interledger/rafiki/pull/2380#issuecomment-2000243396
- [x] jest test should fail immediately when beforeAll errors - noticed when bad config made setupFromSeed fail but it ran every test and it appeared the errors came from the tests. maybe start here https://stackoverflow.com/questions/72027581/how-to-prevent-jest-from-running-tests-when-an-error-occurred
- [x] add `test/integration` readme
- [x] add `mock-account-servicing-lib` readme

PR #2595, opened 2024-03-20
feat: misc integration test improvents

BlairCurrey:

- changes ports used so that tests can be run alongside `localenv`
- `run-tests` now builds images by default, instead of building when run with `-b`/`--build` flag. This is the more common use-case. can now do `--no-build` or `-nb` to skip build. 
- `run-tests` now also builds `mock-account-service-lib` so you dont have to manually do it the first time and after changes to the lib (takes ~1s so why not). Can skip building with `--no-build`/`-nb`
- add `integration` and `mock-acccount-service-lib` package readme

## Context

fixes https://github.com/interledger/rafiki/issues/2529

relies on https://github.com/interledger/rafiki/pull/2608. Will merge that into here before merging this to main. Left them separate for the time being since I figured it would make reviewing easier, and the changes aren't really related (save for branching off this branch and falling under general test cleanup).

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


test/integration/scripts/run-tests.sh
```diff
@@ -7,13 +7,13 @@
...
-build_flag=""
+build_flag="--build"
 
 # Parse cli args
 while [[ $# -gt 0 ]]; do
   case "$1" in
-    -b|--build)
-      build_flag="--build"
+    -nb|--no-build)
+      build_flag=""
```
BlairCurrey, line 16: changed to build by defult. building is the more normal case. no biulding is included because it's nice for developing tests because (changes to tests but not to services we're testing).

test/integration/testenv/docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       volumes:
         - ../dbinit.sql:/docker-entrypoint-initdb.d/init.sql
       ports:
-        - '5432:5432'
+        - '5433:5432'
```
BlairCurrey, line 21: Technically this does not need to be exposed at all. Same for some of the other exposed service ports I think. I left them exposed because I found some of them useful when developing the tests. But perhaps they should not be exposed by default?

packages/mock-account-service-lib/README.md
```diff
@@ -0,0 +1,3 @@
+# Mock Account Service Lib
+
+This is a collection of logic and utilities common to the mock account servicing entities used in our local and test environments.
```
BlairCurrey, line 3: Just didn't find there was really much to say here 🤷 Open to feedback

test/integration/lib/integration-server.ts
```diff
@@ -54,9 +54,7 @@ export class IntegrationServer {
       }
     })
 
-    this.server = this.app.listen(port, () => {
-      console.log(`Integration server listening on port ${port}`)
```
BlairCurrey, line 58: I removed this log as it was unnecessary but it got me thinking about pino here. Frankly I don't think it makes sense to use a structured logger for something that is exclusively going to be viewed from a development/github actions context. Wouldn't be consumed by a log aggregator and just makes it harder to read.

test/integration/lib/mock-ase.ts
```diff
@@ -4,7 +4,7 @@ import {
   createAuthenticatedClient
 } from '@interledger/open-payments'
 import { AccountProvider, setupFromSeed } from 'mock-account-service-lib'
-import { createApolloClient } from './apolloClient'
+import { createApolloClient } from './apollo-client'
```
BlairCurrey, line 7: Changed a few file names to use consistent casing (from camelCase to kebab-case). This leaves the new `integration` and `mock-account-service-lib` packages consistent at least. 

Guess it's a seperate discussion but we do a mix all over the place. I think we have more kebab-case and wouldn't mind standardizing on that going forward, whether we go back and change existing or not.

test/integration/integration.test.ts
```diff
@@ -453,12 +462,14 @@ describe('Integration tests', (): void => {
       })
 
       expect(outgoingPayment_.id).toBe(outgoingPayment.id)
+      expect(outgoingPayment_.receiveAmount.value).toBe(amountValueToSend)
+      expect(outgoingPayment_.sentAmount.value).toBe(amountValueToSend)
```
BlairCurrey, line 466: This is in the final "Get Outgoing Payment step". @mkurapov I think it satisfies your suggestion here https://developers.mews.com/compile-time-functional-programming-in-typescript/ to verify the payment actually completes (already had this check in P2P as it happens).

I also tried checking for the incoming payment webhook but decided against it. I think this mostly checks the same thing and the MASE doesn't need to do anything with the incoming payment completed webhook. The incoming payment completed webhook is very slow (not really sure why that is). I had to poll for 40 seconds just for that webhook, which increases the test runtime by about 4x. If checking the outgoing payment is sufficient then I think we should just limit it to that.

package.json
```diff
@@ -33,8 +33,8 @@
     "@jest/types": "^29.6.3",
     "@swc/jest": "^0.2.36",
     "@types/jest": "^29.5.12",
-    "@typescript-eslint/eslint-plugin": "^7.3.1",
-    "@typescript-eslint/parser": "^7.3.1",
+    "@typescript-eslint/eslint-plugin": "7.2",
+    "@typescript-eslint/parser": "7.2",
```
BlairCurrey, line 37: Need to revisit before merging - I pinned this because a node version issue which was described here: https://interledgerfoundation.slack.com/archives/C06KZCKEEKW/p1710798244722689

BlairCurrey: I figured this would go away when the node 20 upgrade happens (soon?). Otherwise it seems like we can alternatively upgrade to a new node 18 version (not sure what all we would need to specify for that).

test/integration/lib/mock-ase.ts
mkurapov: I'm getting a type error on L45, "...Type 'ApolloLink' is not assignable to type 'ApolloLink | RequestHandler | undefined'." did you encounter this?

test/integration/integration.test.ts
mkurapov: Also I'm glad you thought the article you linked was a suggestion from me because that is pretty cool 😄

BlairCurrey: > Also I'm glad you thought the article you linked was a suggestion from me because that is pretty cool 😄

pffft. wrong link. it was a comment from the previous PR that I meant to link.

That was part of a rabbit hole of trying to figure out how one might define a ts type for a number range. Like:

```
someRgb: IntRange<0, 255> = 0 // OK
someRgb: IntRange<0, 255> = 256 // Error
```

TLDR is **just dont do that**, but it's kinda doable with some major caveats using this recursive method: https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range. Then I was curious about that article I accidentally shared because it could perhaps allow using arithmetic instead ofjust statically defining some upper/lower limit. TS as a full-fledged language is basically voodoo.

package.json
BlairCurrey: waiting on that one is probably a good idea

test/integration/lib/mock-ase.ts
BlairCurrey: I'm seeing it now after a `pnpm i`. Looks like it didnt get bumped (https://github.com/interledger/rafiki/commit/2eb0837392ac87ca997f504f904837b0b4433e7f) with the rest of the packages and there is a type mismatch between the test client version of 3.9.7 the lib (and everywhere else) of 3.9.9. 

I fixed that but I think renovate bot should have picked this up (`mock-account-service-lib` was in main, so tests were too and should have been bumped). I wonder if it's because of the renovate bot config, but I don't see what that would be looking at the `renovate.json`. Especially since nothing was necessary for `mock-account-service-lib`. It is a pnpm package. I wonder if renovate is configured not to look in the test directory? Although I dont see anything like that [in the schema](https://docs.renovatebot.com/renovate-schema.json). Cant find anything on the `config:base` which it apparently extends. https://docs.renovatebot.com/presets-config/

BlairCurrey: yeah, guess its the test dir: https://github.com/renovatebot/renovate/discussions/12755

BlairCurrey: I updated our `ignorePaths` in the renovate config: https://github.com/interledger/rafiki/pull/2595/commits/2f3d6cdcc6867aae908b9a56bddc39cf0fd36130

I don't really think there is a way for me to test but I believe that will work. I wish there was a way to effectively just remove a single item from the preset ignore paths but I dont see a way. I'm not sure if the `ignorePresets: [":ignorePaths"]` is required (maybe the new definition overrides it?) but without a way to test I figured I'd include it to be thorough.

test/integration/integration.test.ts
BlairCurrey: > We could also in addition get the incoming payment, and confirm the receivedAmount there. What do you think?

Yeah that make sense. added.

test/integration/scripts/run-tests.sh
```diff
@@ -4,16 +4,16 @@
 # saves the container logs to a file, and stops the containers.
 # Usage:
 #   ./script.sh            # Run the script with default options
-#   ./script.sh --build    # Re-build the docker images (-b or --build)
+#   ./script.sh --build    # Skip building the docker images and internal test dependencies (-nb or --no-build)
```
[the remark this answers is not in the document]
BlairCurrey: yup, thanks

renovate.json
```diff
@@ -18,5 +19,14 @@
...
     }
+  ],
+  "ignorePaths": [
+    "**/node_modules/**",
+    "**/bower_components/**",
+    "**/vendor/**",
+    "**/examples/**",
+    "**/__tests__/**",
+    "**/tests/**",
+    "**/__fixtures__/**"
```
BlairCurrey, line 30: see: https://github.com/interledger/rafiki/pull/2595/files#r1532698648

basically had to carve out an exception for the test directory. renovate bot was ignoring it because of the defaults (same as here + `**/test/**`)

test/integration/lib/mock-ase.ts
BlairCurrey: looks like it happened again 😆  

https://github.com/interledger/rafiki/commit/0b526f2147d24657dbdd3182929398df03f5c231

test/integration/integration.test.ts
mkurapov: > TS as a full-fledged language is basically voodoo.

I do respect those people trying to do crazy types for fun, but yes, TS is just a DX tool 😆

test/integration/lib/mock-ase.ts
mkurapov: > I wish there was a way to effectively just remove a single item from the preset ignore paths

Would we be able to keep the configuration the same, and extend the `ignorePaths` to use "!" as such: `!test/**`?

test/integration/integration.test.ts
```diff
@@ -453,12 +462,22 @@ describe('Integration tests', (): void => {
...
+    })
+
+    test('Get Incoming Payment', async (): Promise<void> => {
+      const incomingPayment_ = await hlb.opClient.incomingPayment.getPublic({
+        url: `${incomingPayment.id}`
+      })
+      assert(incomingPayment_.receivedAmount)
+      expect(incomingPayment_.receivedAmount.value).toBe(amountValueToSend)
```
mkurapov, line 474: just a flag, for the cross currency test we'd probably want this to be `toBeGreaterThanOrEqual` due to FX rounding

test/integration/lib/mock-ase.ts
BlairCurrey: Maybe... I see someone doing this here https://github.com/equinix/pulumi-equinix/blob/c0e20f3f33c7315032a2f8c459bd354cff050297/renovate.json#L6 but thats the only usage I can find.

Unfortunately I don't see a way to test this... Other than merging into main and seeing what renovate bot picks up when there is a new apollo version.

test/integration/integration.test.ts
BlairCurrey: We have a p2p cross currency case which I added a similar sounding check for (expects higher/lower amount based on exchange rate): https://github.com/interledger/rafiki/pull/2608#discussion_r1543815409

mkurapov (dismissed):

test/integration/lib/mock-ase.ts
BlairCurrey: sure, lets try that.

test/integration/integration.test.ts
```diff
@@ -109,493 +65,260 @@ describe('Integration tests', (): void => {
...
+
+      // Expected amounts depend on the configuration of asset codes, scale, exchange rate, and fees.
+      assert(receiverAssetCode === 'EUR')
+      assert(senderAssetCode === 'USD')
+      assert(
+        receiver.incomingAmount.assetScale === senderWalletAddress.assetScale
+      )
+      assert(senderWalletAddress.assetScale === 2)
+      assert(exchangeRate === 0.91)
+      assert(fee.fixed === 100)
```
mkurapov, line 303: do we need to `assert(fee)` here beforehand? 
I'm getting `'fee' is possibly 'undefined'` here

mkurapov (approved):

test/integration/integration.test.ts
BlairCurrey: fixed [52b9f19](https://github.com/interledger/rafiki/pull/2595/commits/52b9f19e443f5186e778caf0c021ed2d9a885f1c)

mkurapov (approved):

mkurapov:
@BlairCurrey I'll change the netlify node version for these checks to pass, one moment

mkurapov (approved):

mkurapov:
The netlify checks are failing, but I think it's because they can't be retriggered manually, only during a push maybe.

I redeployed the netlify preview, however, - it seems to be working ok

PR #2608, opened 2024-03-25
refactor: test steps into fns and make each flow 1 test

BlairCurrey:
## Changes:
- adds Peer to Peer - Cross Currency flow test
- adds Open Payments with Finish Method flow test (as opposed to the continuation polling)
- squashes each flow into 1 test - see the "Original PR body" collapse section below for more details.
- moves each step in the tests into a `admin` or `openPayments` test action. Makes things a bit more repeatable and the test a bit higher leve - see the "Original PR body" collapse section for more details.

## Context:

This is to merge into [bc/2529/cleanup-integration-test](https://github.com/interledger/rafiki/tree/bc/2529/cleanup-integration-tests), not main. I opened a separate PR in hopes of making these changes more digestable.

I originally opened this PR to get feedback on how I should refactor the tests (leave as-is, squash flows to single test but in-line everything, or squash flows to single test and turn the different steps into functions). I preserved that in the collapse section below, and it's what Max's comment is replying to here https://github.com/interledger/rafiki/pull/2608#issuecomment-2018997919:

### Original PR Body
<details>

<summary>details</summary>

My main question is: how should we structure the tests? I see a couple options:
1. Keep them how they are in main. That is, 1 test per request in the flow and sharing data between them via global scope.
2. 1 test per flow, with all requests/assertions in-lined and variables scoped to that test.
3. 1 test per flow with requests/assertions abstracted into repeatable functions that can be used across flows.

So for <span>#</span>2 or <span>#</span>3, our test would output like this instead of detailing each request in the flow.:

```
  Integration tests
    Requests
      ✓ Can Get Non-Existing Wallet Address (337 ms)
    Flows
      ✓ Open Payments with Continuation via Polling (1 ms)
      ✓ Open Payments with Continuation via "interact_ref"
      ✓ Peer to Peer (1374 ms)
      ✓ Peer to Peer - Cross Currency (1196 ms)
 ```

I'm pursuing <span>#</span>3, wondering if <span>#</span>2 might be better, and want to shared why I don't think <span>#</span>1 is a good option.

These changes start to implement <span>#</span>3.  I converted the `Peer to Peer` flow to 1 test and made a function for what were previously individual tests (`createReceiver`, `createQuote`, `createOutgoingPayment`, `getOutgoingPayment`). Then I implemented the `Peer to Peer - Cross Currency` flow test to see how reusable these functions were - I found that not too much had to be changed. Just the beginning and end of the flow basically. The input for `createReceiver` needs to be passed in wherever it's called, and some of the final assertions of the outgoing payments were flow specific so I pulled some assertions out to the callers. I think this is actually fine, but im wondering if the more these are re-used on different variations of the flows, the more we will have to pull out the assertions and spies and handle them where they are called. In which case this would really be over-abstracted as it would basically just wrap the gql/open payments client requests.

I'm doing this because I think option <span>#</span>1 is a problem basically because the different steps in the flow (create incoming payment, create quote, etc) are not independent. This means:
  - Flows should fail at the first failed test, which is not possible if each step is a test. With a test per flow, the first failed assertion will end the test and move to the next one (another flow). Whereas currently the next test is another step in the flow. No way around this in jest. It's just misusing the concept of tests.
  - We really only need to know if the flow passes/succeeds. We will still be able to see where it fails when it fails which is all that matters.
  - We have to thrust lots of data into the global scope so that tests can share results with subsequent test. This makes it hard to see exactly what each test relies on, and what side effects each test may have. This generally makes the tests a lot harder to understand.

I chose <span>#</span>3 over <span>#</span>2 because there would be a lot of code duplication for doing variations on flows (for example: `Open Payments with Continuation via Polling` and `Open Payments with Continuation via "interact_ref"`). However if it proves that we can't absrtact much of the underling api calls, spying/assertions then they aren't really reusable anyways I guess.

</details>

### DIFF
--- a/test/integration/integration.test.ts
+++ b/test/integration/integration.test.ts
@@ -247,7 +247,7 @@ describe('Integration tests', (): void => {
         outgoingPayment.id,
         value
       )
-      expect(outgoingPayment_.sentAmount.value).toBe(value)
+      expect(outgoingPayment_.sentAmount.value).toBe(BigInt(value))
     })
     test('Peer to Peer - Cross Currency', async (): Promise<void> => {
       const {
@@ -260,7 +260,8 @@ describe('Integration tests', (): void => {
       const senderWalletAddress = await c9.accounts.getByWalletAddressUrl(
         'https://cloud-nine-wallet-test-backend:3100/accounts/gfranklin'
       )
-      assert(senderWalletAddress?.walletAddressID)
+      assert(senderWalletAddress)
+      const senderAssetCode = senderWalletAddress.assetCode
       const senderWalletAddressId = senderWalletAddress.walletAddressID
       const value = '500'
       const createReceiverInput = {
@@ -277,12 +278,29 @@ describe('Integration tests', (): void => {
       }
 
       const receiver = await createReceiver(createReceiverInput)
+      assert(receiver.incomingAmount)
+      assert(
+        receiver.incomingAmount.assetScale === senderWalletAddress.assetScale
+      )
+      const receiverAssetCode = receiver.incomingAmount.assetCode
+
       const quote = await createQuote(senderWalletAddressId, receiver)
       const outgoingPayment = await createOutgoingPayment(
         senderWalletAddressId,
         quote
       )
-      await getOutgoingPayment(outgoingPayment.id, value)
+      const payment = await getOutgoingPayment(outgoingPayment.id, value)
+      const exchangeRate =
+        c9.config.seed.rates[receiverAssetCode][senderAssetCode]
+      const { sentAmount, receiveAmount } = payment
+
+      if (exchangeRate > 1) {
+        expect(sentAmount.value).toBeGreaterThan(receiveAmount.value)
+      } else if (exchangeRate < 1) {
+        expect(sentAmount.value).toBeLessThan(receiveAmount.value)
+      } else {
+        expect(sentAmount.value).toBe(receiveAmount.value)
+      }
     })
   })
 })
--- a/test/integration/integration.test.ts
+++ b/test/integration/integration.test.ts
@@ -35,7 +35,7 @@ describe('Integration tests', (): void => {
   describe('Requests', (): void => {
     test('Can Get Non-Existing Wallet Address', async (): Promise<void> => {
       const notFoundWalletAddress =
-        'https://host.docker.internal:4100/accounts/asmith'
+        'https://happy-life-bank-test-backend:4100/accounts/asmith'
 
       const handleWebhookEventSpy = jest.spyOn(
         hlb.integrationServer.webhookEventHandler,
@@ -83,24 +83,20 @@ describe('Integration tests', (): void => {
       const { consentInteraction } = testActions
 
       const receiverWalletAddressUrl =
-        'http://host.docker.internal:4100/accounts/pfry'
+        'https://happy-life-bank-test-backend:4100/accounts/pfry'
       const senderWalletAddressUrl =
-        'http://host.docker.internal:3100/accounts/gfranklin'
+        'https://cloud-nine-wallet-test-backend:3100/accounts/gfranklin'
       const amountValueToSend = '100'
 
       const receiverWalletAddress = await c9.opClient.walletAddress.get({
         url: receiverWalletAddressUrl
       })
-      expect(receiverWalletAddress.id).toBe(
-        receiverWalletAddressUrl.replace('http', 'https')
-      )
+      expect(receiverWalletAddress.id).toBe(receiverWalletAddressUrl)
 
       const senderWalletAddress = await c9.opClient.walletAddress.get({
         url: senderWalletAddressUrl
       })
-      expect(senderWalletAddress.id).toBe(
-        senderWalletAddressUrl.replace('http', 'https')
-      )
+      expect(senderWalletAddress.id).toBe(senderWalletAddressUrl)
 
       const incomingPaymentGrant = await grantRequestIncomingPayment(
         receiverWalletAddress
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

**Comment by mkurapov** [comment]
@BlairCurrey I think your intuition is right, #3 seems to be the best. 1 integration test = 1 jest test makes sense as well, since it encapsulates a single flow, and each step depends on the success of the previous. 

Likewise, abstracting into functions just makes individual test case easier to read as well :+1:

File: test/integration/lib/test-actions/index.ts [hunk 1]
```diff
@@ -0,0 +1,136 @@
...
+      consentInteraction(deps, outgoingPaymentGrant, senderWalletAddress),
+    consentInteractionWithInteractRef: (
+      outgoingPaymentGrant,
+      senderWalletAddress
+    ) =>
+      consentInteractionWithInteractRef(
+        deps,
+        outgoingPaymentGrant,
+        senderWalletAddress
+      ),
```
mkurapov, line 38: maybe we can pull these out into identity provider actions (or something along those lines)

test/integration/integration.test.ts
```diff
@@ -109,485 +65,222 @@ describe('Integration tests', (): void => {
...
+      const receiver = await createReceiver(createReceiverInput)
+      const quote = await createQuote(senderWalletAddressId, receiver)
+      const outgoingPayment = await createOutgoingPayment(
+        senderWalletAddressId,
+        quote
       )
-      expect(payment.state).toBe(OutgoingPaymentState.Completed)
-      expect(payment.receiveAmount.value).toBe(amountValueToSend)
-      expect(payment.sentAmount.value).toBe(amountValueToSend)
+      await getOutgoingPayment(outgoingPayment.id, value)
```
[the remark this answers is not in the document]
mkurapov: > want to assume a certain conversion rate/fee structure was used

Since we have access to the config file, would we be able to get the rates from there and confirm the amounts that way?

mkurapov (approved):
LGTM, nothing blocking, just some thoughts

test/integration/lib/integration-server.ts
```diff
@@ -1,7 +1,7 @@
+import crypto from 'crypto'
 import Koa from 'koa'
 import bodyParser from '@koa/bodyparser'
 import http from 'http'
-import { v4 as uuid } from 'uuid'
```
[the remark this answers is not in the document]
BlairCurrey: yeah I was using it it validate some id was a uuid but is was unecessary. It was used to generate a uuid but I used `crypto.randomUUID` instead. I guess normally it would be good to use the same uuid package everywhere but it was just for something throwaway in the test so i figured 1 less dependency was preferable.

test/integration/lib/test-actions/index.ts
BlairCurrey: We could. Although I was thinking it would be good to let these just be miscellaneous ones so we dont have to categorize everything. The admin api and open payments api are obvious but beyond that IDK how helpful it is to add the extra scope.

On a similar note, I originally was thinking I could do `AdminTestActions`, `OpenPaymentTestActions`, etc. instead of everything scoped to 1 `TestActions`. But then wondered where to throw these consent interaction ones. I guess `IdentityProviderTestActions` could be the answer but I didn't necessarily want to think of a scope for every potential new one.

test/integration/integration.test.ts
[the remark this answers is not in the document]
BlairCurrey: I think there are fees too, so we'd need to get that from the seed as well. And then duplicate the non-trivial backend logic and live with the fact that it could change, or extract that out and use it here. What do you think? Seem like not great options IMO. I suppose maybe if really want to verify those amounts I can just hardcode what they currently are and state the assumptions in a comment (fee, rates, starting amount)? 

I guess im not sure, but perhaps those details aren't really that critical in this test? We're testing (or should be) those calculations and whether or not they're returned in the backend already. I think the important part of this test is that we can get the payment and see that its complete.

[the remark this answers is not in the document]
njlie: Maybe checking that the conversion went in the right direction would be good enough? Like if we converted between currencies with a 2:1 ratio that the amount went down as opposed to up?

njlie (approved):

test/integration/integration.test.ts
```diff
@@ -109,493 +65,242 @@ describe('Integration tests', (): void => {
...
+      )
+      const payment = await getOutgoingPayment(outgoingPayment.id, value)
+      const exchangeRate =
+        c9.config.seed.rates[receiverAssetCode][senderAssetCode]
+      const { sentAmount, receiveAmount } = payment
+
+      if (exchangeRate > 1) {
+        expect(sentAmount.value).toBeGreaterThan(receiveAmount.value)
+      } else if (exchangeRate < 1) {
+        expect(sentAmount.value).toBeLessThan(receiveAmount.value)
```
mkurapov, line 300: will this always be true though? depending on the fees, you might run into a scenario where the exchange rate is 0.98 or something, but with an added fee, this will fail this assertion. I guess what this means is that:

> I suppose maybe if really want to verify those amounts I can just hardcode what they currently are and state the assumptions in a comment (fee, rates, starting amount).

hardcoding the assumptions will always have to be the case.

With that said, I think we should either directly verify the amount, or not do the check at all. We can also leave it for a future enhancement, if you want. Thoughts?

BlairCurrey: I thought of that so I changed the fee to 0, but I guess we would need to add some assertions checking the fee being 0. Which leads to you point about still having to hardcode assumptions.

I suppose making assertions (not jest `expect`, but `assert`) for fees, exchange rates, asset scale, asset code, and then expecting certain values is probably the better way to go. I'l include it here - this PR is still waiting on approval of the parent and the parents is waiting on the node 20 changes.

BlairCurrey: add [0dda0e4](https://github.com/interledger/rafiki/pull/2608/commits/0dda0e41afc85503c9fc8b5a9864ad91f71141f4)

feel free to give anymore feedback on it but im merging this branch into the parent.

test/integration/integration.test.ts
```diff
@@ -279,28 +279,46 @@ describe('Integration tests', (): void => {
...
         quote
       )
       const payment = await getOutgoingPayment(outgoingPayment.id, value)
+
+      const receiverAssetCode = receiver.incomingAmount.assetCode
       const exchangeRate =
-        c9.config.seed.rates[receiverAssetCode][senderAssetCode]
-      const { sentAmount, receiveAmount } = payment
+        hlb.config.seed.rates[senderAssetCode][receiverAssetCode]
+      const fee = c9.config.seed.fees.find((fee: Fee) => fee.asset === 'USD')
```
BlairCurrey, line 293: did some checks to ensure these are the operative seeds and fees.

PR #2610, opened 2024-03-26
feat(telemetry): track local payments

beniaminmunteanu:

This PR moves the amount metrics collection from the accounting service back to the ILP connector, as an extra middleware.

## Context

All context is provided in the linked issue
fixes #2505

## Checklist

- [x] Related issues: fixes #2505 
- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -208,11 +208,6 @@ export function initIocContainer(
     const knex = await deps.use('knex')
     const config = await deps.use('config')
 
-    let telemetry: TelemetryService | undefined
-    if (config.enableTelemetry && config.openTelemetryCollectors.length > 0) {
-      telemetry = await deps.use('telemetry')
-    }
-
     if (config.useTigerbeetle) {
       container.singleton('tigerbeetle', async (deps) => {
         const config = await deps.use('config')
@@ -226,7 +221,6 @@ export function initIocContainer(
 
       return createTigerbeetleAccountingService({
         logger,
-        telemetry,
         knex,
         tigerbeetle,
         withdrawalThrottleDelay: config.withdrawalThrottleDelay
@@ -235,7 +229,6 @@ export function initIocContainer(
 
     return createPsqlAccountingService({
       logger,
-      telemetry,
       knex,
       withdrawalThrottleDelay: config.withdrawalThrottleDelay
     })
@@ -371,7 +364,8 @@ export function initIocContainer(
       peerService: await deps.use('peerService'),
       ratesService: await deps.use('ratesService'),
       streamServer: await deps.use('streamServer'),
-      ilpAddress: config.ilpAddress
+      ilpAddress: config.ilpAddress,
+      telemetry: await deps.use('telemetry')
     })
   })
 
--- a/packages/backend/src/payment-method/ilp/connector/core/middleware/telemetry.ts
+++ b/packages/backend/src/payment-method/ilp/connector/core/middleware/telemetry.ts
@@ -1,0 +1,22 @@
+import { collectTelemetryAmount } from '../../../../../telemetry/transaction-amount'
+import { ILPContext, ILPMiddleware } from '../rafiki'
+
+export function createTelemetryMiddleware(): ILPMiddleware {
+  return async (
+    { request, services, accounts, state }: ILPContext,
+    next: () => Promise<void>
+  ): Promise<void> => {
+    if (
+      services.telemetry &&
+      Number(request.prepare.amount) &&
+      !state.unfulfillable
+    ) {
+      const { code, scale } = accounts.outgoing.asset
+      collectTelemetryAmount(services.telemetry, services.logger, {
+        amount: BigInt(request.prepare.amount),
+        asset: { code: code, scale: scale }
+      })
+    }
+    await next()
+  }
+}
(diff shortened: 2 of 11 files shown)


packages/backend/src/payment-method/ilp/connector/core/middleware/telemetry.ts
```diff
@@ -0,0 +1,22 @@
...
+      Number(request.prepare.amount) &&
+      !state.unfulfillable
+    ) {
+      const { code, scale } = accounts.outgoing.asset
+      collectTelemetryAmount(services.telemetry, services.logger, {
+        amount: BigInt(request.prepare.amount),
+        asset: { code: code, scale: scale }
+      })
+    }
+    await next()
```
mkurapov, line 20: this will end up counting the amount even if the packet is rejected at the destination.

Maybe we can call `next()` first so that this effectively ends up being the last middleware to run (we can also just put it as the last one in the middleware chain), and check if `ctx.response.fulfill` is defined. This way, only fulfilled packets will be tracked.

packages/backend/src/index.ts
```diff
@@ -371,7 +364,8 @@ export function initIocContainer(
       peerService: await deps.use('peerService'),
       ratesService: await deps.use('ratesService'),
       streamServer: await deps.use('streamServer'),
-      ilpAddress: config.ilpAddress
+      ilpAddress: config.ilpAddress,
+      telemetry: await deps.use('telemetry')
```
mkurapov, line 368: we should check if telemetry is defined here, otherwise `await deps.use('telemetry')` will throw

test/integration/lib/config.ts
```diff
@@ -41,7 +41,9 @@ const loadEnv = (filePath: string): EnvConfig => {
   })
 
   if (missingKeys.length > 0) {
-    const errorMessage = `Missing required environment variable(s): ${missingKeys.join(', ')}`
+    const errorMessage = `Missing required environment variable(s): ${missingKeys.join(
+      ', '
+    )}`
```
mkurapov, line 46: this was changed as a result of your formatter?

mkurapov (commented):
What would you like to do re: 'transactions_total' metric? Should we only call it if the receiver is at a remote rafiki?

test/integration/lib/config.ts
beniaminmunteanu: oh, yes it was.

packages/backend/src/index.ts
beniaminmunteanu: image
It returns Promise<TelemetryService> | undefined, both of which are handled by the optionality of the telemetry dep in createConnectorService.
Or am i getting this wrong, deps.use('someThingThatIsNotRegistered') would throw indefeneatly?

packages/backend/src/payment-method/ilp/connector/core/middleware/telemetry.ts
beniaminmunteanu: I would go for the 2nd option you mentioned.

beniaminmunteanu: ended up going for the 1st 😁

packages/backend/src/payment-method/ilp/connector/core/test/middleware/telemetry.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+  state: {
+    unfulfillable: false,
+    incomingAccount: {
+      quote: 'exists'
+    }
+  }
+})
+
+jest.mock('../../../../../../telemetry/transaction-amount')
```
mkurapov, line 34: do we need this one here? since you are already calling `mockImplementation` when creating the spy

packages/backend/src/index.ts
beniaminmunteanu: actually you were right...noticed it today. Resolved

packages/backend/src/payment-method/ilp/connector/core/test/middleware/telemetry.test.ts
beniaminmunteanu: yes, in order to mock the implementation, the function must be a jest.fn()
the spies no longer work without that mock

mkurapov (commented):
Looks good, just some test comments.

packages/backend/src/payment-method/ilp/connector/core/test/middleware/telemetry.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    let nextCalled = false
+    const next = jest.fn().mockImplementation(() => {
+      nextCalled = true
+      return Promise.resolve()
+    })
+
+    const collectAmountSpy = jest
+      .spyOn(telemetry, 'collectTelemetryAmount')
+      .mockImplementation(() => {
+        expect(nextCalled).toBe(true)
```
mkurapov, line 103: Ah, I understand, this ensures that next was called _before_ the `collectTelemetryAmount` :+1:

I was looking at some more of this, and this is something that exists in [jest-extended](https://jest-extended.jestcommunity.dev/docs/matchers/mock/#tohavebeencalledbefore), but not in the standard jest lib unfourtunately. 
We can keep this test as is.

mkurapov (approved):

beniaminmunteanu:
closed in favor of #2626 (which is a fresh duplicate). Will delete this source branch once that one gets merged

Issue #55, opened 2021-07-21
Expand on "New to interledger" section in readme


cairin:
Should we include links to the Interledger docs or something that introduces people to the different features of Interledger that Rafiki will expose? I like what they did here https://github.com/mozilla/glean.js

vladimirfomene:
@cairin , I think this will help people who are new to the project like me. Knowing what parts of the Interledger protocol suite are implemented will help newbies quickly get a grasp of the codebase.

vladimirfomene:
@cairin is there documentation, where I can get a hold of the Rafiki architecture, understand what services/API it exposes?

cairin:
@vladimirfomene #71 adds the documentation you're looking for.

vladimirfomene:
Thanks @cairin! My weekend will be interesting!

PR #2622, opened 2024-03-30
feat(migrations): support for running database migrations manually

### LINKED ISSUE #2621: [BUG] k8s pod is in running state if database migration fails
## Bug Report

**Describe the bug**
When running in kubernetes and migrations fails, pod will still be alive. Pod should not be alive if prerequsites like database migrations fail. 

Kubernetes way of doing this thing is with [k8s init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/). If this container fails, whole pod should fail because Rafiki in that case is unusable.

**To Reproduce**

1. Install one of the latest rafiki packages (for example auth) to k8s cluster
2. Connect to the database and truncate knex_migrations directory
3. Restart pod

After that, check the logs of the pod. You will see that there are a bunch of error messages due to failing migrations.

**Expected behavior**
If migration fails, do not start pod. In most cases, if pod is not started alert should be triggered.

**Additional context**
In order for this to work, I suppose there should be NPM command that will trigger migrations. Also, to support k8s and bare metal instances, we will probably need another env variable like MANUAL_DB_MIGRATIONS. When this variable is set, that would mean that db migrations will need to be run via npm with command like `npm run db:migrations:run`.

With this command we can also get rid of retry mechanism for migrations, since kubernetes will take over this part.

**Screenshots**

Kubernetes container logs - failing DB migrations
k8s container logs

Kubernetes pod status - 13 minutes without restart despite failing DB migrations
k8s pod status - 13 minutes without restart despite failing db migrations

golobitch:
* Introduce new environmental variable `ENABLE_MANUAL_MIGRATIONS` through which end user can decide if he/she wants to run migrations automatically (default flow as it always was) or manually (`pnpm run knex -- migrate:latest --env production`
* In `knexfile.js` use env variable from the ENV for setting connection string for production environment
* In dockerfile, copy knexfile.js from source to container

With this in place, if Rafiki is installed on kubernetes cluster, it is possible to leverage initContainers for database migrations. This should be backward compatible or in other words, should not change how Rafiki deals with database migrations when default values for this environment variable is set.

## Context

fixes #2621 

After this PR is merged, we will need to implement this ENV into helm charts. After this ENV is added to helm charts, we need to set value to `true`, and we will also need to specify initContainer that will run this migrations "manually" or in other words, automatically in initContainer.

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


packages/auth/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
 FROM base AS prod-deps
 
 COPY package.json pnpm-workspace.yaml .npmrc ./
+COPY packages/auth/knexfile.js ./packages/auth/knexfile.js
```
golobitch, line 20: this line here is not really needed (same for `packages/backend/Dockerfile` however, I like consistency (line #55)

golobitch:
This PR has been approved, however, I am not authorized to merge this pull request.

sabineschaller (approved):
Thank you!

Issue #2629, opened 2024-04-03
Fix single phase transfer accounting


mkurapov:
## Context

After merging in:
- https://github.com/interledger/rafiki/pull/2564

we realized that we are unable to complete incoming payments on a receiving Rafiki, because of the way we are handling single phase transfers.

Basically, during a single phase transfer (without a timeout), we successfully send money to the underlying incoming payment liquidity account when calling `accountingService.createTransfer`, _but_ since we never call `transaction.post` -> we never end up calling `incomingPayment.onCredit()`, which means, we never end up marking the incoming payment as completed, even though it has received the correct amount of funds.

What we should do (IMO), is update the accountingService to split out `createTransfer` into `createSinglePhaseTransfer` and `createTwoPhaseTransfer` to remove any confusion about the current behaviour of `createTransfer`. `createSinglePhaseTransfer` will return `Promise<TransferError | void>`  while `createTwoPhaseTransfer` will have the same signature as `createTransfer`: `Promise<Transaction | TransferError>`, where `Transaction`. 

Both function should call `onDebit` and `onCredit` for the respective destination and source accounts upon the successful completion of the transfer.

Additionally, I think it would make sense to have a base abstract `AccountingService` class, in order to simplify the current `createAccountToAccountTransfer` function:

```ts
export abstract class AccountingService implements AccountingService {
  constructor() {}
  abstract getBalance(id: string): Promise<bigint | undefined>
  abstract getTotalSent(id: string): Promise<bigint | undefined>
  abstract getAccountsTotalSent(ids: string[]): Promise<(bigint | undefined)[]>
  ...

  public async createTwoPhaseTransfer(
    transferArgs: TransferOptions
  ): Promise<TransferError | Transaction> {}
  public async createSinglePhaseTransfer(
    transferArgs: TransferOptions
  ): Promise<TransferError | Transaction> {}
```

This can be done in several PRs, the first one to make the abstract class (while keeping all of the same functionality of `createTransfer`), and a second one to support `createTwoPhaseTransfer` and `createSinglePhaseTransfer`

PR #2630, opened 2024-04-03
chore: Revert "feat(backend): support single phase transfer in connector"

mkurapov:
This reverts commit 81f8a059229d4d8e1ce714cf12605f5a17d127d3.

- Reverting commit that introduced single phase transfers: we ended up having an issue of incoming payments not being able to complete, more details in #2629 
- Adding to integration test to make sure this case is tested

## Context

Progress towards fixing #2629 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


test/integration/integration.test.ts
```diff
@@ -305,21 +316,30 @@ describe('Integration tests', (): void => {
...
         assetScale: 2,
         value: 550n
       })
+
+      const incomingPaymentId = receiver.id.split('/').slice(-1)[0]
+      const incomingPayment = await getIncomingPayment(incomingPaymentId)
+      expect(incomingPayment.receivedAmount).toMatchObject({
+        assetCode: 'EUR',
+        assetScale: 2,
+        value: 501n
```
mkurapov, line 340: I understand that the receivedAmount may be greater than the sentAmount during cross currency transactions (i.e. you need 0.02 USD to reach a minimum of 0.01 EUR).

However, this still feels wrong to me:

original receiveAmount = 500
sentAmount = 550
receivedAmount = 501

with an exchange rate of 0.91 as tested above, we have 550 * 0.91 = 500.5, which is under the received amount of 501. This signifies somewhere in ILP pay there is a `Math.ceil` that ends up making money out of nothing. Are my assumptions correct in the fact that _this_ seems like a problem?

sabineschaller: I was about to say that this could be because the stream was closed too late but given that the packet size is huge, that can't be it. I guess we have a rounding problem somewhere.

sabineschaller (commented):
Thanks for adding the additional tests!

sabineschaller (approved):
I don't know why I didn't approve earlier... 🙃

test/integration/integration.test.ts
mkurapov: Unfourtunately will need to dive in deeper into ILP pay, since this is just the default behaviour. Will merge in for now, since it is the status quo at this point anyway.

PR #2632, opened 2024-04-03
feat(backend,frontend,mock-ase): use hmac signature to secure admin api

### LINKED ISSUE #2218: Design auth mechanism for admin graphql API
## Context

Right now, there is no auth on the admin API endpoints and we tell integrators to not expose the graphql port. We want to add some auth mechanism, probably, API keys, to secure the admin APIs.

### LINKED ISSUE #2568: Add Admin API authorization capability to Bruno collection

### LINKED ISSUE #2596: Secure Admin API with HMAC signature
## Context
To follow established patterns in other services and use a low-overhead solution, it has been decided to secure the Admin API with an HMAC signature, using a shared secret between it and ASEs/the Admin UI.

Signature generation can be duplicated from:
https://github.com/interledger/rafiki/blob/5c0ee4b76550aab40c77eed1a5338004ccd52823/packages/backend/src/webhook/service.ts#L218-L235

## Todos
- [ ] Add HMAC signature verification to backend
- [ ] Add shared secrets to backend, mock-ase, and frontend
- [ ] Add signature generation to mock-ase and frontend
- [ ] Add signature generation to integration tests

njlie:

- Adds middleware to the `backend` Admin Server that validates a signature comprised of a timestamp and a HMAC digest of the request body.
- Adds an `ApolloLink` to the Apollo Clients of `frontend` and `mock-ase` that generates a signature from a timestamp and the request body.

## Context

Fixes #2218 and fixes #2596 and fixes #2568.

This signature scheme is pulled from the one used for the webhook service on the backend: 
https://github.com/interledger/rafiki/blob/main/packages/backend/src/webhook/service.ts#L218-L235

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


bruno/collections/Rafiki/scripts.js
```diff
@@ -1,11 +1,16 @@
...
 const url = require('url')
 
 const scripts = {
   sanitizeUrl: function () {
     return req
       .getUrl()
-      .replace(/{{([A-Za-z]\w+)}}/g, (_, key) => bru.getEnvVar(key))
+      .replace(
+        /{{([A-Za-z]\w+)}}/g,
+        (_, key) => bru.getVar(key) || bru.getEnvVar(key)
```
njlie, line 12: Bruno Collection variables take precedence over Bruno Environment Variables, [per the docs.](https://docs.usebruno.com/scripting/javascript-reference#collection-variables)

sabineschaller: But we don't have collection variables, do we?

packages/backend/src/config/app.ts
```diff
@@ -138,6 +138,8 @@ export const Config = {
   signatureSecret: process.env.SIGNATURE_SECRET, // optional
   signatureVersion: envInt('SIGNATURE_VERSION', 1),
 
+  apiSecret: process.env.API_SECRET, // optional
```
BlairCurrey, line 141: My understanding is that the `apiSecret` is an optional environment variable that secures the admin api if provided, and if not, it's unsecured. I'm just wondering if we want it to be secure by default instead? I guess this way is more backwards compatible I just find my initial assumptions were secure by default.

Noticed when I was surprised the integration tests didn't fail.

packages/backend/src/app.ts
```diff
@@ -367,6 +370,13 @@ export class App {
       }
     )
 
+    koa.use(async (ctx, next: Koa.Next): Promise<void> => {
+      if (!verifyApiSignature(ctx, this.config) && this.config.apiSecret) {
```
BlairCurrey, line 374: nit:

Perhaps only add the middleware if `this.config.apiSecret` then always verify in the middleware? 1 less middleware to traverse when not secured and its not like the config will change during runtime.

packages/frontend/app/lib/apollo.server.ts
```diff
@@ -16,18 +26,48 @@ BigInt.prototype.toJSON = function (this: bigint) {
...
+      ...headers,
+      signature: `t=${timestamp}, v${version}=${digest}`
+    }
+  }
+})
+
+const httpLink = createHttpLink({
+  uri: process.env.GRAPHQL_URL ?? 'http://localhost:3001/graphql'
+})
+
```
BlairCurrey, line 58: It would be nice if we could share this between `frontend` and `mock-account-servicing-entity`. And potentially `integration` in the future. I recall @sabineschaller brought up a similar need for something else recently (I don't recall what exactly).

I could definitely see a case for a `lib`, `rafiki-lib`, or `internal` (to borrow [a Go convention](https://github.com/golang-standards/project-layout?tab=readme-ov-file#internal)) where this and anything else we wanted to share internally (standardized pino log factory function?).

I guess we dont necessarily need to do that in this PR but but this certainly illustrates the point.

bruno/collections/Rafiki/scripts.js
njlie: I didn't check _every_ API request for this, but _Deposit Asset Liquidity_ definitely does with `transferId` and `idempotencyKey`.

packages/backend/src/config/app.ts
njlie: I couldn't remember where I heard this, but I had thought I'd heard at some point that securing the admin apio should be optional.

BlairCurrey: Yeah I recall that as well. Just wondering if it should be opt-in or opt-out. It's not necessarily a suggestion, just thinking out loud. 

I generally assume people should have to explicitly opt-out of the more secure way but perhaps in this specific case it doesnt matter? Since its currently insecure and not meant to be exposed to the internet. I think the configuration (how you have it now) is probably simpler than enabling by default (by expecting `apiSecret` unless some additional env var is set to bypass it or something).

bruno/collections/Rafiki/Examples/Admin API -  only locally/Peer-to-Peer Cross Currency Payment/Create Receiver (remote Incoming Payment).bru
```diff
@@ -55,6 +55,31 @@ body:graphql:vars {
...
+}
+
+script:pre-request {
+  const { createHmac } = require('crypto');
+  const { canonicalize } = require('json-canonicalize');
+  
+  const body = req.getBody();
+  
+  //console.log(body);
+
```
BlairCurrey, line 70:
```suggestion
```

bruno/collections/Rafiki/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "rafiki-bruno-collection",
+  "version": "1.0.0",
+  "dependencies": {
+    "json-canonicalize": "^1.0.6"
+  }
+}
```
BlairCurrey, line 7: I think we should probably add this as a package to our pnpm workspace so `pnpm i` will install these dependencies. Took me some time to figure out why bruno couldn't find `json-canonicalize`. 

In fact I'm not even sure how to install these with pnpm otherwise... doing `pnpm i` from `bruno/collections/Rafiki` just installs the workspace packages (and not this).

njlie: I think I managed it by running `pnpm init` inside the directory to generate the `package.json` on my machine. But you're right, it should be listed in the pnpm workspace.

pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ packages:
   - 'packages/*'
   - 'localenv/mock-account-servicing-entity'
   - 'test/integration'
+  - 'bruno/'
```
BlairCurrey, line 5: this isn't working for me. doesn't find the package. think it should be:

```yaml
  - 'bruno/collections/Rafiki'
```

packages/backend/src/shared/utils.ts
```diff
@@ -91,3 +95,22 @@ export async function poll<T>(args: PollArgs<T>): Promise<T> {
     // eslint-disable-next-line no-constant-condition
   } while (true)
 }
+
+export function verifyApiSignature(ctx: Context, config: IAppConfig): boolean {
```
JoblersTune, line 99: Should we perhaps include a signature version check in here as well?

njlie: Added an env var for the backend to set its signature version for the api signature :+1:

BlairCurrey (approved):

BlairCurrey (approved):

Issue #1905, opened 2023-09-19
Standardize Open Payment errors


mkurapov:
We should standardize errors that get thrown in the Open Payments API, so we can return proper error codes and messages in the API response. 

Potentially, we could do something like: 

```ts
class OpenPaymentsError extends Error { 
    constructor(message: string, status: number) {
      super(`Open Payments API error: ${message}`);
      this.name = 'OpenPaymentsError';
      this.status = status
    }
}


throw new OpenPaymentsError('Could not get Incoming Payment', 404); // usage
```
and then use a middleware to properly format those errors when thrown in the Open Payment routes.
This will also help standardize usage for the Open Payments client: https://github.com/interledger/open-payments/tree/main/packages/open-payments

anurag629:
@sabineschaller Please assign this to me.

Jacintomuambumba:
Can you assign this issue to me so I can work on it?

mkurapov:
Hey @anurag629 & @Jacintomuambumba, did you end up starting on this issue?

Jacintomuambumba:
Iám working

mkurapov:
Reopening while I add additional related PRs

PR #2635, opened 2024-04-04
feat(backend): adding OpenPaymentsServerRouteError

mkurapov:

- Adding `OpenPaymentsServerRouteError` that is caught by a new `openPaymentsServerErrorMiddleware` (this middleware runs as the last middleware), which will properly handle all of errors thrown in the Open Payments Resource Server routes. This means:
    - we wont ever need to use `ctx.throw` in the routes, we just throw `OpenPaymentsServerRouteError` which will automatically be handled in the middleware
    - all logging is standardized for errors: response code + request body, other details are properly logged
    - The new `@interledger/openapi` package exports `OpenAPIValidatorMiddlewareError`, which is now also handled explicitly in the middleware

## Context

Work toward #1905 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -85,47 +87,64 @@ describe('Auth Middleware', (): void => {
...
-      ctx.throw = jest.fn().mockImplementation(() => {
-        throw error
-      }) as never
 
-      await expect(middleware(ctx, next)).rejects.toBe(error)
+      jest
+        .spyOn(tokenIntrospectionClient, 'introspect')
+        .mockResolvedValueOnce({ active: true, access: {} } as TokenInfo) // causes an error other than OpenPaymentsServerRouteError
+
+      expect.assertions(3)
```
mkurapov, line 108: Handy `expect.assertions(num)` method when testing errors in `try/catch` to make sure the number assertions actually happened (instead of being ignored)

packages/backend/src/open_payments/errors.ts
```diff
@@ -0,0 +1,65 @@
...
+  } catch (err) {
+    const logger = await ctx.container.use('logger')
+
+    const baseLog = {
+      method: ctx.req.method,
+      path: ctx.path
+    }
+
+    if (err instanceof OpenPaymentsServerRouteError) {
+      logger.info(
```
mkurapov, line 35: info because its "expected" behavior

packages/backend/src/open_payments/payment/outgoing/routes.test.ts
```diff
@@ -129,38 +125,6 @@ describe('Outgoing Payment Routes', (): void => {
     })
   })
 
-  describe('get', () => {
-    test('returns 500 for unexpected error', async (): Promise<void> => {
```
mkurapov, line 133: all properly handled in `openPaymentsServerErrorMiddleware`

packages/backend/src/open_payments/payment/outgoing/routes.ts
```diff
@@ -151,3 +133,13 @@ function outgoingPaymentToBody(
...
 }
+
+function handleMissingWalletAddress(
+  deps: ServiceDependencies,
+  outgoingPayment: OutgoingPayment
+): never {
+  const errorMessage =
+    'Outgoing payment does not have wallet address. This should not be possible.'
+  deps.logger.error({ outgoingPaymentId: outgoingPayment.id }, errorMessage)
+  throw new Error(errorMessage)
```
mkurapov, line 144: Just throwing normal error here since it's something quite unexpected, and shouldn't ever actually happen

BlairCurrey: definitely agree with your reasoning here.

this is not a suggestion, just a stackoverflow post with some interesting food for thought regarding logging errors and throwing them: https://stackoverflow.com/questions/22072343/throw-exception-vs-logging

I did some similar work previously in cleaning up errors/logs and this is triggering flashbacks. One of the things I remember was our GCP error reports would include duplicate (but not obviously duplicate) errors because of this. Could be dependent on structured logger configuration and cloud error reporting/logging services. In these scenarios I changed them to warnings so they didn't show in error reports but could still be filtered in the log viewer. 

I'm just throwing out 1 anecdote and not claiming this is ideal. There are dissenting opinions in that post as well. Just thought it was relevant.

packages/backend/src/open_payments/errors.ts
BlairCurrey: just digesting this out loud... looks like the paths are:

error is `OpenPaymentsServerRouteError`:
- logs at info level with details
- throws with status and message

error is `OpenAPIValidatorMiddlewareError`:
- logs at info level with details
- throws with status and message

anything else:
- logs at error level with details
- throws generic 500

packages/backend/src/open_payments/payment/outgoing/routes.ts
BlairCurrey: Looks like I should also look at the middleware handling this...

I guess in this case it will:

- log at error level
- log again at error level in `openPaymentsServerErrorMiddleware` with some additional generic details
- throw generic 500

mkurapov: @BlairCurrey definitely makes sense, and those order of events are correct. 

I think I would also prefer just to keep only the "main" log in the middleware, but I think the only reason I included this was because I also wanted to log the outgoingPaymentId, which otherwise wouldn't get logged. I don't have a nice solution of propagating that id out to the log in the middleware (particularly for the create routes)

BlairCurrey: yeah... that seems fair. I dont think we can expect to log all the context in some central place away from where the actually error happens.

packages/backend/src/app.ts
```diff
@@ -403,17 +405,26 @@ export class App {
     const { resourceServerSpec, walletAddressServerSpec } =
       await this.container.use('openApi')
 
+    const validatorMiddlewareOptions = {
+      validateRequest: true,
+      validateResponse: process.env.NODE_ENV !== 'production'
```
mkurapov, line 410: This is what the `createValidatorMiddleware` function was doing previously, now to keep the same behaviour, it has to be stated explicitly 

https://github.com/interledger/open-payments/releases/tag/%40interledger%2Fopenapi%402.0.0

packages/backend/src/open_payments/wallet_address/model.ts
```diff
@@ -229,3 +230,15 @@ export abstract class WalletAddressSubresource extends BaseModel {
...
+  deps: { logger: Logger },
+  resource: WalletAddressSubresource
+): asserts resource is WalletAddressSubresource &
+  Required<Pick<WalletAddressSubresource, 'walletAddress'>> {
+  if (!resource.walletAddress) {
+    const errorMessage = `${resource.$modelClass.name} does not have wallet address. This should be investigated.`
+    deps.logger.error({ id: resource.id }, errorMessage)
+    throw new Error(errorMessage)
+  }
+}
```
mkurapov, line 244: Function checks whether incoming payment, quote or outgoing payment has defined `walletAddress` property. Maybe the log message could be less about investigation though? The only way this would happen if we don't add `walletAddress` wasn't added to the `withGraphFetched()` function when getting the resource.

Ideally this would all be resolved with types (ie. we could specify that if `walletAddress` is in the `withGraphFetched` then we can have that directly in the type -> but that would be a pretty big change, although it would be handy I think?

packages/backend/src/open_payments/route-errors.ts
```diff
@@ -0,0 +1,79 @@
+import { Logger } from 'pino'
+import { AppContext } from '../app'
+import { OpenAPIValidatorMiddlewareError } from '@interledger/openapi'
+import { WalletAddressSubresource } from './wallet_address/model'
+
+export class OpenPaymentsServerRouteError extends Error {
```
mkurapov, line 6: @njlie what do you think about having something like this in the `auth` package where we would define a custom `GNAPError` class and catch it similarily in a middleware?

packages/backend/src/open_payments/wallet_address/model.ts
BlairCurrey: > Maybe the log message could be less about investigation though

I guess the pertinent thing is that its just (I assume) completely unexpected.

Is this based off any experience with a missing wallet address? I dont see this used so not exactly sure what the intention is and wondering why we need it.

BlairCurrey:
Is the open api request validation not working for you? I deleted the wallet address from the create incoming payment body in bruno and it makes it all the way to objecting before erroring because its missing

image

mkurapov:
@BlairCurrey 
That is coming from `createWalletAddressMiddleware`, which is called before the validator middleware. 
Unfourtunatetly I still need to update `createWalletAddressMiddleware` to properly throw the new error type, but I was thinking of doing it in a separate PR since this already is getting a bit too big.

If you try changing, or passing an incorrect value for a different field, you should see the correct validation error:



```
{"level":30,"time":1712933544211,"pid":1,"hostname":"happy-life-bank-backend","method":"POST","path":"/incoming-payments","message":"Received error validating OpenAPI request: body must NOT have additional properties: sadfdsa","status":400,"msg":"Received OpenAPI validation error when handling Open Payments request"}
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -130,21 +138,19 @@ export function createTokenIntrospectionMiddleware({
...
-        ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
-      } else {
+      ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
+
+      if (!bypassError) {
         throw err
       }
     }
+
+    await next()
```
BlairCurrey, line 153: no feedback, just a little retrospective: I wish I would have not added `bypassError` flag here and instead factored that use-case out into a separate function (kinda similar to `throwIfSignatureInvalid` and `authenticatedStatusMiddleware`). makes changes like this a lot harder to reason about. I added when implementing the public incoming payment.

Changes look fine though... getting public payment works as expected.

packages/backend/src/open_payments/payment/incoming/routes.test.ts
```diff
@@ -126,23 +121,6 @@ describe('Incoming Payment Routes', (): void => {
...
-        reqOpts: {
-          headers: { Accept: 'application/json' }
-        },
-        walletAddress
-      })
-      await expect(incomingPaymentRoutes.list(ctx)).rejects.toMatchObject({
-        status: 500,
-        message: `Error trying to list incoming payments`
-      })
-    })
```
BlairCurrey, line 145: are we removing this because the case is already covered by the general error handling middleware? I realize the rejects assertions here are no longer correct but I'm just trying to verify the new behavior is tested.

BlairCurrey (dismissed):

packages/backend/src/open_payments/wallet_address/model.ts
mkurapov: It's only to make typescript happy:

The  `WalletAddressSubresource` model:
```ts
export abstract class WalletAddressSubresource extends BaseModel {
  ...
  public readonly walletAddressId!: string
  public walletAddress?: WalletAddress
```

In order to have `walletAddress` defined we need to include it in any `withGraphFetched` requests. However, Objection of course wouldn't know anything about typescript types, so we're left with `walletAddress?: WalletAddress`. In our routes, we need the `walletAddress` to be defined to do e.g.:
```
  ctx.body = incomingPayment.toOpenPaymentsTypeWithMethods(
    incomingPayment.walletAddress,
    streamCredentials
  )
```

when calling function we just assert that the wallet subresource will for sure have walletAddress defined.

BlairCurrey: Ah, I see. For some reason I didn't see the usage until now.

Would we save ourselves some trouble by doing it in the service instead? where we are specifying `.withGraphFetched('[asset, walletAddress]')`? Presumably it's always a problem if we dont find the `walletAddress` in such cases.

Or is it acting as a typegaurd that wouldn't work if used like that? It kinda looks that way but commenting out it's usage doesn't seem to result in any errors for me.

sabineschaller (approved):

PR #2643, opened 2024-04-05
feat(walletAddressKeys): list wallet address keys

### LINKED ISSUE #2631: [FEATURE REQUEST] Admin API query for fetching walletAddressKeys
## Feature Request

**Describe the feature you'd like to request**

I would like to have query that would return walletAddressKeys.
Use case is actually quite simple: 
* For Key management

If ASE wants to implement Key management, one of the important thing is also to show keys. This is currently possible by performing following call:
```
GET {walletAddress}/jwks.json
```

This one is fine, but apparently `jwks.json` does not include id, which is needed if ASE wants to revoke key.

**Describe the solution you'd like**
Probably the best solution is to have FieldResolver inside of the WalletAddress object. Following query would be perfect IMHO

```
query WalletAddressData(id: String!) {
    walletAddress(id: $id) {
        id
        url
        keys: {
            id
            name
            createdAt
            publicKey
        }
    }
}
```

Something like this :)

**Describe alternatives you've considered**
Alternative solution for ASE is to save keys in own database / table / whatever. Or maybe save mapping between walletAddressKey and own key data for the purpose of having all the data in some persistent storage.

**Additional context**
N/A

golobitch:
* Add API for fetching wallet address keys

## Context

fixes #2631 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/resolvers/walletAddressKey.ts
```diff
@@ -4,10 +4,46 @@ import {
...
+    parent,
+    args,
+    ctx
+  ): Promise<ResolversTypes['WalletAddressKeyConnection']> => {
+    const walletAddressKeyService = await ctx.container.use(
+      'walletAddressKeyService'
+    )
+
+    const getPageFn = (pagination_: Pagination) => {
+      if (!parent.id) throw new Error('missing wallet address id')
```
mkurapov, line 26: could we do this check as the first thing in the resolver? (then we could probably get rid of `getPageFn` altogether)

packages/backend/src/graphql/resolvers/walletAddressKey.ts
```diff
@@ -4,10 +4,46 @@ import {
...
+      return walletAddressKeyService.getPage(parent.id, pagination_)
+    }
+
+    const { ...pagination } = args
+
+    const walletAddressKeys = await getPageFn(pagination)
+    ctx.logger.info({ walletAddressKeys }, 'got walletAddressKeys')
```
mkurapov, line 33:
```suggestion
```

packages/backend/src/graphql/resolvers/walletAddressKey.ts
```diff
@@ -4,10 +4,46 @@ import {
...
-  Crv
+  Crv,
+  WalletAddressResolvers
 } from '../generated/graphql'
 import { ApolloContext } from '../../app'
 import { WalletAddressKey } from '../../open_payments/wallet_address/key/model'
+import { Pagination } from '../../shared/baseModel'
+import { getPageInfo } from '../../shared/pagination'
+
+export const getWalletAddressKeys: WalletAddressResolvers<ApolloContext>['walletAddressKeys'] =
```
mkurapov, line 15: we should write a test for this. see if you can repurpose `getPageTests` (`packages/backend/src/shared/baseModel.test.ts`) maybe.

Would be good to test `getWalletAddressKeyPage` too.

packages/backend/src/graphql/resolvers/walletAddressKey.ts
BlairCurrey: @tadejgolobic would you recommend applying this logging pattern more generally to other resources? in our http routes/gql resolvers? 

I know Max is already working on stuff, but at that point I wonder about simply adding a logging middleware that logs the response (and request for the matter). Although perhaps this is upstream enough from that that doing it here is still useful too.

golobitch: For me, this was actually a debug log, however, it wouldn't hurt to have this kind of logs. So basically on every request, to log URL and any parameters / body that will be passed and ofc response. This would show only if the log level is set to debug.

However, we should probably make sure, that we are not exposing some sensitive data in the log. This is probably the hardest thing to do

packages/backend/src/graphql/resolvers/walletAddressKey.ts
golobitch: Yes, I was able to use `getPageTests` and I also wrote tests for `getWalletAddressKeyPage`

packages/backend/src/graphql/resolvers/walletAddressKey.ts
sabineschaller: > For me, this was actually a debug log, however, it wouldn't hurt to have this kind of logs. So basically on every request, to log URL and any parameters / body that will be passed and ofc response. This would show only if the log level is set to debug.

I'm currently working on this.

packages/backend/src/graphql/resolvers/walletAddressKey.ts
```diff
@@ -4,10 +4,51 @@ import {
...
+    const order = sortOrder === 'ASC' ? SortOrder.Asc : SortOrder.Desc
+
+    const walletAddressKeys = await walletAddressKeyService.getPage(
+      parent.id,
+      pagination,
+      order
+    )
+    const pageInfo = await getPageInfo({
+      getPage: (pagination_?: Pagination, sortOrder_?: SortOrder) =>
+        walletAddressKeyService.getPage(parent.id!, pagination_, sortOrder_),
```
mkurapov, line 40: I'm definitely confused why `parent.id` is defined after explicitly checking for it, but I guess that's some weird TS shenanigans 🤷‍♂️

mkurapov (commented):
Looks good :+1: .
Last thing, we should add a Bruno query for `Get Wallet Address Keys`

packages/backend/src/graphql/resolvers/walletAddressKey.ts
golobitch: @mkurapov apparently `parent.id` is of type `(property) id?: string | undefined`. And even though I am checking if `parent.id` is undefined and then throwing error, without this explicitly marking as defined, it doesn't go through.

I have also updated bruno collection

golobitch:
> Looks good 👍 . Last thing, we should add a Bruno query for `Get Wallet Address Keys`

@mkurapov I added Bruno query for `Get Wallet Address Key` instead of a plural form, just to be consistent with other examples. There is also field resolver for incoming and outgoing payments.

sabineschaller:
> @mkurapov I added Bruno query for Get Wallet Address Key instead of a plural form, just to be consistent with other examples. There is also field resolver for incoming and outgoing payments.

We do have "Get Peers" and "Get Payments" so I think the plural would actually be consistent since this uses pagination?

golobitch:
@mkurapov I have added new Bruno query for fetching wallet addresses keys. We can do another iteration of this PR.

bruno/collections/Rafiki/Rafiki Admin APIs/Get Wallet Addresses Keys.bru
```diff
@@ -0,0 +1,43 @@
...
+body:graphql {
+  query GetWalletAddressesKeys {
+      walletAddresses {
+          edges {
+              cursor
+              node {
+                  id
+                  publicName
+                  url
+                	walletAddressKeys {
```
mkurapov, line 22: nitpick: formattin

(don't think prettier knows what .bru files are yet)

golobitch: Fixed.

mkurapov (approved):

PR #2663, opened 2024-04-16
fix(webhook): move max retry check in query

### LINKED ISSUE #2664: [BUG] Webhook check for max retries causes processing of the same event
Thanks for taking the time to file a bug report! Please fill out this form as completely as possible.

## Bug Report
In recent release a new check for webhooks attempts was introduced that checks for max retries attempts ([PR](https://github.com/interledger/rafiki/pull/2541/files#diff-efa066cc9f163e17bcc35bfaa5a4bc80a60c56789c87f6b2e3dde96f5910013c)).
As the check is happening in the code that is causing the processing of the same events that reached the max attempts and does not go further to newly created webhooks. That is causing the system to be stuck and not processing new payments as the webhooks are not sent.
The events that reached max attempts should be ignored and not processed anymore.

dragosp1011:

- 

## Context

fixes #2664

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/webhook/service.test.ts
+++ b/packages/backend/src/webhook/service.test.ts
@@ -311,7 +311,10 @@ describe('Webhook Service', (): void => {
       })
     })
 
-    function mockWebhookServer(status = 200): Scope {
+    function mockWebhookServer(
+      status = 200,
+      expectedEvent: WebhookEvent = event
+    ): Scope {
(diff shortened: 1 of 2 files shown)


mkurapov (approved):

mkurapov (approved):
Test looks great, thanks for adding it & fixing the bug!

PR #2665, opened 2024-04-16
feat(auth)!: move consent interaction to different port

### LINKED ISSUE #2649: [Auth] Move interaction routes to a different port
## Context

The auth package currently has the GNAP routes and the interaction routes on the same port. The GNAP routes need to be exposed to the world while the interaction routes do not. Hence, we'd like to move the interaction routes to their own port.

To clarify: 
- all the routes that are specified [here](https://openpayments.dev/apis/auth-server/) stay on the current port
- all the remaining routes (those used for interacting with the grant) move to a new port

## Todos
- [ ] move interaction routes to own port

golobitch:

- Check issue

## Context

fixes #2649

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/auth/src/app.ts
```diff
@@ -103,6 +103,7 @@ export type AppContainer = IocContract<AppServices>
 
 export class App {
   private authServer!: Server
+  private authChoiceServer!: Server
```
sabineschaller, line 106: Can we call this `interactionServer`?

packages/auth/src/app.ts
```diff
@@ -455,6 +487,10 @@ export class App {
     return this.getPort(this.authServer)
   }
 
+  public getAuthChoicePort(): number {
```
sabineschaller, line 490: Can we call this `getInteractionPort`?

test/integration/lib/test-actions/index.ts
```diff
@@ -119,8 +119,11 @@ async function _startAndAcceptInteraction(
   assert(interactId)
 
   // Accept
+  const url = new URL(senderWalletAddress.authServer)
+  url.port = '3109'
+
   const acceptResponse = await fetch(
-    `${senderWalletAddress.authServer}/grant/${interactId}/${nonce}/accept`,
+    `${url.toString()}grant/${interactId}/${nonce}/accept`,
```
sabineschaller, line 126: I'm not sure the interaction url should be derived from the auth server url 🤔

golobitch: I mean, this is used only for integration tests, and tests are designed so that on the same server, there is auth server running and interaction server running, so it should not be a problem for tests. But for ASE, they will need to specify this address in their config I assume

sabineschaller: Fair enough 🙂

packages/auth/src/app.ts
```diff
@@ -391,6 +382,44 @@ export class App {
...
+    const router = new Router<DefaultState, AppContext>()
+    router.use(bodyParser())
+
+    const openApi = await this.container.use('openApi')
+    const interactionRoutes = await this.container.use('interactionRoutes')
+
+    // Grant accept/reject
+    router.post<DefaultState, ChooseContext>(
+      '/grant/:id/:nonce/:choice',
+      createValidatorMiddleware<ChooseContext>(openApi.idpSpec, {
```
njlie, line 397: I would consider any route that pulls from the Open Payments IDP spec to be part of this move. So like lines 304 to 332.

golobitch: You mean to also have following routes on this port:
* Start interaction
* Finish interaction
* Grant lookup

Normally, this issue was created due to security concerns regarding accepting and rejecting grants on behalf of xy user. So I would normally create separate ticket and also a separate PR.

But from my understanding Grant lookup should be publicly accessible right? Also for interaction start agreed? Because if interaction start is not publicly accessible then that would mean that you are not redirected to IDP site. So ASE would need to implement also some own startInteraction route and call then start interaction on Rafiki Auth ... Would that also mean to change the start interaction API to return url where the ASE should redirect user? Basically return 301 with IDP_URL ?

Same question for finish ... what finish does right now is that it finished the interaction and redirects the browser session to client callback. So we would also need to change this API right? And also from my memory I remember that something needs to be passed to the finish interaction route? Something that is in cookie right?

IMHO that would be quite an overhaul so if we would go down this road I would like to have separate ticket for this improvement. What do you think @sabineschaller @mkurapov ?

njlie: Grant lookup does not need to be publicly accessible - at least, not the one that's implemented here. GNAP doesn't specify that grants should be publically queried and the purpose that endpoint in particular is so that the IDP hosting the interaction can lookup the grant - there isn't any wider audience that endpoint is intended for.

I'll walk back the suggestion to move the start/finish interaction, those are meant to be public.

golobitch:
@njlie @mkurapov I have also moved the Grant lookup to the interaction server.

test/integration/lib/test-actions/index.ts
```diff
@@ -119,8 +119,11 @@ async function _startAndAcceptInteraction(
   assert(interactId)
 
   // Accept
+  const url = new URL(senderWalletAddress.authServer)
+  url.port = '3109'
```
mkurapov, line 123: I think this port would depend on the ASE itself, right? It's either 3109 for cloud-nine-wallet or 4109 for happy-life-bank

packages/auth/src/index.ts
```diff
@@ -293,6 +293,9 @@ export const start = async (
   await app.startAuthServer(config.authPort)
   logger.info(`Auth server listening on ${app.getAuthPort()}`)
 
+  await app.startInteractionServer(config.interactionPort)
+  logger.info(`Auth choice server listening on ${app.getInteractionPort()}`)
```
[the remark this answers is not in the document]
golobitch: Ah yes, first, I named it `authChoiceServer` and then renamed it to interactionServer and failed to alos update it here in log.

test/integration/lib/test-actions/index.ts
mkurapov: I think since you have added the `interactionServer` in the env file/config, you can just pass it through, pass `deps: TestActionDeps` into the function and use  `deps.sendingASE.config.interactionServer` directly

njlie (approved):

test/integration/lib/test-actions/index.ts
```diff
@@ -119,11 +119,8 @@ async function _startAndAcceptInteraction(
   assert(interactId)
 
   // Accept
-  const url = new URL(senderWalletAddress.authServer)
-  url.port = '3109'
-
   const acceptResponse = await fetch(
-    `${url.toString()}grant/${interactId}/${nonce}/accept`,
+    `${deps.sendingASE.config.interactionServer}/grant/${interactId}/${nonce}/accept`,
```
mkurapov, line 123: final thing, I think we just need to expose `interactionServer` on the `TestConfig` interface

localenv/happy-life-bank/docker-compose.yml
```diff
@@ -80,7 +80,11 @@ services:
     ports:
       - '4003:3003'
       - '4006:3006'
+<<<<<<< HEAD
```
sabineschaller, line 83: Some rebase comments are left here.

mkurapov (dismissed):

sabineschaller (approved):

PR #2676, opened 2024-04-24
feat(auth)!: add logs to graphql resolvers

### LINKED ISSUE #2574: Add logs in graphql resolvers - auth

njlie:

- Adds logs to the admin graphql resolvers in the auth server.

## Context

Fixes #2574.

Based largely off of https://github.com/interledger/rafiki/pull/2667. Thanks @sabineschaller!

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/graphql/plugin/index.ts
+++ b/packages/auth/src/graphql/plugin/index.ts
@@ -37,7 +37,7 @@ export class LoggingPlugin implements ApolloServerPlugin {
         if (context.errors) {
           context.errors.forEach((error) => {
             if (
-              error.extensions.code === GraphQLErrorCode.InternalServerError
+              Object.keys(error.extensions).length === 0 || error.extensions.code === GraphQLErrorCode.InternalServerError
             ) {
               logger.error({
                 requestId,
--- a/packages/auth/src/graphql/plugin/index.ts
+++ b/packages/auth/src/graphql/plugin/index.ts
@@ -37,7 +37,8 @@ export class LoggingPlugin implements ApolloServerPlugin {
         if (context.errors) {
           context.errors.forEach((error) => {
             if (
-              Object.keys(error.extensions).length === 0 || error.extensions.code === GraphQLErrorCode.InternalServerError
+              Object.keys(error.extensions).length === 0 ||
+              error.extensions.code === GraphQLErrorCode.InternalServerError
             ) {
               logger.error({
                 requestId,
--- a/packages/auth/src/graphql/plugin/index.ts
+++ b/packages/auth/src/graphql/plugin/index.ts
@@ -1,0 +1,59 @@
+import {
+  ApolloServerPlugin,
+  GraphQLRequestContextDidEncounterErrors,
+  GraphQLRequestContextDidResolveOperation,
+  GraphQLRequestListener
+} from '@apollo/server'
+import { Logger } from 'pino'
+import { v4 as uuid } from 'uuid'
+
+import { GraphQLErrorCode } from '../errors'
+
+export class LoggingPlugin implements ApolloServerPlugin {
+  private logger
+
+  constructor(logger: Logger) {
+    this.logger = logger
+  }
+
+  async requestDidStart(): Promise<GraphQLRequestListener<never>> {
+    const requestId = uuid()
+    const logger = this.logger
+    let operation: string | null
+
+    return {
+      async didResolveOperation(
+        context: GraphQLRequestContextDidResolveOperation<never>
+      ) {
+        operation = context.operationName
+        logger.info({
+          requestId,
+          operation
+        })
+      },
+      async didEncounterErrors(
+        context: GraphQLRequestContextDidEncounterErrors<never>
+      ): Promise<void> {
+        if (context.errors) {
+          context.errors.forEach((error) => {
+            if (
+              error.extensions.code === GraphQLErrorCode.InternalServerError
+            ) {
+              logger.error({
+                requestId,
+                variables: context.request.variables,
+                error
+              })
+            } else {
+              logger.info({
+                requestId,
+                variables: context.request.variables,
+                error
+              })
+            }
+          })
+        }
+      }
+    }
+  }
+}
--- a/packages/auth/src/graphql/resolvers/grant.test.ts
+++ b/packages/auth/src/graphql/resolvers/grant.test.ts
@@ -482,8 +482,9 @@ describe('Grant Resolvers', (): void => {
     })
 
     test('Returns error for unknown grant', async (): Promise<void> => {
+      let gqlQuery
       try {
-        const gqlQuery = appContainer.apolloClient
+        gqlQuery = appContainer.apolloClient
           .mutate({
             mutation: gql`
               query GetGrant($id: ID!) {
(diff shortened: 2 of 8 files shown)


packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -484,7 +484,7 @@ describe('Grant Resolvers', (): void => {
       const gqlQuery = appContainer.apolloClient
         .mutate({
           mutation: gql`
-            query GetGrant($id: String!) {
+            query GetGrant($id: ID!) {
```
njlie, line 487: This test was passing because it was throwing an `ApolloError`, but not for the reason this test was looking for - it was erroring over a `String!` being passed into a parameter that expected an `ID!` type instead.

packages/auth/src/graphql/plugin/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    const requestId = uuid()
+    const logger = this.logger
+    let operation: string | null
+
+    return {
+      async didResolveOperation(
+        context: GraphQLRequestContextDidResolveOperation<never>
+      ) {
+        operation = context.operationName
+        logger.info({
```
mkurapov, line 29: should this be a debug log, and should it include the input variables?
CC @sabineschaller (since its related to your change for `backend` as well)

mkurapov (commented):
We also need to change the Revoke Grant bruno example, because it is trying to resolve fields that no longer exist

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -582,24 +578,21 @@ describe('Grant Resolvers', (): void => {
           }
         })
 
-      expect(response.success).toBe(false)
-      expect(response.code).toBe('401')
-      expect(response.message).toBe('Grant Id is not provided')
+      expect(gqlQuery).rejects.toThrow(ApolloError)
+      expect(gqlQuery).rejects.toThrow('Grant id is not provided')
```
mkurapov, line 582: should we check that the correct `error.extensions.code` was thrown as well?

packages/auth/src/graphql/plugin/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (context.errors) {
+          context.errors.forEach((error) => {
+            if (
+              error.extensions.code === GraphQLErrorCode.InternalServerError
+            ) {
+              logger.error({
+                requestId,
+                variables: context.request.variables,
+                error
+              })
```
mkurapov, line 46: Do `error.extensions.code` actually get exposed in here, or is it only exposed to the client? I can't seem to trigger this error log by throwing an early error in one of our resolvers

packages/auth/src/graphql/plugin/index.ts
BlairCurrey, line 1: seems like another candidate to move to a shared `rafiki-lib` (or similar) package, if we ever do that. Seems like its basically the same as the one being added to `backend`. https://github.com/interledger/rafiki/pull/2667/files#diff-76710b5ac3ea0b241c0ca1d25ee4fb8dae45d1b1e5a4b1366a03fe62ba5b8551

packages/auth/src/graphql/plugin/index.ts
BlairCurrey: I would expect this to stay info personally, especially since the `requestId` is referenced in the `didEncounterErrors` logs. Helps tell the whole story I think.

packages/auth/src/graphql/plugin/index.ts
BlairCurrey: @mkurapov  are you testing with a `GraphQLError` like this? Im seeing this when doing this in the `getGrants` resolver.

```ts
 throw new GraphQLError('Something went wrong', {
   extensions: {
     code: GraphQLErrorCode.InternalServerError
   }
 })
 ```

Comment (mkurapov) [reply packages/auth/src/graphql/plugin/index.ts:46]: That's right @BlairCurrey, was throwing a normal error. What if we add a condition to check whether `error.extensions` exist in the first place? If that happens, we could probably assume it was an unhandled "normal" error.

File: packages/auth/src/graphql/resolvers/grant.test.ts [hunk 5]
```diff
@@ -481,38 +482,46 @@ describe('Grant Resolvers', (): void => {
...
+            if (query.data) {
+              return query.data.grant
+            } else {
+              throw new Error('Data was empty')
+            }
+          })
 
-      await expect(gqlQuery).rejects.toThrow(ApolloError)
+        await expect(gqlQuery).rejects.toThrow(ApolloError)
+        await expect(gqlQuery).rejects.toThrow('No grant')
```
mkurapov, line 518: would it makes sense to just do these assertions in the catch?

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -481,38 +482,46 @@ describe('Grant Resolvers', (): void => {
...
-      const gqlQuery = appContainer.apolloClient
-        .mutate({
-          mutation: gql`
-            query GetGrant($id: String!) {
-              grant(id: $id) {
-                id
-                client
-                state
-                access {
+      try {
```
mkurapov, line 485:
```suggestion
      expect.assertions(3)
      try {
```
to make sure we don't end up ignoring the expect's in the try

packages/auth/src/graphql/resolvers/grant.test.ts
njlie: It would still work, and I personally think it's cleaner to have one level of assertion, rather than the assertion-of-assertions approach of https://github.com/interledger/rafiki/pull/2676#discussion_r1586408057

packages/auth/src/graphql/plugin/index.ts
sabineschaller: I looked at the [diagram Blair shared with us](https://i.stack.imgur.com/z5Fim.png) and I agree. That should be `info`.

packages/auth/src/graphql/resolvers/grant.test.ts
sabineschaller: Tbh, I also skipped that because I didn't know how to test test.

packages/auth/src/graphql/plugin/index.ts
sabineschaller: That is correct

packages/auth/src/graphql/resolvers/grant.test.ts
```diff
@@ -481,38 +482,47 @@ describe('Grant Resolvers', (): void => {
...
+            } else {
+              throw new Error('Data was empty')
+            }
+          })
 
-      await expect(gqlQuery).rejects.toThrow(ApolloError)
+        await gqlQuery // throw error to inspect error object
+      } catch (err) {
+        await expect(gqlQuery).rejects.toThrow(ApolloError)
+        await expect(gqlQuery).rejects.toThrow('No grant')
```
mkurapov, line 521:
```suggestion
        assert.ok(error instanceof ApolloError)
        expect(error.message).toBe('No grant')
```
since we already have the error & to not need to make the gqlQuery request multiple times

BlairCurrey (approved):
looks good to me

mkurapov (approved):

Issue #2489, opened 2024-03-05
Logging & Error Handling improvements


mkurapov:
## Context

We do not log a whole lot of things right now, which makes it hard to debug. Furthermore, we want to make Rafiki ready to work with log management in the future. So let's add some logs! 

More is more, we can always get rid of logs later. Add logs wherever you would want to see them. 

## Todos
- [x] #1905 
- [x] #2575
- [x] #2574
- [x] #2576
- [x] #2577
- [x] #2578
- [x] #2579
- [x] #2580
- [x] #2581
- [x] #2582

PR #2680, opened 2024-04-28
feat(backend): improve ilp connector logs

BlairCurrey:

changes primarily consist of
- adding logs where there are errors to capture the context
- adding context to existing logs
- added some debug/trace logs
- updated level of some existing logs

## Context

The connector services already follows a similar pattern as Open Payments does for error handling with the `createIncomingErrorHandlerMiddleware` which catches all errors and logs and handles them appropriately. 

Fixes: #2578

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/payment-method/ilp/connector/core/controllers/stream.ts
```diff
@@ -53,11 +53,12 @@ export function createStreamController(): ILPMiddleware {
       } else {
         logger.warn(
           {
+            connectionKey,
             totalReceived,
             err,
             err2
           },
-          'error incrementing stream totalReceived'
+          'failed to increment stream totalReceived'
```
BlairCurrey, line 61: no error actually being throw here, so I changed the wording

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
```diff
@@ -23,7 +23,7 @@ export function createIncomingErrorHandlerMiddleware(
       if (!err || typeof err !== 'object') {
         err = new Error('Non-object thrown: ' + e)
       }
-      ctx.services.logger.debug({ err }, 'Error thrown in incoming pipeline')
+      ctx.services.logger.info({ err }, 'Error thrown in incoming pipeline')
```
BlairCurrey, line 26: I figured we don't want to limit the error logs to development so I changed to info (which matches what we do in the open payments server error handler as well.)

packages/backend/src/payment-method/ilp/connector/core/middleware/ildcp.ts
```diff
@@ -22,7 +22,7 @@ export function createIldcpMiddleware(serverAddress: string): ILPMiddleware {
 
     const clientAddress = incoming.staticIlpAddress
     if (!clientAddress) {
-      logger.warn(
+      logger.error(
```
BlairCurrey, line 25: For logging where there are errors, I generally made them error level like this. There were a few exceptions like the rate limiting errors. Those are currently warnings and I left them alone - warning felt appropriate there.

packages/backend/src/payment-method/ilp/connector/core/middleware/reduce-expiry.ts
```diff
@@ -24,7 +24,7 @@ function getDestinationExpiry(
   const sourceExpiryTime = request.expiresAt.getTime()
 
   if (sourceExpiryTime < Date.now()) {
-    log.trace('incoming packet has already expired', { request })
+    log.debug({ request }, 'incoming packet has already expired')
```
BlairCurrey, line 27: pino doesnt complain, but `{request}` did not end up getting logged when put in the original order.

BlairCurrey (commented):
I scrubbed an auth token from one of the logs but didn't see anything else obviously sensitive. Wondered about these sort of details in `getAccountByDestinationAddress`.

```js
{
  streamDestination: '5cae9d2a-fe66-43a5-b9e4-0caf6ad817c7',
  destinationAddress: 'test.happy-life-bank-test.Uq3iqJnztJuzIQSSCXQs6qkXoslEMKjBiLA6HJf5lNU1yXKJwZ4V-r4WXULwMWNz16fGu_YiM3jvkezCx94ysV9t0kNaC7I'
}
```

packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
```diff
@@ -56,7 +58,15 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
               IncomingPaymentState.Expired
             ].includes(incomingPayment.state)
           ) {
-            throw new Errors.UnreachableError('destination account is disabled')
+            const errorMessage = 'destination account is disabled'
```
mkurapov, line 61: maybe "destination account is in an incorrect state" makes more sense?

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
mkurapov: Maybe we should have an error log in the unexpected internal error case down below?

packages/backend/src/payment-method/ilp/connector/core/middleware/max-packet-amount.ts
```diff
@@ -14,7 +14,7 @@ export function createIncomingMaxPacketAmountMiddleware(): ILPMiddleware {
     if (maxPacketAmount) {
       const amount = request.prepare.intAmount
       if (amount > maxPacketAmount) {
-        logger.warn(
+        logger.error(
```
mkurapov, line 17: I was back and forth on this one, but I _think_ error is correct here, since it would probably require actioning to update the configuration for the peer

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
```diff
@@ -51,20 +60,35 @@ export function createBalanceMiddleware(): ILPMiddleware {
...
 
       if (isTransferError(trxOrError)) {
+        const safeLogTransferError = (transferOptions: TransferOptions) => {
+          if (transferOptions.destinationAccount?.http?.outgoing.authToken) {
+            // @ts-expect-error - "The operand of a 'delete' operator must be optional"
+            delete transferOptions.destinationAccount.http.outgoing.authToken
+          }
+          services.logger.error(
+            { transferOptions },
+            'Could not create transfer'
```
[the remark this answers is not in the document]
BlairCurrey: I will try to use that instead - looks handy. I found this surprisingly cumbersome. I think the `delete` with `@ts-expect-error` is about the best we can do and I dont love it.

packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
BlairCurrey: ya thats better, updated

packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
```diff
@@ -67,6 +77,7 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
               LiquidityAccountType.INCOMING
             )
           }
+          ctx.services.logger.trace('got account from incoming payment')
```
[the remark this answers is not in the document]
BlairCurrey: I'm interpreting "id of the account" to mean `incomingPayment` in this case. As opposed to the created liquidity account (not sure if that id would be the same or not). Correct me if needed.

I updated this and similar to log the underlying account id (in this case incoming payment).

packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
```diff
@@ -106,7 +119,15 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
...
+        {
+          streamDestination: ctx.state.streamDestination,
+          destinationAddress: ctx.request.prepare.destination
+        },
+        errorMessage
+      )
+      throw new Errors.UnreachableError(errorMessage)
     }
     ctx.accounts = {
       get incoming(): IncomingAccount {
```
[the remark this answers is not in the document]
BlairCurrey: yeah that makes sense, missed that

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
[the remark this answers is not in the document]
BlairCurrey: updated to log the `trxOrError` and used `redact` instead

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
BlairCurrey: there is one but it looks like its basically null if its not an `HttpError`. changed it to always log with the original error.

mkurapov (commented):
just a single comment about the redaction

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
```diff
@@ -20,6 +20,12 @@ export function createBalanceMiddleware(): ILPMiddleware {
     next: () => Promise<void>
   ): Promise<void> => {
     const { amount } = request.prepare
+    const logger = services.logger.child(
+      { module: 'balance-middleware' },
+      {
+        redact: ['transferOptions.destinationAccount.http.outgoing.authToken']
```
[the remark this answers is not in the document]
BlairCurrey: I would say yes but the wildcard `*token` doesnt work like that. https://github.com/davidmarkclements/fast-redact?tab=readme-ov-file#wildcards

It doesn't match `token`, `authToken`, etc. [nor would it look for tokens at all depths](https://github.com/davidmarkclements/fast-redact/issues/5).

From what I can tell there isn't a way to configure redact to redact all values for a key matching `*token` at any depth, nor even a key matching `*token` at a specific depth.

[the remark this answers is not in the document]
mkurapov: Makes sense, I'm guessing recursive redacting is quite costly. In that case, maybe we just do what you had previously, or instead of deleting keys how you had before, we can just selectively log a few properties of this object. Sorry for the back-and-forth, thought this was recursive from the beginning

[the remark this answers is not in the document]
BlairCurrey: As we talked off github, it is probably easier to blacklist sensitive info than whitelist useful info here because there is a fair amount and not exactly sure what will matter. So it currently redacts just the `authToken`.

packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
```diff
@@ -29,17 +29,25 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
         )
       } catch (err) {
         // Don't complain if liquidity account already exists.
-        if (err instanceof AccountAlreadyExistsError) {
-          // Do nothing.
-        } else {
+        if (!(err instanceof AccountAlreadyExistsError)) {
```
njlie, line 32: I can't comment on the exact line where this could be added, but in the spirit of the directive in https://github.com/interledger/rafiki/issues/2489  for "more is more", but might it be useful for a debug or trace-level log when liquidity accounts are successfully created

Issue #2673, opened 2024-04-23
Move auth's koa session state out of service


BlairCurrey:
## Feature Request

**Describe the feature you'd like to request**

Dont keep state in-memory for `koa-session` in `auth`.

**Describe the solution you'd like**

Use redis as external session store for `koa-session` in `auth`. https://github.com/koajs/session?tab=readme-ov-file#external-session-stores. Note that `koa-session` is also used in a test context, which this doesn't really need to change.

**Describe alternatives you've considered**

In light of discussions about migrating away from redis I also considered storing it in postgres instead. However even if we migrate away from redis, I think migrating from a redis to a redis-like service will be much simpler (probably just changing image in docker compose, no code changes) and just generally more fit for this purpose.

PR #2693, opened 2024-05-01
feat: use redis as backend for koa-session

BlairCurrey:

- use redis for `koa-session` backend

## Context

fixes https://github.com/interledger/rafiki/issues/2673

corresponding helm-chart issue created as well: https://github.com/interledger/helm-charts/issues/38

## Checklist

- [x] Tests added/updated
- [x] Documentation added


BlairCurrey (commented):
I believe I will need to update [this repo's helm values.yaml](https://github.com/interledger/rafiki/blob/main/infrastructure/helm/rafiki/values.yaml) after they are added to our helm chart repo. Assuming it mirrors our backend, I believe we just need to add the same `redis` config that we have for `rafiki-backend`.

packages/auth/src/app.ts
```diff
@@ -342,12 +344,26 @@ export class App {
...
           maxAge: 60 * 1000,
-          signed: true
+          signed: true,
+          store: {
+            async get(key) {
+              const data = await redis.get(key)
+              return data ? JSON.parse(data) : null
+            },
+            async set(key, session) {
+              await redis.set(key, JSON.stringify(session))
```
mkurapov, line 361: How does expiry/TTL of the session get saved in redis?

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -85,8 +85,10 @@ services:
       TRUST_PROXY: ${TRUST_PROXY}
       AUTH_DATABASE_URL: postgresql://cloud_nine_wallet_auth:cloud_nine_wallet_auth@shared-database/cloud_nine_wallet_auth
       AUTH_SERVER_DOMAIN: ${CLOUD_NINE_AUTH_SERVER_DOMAIN:-http://localhost:3006}
+      REDIS_URL: redis://shared-redis:6379/0
```
mkurapov, line 88: do we want to distinguish this from the `cloud-nine-backend` one?

packages/auth/src/app.ts
```diff
@@ -342,12 +344,26 @@ export class App {
 
     koa.use(cors())
     koa.keys = [this.config.cookieKey]
+
+    const redis = await this.container.use('redis')
     koa.use(
       session(
         {
           key: 'sessionId',
```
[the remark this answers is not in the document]
BlairCurrey: This is the original configuration and from what I can tell it should have nothing to do with redis, it's just for the http cookie returned to the client:

image

The value there would be the UUID key you see in redis.

localenv/cloud-nine-wallet/docker-compose.yml
BlairCurrey: That's a good question. Not sure how much it matters TBH but I changed it to:

- c9 backend: 0
- c9 auth: 1
- hlb backend: 2
- hlb auth: 3

packages/auth/src/app.ts
BlairCurrey: Testing the flow showed that it was being expired but it looks like it was just the http cookie. Redis key was not actually expiring. Changed it to make the record expire in redis. [05fcb02](https://github.com/interledger/rafiki/pull/2693/commits/05fcb02334c3cba30dceaf5d05472d59cfe5da3d)

golobitch: This does not set expiry on redis key. With set command, that would be done either via EX options of set command or through some expiry command. If we know the TTL of the session I would rewrite this code to be like this:
```
const op = redis.multi()
op.hmset(key, session)
op.expire(key, expirationInMilliseconds)

await op.exec()
```

packages/auth/src/app.ts
```diff
@@ -342,12 +344,29 @@ export class App {
...
-          maxAge: 60 * 1000,
-          signed: true
+          maxAge: maxAgeMs,
+          signed: true,
+          store: {
+            async get(key) {
+              const data = await redis.get(key)
+              return data ? JSON.parse(data) : null
+            },
+            async set(key, session) {
```
mkurapov, line 361: I see there is a third parameter `maxAge` possible in this function. Do we want to include that here?

packages/auth/src/app.ts
```diff
@@ -342,12 +344,29 @@ export class App {
...
+              // Add a delay to cookie age to ensure redis record expires after cookie
+              const expireInMs = maxAgeMs + 10 * 1000
+              await redis.psetex(key, expireInMs, JSON.stringify(session))
+            },
```
mkurapov, line 365:
```suggestion
              await redis.set(key, JSON.stringify(session), 'PX', expireInMs)
```
I think `psetex` is deprecated based on the docs

packages/auth/src/app.ts
```diff
@@ -342,12 +344,29 @@ export class App {
...
+          maxAge: maxAgeMs,
+          signed: true,
+          store: {
+            async get(key) {
+              const data = await redis.get(key)
+              return data ? JSON.parse(data) : null
+            },
+            async set(key, session) {
+              // Add a delay to cookie age to ensure redis record expires after cookie
+              const expireInMs = maxAgeMs + 10 * 1000
```
mkurapov, line 363: If I understand correctly, this is to possibly allow the "final" request with the to-be-expired cookie parameter to complete before deleting it in redis?

packages/auth/src/app.ts
golobitch: `psetex` is not deprecated per my knowledge, but it is not wise to use it, since it will be deprecated and removed.
From the docs:
```
Note: Since the SET command options can replace SETNX, SETEX, PSETEX, GETSET, it is possible that in future versions of Redis these commands will be deprecated and finally removed.
```

golobitch: BTW, no need to do stringify on session object. Instead, you can just use hmset(key, session)

packages/auth/src/app.ts
BlairCurrey: Yeah basically. could potentially be shorter I guess - just needs to be long enough to cover the time between finding the non-expired koa-session in the server and retrieving the redis record. no harm in having a little extra buffer though I figure.

packages/auth/src/app.ts
BlairCurrey: Updated to this ^ except `hset`. apparently `hmset`, like `psetex` is marked as deprecated in `redis` docs https://redis.io/docs/latest/commands/hmset/

[dda0079](https://github.com/interledger/rafiki/pull/2693/commits/dda0079fafeb6f6e8da4206e3e90ee94f556c839)

mkurapov (approved):

njlie (approved):
LGTM

PR #2694, opened 2024-05-01
feat(api): cancel outgoing payment

### LINKED ISSUE #2688: Add `cancelOutgoingPayment` GraphQL API
## Context
Currently, when we create an outgoing payment, we send a `outgoing_payment.created` webhook to the ASE such that the ASE can check the user account, place a hold on the account for the payment amount, and then call `fundOutgoingPayment`, to deposit liquidity into the payment. Then, the ILP payment starts sending into the incoming payment.

Sometimes, however, the sending user account may not have enough balance to fulfill the outgoing payment amount. Rafiki does not provide a way for the ASE to notify about this possibility, and as a result, the outgoing payment will indefinitely stay in a FUNDING state. 

We should allow the ASE to notify Rafiki about this possibility through a new Admin API, `cancelOutgoingPayment` (name always up for discussion). `cancelOutgoingPayment` will take in the `outgoingPaymentId` and update the state of the outgoing payment to a new state `CANCELLED`. 

The only time an outgoing payment can be marked CANCELLED is if it is in a FUNDING state. 
Make sure to also update the `OutgoingPayment.failed` getter in the model to include this state, so that we properly return `failed: true` when it is fetched via Open Payments. 
Also, there is no need to send a webhook when an outgoing payment has been cancelled.

## Todos
- [ ] Create `cancelOutgoingPayment` mutation
- [ ] Update the `OutgoingPaymentState` enum in the GraphQL API
- [ ] Update the mock ASE to call this mutation when a user does not have enough balance.
- [ ] Make sure to add description to the mutation

golobitch:
- feat(localenv): cancel outgoing payment on created webhook
- feat(graphql): cancel outgoing payment api
- feat(outgoing-payment): cancel outgoing payment service implementation

Add an API for canceling outgoing payment. This can be possible only if outgoing payment is in funding state. Normally, this API would be called by ASE to cancel the outgoing payment. It can be canceled due too numerous reasons like account not having enough balance or like his KYC is not valid, etc. ASE can also specify cancelation reason, and it will be written in meta under cancellationReason key. 

## Context

fixes #2688

## Checklist

- [x] Tests added/updated
- [x] Documentation added

### DIFF
--- a/localenv/cloud-nine-wallet/seed.yml
+++ b/localenv/cloud-nine-wallet/seed.yml
@@ -40,6 +40,12 @@ accounts:
     path: accounts/wbdc
     brunoEnvVar: wbdcWalletAddress
     assetCode: USD
+  - name: "Broke Account"
+    id: 5a95366f-8cb4-4925-88d9-ae57dcb444bb
+    initialBalance: 50
+    path: accounts/broke
+    brunoEnvVar: brokeWalletAddress
+    assetCode: USD
 rates:
   EUR:
     MXN: 18.78
--- a/localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
+++ b/localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
@@ -69,6 +69,28 @@ export async function handleOutgoingPaymentCreated(wh: Webhook) {
   }
 
   const amt = parseAmount(payment['debitAmount'] as AmountJSON)
+  const accBalance = BigInt(acc.creditsPosted) - BigInt(acc.debitsPosted)
+
+  if (accBalance < amt.value) {
+    await apolloClient.mutate({
+      mutation: gql`
+        mutation CancelOutgoingPayment($input: CancelOutgoingPaymentInput!) {
+          cancelOutgoingPayment(input: $input) {
+            code
+            success
+            message
+          }
+        }
+      `,
+      variables: {
+        input: {
+          id: payment.id,
+          reason: 'Account does not have enough balance.'
+        }
+      }
+    })
+    return
+  }
 
   await mockAccounts.pendingDebit(acc.id, amt.value)
 
--- a/localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
+++ b/localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
@@ -69,9 +69,10 @@ export async function handleOutgoingPaymentCreated(wh: Webhook) {
   }
 
   const amt = parseAmount(payment['debitAmount'] as AmountJSON)
-  const accBalance = BigInt(acc.creditsPosted) - BigInt(acc.debitsPosted)
 
-  if (accBalance < amt.value) {
+  try {
+    await mockAccounts.pendingDebit(acc.id, amt.value)
+  } catch (err) {
     await apolloClient.mutate({
       mutation: gql`
         mutation CancelOutgoingPayment($input: CancelOutgoingPaymentInput!) {
@@ -91,8 +92,6 @@ export async function handleOutgoingPaymentCreated(wh: Webhook) {
     })
     return
   }
-
-  await mockAccounts.pendingDebit(acc.id, amt.value)
 
   // notify rafiki
   await apolloClient
--- a/packages/backend/src/graphql/resolvers/outgoing_payment.test.ts
+++ b/packages/backend/src/graphql/resolvers/outgoing_payment.test.ts
@@ -539,101 +539,115 @@ describe('OutgoingPayment Resolvers', (): void => {
     })
   })
 
-  describe('Mutation.cancelOutgoingPayment', (): void => {
+  describe.only('Mutation.cancelOutgoingPayment', (): void => {
     let payment: OutgoingPaymentModel
-
-    beforeEach(async (): Promise<void> => {
-      const { id: walletAddressId } = await createWalletAddress(deps, {
+    beforeEach(async () => {
+      const walletAddress = await createWalletAddress(deps, {
         assetId: asset.id
       })
-      payment = await createPayment({ walletAddressId })
+      
+      payment = await createPayment({ walletAddressId: walletAddress.id })
     })
 
-    const states: [string, OutgoingPaymentError | null][] = Object.values(
-      OutgoingPaymentState
-    ).flatMap((state) => [
-      [
-        'Not enough balance',
-        state == OutgoingPaymentState.Funding
-          ? null
-          : OutgoingPaymentError.WrongState
-      ],
-      [
-        'Missing KYC',
-        state == OutgoingPaymentState.Funding
-          ? null
-          : OutgoingPaymentError.WrongState
-      ]
-    ])
-    test.each(states)(
-      '200 - %s, error: %s',
-      async (reason, error): Promise<void> => {
-        jest
+    const reasons: (string | undefined)[] = [undefined, "Not enough balance"]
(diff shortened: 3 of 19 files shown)


packages/backend/src/graphql/resolvers/outgoing_payment.test.ts
```diff
@@ -539,6 +539,103 @@ describe('OutgoingPayment Resolvers', (): void => {
...
+          : OutgoingPaymentError.WrongState
+      ],
+      [
+        'Missing KYC',
+        state == OutgoingPaymentState.Funding
+          ? null
+          : OutgoingPaymentError.WrongState
+      ]
+    ])
+    test.each(states)(
```
mkurapov, line 568: I think we should do this kind of testing in the outgoing payment service tests instead (since that's where the logic for checking the status lives), while these resolver tests can just worry about a happy path and a single error path where the correct error result gets returned

localenv/cloud-nine-wallet/seed.yml
```diff
@@ -40,6 +40,12 @@ accounts:
     path: accounts/wbdc
     brunoEnvVar: wbdcWalletAddress
     assetCode: USD
+  - name: "Broke Account"
```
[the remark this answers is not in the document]
mkurapov: I think we should differentiate the paths, and the names across the two localenv instances, to make it a bit more clear?

packages/documentation/src/content/docs/integration/event-handlers.mdx
```diff
@@ -75,6 +77,18 @@ Example: An outgoing payment for $12 has been created.
...
+Example: Cancel outgoing payment due to low funds
+
+<Mermaid
+  graph={`sequenceDiagram
+    participant ASE as Account Servicing Entity
+    participant R as Rafiki
+    R->>ASE: webhook event: outgoing payment created,<br>debit amount: $12
+    ASE->>ASE: check if account has enough balance
+    ASE->>R: admin API call: CancelOutgoingPayment<br>reason: Not enough balance
+`}
```
mkurapov, line 89: we could probably avoid having to have a second diagram by using `alt` to describe the two possible flows in the original diagram

packages/documentation/src/content/docs/integration/event-handlers.mdx
```diff
@@ -59,7 +59,9 @@ Example: An incoming payment has expired and received $2.55.
 
 ## `outgoing_payment.created`
 
-The `outgoing_payment.created` event indicates that an outgoing payment has been created and is awaiting liquidity. The Account Servicing Entity should put a hold on the sender's account and deposit the funds into Rafiki.
+The `outgoing_payment.created` event indicates that an outgoing payment has been created and is awaiting liquidity. The Account Servicing Entity should perform a check here to see if there is enough balance, or some other kind of checks that must pass in order to fullfill this outgoing payment.
```
mkurapov, line 62:
```suggestion
The `outgoing_payment.created` event indicates that an outgoing payment has been created and is awaiting liquidity. The Account Servicing Entity should verify the user account balance (and perform any other kinds of checks necessary) before funding or cancelling the outgoing payment.
``` 
just to explain that you can do either or in this event handler

mkurapov (commented):
I tested this & it worked for me :+1:

localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
```diff
@@ -69,6 +69,28 @@ export async function handleOutgoingPaymentCreated(wh: Webhook) {
   }
 
   const amt = parseAmount(payment['debitAmount'] as AmountJSON)
+  const accBalance = BigInt(acc.creditsPosted) - BigInt(acc.debitsPosted)
```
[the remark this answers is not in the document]
golobitch: Yes, much better!

localenv/cloud-nine-wallet/seed.yml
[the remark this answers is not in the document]
golobitch: Not really sure what you meant by that. But this account is only on one instance, named `cloud-nine-wallet`. So it shouldn't be a problem right?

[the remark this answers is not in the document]
mkurapov: Ah, I saw the change in `localenv/mock-account-servicing-entity/seed.example.yml` and thought it was for happy-life-bank as well, but it's just the seed.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -118,6 +127,35 @@ export function isCreateFromIncomingPayment(
...
+
+    payment = await payment
+      .$query(trx)
+      .patchAndFetch({
+        state: OutgoingPaymentState.Cancelled,
+        metadata: {
+          ...payment.metadata,
+          cancellationReason: options.reason
+        }
+      })
```
mkurapov, line 152:
```suggestion
        metadata: {
          ...payment.metadata,
          ...(options.reason ? { cancellationReason: options.reason } : {})
        }
      })
```
so we don't set `cancellationReason: undefined`

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -388,6 +389,73 @@ describe('OutgoingPaymentService', (): void => {
...
+
+        await outgoingPayment.$query(knex).patch({ state })
+
+        const response = await outgoingPaymentService.cancel({
+          id: outgoingPayment.id,
+          reason
+        })
+
+        if (!outgoingPaymentError) {
+          const dataCheck = response as OutgoingPayment
```
mkurapov, line 446:
```suggestion
          assert.ok(response instanceof OutgoingPayment)
```
nit: if you want to avoid a new variable and `as`

packages/backend/src/graphql/schema.graphql
```diff
@@ -922,6 +929,13 @@ input CreateOutgoingPaymentInput {
   idempotencyKey: String
 }
 
+input CancelOutgoingPaymentInput {
+  "Outgoing payment id"
+  id: ID!
+  "Reason why this Outgoing Payment has been cancelled"
```
mkurapov, line 935: We should mention something along the lines that if this outgoing payment is requested via Open Payments, the reason will be publicly visible in the `metadata` field

mkurapov (commented):
A few minor comments

packages/backend/src/graphql/resolvers/outgoing_payment.test.ts
```diff
@@ -539,99 +539,118 @@ describe('OutgoingPayment Resolvers', (): void => {
     })
   })
 
-  describe('Mutation.cancelOutgoingPayment', (): void => {
+  describe.only('Mutation.cancelOutgoingPayment', (): void => {
```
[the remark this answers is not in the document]
tadejgolobic: Ups :)

packages/backend/src/graphql/schema.graphql
golobitch: Yes, I agree. Thank you.

packages/backend/src/open_payments/payment/outgoing/service.ts
mkurapov: > if the field is undefined, then it will not be included in the metadata when doing serialization to database right

This one I'm not sure about since `metadata` is stored as JSON in the DB, and maybe it doesn't mind saving values as undefined

mkurapov (approved):

PR #2697, opened 2024-05-02
feat(auth): add logs to routes

### LINKED ISSUE #2577: Add logs in Auth routes

njlie:

- Adds logging to route handlers on the auth server
  - Includes request method, path, and a unique request id in each log
  - Includes a breakdown of the request body, params, and headers (that are applicable) in each log
  - If applicable, any other relevant objects that were created/retrieved in the route business logic (e.g. a grant or an interaction)

## Context

Closes #2577.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -10,7 +10,7 @@ import { AccessToken, toOpenPaymentsAccessToken } from './model'
 import { AccessService } from '../access/service'
 import { TransactionOrKnex } from 'objection'
 import { GrantService } from '../grant/service'
-import { GNAPErrorCode, throwGNAPError } from '../shared/gnapErrors'
+import { GNAPErrorCode, GNAPServerRouteError } from '../shared/gnapErrors'
 import { generateRouteLogs } from '../shared/utils'
 
 export type TokenHttpSigContext = AppContext & {
@@ -159,14 +159,11 @@ async function rotateToken(
     await trx.rollback()
     const errorMessage =
       error instanceof Error ? error.message : 'Could not rotate token'
-    deps.logger.error(
-      {
-        ...generateRouteLogs(ctx),
-        err: error instanceof Error && error.message
-      },
+    throw new GNAPServerRouteError(
+      400,
+      GNAPErrorCode.InvalidRotation,
       errorMessage
     )
-    throwGNAPError(ctx, 400, GNAPErrorCode.InvalidRotation, errorMessage)
   }
 
   deps.logger.debug(
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -61,9 +61,19 @@ export interface AccessTokenRoutes {
 export function createAccessTokenRoutes(
   deps_: ServiceDependencies
 ): AccessTokenRoutes {
-  const logger = deps_.logger.child({
-    service: 'AccessTokenRoutes'
-  })
+  const logger = deps_.logger.child(
+    {
+      service: 'AccessTokenRoutes'
+    },
+    {
+      redact: [
+        'requestBody.access_token',
+        'accessToken.value',
+        'headers.authorization',
+        'grant.continueToken'
+      ]
+    }
+  )
   const deps = { ...deps_, logger }
   return {
     introspect: (ctx: IntrospectContext) => introspectToken(deps, ctx),
@@ -116,7 +126,8 @@ async function revokeToken(
 
   deps.logger.debug(
     {
-      ...generateRouteLogs(ctx)
+      ...generateRouteLogs(ctx),
+      accessToken
     },
     'revoked access token'
   )
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -86,7 +86,7 @@ async function introspectToken(
   deps.logger.debug(
     {
       ...generateRouteLogs(ctx),
-      grant: grant ?? 'n/a', // grant is undefined if access token is invalid
+      grant,
       accessToken
     },
     'introspected access token'
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -86,8 +86,7 @@ async function introspectToken(
   deps.logger.debug(
     {
       ...generateRouteLogs(ctx),
-      grant,
-      accessToken
+      grant
     },
     'introspected access token'
   )
@@ -117,8 +116,7 @@ async function revokeToken(
 
   deps.logger.debug(
     {
-      ...generateRouteLogs(ctx),
-      accessToken
+      ...generateRouteLogs(ctx)
     },
     'revoked access token'
   )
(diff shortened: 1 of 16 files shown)


packages/auth/src/shared/utils.ts
```diff
@@ -7,3 +9,21 @@ export function generateNonce(): string {
...
+export function generateRouteLogs(ctx: AppContext): {
+  requestId: string
+  route: typeof ctx.path
+  method: typeof ctx.method
+  params: typeof ctx.params
+  headers: typeof ctx.headers
+  requestBody: typeof ctx.request.body
+} {
+  return {
+    requestId: v4(),
```
mkurapov, line 22: This would show a different uuid for each log, right? 
I was looking into a way to have a common request/traceId as a property on a child pino logger, but it is a bit difficult to do that with how we've set up dependency injection with adonis fold library. I'll take another attempt at addressing this in my upcoming changes.

packages/auth/src/accessToken/routes.ts
```diff
@@ -101,7 +113,15 @@ async function revokeToken(
   deps: ServiceDependencies,
   ctx: RevokeContext
 ): Promise<void> {
-  await deps.accessTokenService.revoke(ctx.accessToken.id)
+  const accessToken = await deps.accessTokenService.revoke(ctx.accessToken.id)
+
+  deps.logger.debug(
+    {
+      ...generateRouteLogs(ctx),
+      accessToken
```
mkurapov, line 121: should we be logging out tokens such as accessToken?

mkurapov (commented):
Should we create an `openPaymentsServerErrorMiddleware` like we do in backend to catch all of the GNAPErrors to have a single place for a standardized log for each error? (Since most of the logs are added right before the throw).

packages/auth/src/shared/utils.ts
njlie: I figured it would be workable for now as the routes log once per request, but I admit it's wishful thinking

packages/auth/src/accessToken/routes.ts
njlie: Probably not

BlairCurrey: Does it make any difference that its a debug log? I guess we can say these _can_ but _shouldnt_ be logged in production. Had the same thought about `grant.continueToken` elsewhere.

packages/auth/src/shared/utils.ts
```diff
@@ -7,3 +8,19 @@ export function generateNonce(): string {
...
+  requestBody: typeof ctx.request.body
+} {
+  return {
+    method: ctx.method,
+    route: ctx.path,
+    headers: ctx.headers,
+    params: ctx.params,
+    requestBody: ctx.request.body
+  }
+}
```
BlairCurrey, line 26: I imagine there is probably some sensitive info in here somewhere.

for example from a `POST continue/{{continueId}}`:
- `requestBody.continue.access_token.value`
- `requestBody.access_token.value`

And is everything in headers OK to log or should any of it be redacted?

```json
  "headers": {
    "accept": "application/json",
    "content-type": "application/json",
    "content-digest": "sha-512=:wJHRy2GvqCYWZhLTIAdtBaDT9MRO69ldMWgdld0GO8cYVu8uop7pmilYh1xmlVDiiT8d2P1CSdyYgnbCsJIrNA==:",
    "content-length": "174",
    "signature": "sig1=:M5wb//IFQa43pZ6DfUrxcx7RCf3aQWXApk28TmhqgiUrzBbhlu0hs2/zVzXuiB7kEa1LZuq3taQT1XT5wuBDDg==:",
    "signature-input": "sig1=(\"@method\" \"@target-uri\" \"content-digest\" \"content-length\" \"content-type\");keyid=\"keyid-97a3a431-8ee1-48fc-ac85-70e2f5eba8e5\";created=1715013373",
    "user-agent": "axios/1.6.8",
    "accept-encoding": "gzip, compress, deflate, br",
    "host": "happy-life-bank-test-auth:4106",
    "connection": "keep-alive"
  },
```

If we need to [configure pino to redact](https://github.com/pinojs/pino/blob/main/docs/redaction.md) I imagine it might make sense to do it where we initialize the logger.

BlairCurrey: This seems like a good pattern btw though. Also wondering if you think it would make sense to desstructure this where we init the pino logger so we dont have to do `...generateRouteLogs(ctx)` everywhere? Or if we dont want to always log context do a new `loggerWithCtx` or something?

mkurapov:
What do you think about original PR review comment?

> Should we create an openPaymentsServerErrorMiddleware like we do in backend to catch all of the GNAPErrors to have a single place for a standardized log for each error? (Since most of the logs are added right before the throw).

Might be nice to match the behaviour in `backend`

njlie:
> What do you think about original PR review comment?
> 
> > Should we create an openPaymentsServerErrorMiddleware like we do in backend to catch all of the GNAPErrors to have a single place for a standardized log for each error? (Since most of the logs are added right before the throw).
> 
> Might be nice to match the behaviour in `backend`

I think it's possible to have this PR also match that pattern. It'll be slightly different since it's supposed to also return [GNAP error codes](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-18#name-error-response), but will be mostly similar.

packages/auth/src/signature/middleware.ts
```diff
@@ -146,21 +153,15 @@ export async function tokenHttpsigMiddleware(
...
       GNAPErrorCode.InvalidClient,
       'invalid access token'
     )
   }
 
   if (!accessToken.grant) {
-    const logger = await ctx.container.use('logger')
-    logger.error(
-      `access token with management id ${ctx.params['id']} has no grant associated with it.`
-    )
```
mkurapov, line 161: is this log worth keeping still?

packages/auth/src/logger/service.ts
```diff
@@ -3,7 +3,14 @@ import createLogger, { Logger as PinoLogger } from 'pino'
 import { Config } from '../config/app'
 
 function initLogger(): PinoLogger {
-  const logger = createLogger()
+  const logger = createLogger({
```
mkurapov, line 6: do we need to use this when instantiating the logger singleton in index.ts?

packages/auth/src/shared/gnapErrors.ts
```diff
@@ -19,11 +19,53 @@ export interface GNAPErrorResponse {
...
+  public code: GNAPErrorCode
+
+  constructor(status: number, code: GNAPErrorCode, message?: string) {
+    super(message)
+    this.status = status
+    this.code = code
+  }
+}
+
+export async function gnapServerErrorMiddleware(
```
mkurapov, line 33: We should also check for `OpenAPIValidatorMiddlewareError` here as well, similar to `backend`, so we get a nicely formatted error when getting validation errors as well. This means we need to update `@interledger/openapi` to 2.0.1

mkurapov (commented):
Looks good, just a few minor comments

packages/auth/src/signature/middleware.ts
njlie: Probably...in hindsight that's a pretty severe case that shouldn't happen - we should know about if it does for some reason

BlairCurrey (approved):

mkurapov (commented):
just one comment

mkurapov (approved):

PR #2699, opened 2024-05-03
feat(2589): soft delete unused asset

lengyel-arpad85:
Soft delete asset including the Admin frontend part, by adding a deletedAt date and filtering get & getAll calls for deletedAt being NULL
This approach was adopted instead of hard delete since TigerBeetle has no support for deleting rows

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/resolvers/asset.ts
```diff
@@ -192,6 +197,78 @@ export const getFees: AssetResolvers<ApolloContext>['fees'] = async (
   }
 }
 
+export const deleteAsset: MutationResolvers<ApolloContext>['deleteAsset'] =
```
mkurapov, line 200: I think we should try to keep the resolvers as lean as possible. Maybe, for this particular function (since it's relatively small), we just use `WalletAddress` and `Peer` models directly in the `assetService.delete` function. This would move the business logic to the service instead, and would let you avoid adding new functions to peer and walletAddres services.

packages/backend/src/asset/errors.ts
```diff
@@ -1,8 +1,25 @@
...
+  [AssetError.DuplicateAsset]: 400,
+  [AssetError.InuseAsset]: 400
+}
+
+export const errorToMessage: {
+  [key in AssetError]: string
+} = {
+  [AssetError.UnknownAsset]: 'unknown asset',
+  [AssetError.DuplicateAsset]: 'Asset already exists',
+  [AssetError.InuseAsset]: 'Asset in use'
```
mkurapov, line 24: When reading, 'Asset in use' it doesn't seem like an error to me right away. Maybe this can include something about the fact that you couldn't delete it?

mkurapov (commented):
My main question is, what should we do when we delete an asset, then try to re-create another one with the same code and scale (via `createAsset` mutation)?

Right now, we will get a "duplicate asset" error, but that might be a bit confusing given the asset isn't visible in the UI. Maybe we can just set deletedAt to be null again, and return the asset?

lengyel-arpad85:
> My main question is, what should we do when we delete an asset, then try to re-create another one with the same code and scale (via `createAsset` mutation)?
> 
> Right now, we will get a "duplicate asset" error, but that might be a bit confusing given the asset isn't visible in the UI. Maybe we can just set deletedAt to be null again, and return the asset?

yes, that might work, but if the scale is different from what we had before, do we just update the scale to the new one ?

mkurapov:
@arpad-lengyel 

> yes, that might work, but if the scale is different from what we had before, do we just update the scale to the new one ?

An asset is unique on scale and code, so that would signify a new/different asset

mkurapov:
>  It's an unused asset, so in theory it should not matter what the previous scale was, it was never used, so it won't have an effect on anything, no ?

Right, however I would say to be safe, we should treat e.g. USD/9 vs USD/2 as completely different assets (even if they have the same asset code). This is reflected in the DB as well:

https://github.com/interledger/rafiki/blob/8835414f0f478b30ab7ee80a54e8af39ac3b1904/packages/backend/migrations/20210422194130_create_assets_table.js#L15

packages/backend/src/asset/service.ts
```diff
@@ -111,21 +132,60 @@ async function updateAsset(
...
+    const peer = await Peer.query(deps.knex)
+      .withGraphJoined('asset')
+      .where('assetId', id)
+      .first()
+    if (peer) {
+      return AssetError.InuseAsset
+    }
+
+    const walletAddress = await WalletAddress.query(deps.knex)
+      .withGraphJoined('asset')
```
mkurapov, line 154:
```suggestion
```
since we just need to check the `assetId` column, we don't need to fetch the asset from the assets table. (same above)

packages/backend/src/graphql/resolvers/asset.ts
```diff
@@ -192,6 +197,58 @@ export const getFees: AssetResolvers<ApolloContext>['fees'] = async (
...
+  async (
```
mkurapov, line 201: Let's add a Bruno request for this new API

packages/backend/src/asset/service.ts
```diff
@@ -21,10 +23,15 @@ export interface UpdateOptions {
   withdrawalThreshold: bigint | null
   liquidityThreshold: bigint | null
 }
+export interface DeleteOptions {
+  id: string
+  deletedAt: string
```
mkurapov, line 28:
```suggestion
  deletedAt: Date
```
ObjectionJS can handle the casting, I believe we don't need to call `.toISOString` manually before inserting into the DB

packages/backend/src/graphql/resolvers/asset.ts
```diff
@@ -192,6 +197,58 @@ export const getFees: AssetResolvers<ApolloContext>['fees'] = async (
...
+            }
+          case AssetError.InuseAsset:
+            return {
+              code: errorToCode[AssetError.InuseAsset].toString(),
+              message: errorToMessage[AssetError.InuseAsset],
+              success: false
+            }
+          default:
+            throw new Error(`AssetError: ${assetOrError}`)
+        }
```
mkurapov, line 229:
```suggestion
        return {
          code: errorToCode[assetOrError].toString(),
          message: errorToMessage[assetOrError],
          success: false
        }
```
I think you can get away with this

packages/backend/src/asset/errors.ts
```diff
@@ -1,8 +1,25 @@
...
+  [AssetError.DuplicateAsset]: 400,
+  [AssetError.InuseAsset]: 400
+}
+
+export const errorToMessage: {
+  [key in AssetError]: string
+} = {
+  [AssetError.UnknownAsset]: 'unknown asset',
+  [AssetError.DuplicateAsset]: 'Asset already exists',
+  [AssetError.InuseAsset]: 'Cannot delete! Asset in use.'
```
mkurapov, line 24:
```suggestion
  [AssetError.CannotDeleteInUseAsset]: 'Cannot delete! Asset in use.'
```
nit

lengyel-arpad85:
> > It's an unused asset, so in theory it should not matter what the previous scale was, it was never used, so it won't have an effect on anything, no ?
> 
> Right, however I would say to be safe, we should treat e.g. USD/9 vs USD/2 as completely different assets (even if they have the same asset code). This is reflected in the DB as well:
> 
> https://github.com/interledger/rafiki/blob/8835414f0f478b30ab7ee80a54e8af39ac3b1904/packages/backend/migrations/20210422194130_create_assets_table.js#L15

yes, you are right, I've corrected the condition in my code

packages/backend/src/graphql/resolvers/asset.ts
lengyel-arpad85: not really :) 
that is an unreachable code section because of the switch: default, and assetOrError is an object so I cannot use it as an index

packages/backend/src/asset/service.ts
```diff
@@ -50,6 +57,7 @@ export async function createAssetService({
   return {
     create: (options) => createAsset(deps, options),
     update: (options) => updateAsset(deps, options),
+    delete: (options) => deleteAsset(deps, options),
```
mkurapov, line 60: we should add tests for this in `service.test`

packages/backend/src/graphql/resolvers/asset.ts
mkurapov: Oh maybe the diff didn't show well, but I meant to replace the whole switch statement with that block. Typescript knows after `isAssetError` type guard that `assetOrError` is a string

packages/frontend/app/routes/assets.$assetId.tsx
```diff
@@ -289,12 +337,44 @@ export async function action({ request }: ActionFunctionArgs) {
...
+          location: '.'
+        })
+      }
+
+      const response = await deleteAsset({ id: result.data.id })
+      if (!response?.success) {
+        return setMessageAndRedirect({
+          session,
+          message: {
+            content: 'Could not delete asset.',
```
mkurapov, line 358: Non blocker, but maybe we can service up the error from the API here, to show why the asset could not be deleted.

Or we can add somewhere in the danger box explaining that an asset can only be deleted when not in use.

packages/backend/src/asset/service.test.ts
```diff
@@ -230,4 +234,50 @@ describe('Asset Service', (): void => {
...
+
+      const restoredAsset = await assetService.create({ code, scale })
+      assert.ok(!isAssetError(restoredAsset))
+      expect(restoredAsset.id).toEqual(assetId)
+      expect(restoredAsset.code).toEqual(code)
+      expect(restoredAsset.scale).toEqual(scale)
+      expect(restoredAsset.deletedAt).toBeNull()
+    })
+
+    test('Cannot delete in use asset', async (): Promise<void> => {
```
mkurapov, line 266: we should check the peer scenario as well

mkurapov (commented):
Looks good, just minor test comment

mkurapov (approved):

PR #2708, opened 2024-05-07
chore(backend): remove wallet address middleware

mkurapov:

To finalize the standardization of Open Payments route errors in `backend`, this PR removes `walletAddressMiddleware`. 
The benefits:
- We now handle all possible errors properly in the `openPaymentsServerErrorMiddleware` (i.e. we don't use `ctx.throw` at all in other places)
- `walletAddressUrl` is determined as an inline-middleware per route, meaning the complex logic `walletAddressMiddleware` can be removed
- We don't need to fetch the walletAddress until each route needs it. This is nice for a few reasons:
    - we can control whether a route will poll (or not) the wallet address creation at the ASE (i.e. we can choose whether a route will call `walletAddressService.getOrPollByUrl` or just `getByUrl`)
    - we can eliminate a request to get the wallet address before other checks are ran (like token introspection, http signature middleware, Open Api validation, etc)

## Context

Completion of `backend` changes for #1905

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -181,12 +188,13 @@ async function getWalletAddress(
 
 async function getOrPollByUrl(
   deps: ServiceDependencies,
-  url: string
+  url: string,
+  options: GetOrPollByUrlOptions = { mustBeActive: true }
```
mkurapov, line 192: Since we primarily use this method when resolving the wallet address in Open Payments routes, we should expect it to always be active

packages/backend/src/app.ts
```diff
@@ -547,7 +554,24 @@ export class App {
...
+        })
+
+        if (!incomingPayment?.walletAddress) {
+          throw new OpenPaymentsServerRouteError(401, 'Unauthorized', {
+            description: 'Failed to get wallet address from incoming payment',
+            id: ctx.params.id
+          })
+        }
+
+        ctx.walletAddressUrl = incomingPayment.walletAddress.url
```
mkurapov, line 572: We do not get a specific wallet address in the following OP requests (either in the query params or request body):

GET /incoming-payments/:id
GET /outgoing-payments/:id
GET /quotes/:id
POST /incoming-payments/:id/complete

We need the `walletAddressUrl` in order to pass down into `createTokenIntrospectionMiddleware`, as the middleware needs to [determine](https://github.com/interledger/rafiki/pull/2708/files#diff-efcf49ef343448125aefe358a9de22df56ac089b07acbc9d120ea35192d26ca0L98) whether the possible `identifier` on the grant matches up to the resource for the particular wallet address. 

This is why we still need to request the resource first, before we can proceed with handling the request. The status quo is returning a 401/Unauthorized response code (to not give away information about whether resource exists or not). Maybe this should be 400, but I'm not entirely sure.

mkurapov: I need to add tests for this, but I would need to do some work about splitting apart these in-line middlewares into separate functions, so I can handle it in a separate PR where I do a little more refactoring.

packages/backend/src/open_payments/payment/outgoing/routes.ts
```diff
@@ -100,9 +102,18 @@ async function createOutgoingPayment(
   deps: ServiceDependencies,
   ctx: CreateContext<CreateBody>
 ): Promise<void> {
+  const walletAddress = await deps.walletAddressService.getOrPollByUrl(
```
mkurapov, line 105: Do we want to potentially poll here? Typically, the polling probably necessary only on the receiving side, i.e. incoming payment creation only. (since that is the most likely place where we would need to allow the ASE to make a request to their backend to create the corresponding wallet addresses.

packages/backend/src/payment-method/ilp/spsp/middleware.ts
```diff
@@ -41,11 +38,11 @@ const spspMiddleware = async (
   if (ctx.accepts('application/spsp4+json')) {
     const walletAddressService = await ctx.container.use('walletAddressService')
 
-    const walletAddress = await walletAddressService.getByUrl(
+    const walletAddress = await walletAddressService.getOrPollByUrl(
       ctx.walletAddressUrl
     )
 
-    if (!walletAddress?.isActive) {
```
mkurapov, line 48: already handled in `getOrPollByUrl` by default. However, this comment still stands: https://github.com/interledger/rafiki/pull/2708/files#r1592829255

packages/backend/src/payment-method/ilp/spsp/middleware.test.ts
```diff
@@ -107,27 +107,4 @@ describe('SPSP Middleware', (): void => {
       expect(spspGetRouteSpy).not.toHaveBeenCalled()
     }
   })
-
-  test('throws error if inactive wallet address', async () => {
```
mkurapov, line 111: Already handled internally in `getOrPollByUrl`

packages/backend/src/app.ts
mkurapov: For the following:

GET /incoming-payments/:id
GET /outgoing-payments/:id
GET /quotes/:id

we _could_ just store these on `ctx.state` and not have to call the `<resource>Routes.get`/fetch the resource from the DB again, but a) could be done in separate PR, or b) not worth doing at all for the consistency of having the routes as they are

packages/backend/src/shared/pagination.ts
```diff
@@ -20,14 +20,12 @@ export function parsePaginationQueryParameters({
 type GetPageInfoArgs<T extends BaseModel> = {
   getPage: (pagination: Pagination, sortOrder?: SortOrder) => Promise<T[]>
   page: T[]
-  walletAddress?: string
```
mkurapov, line 23: was unused

PR #2709, opened 2024-05-07
feat(auth): secure admin api with hmac signatures

### LINKED ISSUE #2704: Secure Auth Server Admin API with signatures
As was done with the [backend server's admin API](https://github.com/interledger/rafiki/pull/2632), the auth server's admin API needs to be secured in the same way.

njlie:

- Added signature header verification to Auth Server Admin API

## Context

Closes #2704.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/jest.setup.js
+++ b/packages/auth/jest.setup.js
@@ -73,6 +73,6 @@ module.exports = async (globalConfig) => {
       .start()
 
     global.__AUTH_REDIS__ = redisContainer
-    process.env.REDIS_URL = `redis://localhost${redisContainer.getMappedPort(REDIS_PORT)}`
+    process.env.REDIS_URL = `redis://localhost:${redisContainer.getMappedPort(REDIS_PORT)}`
   }
 }
--- a/packages/auth/jest.setup.js
+++ b/packages/auth/jest.setup.js
@@ -4,6 +4,7 @@ const { knex } = require('knex')
 const { GenericContainer, Wait } = require('testcontainers')
 
 const POSTGRES_PORT = 5432
+const REDIS_PORT = 6379
 
 module.exports = async (globalConfig) => {
   const workers = globalConfig.maxWorkers
@@ -66,4 +67,12 @@ module.exports = async (globalConfig) => {
   }
 
   global.__AUTH_KNEX__ = db
+  if (!process.env.REDIS_URL) {
+    const redisContainer = await new GenericContainer('redis:7')
+      .withExposedPorts(REDIS_PORT)
+      .start()
+
+    global.__AUTH_REDIS__ = redisContainer
+    process.env.REDIS_URL = `redis://localhost${redisContainer.getMappedPort(REDIS_PORT)}`
+  }
 }
--- a/packages/auth/src/config/app.ts
+++ b/packages/auth/src/config/app.ts
@@ -44,8 +44,9 @@ export const Config = {
   identityServerUrl: envString('IDENTITY_SERVER_URL'),
   identityServerSecret: envString('IDENTITY_SERVER_SECRET'),
   authServerUrl: envString('AUTH_SERVER_URL'),
-  apiSecret: process.env.API_SECRET, // optional
-  apiSignatureVersion: envInt('API_SIGNATURE_VERSION', 1),
+  adminApiSecret: process.env.ADMIN_API_SECRET, // optional
+  adminApiSignatureVersion: envInt('ADMIN_API_SIGNATURE_VERSION', 1),
+  adminApiSignatureTtl: envInt('ADMIN_API_SIGNATURE_TTL_SECONDS', 30),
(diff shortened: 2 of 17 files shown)


packages/auth/src/config/app.ts
```diff
@@ -45,6 +45,8 @@ export const Config = {
     'AUTH_SERVER_DOMAIN',
     `http://localhost:${envInt('AUTH_PORT', 3006)}`
   ),
+  apiSecret: process.env.API_SECRET, // optional
+  apiSignatureVersion: envInt('API_SIGNATURE_VERSION', 1),
```
mkurapov, line 49: Not related to this PR necessarily, but what is the benefit of having version? just to track how many times a key has been rotated, basically?

Also, should we call this "Admin API secret"?

packages/auth/src/shared/utils.ts
```diff
@@ -7,3 +10,28 @@ export function generateNonce(): string {
...
+
+export function verifyApiSignature(ctx: Context, config: IAppConfig): boolean {
+  const { headers, body } = ctx.request
+  const signature = headers['signature']
+  if (!signature) {
+    return false
+  }
+
+  const signatureParts = (signature as string)?.split(', ')
+  const timestamp = signatureParts[0].split('=')[1]
```
golobitch, line 22: Replay attack? So nothing here prevents us to get the same and same request over and over again, correct. As long that once it was correct, we can get it n number of times.

I would do it like this:
1. Save the signature to redis for example with key: `signature:${headers['signature']}`
2. On line 20 check if this key exists. If exists, return Unauthorized
3. Check the validity of request ... basically...how long after the request has been signed, we can still process this request? Let's say 30 seconds? Or even less? Also note, that this value is also the TTL of the above signature key

packages/auth/src/config/app.ts
njlie: Not sure what the answer to having version is, but I'm down to rename the env var

packages/auth/src/shared/utils.ts
njlie: Need to do this on the `backend` package as well. Going to capture in https://github.com/interledger/rafiki/issues/2725

packages/auth/src/shared/utils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  let redis: Redis
+
+  describe('admin api signatures', (): void => {
+    beforeAll(async (): Promise<void> => {
+      deps = initIocContainer({
+        ...Config,
+        adminApiSecret: 'test-secret'
+      })
+      appContainer = await createTestApp(deps)
+      const redis = await deps.use('redis')
```
BlairCurrey, line 24: (doesn't fix the `MaxRetriesPerRequestError` error)

```suggestion
     redis = await deps.use('redis')
```

packages/auth/jest.setup.js
```diff
@@ -66,4 +67,12 @@ module.exports = async (globalConfig) => {
   }
 
   global.__AUTH_KNEX__ = db
+  if (!process.env.REDIS_URL) {
+    const redisContainer = await new GenericContainer('redis:7')
+      .withExposedPorts(REDIS_PORT)
+      .start()
+
+    global.__AUTH_REDIS__ = redisContainer
+    process.env.REDIS_URL = `redis://localhost${redisContainer.getMappedPort(REDIS_PORT)}`
```
BlairCurrey, line 76: missing colon

```suggestion
    process.env.REDIS_URL = `redis://localhost:${redisContainer.getMappedPort(REDIS_PORT)}`
```

BlairCurrey (dismissed):

packages/auth/src/shared/utils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      const requestBody = { test: 'value' }
+      const signature = generateApiSignature(
+        'test-secret',
+        Config.adminApiSignatureVersion,
+        requestBody
+      )
+      const key = `signature:${signature}`
+      const op = redis.multi()
+      op.set(key, signature)
+      op.expire(key, signature)
```
mkurapov, line 90: should the expiry take in a num as the second argument?

packages/auth/src/shared/utils.ts
```diff
@@ -24,3 +27,90 @@ export function generateRouteLogs(ctx: AppContext): {
...
+  logger.info(
+    {
+      currentTime,
+      signatureTime,
+      diff: currentTime - signatureTime,
+      ttlMilliseconds
+    },
+    'time differential'
+  )
+  if (currentTime - signatureTime > ttlMilliseconds) return false
```
mkurapov, line 88: I think we should add a test for this as well

BlairCurrey (approved):

mkurapov (approved):

PR #2716, opened 2024-05-09
feat(frontend): admin ui changes

### LINKED ISSUE #2681: Admin UI changes from workweek
## Context
In Cluj for the breakpoint work week, we ended up going over the current state of the Admin UI, and came up with a few changes that we should address:

## Todos
- [ ] Rename “Webhooks” tab to “Webhook Events”
- [ ] Swap assets tab with peer tab (since we need Assets first to have Peers)
- [ ] Tooltips/descriptions over each of the input fields in the peering page (in general, more descriptions/links out to the docs)

njlie:

- Updated the Admin UI with the following:
  - Changed `Webhook` to `Webhook Data`.
  - Swapped `Asset` and `Peer` tabs.
  - Added tooltips to `Peer` section with links to documentation.

## Context

Fixes #2681.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/frontend/app/components/ui/Input.tsx
+++ b/packages/frontend/app/components/ui/Input.tsx
@@ -8,10 +8,11 @@ type InputProps = Omit<ComponentPropsWithoutRef<'input'>, 'className'> & {
   label?: string
   error?: string | string[]
   addOn?: ReactNode
+  description?: ReactNode
 }
 
 export const Input = forwardRef<HTMLInputElement, InputProps>(
-  ({ addOn, label, type, error, id, ...props }, ref) => {
+  ({ description, addOn, label, type, error, id, ...props }, ref) => {
     const internalId = id ?? useId()
 
     return (
@@ -38,6 +39,9 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
             {...props}
           />
         </div>
+        {description ? (
+          <div className='font-medium text-sm'>{description}</div>
+        ) : null}
         <FieldError error={error} />
       </div>
     )
--- a/packages/frontend/app/routes/peers.$peerId.tsx
+++ b/packages/frontend/app/routes/peers.$peerId.tsx
@@ -121,6 +121,18 @@ export default function ViewPeerPage() {
                     defaultValue={peer.name ?? ''}
                     placeholder='Peer name'
                     error={response?.errors.general.fieldErrors.name}
+                    description={
+                      <>
+                        The name of the{' '}
+                        <a
+                          className='tooltip'
+                          href='https://rafiki.dev/concepts/interledger-protocol/peering/'
+                        >
+                          peer
+                        </a>
+                        .
+                      </>
+                    }
                   />
                   <Input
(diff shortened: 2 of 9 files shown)


packages/frontend/app/routes/peers.$peerId.tsx
```diff
@@ -121,6 +121,18 @@ export default function ViewPeerPage() {
                     defaultValue={peer.name ?? ''}
                     placeholder='Peer name'
                     error={response?.errors.general.fieldErrors.name}
+                    description={
+                      <>
+                        The name of the{' '}
+                        <a
+                          className='tooltip'
+                          href='https://rafiki.dev/concepts/interledger-protocol/peering/'
+                        >
```
BlairCurrey, line 130: This would apply to all these links:

Should it open in a new tab? user likely wants to resume whatever they were doing here after visiting this site for more context.

BlairCurrey (commented):
Non-blocking, just some food for thought.

From the task I was expecting a tooltip more [like this](https://www.w3schools.com/css/css_tooltip.asp). Perhaps in an info icon like in [this example](https://dribbble.com/shots/22959995-Number-Inputs). As opposed to descriptions under the field. I guess I can see a case for that too though, if we wanted the extra information to be more prominent. An info icon tooltip would be a bit less busy though IMO and keep the error message closer to the input as well. How it currently looks, for context:

image

packages/frontend/app/styles/tailwind.css
```diff
@@ -1,3 +1,8 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+a.tooltip {
+  color: revert;
+  text-decoration: revert;
+}
```
BlairCurrey, line 8: My understanding is this restores some original styling for `a` links that tailwind strips out.

This seems like a good way to do it. We've handled this differently elsewhere (a bunch of tailwind utility classnames) but I think what you've done here is simpler and easier to use. 

If it makes sense to you, please merge this PR which standardizes on this method: https://github.com/interledger/rafiki/pull/2719. I also renamed the class to `default-link` and updated reference since it would apply to more than tooltips.

BlairCurrey: thanks for the review, merged in

BlairCurrey (approved):
Looks good to me. Remaining comments aren't really blocking.

packages/frontend/app/components/ui/Input.tsx
```diff
@@ -38,6 +39,9 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
             {...props}
           />
         </div>
+        {description ? (
+          <div className='font-medium text-sm'>{description}</div>
```
mkurapov, line 43:
```suggestion
          <div className='font-light italic mt-1 text-sm'>{description}</div>
```
maybe? to give a bit more visual separation

Screenshot 2024-05-13 at 19 09 39

mkurapov (approved):

PR #2722, opened 2024-05-13
chore(backend): update wallet address middleware & error handling

### LINKED ISSUE #1905: Standardize Open Payment errors

mkurapov:

To finalize the standardization of Open Payments route errors in `backend`, this PR splits up `walletAddressMiddleware` into several separate middlewares. One group of them deals with adding the walletAddressUrl onto the ctx (from several resources), while `getWalletAddressForSubresource` simply polls for the active wallet address by using the `ctx.walletAddressUrl` itself. Each of the middlewares is then added explicity to the route.

The benefits:
- We now handle all possible Open Payments API errors properly in the `openPaymentsServerErrorMiddleware` (i.e. we don't use `ctx.throw` at all in other places)
- `walletAddressUrl` is added to the `ctx` as a middleware specific to the route route, meaning we can remove all-encompassing logic in `walletAddressMiddleware`. Likewise, we don't need to actually fetch the wallet address before doing token introspection and http sig verification

Other changes:
- When wallet address is requested, return the Open Payments one by default. 
- SPSP payment pointer can only be fetched when setting `accept` header to include `'application/spsp4+json'` explicitly 

## Context

Fixes #1905 for backend

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/app.ts
```diff
@@ -622,43 +632,44 @@ export class App {
...
+      createValidatorMiddleware(
         walletAddressServerSpec,
         {
           path: '/',
           method: HttpMethod.GET
         },
         validatorMiddlewareOptions
       ),
+      getWalletAddressUrlFromPath,
+      createSpspMiddleware(this.config.spspEnabled),
```
mkurapov, line 672: do not call `getWalletAddressForSubresource` here and in the jwks.json route since we have different behaviours in these routes for handling `deps.config.walletAddressUrl === ctx.walletAddressUrl`

BlairCurrey:
> * `walletAddressUrl` is added to the `ctx` as a middleware specific to the route route, meaning we can remove all-encompassing logic in `walletAddressMiddleware`. Likewise, we don't need to actually fetch the wallet address before doing token introspection and http sig verification

This is much better. Low coupling/high cohesion. 👍

packages/backend/src/open_payments/wallet_address/routes.ts
```diff
@@ -1,48 +1,60 @@
...
 }
 
-// Spec: https://docs.openpayments.guide/reference/get-public-account
 export async function getWalletAddress(
   deps: ServiceDependencies,
-  ctx: WalletAddressContext
+  ctx: WalletAddressUrlContext
 ): Promise<void> {
-  if (!ctx.walletAddress) {
+  // The instance's wallet address is only relevant for the /jwks.json route, so "hide" it when directly requested
```
BlairCurrey, line 36: Does this mean we just shouldn't have a `/GET` route for the wallet address? If we still need this route then when would we need it? Comment seems to suggest we only need the `/jwks.json` route IMO but maybe im not following.

BlairCurrey: When I call the `Examples > Open Payments > Get receiver wallet address` bruno request (which creates the wallet address) then call `Open Payments APIs > Get a Wallet Address` I receive the the wallet address.

If I copy/paste the url into the browser `http://localhost:4000/accounts/asmith` then I see `Could not get wallet address`.

How should it be working? It seems like the bruno request should not return the wallet  address according to this comment.

mkurapov: @BlairCurrey

`deps.config.walletAddressUrl` is the configured wallet address for the Rafiki instance itself, i.e. its the wallet address used to get incoming payments/incoming payment grants when creating "remote" incoming payments at other Rafikis. e.g. https://ilp.rafiki.money/.well-known/pay/. 

Maybe I can word this comment better, but we actually don't want to resolve https://ilp.rafiki.money/.well-known/pay/ (this shouldn't be a wallet address which resources can be created on, its not for an actual user account). As a result, we only care about the https://ilp.rafiki.money/.well-known/pay/jwks.json path, since it gets looked up to check the keys of the rafiki instance when validating the grant on another rafiki.

mkurapov: All other "normal" wallet addresses, should get resolved as usual with the correct asset, publicName, auth/resourceServer urls (since they represent actual user accounts)

BlairCurrey: Thank you for the context.

Same result with `http://localhost:3000/accounts/gfranklin` in the browser, but not with bruno. This is not the case on main - this request works as expected in the browser. The only difference between bruno and the browser I see is the host header that gets added in the pre-request script. Not sure if that's the actual difference.

Here is the error:

```
{"level":30,"time":1715787978022,"pid":1,"hostname":"cloud-nine-wallet-backend","method":"GET","path":"/accounts/gfranklin","message":"Could not get wallet address","msg":"Received error when handling SPSP request"}
```

Looks like its failing at the `if (!walletAddress?.isActive) {` check of the `spspsMiddleware`.

BlairCurrey: going back just 1 step further it looks like `if (ctx.accepts('application/spsp4+json')) {` resolves to false with the bruno request, and true for the browser.

mkurapov: What happens when we call `http://localhost:3000/accounts/gfranklin` in the browser, or without the host header in bruno, is we actually end up updating the wallet address via the `wallet_address.not_found` webhook event. (ie if you request `http://localhost:3000/accounts/gfranklin` on main, the wallet address will end up changing from `https://cloud-nine-wallet-backend/accounts/gfranklin`  to `http://localhost:3000/accounts/gfranklin`. 

This PR is different for SPSP request since we use `walletAddressService.getByUrl` instead of `walletAddressService.getOrPollByUrl`. However, I think this PR has same behaviour when we request a wallet address _without_ going through SPSP as well just by requesting the wallet address with `Accept: application/json`. I think we should update the MASE code to prevent this, and seems like a separate issue in general.

I'm not sure why I decided to use `walletAddressService.getByUrl`  in SPSP instead of polling (maybe I assume if you are using SPSP then we should already have a wallet address)?

This makes me wonder if we should default to returning the Open Payments wallet address actually, and only return the SPSP details if explicitly requested?

mkurapov: > This makes me wonder if we should default to returning the Open Payments wallet address actually, and only return the SPSP details if explicitly requested?

Chatted w Sabine regarding this, decision made to return Open Payments wallet address by default when fetching it. Also made a change to return SPSP wallet address only if `'application/spsp4+json'` is present in accept header explicitly.

packages/backend/src/app.ts
```diff
@@ -622,36 +632,37 @@ export class App {
...
+      getWalletAddressUrlFromPath,
+      walletAddressKeyRoutes.get
     )
 
     // Add the wallet address query route last.
     // Otherwise it will be matched instead of other Open Payments endpoints.
     router.get(
       WALLET_ADDRESS_PATH,
-      createWalletAddressMiddleware(),
+      getWalletAddressUrlFromPath,
```
mkurapov, line 663: Basically, if we are fetching the SPSP payment pointer, we don't need to call the validator middleware since its not Open Payments related

packages/backend/src/payment-method/ilp/spsp/middleware.ts
```diff
@@ -38,7 +39,7 @@ const spspMiddleware = async (
   ctx: SPSPWalletAddressContext,
   next: () => Promise<unknown>
 ): Promise<void> => {
-  if (ctx.accepts('application/spsp4+json')) {
+  if (ctx.request.header.accept?.includes(SPSP_CONTENT_TYPE_V4)) {
```
mkurapov, line 42: this `ctx.accepts` was the reason why by default when fetching the wallet address, the SPSP details were returned, since usually, the `accept` header included catch all `*/*` route, which to `ctx.accepts` would be truthy. 

Now, we need to explicitly add `application/spsp4+json` in the header to get the SPSP details

BlairCurrey (approved):

mkurapov:
Going to close and re-open to see if I can get past the stuck actions (I think affected by their degradation [here](https://www.githubstatus.com/incidents/nj47vccwm2zj))

Issue #2707, opened 2024-05-07
Investigate Shared Logger Package


BlairCurrey:
Look into creating a shared logging package with to be used across our packages. May include graphql logging plugin and similar (if any). Is there a value-add there?

BlairCurrey:
## Results:

There is practically no configuration for all of our loggers. Basically just the level. In light of that I don't really see any value-add for making a shared logger. If in the future we ever say "ensure all of our loggers are configured in [X] way" then it would probably be a good idea to make them shared then.

However, while investigating I found a (practically) unused `Logger` in `backend` and `auth`. I figure I might as well address that here.

PR #2724, opened 2024-05-15
refactor: use pino logger type directly, rm unused logger

BlairCurrey:

-  removes a couple mostly unused `Logger`s. They were only used for the return type so I just used the return type directly instead.

## Context

fixes https://github.com/interledger/rafiki/issues/2707. See comment in issue.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/auth/src/logger/service.ts
```diff
@@ -1,11 +0,0 @@
-import createLogger, { Logger as PinoLogger } from 'pino'
```
mkurapov, line 1: Might still be relevant for https://github.com/interledger/rafiki/pull/2697/files#diff-9c3c6613a84f50df315ff190bc76ca450044920269a9467a6ad13e0bf10753c5

njlie: I think it's not, actually. Just me covering all the bases but I don't plan on expanding its usage.

njlie (approved):

Issue #2745, opened 2024-05-24
[BUG] Internal Server Error on Interact Redirect URL


BlairCurrey:
**Description**

When navigating to the interact url there is an `Request failed with status code 500` and I see this error in auth:

```
  "err": {
    "type": "RangeError",
    "message": "Input buffers must have the same byte length",
    "stack": "RangeError: Input buffers must have the same byte length\n    at getGrantDetails (/home/rafiki/packages/auth/src/interaction/routes.ts:122:13)\n    at details (/home/rafiki/packages/auth/src/interaction/routes.ts:105:35)\n    at dispatch (/home/rafiki/node_modules/.pnpm/koa-compose@4.1.0/node_modules/koa-compose/index.js:42:32)\n    at /home/rafiki/node_modules/.pnpm/@interledger+openapi@2.0.1/node_modules/@interledger/openapi/dist/middleware.js:27:15\n    at dispatch (/home/rafiki/node_modules/.pnpm/koa-compose@4.1.0/node_modules/koa-compose/index.js:42:32)\n    at /home/rafiki/node_modules/.pnpm/@koa+router@12.0.0/node_modules/@koa/router/lib/router.js:425:16\n    at next (/home/rafiki/node_modules/.pnpm/koa-compose@4.1.0/node_modules/koa-compose/index.js:42:32)\n    at gnapServerErrorMiddleware (/home/rafiki/packages/auth/src/shared/gnapErrors.ts:38:11)\n    at dispatch (/home/rafiki/node_modules/.pnpm/koa-compose@4.1.0/node_modules/koa-compose/index.js:42:32)\n    at /home/rafiki/node_modules/.pnpm/@koa+router@12.0.0/node_modules/@koa/router/lib/router.js:425:16",
    "code": "ERR_CRYPTO_TIMING_SAFE_EQUAL_LENGTH"
  },
```

It does not happen in the integration tests. 

Looks like the root cause are some env value updates - probably just need to adjust the `x-idp-secret` header in the mock ase like this one was in the tests: https://github.com/interledger/rafiki/commit/99e43d7901de35f9eb07825502f9d6e1080ef2c9#diff-7d108c79df02eb44254b549745bd165705873ac17551f3f4b66c1e409bb82b04R60

**To Reproduce**

1. Run bruno "Open Payments" requests through `Grant Request Outgoing Payment`
2. Visit the interact.redirect url from the `Grant Request Outgoing Payment` response.
[...]

PR #2746, opened 2024-05-24
fix(mock-ase): interact bug by updating idp secret

BlairCurrey:

- updates `x-idp-secret` header in `mock-account-servicing-entity` to fix interact bug

## Context

fixes https://github.com/interledger/rafiki/issues/2745

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/localenv/mock-account-servicing-entity/app/lib/apiClient.ts
+++ b/localenv/mock-account-servicing-entity/app/lib/apiClient.ts
@@ -8,8 +8,6 @@ export const StepNames = {
   endInteraction: 3
 }
 
-const IDP_SECRET = '2pEcn2kkCclbOHQiGNEwhJ0rucATZhrA807HTm2rNXE='
-
 /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
 export type ApiResponse<T = any> = (
   | {
@@ -34,7 +32,8 @@ export class ApiClient {
    */
 
   public static async getGrant(
-    params: Record<string, string>
+    params: Record<string, string>,
+    idpSecret: string
   ): Promise<ApiResponse> {
     // get grant --> GET /grant/:id/:nonce
     const { interactId, nonce } = params
@@ -42,7 +41,7 @@ export class ApiClient {
       `http://localhost:3006/grant/${interactId}/${nonce}`,
       {
         headers: {
-          'x-idp-secret': IDP_SECRET
+          'x-idp-secret': idpSecret
         }
       }
     )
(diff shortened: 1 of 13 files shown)


localenv/mock-account-servicing-entity/app/lib/apiClient.ts
```diff
@@ -8,6 +8,8 @@ export const StepNames = {
   endInteraction: 3
 }
 
+const IDP_SECRET = '2pEcn2kkCclbOHQiGNEwhJ0rucATZhrA807HTm2rNXE='
```
mkurapov, line 11: can we pass this through `process.env`?

BlairCurrey: it was a bit involved but yes [15178f6](https://github.com/interledger/rafiki/pull/2746/commits/15178f6f23941359e450d97199c3aa4d5845696a)

golobitch:
Yes, apparently I introduced this bug when working on auth env variables. Thank you for spotting and fixing this @BlairCurrey

localenv/mock-account-servicing-entity/app/lib/parse_config.server.ts
```diff
@@ -3,6 +3,10 @@ import { readFileSync } from 'fs'
 import { loadOrGenerateKey } from '@interledger/http-signature-utils'
 import type { Config } from 'mock-account-service-lib'
 
+if (!process.env.IDP_SECRET) {
+  throw new Error('environment variable IDP_SECRET is required')
+}
```
mkurapov, line 8: Should we do the same for the other env vars? Most of them seem mandatory

localenv/mock-account-servicing-entity/app/routes/mock-idp._index.tsx
```diff
@@ -1,5 +1,13 @@
 import ConsentScreen from '~/routes/consent-screen'
+import { json } from '@remix-run/node'
+import { useLoaderData } from '@remix-run/react'
+import { CONFIG } from '~/lib/parse_config.server'
+
+export function loader() {
+  return json({ idpSecret: CONFIG.idpSecret })
+}
```
mkurapov, line 8: Just thinking out loud, should some of these requests (ie, those requested via ApiClient.*) be handled in the `action()` handler on the server side instead? Maybe I'm just not used to remix, but seems a bit odd to have to return a env variable (that's probably best to keep secret) to the front-end to be used for admin API requests.

mkurapov (approved):
Non blocking comments, just worth consideration

PR #2748, opened 2024-05-27
chore(backend): allow asset scale 2 to asset scale 9 payments

### LINKED ISSUE #2747: Cannot send from asset scale 2 to asset scale 9
## Context

We are unable to send payments between asset scale 2 and asset scale 9 wallet addresses. This is because of how we are storing the results of the ilp quote exchange rates in the DB.

First, the ILP pay library starts the rate probe from 10^12 value packets:
https://github.com/interledgerjs/interledgerjs/blob/9b740a831c5221172230dfb43d0f5198d3b56848/packages/pay/src/senders/rate-probe.ts#L35

and because the resulting exchange rate ends up being 1.00^7, ie 1^(9-2), we try to store 10000000000000000001 as the value for highEstimatedExchangeRate in the DB, and end up failing because highEstimatedExchangeRateNumerator (just like all of the other *numerator/denominator fields in the quotes table) is a type of `bigint`, with a max value of 9223372036854775807.

Additional details from @raducristianpopa's [document](https://docs.google.com/document/d/181RrPtkMVieRHjKw39JakRUzRNXOsqI90NPWUKB6dWo/edit?usp=sharing): 

> I tried sending money from asset scale 2 to asset scale 9.

> In WM, the extension/browser implementation should be able to send money without worrying about the scale and it just needs to make some adjustments. What I am currently testing with:
Rate of pay (USD/h): $0.60/h
Connected wallet address: USD, asset scale 2
Monetization link href (wallet address in a webpage): USD, asset scale 9

> Since the connected wallet cannot send a payment every second (0.60 / 3600s = 0.000166…..), we need to adjust the amount that needs to be sent. The minimum amount for this use-case is 1 unit of asset scale 2 ($0.01), to which we add another unit, because @interledger/pay subtracts 1 unit when using a debit amount.

> With the mentions from above we have:
Minimum amount: $0.02 - 2 units in asset scale 2
Interval: not really important for the current issue, but it needs to be adjusted as well

> Now, since we know how much we need to send through WM we can proceed by creating the quote + outgoing payment. When creating the quote, the debit amount is:

```js
debitAmount: {
  value: "2",
  assetCode: "USD",
  assetScale: 2
}
```

> But … when making the POST request to create it, I am getting back an Internal Server Error. When inspecting the logs, I see the following error:

```
…
{"type":"DataError","message":"insert into \"quotes\" (\"additionalFields\", \"assetId\", \"client\", \"debitAmountValue\", \"estimatedExchangeRate\", \"expiresAt\", \"feeId\", \"highEstimatedExchangeRateDenominator\", \"highEstimatedExchangeRateNumerator\", \"id\", \"lowEstimatedExchangeRateDenominator\", \"lowEstimatedExchangeRateNumerator\", \"maxPacketAmount\", \"minExchangeRateDenominator\", \"minExchangeRateNumerator\", \"receiveAmountAssetCode\", \"receiveAmountAssetScale\", \"receiveAmountValue\", \"receiver\", \"walletAddressId\") values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19, $20) returning \"id\" - value \"10000000000000000001\" is out of range for type bigint"
…
```

> The big number represents the low/high estimated exchange rate. 
Output from Pay.startQuote (see additionalFields):

```js
{
  "receiver": {
    "ilpAddress": "test.net.G7Q7M9cJOmrq64SnMqFKp_9obMXoYi7NlJnNCYRL1qS_8LH0rZQVKWMF8KXylnAzWuMkRzYrin2XAEnuxqmhoYTAZlJFYRc",
    "sharedSecret": { <redacted> },
    "assetCode": "USD",
    "assetScale": 9,
    "incomingPayment": {
      "id": "https://ilp.rafiki.money/incoming-payments/582a4e08-c284-44f6-bb07-6e0f9ea973bd",
      "walletAddress": "https://ilp.rafiki.money/radu-wm",
      "receivedAmount": {
        "value": "0",
        "assetCode": "USD",
        "assetScale": 9
      },
      "completed": false,
      "metadata": {
        "source": "Web Monetization"
      },
      "createdAt": "2024-05-23T05:44:00.789Z",
      "updatedAt": "2024-05-23T05:44:00.789Z",
      "expiresAt": "2024-05-23T06:44:00.473Z",
      "methods": <redacted>
    }
  },
  "walletAddress": <redacted>,
  "estimatedExchangeRate": 10000000,
  "debitAmount": {
    "value": "2",
    "assetCode": "USD",
    "assetScale": 2
  },
  "receiveAmount": {
    "value": "9900000",
    "assetCode": "USD",
    "assetScale": 9
  },
  "additionalFields": {
    "lowEstimatedExchangeRate": [
      "10000000000000000000",
      "1000000000000"
    ],
    "highEstimatedExchangeRate": [
      "10000000000000000001",
      "1000000000000"
    ],
    "minExchangeRate": [
      "9900000",
      "1"
    ],
    "maxPacketAmount": "18446744073709551615"
  }
}
```

## Todos
- [ ] Fix quotes table to allow making payments between asset scale 2 and asset scale 9 wallet addresses

mkurapov:

- Fix issue where we are unable to send payments between asset scale 2 and asset scale 9 wallet addresses because the ILP rate probe exchange rate numerators were too big to store in the DB as bigints. See details in issue. 
- Update small logic where we allow fees to make up more than 50% of receive amount: https://github.com/interledger/rafiki/pull/2748/files#r1616401580
- Some minor error/logging additions, refactoring create quote function slightly

## Context

Fixes #2747 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -195,6 +196,10 @@ async function pay(
   try {
     const receipt = await Pay.pay({ plugin, destination, quote })
 
+    if (receipt.error) {
+      throw receipt.error
+    }
```
mkurapov, line 201: move this check before the debug logger, so we don't log "'ILP payment completed'" when there is an error

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -131,20 +131,17 @@ async function createQuote(
...
             .lowEstimatedExchangeRate as Pay.Ratio,
           highEstimatedExchangeRate: quote.additionalFields
             .highEstimatedExchangeRate as Pay.PositiveRatio,
-          // Patch using createdAt below
-          expiresAt: new Date(0),
+          expiresAt: new Date(0), // expiresAt is patched in finalizeQuote
           client: options.client,
           feeId: sendingFee?.id,
-          estimatedExchangeRate: quote.estimatedExchangeRate,
-          additionalFields: quote.additionalFields
```
mkurapov, line 147: don't store `additionalFields` anymore

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -245,20 +246,6 @@ async function pay(
   }
 }
 
-function fromJSONtoRatio(ratio: unknown): Pay.Ratio {
```
mkurapov, line 248: not needed anymore, we don't use `additionalFields` quote field to get these ratio amounts

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -195,55 +192,87 @@ export async function resolveReceiver(
...
+      { fees, exchangeAdjustedFees, estimatedExchangeRate, receiveAmountValue },
+      'Negative receive amount when calculating quote amount'
     )
-    receiveAmountValue -= exchangeAdjustedFees
+    throw QuoteError.NegativeReceiveAmount
+  }
 
-    if (receiveAmountValue <= exchangeAdjustedFees) {
-      throw QuoteError.NegativeReceiveAmount
-    }
```
mkurapov, line 234: Behaviour change:
We used to throw if the receiveAmount was less than the fees: e.g. if receiveAmountValue = 100, and exchangeAdjustedFees were 60, we would throw, even though receiveAmountValue of 40 seems like its ok (just more than 50% of the transfer goes to fees, which wouldn't be nice for the user, but not necessarily a problem)

After a [discussion](https://interledgerfoundation.slack.com/archives/C06KZCKEEKW/p1716562815852079),

packages/backend/migrations/20240524102509_drop_quote_ratio_columns.js
```diff
@@ -0,0 +1,33 @@
...
+exports.up = function (knex) {
+  return knex.schema.alterTable('quotes', (table) => {
+    table.string('minExchangeRateNumerator').alter()
+    table.string('minExchangeRateDenominator').alter()
+    table.string('lowEstimatedExchangeRateNumerator').alter()
+    table.string('lowEstimatedExchangeRateDenominator').alter()
+    table.string('highEstimatedExchangeRateNumerator').alter()
+    table.string('highEstimatedExchangeRateDenominator').alter()
+
+    table.dropColumn('additionalFields')
```
mkurapov, line 14: Originally, `additionalFields` was a temp solution when building out support for multiple payment methods. Now though, since we will not be doing that, I decided to get rid of this extra field.

packages/backend/migrations/20240524102509_drop_quote_ratio_columns.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('quotes', (table) => {
+    table.decimal('minExchangeRateNumerator', 64, 0).alter()
+    table.decimal('minExchangeRateDenominator', 64, 0).alter()
+    table.decimal('lowEstimatedExchangeRateNumerator', 64, 0).alter()
+    table.decimal('lowEstimatedExchangeRateDenominator', 64, 0).alter()
+    table.decimal('highEstimatedExchangeRateNumerator', 64, 0).alter()
+    table.decimal('highEstimatedExchangeRateDenominator', 64, 0).alter()
```
mkurapov, line 12: needed something bigger than 20, (I don't think we will ever need more than 20 ever), but just being generous so we don't run into this issue again

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -195,55 +192,87 @@ export async function resolveReceiver(
...
-    // FixedDelivery
-    const fees = quote.fee?.calculate(debitAmountValue) ?? 0n
-    debitAmountValue = BigInt(debitAmountValue) + fees
-    if (
-      debitAmountValue < quote.debitAmount.value ||
-      debitAmountValue <= BigInt(0)
-    ) {
-      throw QuoteError.InvalidAmount
-    }
-  } else {
```
mkurapov, line 221: split these up into functions, logic remains the same other than the comment below

packages/backend/src/payment-method/ilp/service.test.ts
```diff
@@ -681,41 +697,5 @@ describe('IlpPaymentService', (): void => {
         expect((error as PaymentMethodHandlerError).retryable).toBe(false)
       }
     })
-
-    test('throws if invalid quote data', async (): Promise<void> => {
```
mkurapov, line 685: we don't use `additionalFields` so parsing JSON is not required, so this test is not necessary

packages/backend/src/payment-method/ilp/service.test.ts
```diff
@@ -272,6 +284,7 @@ describe('IlpPaymentService', (): void => {
         minDeliveryAmount: -1n
       } as Pay.Quote)
 
+      expect.assertions(4)
```
mkurapov, line 287: to make sure exceptions in catch are tracked

packages/backend/migrations/20240524102509_drop_quote_ratio_columns.js
BlairCurrey: Seems reasonable. Inspecting the db shows these are stored as `numeric` types (same as decimal) in postgres. Found this in the docs about how much storage this would take:

> Numeric values are physically stored without any extra leading or trailing zeroes. Thus, the declared precision and scale of a column are maximums, not fixed allocations. (In this sense the numeric type is more akin to varchar(n) than to char(n).) The actual storage requirement is two bytes for each group of four decimal digits, plus three to eight bytes overhead.

https://www.postgresql.org/docs/current/datatype-numeric.html

So it varies on actual size of the data, not just this declaration. So I don't really see a downside of leaving lots of extra headroom.

BlairCurrey: Were you able to test this migration against any data? Since we're going from lower to higher precision I imagine it probably works fine but figure it's worth testing.

mkurapov: @BlairCurrey I just made some transactions on `main` then rebuilt the backend and continued making transactions without issue

BlairCurrey (approved):

Issue #2739, opened 2024-05-22
Allow `createWalletAddress` mutation to take in additional properties


mkurapov:
### Context
During wallet address creation, we want to let the ASE allow to add additional properties to be associated with the wallet address.

We need to update the `createWalletAddress` mutation input to take in additionalProperties.
These properties would be optionally visible when requesting a wallet address through Open Payments. The input is an optional array of objects with `key` & `value` and `visibleInOpenPayments` (suggested name), a flag that enables or disables this field from appearing in Open Payments Wallet Address lookup.

### Tasks
- [x] Update `createWalletAddress` mutation input
- [x] Store these fields in a new table with structure: `walletAddressId: FK` `fieldKey: string` `fieldValue: string` `visibleInOpenPayments: bool`, `createdAt: Date`, `updatedAt: Date`)
- [x] Expose these new fields when fetching wallet address via Open Payments (making sure to show only fields that are indeed enabled for visibility for Open Payments)
- [x] Expose these new fields when fetching wallet address via the Admin API
- [x] Update the documentation

### PR
https://github.com/interledger/rafiki/pull/2752

PR #2752, opened 2024-05-29
feat(2739)!: additional properties for createwalletaddress

koekiebox:
# Story
https://github.com/interledger/rafiki/issues/2739

- Add additionalProperty support for `createWalletAddress`.

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -170,22 +208,75 @@ async function updateWalletAddress(
...
+
+  if (includeAddProps && returnVal) {
+    returnVal.additionalProperties = await getWalletAdditionalProperties(
+      deps,
+      returnVal.id,
+      includeVisibleOnlyAddProps
+    )
+  }
+  return returnVal
+}
```
BlairCurrey, line 244: Not blocking and maybe just my opinion, but I think it would be cleaner to just have separate `getWalletAdditionalProperties` and `getWalletAddress` methods without this central one.

packages/backend/src/graphql/schema.graphql
```diff
@@ -33,8 +33,11 @@ type Query {
     sortOrder: SortOrder
   ): PeersConnection!
 
-  "Fetch a wallet address"
-  walletAddress(id: String!): WalletAddress
+  "Fetch a wallet address, and optionally additional properties."
+  walletAddress(
+    id: String!
+    includeAdditionalProperties: Boolean
+  ): WalletAddress
```
BlairCurrey, line 40: I think controlling the return of `additionalProperties` with `includeAdditionalProperties` here is a kind of gql anti-pattern.

Consider this is how we query for it:

```ts
query WalletAddress($walletAddressId: String!) {
  walletAddress(
    id: $walletAddressId
    includeAdditionalProperties: true
  ) {
    id
    liquidity
    asset {
      code
      scale
    }
    url
    publicName
    status
    additionalProperties {
      properties {
        key
        value
        visibleInOpenPayments
      }
    }
  }
}
```

We are already specifying that we want `additionalProperties` by the query itself. That should drive whether or not we fetch/return them - the argument is superfluous and complicating (what happens if we specify the field in the query but `includeAdditionalProperties` is `false`? and vice-versa). 

Instead of including this argument, we can make a separate resolver for `additionalProperties`, so the `rafiki/packages/backend/src/graphql/resolvers/index.ts` would look like:

```ts
  WalletAddress: {
    liquidity: getWalletAddressLiquidity,
    incomingPayments: getWalletAddressIncomingPayments,
    outgoingPayments: getWalletAddressOutgoingPayments,
    quotes: getWalletAddressQuotes,
    walletAddressKeys: getWalletAddressKeys,
    additionalProperties: getAdditionalProperties
  },
```

This way, every time the Wallet Address resolves, it will then resolve the additionalProperties if it was specified in the query. Like it does for `liquidity`, `incomingPayments`, etc.

BlairCurrey:
I think I may have a different interpretation of the "optional" aspect described in the task. It looks like these changes include 2 layers of options:

- `includeAdditionalProperties` controls if additional properties are fetched
- `visibleInOpenPayments` controls if properties to be fetched should include visible/non-visible

Maybe @mkurapov  can chime in on his original intentions but my interpretation was that we just always fetch the additional properties where `visibleInOpenPayments` was set to `true`. So removing the `includeAdditionalProperties` controlling if we fetch them.

bruno/collections/Rafiki/Rafiki Admin APIs/Get Wallet Address.bru
```diff
@@ -45,27 +45,34 @@ body:graphql {
...
             }
           }
           status
+      	  additionalProperties {
+            properties {
+              key
+              value
+              visibleInOpenPayments
+            }
+          }
```
BlairCurrey, line 75: can we get rid of the properties layer? just making `additionalProperties` the array of `{ key, value, visibleInOpenPayments }` ?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -123,13 +139,35 @@ async function createWalletAddress(
...
+          itm.fieldKey.trim().length == 0 || itm.fieldValue.trim().length == 0
+        )
+      })
+      // set defaults:
+      const now = new Date()
+      for (const prop of addProperties) {
+        prop.walletAddressId = wallet.id
+        prop.createdAt = now
+        prop.updatedAt = now
+      }
```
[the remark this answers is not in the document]
BlairCurrey: I guess we'd still need to set the `walletAddressId`... Could do that inside the iteration above (although it would need to be changed to a for loop that conditionally pushes to `addProperties`) but I guess these arrays are never going to be very big so maybe this isn't really a big deal to be honest.

packages/backend/src/open_payments/wallet_address/routes.ts
```diff
@@ -41,6 +41,13 @@ export async function getWalletAddress(
   const walletAddress = await deps.walletAddressService.getOrPollByUrl(
     ctx.walletAddressUrl
   )
+  if (walletAddress && ctx.fetchAdditionalProperties) {
+    walletAddress.additionalProperties =
```
BlairCurrey, line 45: we are setting `additionalProperties` here but then it gets lost later when we do `walletAddress.toOpenPaymentsType()`. the `toOpenPaymentsType` method on the `WalletAddress` model should be updated to include `additionalProperties`.

bruno/collections/Rafiki/Rafiki Admin APIs/Get Wallet Address.bru
koekiebox: @BlairCurrey sorted already.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -123,13 +139,35 @@ async function createWalletAddress(
...
+      if (addProperties.length) {
+        await WalletAddressAdditionalProperty.query(deps.knex).insert(
+          addProperties
+        )
+      }
+    }
+    return wallet
```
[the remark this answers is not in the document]
koekiebox: Ah, I think what I did works nicely. Please have a look @BlairCurrey

packages/backend/src/open_payments/wallet_address/routes.ts
koekiebox: Good catch! I also added a test case for this in `routes.test.ts`.

koekiebox: Test case is catching it. 

## Run:
https://github.com/interledger/rafiki/actions/runs/9400698822/job/25890897174?pr=2752

packages/backend/src/graphql/schema.graphql
```diff
@@ -1025,6 +1045,14 @@ input CreateWalletAddressInput {
   publicName: String
   "Unique key to ensure duplicate or retried requests are processed only once. See [idempotence](https://en.wikipedia.org/wiki/Idempotence)"
   idempotencyKey: String
+  "Additional properties associated with the [walletAddress]."
+  additionalProperties: [AdditionalPropertyInput!]!
```
BlairCurrey, line 1049: I think the `additionalProperties` field on the wallet address create input should be optional. just makes it easier on the caller. So this would be:

```gql
additionalProperties: [AdditionalPropertyInput!]
```

then we can remove the empty `additionalProperties` arrays where the mutation is called. like `test/integration/lib/integration-server.ts`, `packages/mock-account-service-lib/src/requesters.ts`, `packages/frontend/app/routes/wallet-addresses.create.tsx`

packages/backend/src/open_payments/wallet_address/model.ts
```diff
@@ -102,14 +112,17 @@ export class WalletAddress
...
+    const returnVal: OpenPaymentsWalletAddress = {
       id: this.url,
       publicName: this.publicName,
       assetCode: this.asset.code,
       assetScale: this.asset.scale,
       authServer,
       resourceServer
     }
+    if (this.additionalProperties)
+      returnVal.additionalProperties = this.additionalProperties
```
BlairCurrey, line 124: Could change this to `if (this.additionalProperties.length)` to fix the current CI error.

Although I think another option would be returning `undefined` from the `getAdditionalProperties` service method. Whatever you think makes more sense.

BlairCurrey (approved):

Issue #2678, opened 2024-04-25
Return amount spent on the outgoing payment grant during outgoing payment creation


mkurapov:
## Context
For the web monetization extension (and very likely other use cases), we should know how much has already been spent on the outgoing payment grant. 

Whenever we create an outgoing payment, we determine whether the amount for the payment fits within the limits of the issued outgoing payment grant (by summing up all of the other payment that have been created under that said grant). If not, we reject the payment, otherwise, we create it. This accounting is done in the Open Payment resource server (the `backend` package), because that is where all of the outgoing payments are stored. In the auth server, we only know what the limit for the grant is, but not how much has been spent on it already.

This concerns the Open Payments outgoing payment, meaning there will be no changes in the Admin GraphQL API.

## Todos
- [x] https://github.com/interledger/open-payments/issues/469
- [ ] Return `grantSpentDebitAmount: Amount` and `grantSpentReceiveAmount: Amount` when creating outgoing payment

BlairCurrey:
We talked about the reason for only returning on outgoing payment create, and that is because that is where the accounting is happening. We don't want to repeat that logic in a GET (especially for an entire page). We also concluded that returning in the create is sufficient for the webmon use-case.

PR #2753, opened 2024-05-29
feat: add spent amounts to POST /outgoing-payment

BlairCurrey:

- adds `grantSpentDebitAmount` and `grantSpentReceiveAmount` to outgoing payment create response
- adds "Web Monetization" example to Bruno collection

## Context

https://github.com/interledger/rafiki/issues/2678

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -491,6 +494,60 @@ describe('OutgoingPaymentService', (): void => {
...
+          for (let i = 0; i < 3; i++) {
+            const payment = await outgoingPaymentService.create(options)
+            assert.ok(!isOutgoingPaymentError(payment))
+            expect(payment.grantSpentDebitAmount.value).toBe(
+              BigInt(debitAmount.value * BigInt(i))
+            )
+            expect(payment.grantSpentReceiveAmount.value).toBe(
+              // Must account for interledger/pay off-by-one issue (even with 0 slippage/fees)
+              BigInt((debitAmount.value - BigInt(1)) * BigInt(i))
+            )
```
BlairCurrey, line 545: didn't see an issue for this quirk identified in the comment (not sure where we left it) but figured id note why receive amount should be 1 less than the debit amount here

bruno/collections/Rafiki/Examples/Web Monetization/Grant Request Outgoing Payment.bru
BlairCurrey, line 1: I guess with the consent interaction after this step it's not _exactly_ what the web monetization flow will look like. Otherwise I think it's similar? Not sure if there is a better name but I felt like there were enough changes compared to the Open Payments flow to include this example.

njlie (approved):

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -374,11 +392,21 @@ async function validateGrant(
...
-  let receivedAmount = BigInt(0)
+  const amounts = {
+    sent: {
+      assetCode: payment.asset.code,
+      assetScale: payment.asset.scale,
+      value: BigInt(0)
+    },
+    received: {
+      assetCode: payment.asset.code,
+      assetScale: payment.asset.scale,
```
mkurapov, line 406: I think the received amount should be the asset of the receiver itself, ie. it can be a different asset than the payment. In this case, we should use `grant.limits` to get this asset information instead.

I think debitAmount should always be the outgoing payment asset, however.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -170,6 +170,7 @@ export async function resolveReceiver(
 ): Promise<Receiver> {
   const receiver = await deps.receiverService.get(options.receiver)
   if (!receiver) {
+    deps.logger.info({ receiver: options.receiver }, 'Receiver not found')
```
mkurapov, line 173: Maybe we can add something like "could not create quote" in the log message

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -491,6 +494,60 @@ describe('OutgoingPaymentService', (): void => {
...
+          }
```
mkurapov, line 546: Maybe it makes sense to split the test for grantSpentDebitAmount and grantSpentReceiveAmount, since we want to allow a grant to have either or only, not both:
- https://github.com/interledger/open-payments/issues/471

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -63,6 +66,30 @@ export class OutgoingPayment
...
+  public set grantSpentDebitAmount(amount: Amount) {
+    this.grantSpentDebitAmountValue = amount.value
+  }
+
+  private grantSpentReceiveAmountValue?: bigint
+  public get grantSpentReceiveAmount(): Amount {
+    return {
+      value: this.grantSpentReceiveAmountValue || BigInt(0),
+      assetCode: this.asset.code,
+      assetScale: this.asset.scale
```
mkurapov, line 86: same comment as in the service file. The asset code may be different than the payment itself.

bruno/collections/Rafiki/Examples/Web Monetization/Continuation Request.bru
```diff
@@ -0,0 +1,33 @@
+meta {
```
mkurapov, line 1: I'm trying to think whether we need a separate folder for this. Technically, we can get a large grant, and then keep re-running the same "Create Outgoing Payment" request. Maybe it's good to have, however, just thinking out loud

BlairCurrey: > I'm trying to think whether we need a separate folder for this

In other words, maybe not adding this new "Web Monetization" example? 

I mentioned my reasoning a bit in this comment (while questioning the name) https://github.com/interledger/rafiki/pull/2753#discussion_r1625241575. Basically I didn't find it trivial to alter the existing Open Payments one to do this. It took me some time to put together and I dont even remember all the changes off the top of my head (I would not want to have to re-figure it out). Feel like it could save myself and other some time.

test/integration/package.json
```diff
@@ -16,7 +16,7 @@
   "devDependencies": {
     "@apollo/client": "^3.10.4",
     "@interledger/http-signature-utils": "2.0.2",
-    "@interledger/open-payments": "^6.11.1",
+    "@interledger/open-payments": "6.11.1",
```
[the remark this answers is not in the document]
BlairCurrey: As I recall I did this in the middle of fixing the open payments issue we saw related to the ESM import. I noticed the others were pinned as well so I changed this one but at this point it may not matter.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
BlairCurrey: yeah that does make sense. split

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: updated: https://github.com/interledger/rafiki/pull/2753/commits/fc4adba14de448fde3d5e545aa70aac9e41e28de

I used the `paymentLimits` (comes from grant.limits) and added some typeguards. Otherwise its a bit tedious to check which of the amounts are defined.

packages/backend/src/open_payments/payment/outgoing/model.ts
BlairCurrey: I guess this means we dont have any sort of default for the `grantSpentReceiveAmount` since this depends on being created with a grant limit. So we simply cant return `grantSpentReceiveAmount` for outgoing payments created in this case.

I guess that's fine... it just seems incongruent with the `grantSpentDebitAmount`, which we can return a default value for (0 with outgoing payment's `assetCode` and `assetScale`). I guess maybe that's OK, but it seems non-obvious from the caller's perspective. 

I guess we could:
- just leave as-is
- not return any defualt/null version  of `grantSpentDebitAmount` either
- return something like `{value: 0, assetCode: "", assetScale: -1 }` for both `grantSpentDebitAmount` and `grantSpentReceiveAmount` in this case

BlairCurrey: or maybe we can get it from somewhere other than `grant.limits`...

mkurapov: What if instead of storing `grantSpentReceiveAmountValue: bigint` and `grantSpentDebitAmountValue: bigint`, we just allow setting the whole `Amount`? Otherwise if not set, it just gets returned as undefined?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -338,23 +365,32 @@ async function validateGrant(
...
+
+  let paymentLimitAmount: Amount
+
+  if (hasDebitAmount(paymentLimits)) {
+    paymentLimitAmount = paymentLimits.debitAmount
+  } else if (hasReceiveAmount(paymentLimits)) {
+    paymentLimitAmount = paymentLimits.receiveAmount
+  } else {
+    return { isValid: true }
   }
```
mkurapov, line 385: I think we can only do this once
- https://github.com/interledger/open-payments/issues/471

is merged? Since now, we _can_ have both

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -187,6 +191,13 @@ export async function resolveReceiver(
       }
     }
   } else if (!options.debitAmount && !receiver.incomingAmount) {
+    deps.logger.info(
+      {
+        debitAmount: options.debitAmount,
+        incomingAmount: receiver.incomingAmount
+      },
+      'debitAmount or incomingAmount required'
```
mkurapov, line 199: Same thing with "could not create quote"

test/integration/package.json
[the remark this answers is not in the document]
mkurapov: I see, they were always pinned. I guess it's ok.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
mkurapov: It's a feature, not a bug 😆

packages/backend/src/open_payments/payment/outgoing/model.ts
BlairCurrey: I've changed it set the whole amount locally already so in full agreement there.

After sitting on this for a bit, I think `undefined` and then not returning them in the `/POST /outgoing-payment` seems fine. I guess my default expectation is to not include/exclude properties when they're in a sort of null state but in this case I think it may be the most straightforward.

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: I renamed `paymentLimit` to `amountForReceivedAmount` but the logic should be the same as it was before. I think the `paymentLimit` was misleading - it was only being used to get the asset code/scale (per your request to get it from grant limits). I also removed the typeguards (had bigger plans for them originally, but doing nothing compared to `paymentLimit.debitAmount` etc. checks now).

packages/backend/src/open_payments/payment/outgoing/model.ts
BlairCurrey: updated to above ^ [710897d](https://github.com/interledger/rafiki/pull/2753/commits/710897d194c4021316c4d13ae52e6415afe3222e)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: [ccc2dc2](https://github.com/interledger/rafiki/pull/2753/commits/ccc2dc22ce300ec311465306324684b40a509659)

mkurapov: Actually, looking at it, we can actually do something easier: I think we can use `payment.receiveAmount` -> that will have the asset code & scale of the receiver (since it uses the quote.receiveAmount) under the hood for it

mkurapov (approved):

Issue #2758, opened 2024-06-04
Admin UI payments page does not save type filter when navigating across pages


mkurapov:
## Context
Unlike the webhook events page, the payments page in the Admin UI does not persist the type when navigating across different pages:

https://github.com/interledger/rafiki/assets/15069181/61758a38-110f-4b7f-a92e-d6f10c2f2632

## Expected Behaviour
When selecting the type in the filter of the payments page, it should persist when navigating across pages of results.

sarthakChauhan12:
Can I work on this issue?

JoblersTune:
@sarthakChauhan12 I actually just added this fix while working on a related issue.

PR #2766, opened 2024-06-05
feat: filter payments per wallet address ID

### LINKED ISSUE #2689: [Admin UI] Display payments under each of the wallet addresses
From the breakpoint workweek, we determined that it would be a nice feature to have in the Admin UI to display all of the payments under a wallet address (in the wallet address details screen). We already have the resolver for this, we just need to expose it on the frontend, together with pagination.

JoblersTune:

- Implemented filtering of payments based on walletAddressId.
- Maintained other filter parameters (e.g., type, pagination) when updating walletAddressId.
- Included a button in the wallet Address page to navigate directly to view the payments page filtering for that wallet's address ID.

## Context

- fixes #2689
- fixes ##2758

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/frontend/app/routes/payments._index.tsx
+++ b/packages/frontend/app/routes/payments._index.tsx
@@ -1,8 +1,20 @@
-import { json, type LoaderFunctionArgs } from '@remix-run/node'
-import { useLoaderData, useNavigate, useSearchParams } from '@remix-run/react'
+import { useState } from 'react'
+import {
+  json,
+  type LoaderFunctionArgs,
+  type ActionFunctionArgs,
+  redirect
+} from '@remix-run/node'
+import {
+  useLoaderData,
+  useNavigate,
+  useSearchParams,
+  Form,
+  useActionData
+} from '@remix-run/react'
 import { Badge, PageHeader } from '~/components'
 import { PopoverFilter } from '~/components/Filters'
-import { Button, Table } from '~/components/ui'
+import { Button, Table, ErrorPanel, FieldError } from '~/components/ui'
 import { listPayments } from '~/lib/api/payments.server'
 import { paymentsSearchParams } from '~/lib/validate.server'
 import { PaymentType } from '~/generated/graphql'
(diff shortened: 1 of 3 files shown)


koekiebox (dismissed):
I am not the best to review, but it looks good to me.

mkurapov:
Everything works well for me :+1: 
going to come back to this and review the code

mkurapov:
I also think this fixes
- #2758

packages/frontend/app/routes/payments._index.tsx
```diff
@@ -95,27 +128,94 @@ export default function PaymentsPage() {
...
+                  action: () => setTypeFilterParams(value)
                 }))
               ]}
             />
+            <Form
+              method='post'
+              className='relative mt-2'
+              onSubmit={(e) => {
+                if (!walletId) {
+                  console.log('ESCAPE')
```
mkurapov, line 162:
```suggestion
                  
```

packages/frontend/app/routes/payments._index.tsx
```diff
@@ -31,57 +50,71 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
...
 }
 
 export default function PaymentsPage() {
-  const { payments, previousPageUrl, nextPageUrl, type } =
-    useLoaderData<typeof loader>()
-  const setSearchParams = useSearchParams()[1]
-
+  const { payments, type, walletAddressId } = useLoaderData<typeof loader>()
+  const [searchParams, setSearchParams] = useSearchParams()
+  const [walletId, setWalletId] = useState(walletAddressId || '')
```
mkurapov, line 79: My understanding `walletId` is necessary as a "local" variable in order to make sure it's valid via the form submit before we actually set it in the searchParams?

JoblersTune: Yes. It is also means we can keep the input field nicely reactive. And if the wallet address is being removed it means we can skip the validation and simply redirect immediately since all we're doing is resetting the paging and removing the wallet address filter. And it makes it easy to capture it from the URL. Sometimes we get here not by form submission but actually by clicking the view paymnents page button when looking at a wallet address and then the walletAddressId is already in the URL. And also makes it easy to remove this value from the input field when we reset the filters.

JoblersTune:
Yes I didn't see we had an issue for #2758 but it did occur to me that was a problem while I was adding this in.

mkurapov (approved):

PR #2767, opened 2024-06-05
fix: put additional properties resolver under wallet address

BlairCurrey:
Moves the `additionalProperties` resolver under the `WalletAddress` resolver. This allows us to query this field on wallet address as opposed to calling it separately with the wallet address id, but keeps the fetching outside of the wallet address resolver/service method itself. This means we can specify it anywhere the wallet address is returned (the get wallet address query, create wallet address mutation, etc.)

See test and bruno collection changes to see how its called.

To be merged into https://github.com/interledger/rafiki/pull/2752

### DIFF
--- a/packages/backend/src/graphql/resolvers/walletAddressAdditionalProperties.ts
+++ b/packages/backend/src/graphql/resolvers/walletAddressAdditionalProperties.ts
@@ -1,33 +1,31 @@
 import {
   ResolversTypes,
   AdditionalProperty,
-  QueryResolvers
+  WalletAddressResolvers
 } from '../generated/graphql'
 import { ApolloContext } from '../../app'
 import { WalletAddressAdditionalProperty } from '../../open_payments/wallet_address/additional_property/model'
 
-export const getWalletAddressAdditionalProperties: QueryResolvers<ApolloContext>['additionalProperties'] =
+export const getWalletAddressAdditionalProperties: WalletAddressResolvers<ApolloContext>['additionalProperties'] =
   async (
     parent,
     args,
     ctx
-  ): Promise<ResolversTypes['AdditionalPropertyConnection']> => {
-    if (!args.walletAddressId) {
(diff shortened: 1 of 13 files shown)


packages/backend/src/graphql/resolvers/walletAddressAdditionalProperties.ts
```diff
@@ -1,33 +1,31 @@
...
-export const getWalletAddressAdditionalProperties: QueryResolvers<ApolloContext>['additionalProperties'] =
+export const getWalletAddressAdditionalProperties: WalletAddressResolvers<ApolloContext>['additionalProperties'] =
   async (
     parent,
     args,
     ctx
-  ): Promise<ResolversTypes['AdditionalPropertyConnection']> => {
-    if (!args.walletAddressId) {
+  ): Promise<Array<ResolversTypes['AdditionalProperty']>> => {
+    const { id: walletAddressId } = parent
```
BlairCurrey, line 15: Since its under the `WalletAddress` resolver, we can get the id off the parent (the wallet address) instead of passing it in as an arg.

packages/backend/src/graphql/schema.graphql
```diff
@@ -698,10 +695,9 @@ type WalletAddress implements Model {
     "Ascending or descending order of creation."
     sortOrder: SortOrder
   ): WalletAddressKeyConnection
-}
 
-type AdditionalPropertyConnection {
-  properties: [AdditionalProperty!]!
+  "List additional properties associated with this wallet address."
+  additionalProperties: [AdditionalProperty]!
```
BlairCurrey, line 700: No additional properties layer (since we aren't gonig to be doing any paging anyways), so its just:


```gql
additionalProperties {
  key
  value
visibleInOpenPayments
}
```

koekiebox (approved):
LGTM! Thank you.

PR #2768, opened 2024-06-05
feat(backend)!: logging plugin for graphql resolvers, remove error resolver responses

### LINKED ISSUE #2575: Add logs in graphql resolvers - backend
### Context

We want to add logs to all the graphql resolvers on the backend.

UPDATE:
When working on this, @sabineschaller realized that we don't handle errors the same way in queries and mutations. In queries, we throw errors, whereas in mutations, we treat them as data. To be consistent and to be able to use the apollo plugin for logging, we should move away from the error as data approach and just throw `GraqhQLError`s when something goes wrong in the resolvers. 

### Todos
- [x] add apollo plugin to log errors
- [x] update GraphQL schema
- [ ] update resolvers to throw instead of treating errors as data
- [ ] update MASE to handle new resolver behavior
- [ ] update docs
- [ ] potentially update bruno collection (scripts/tests)

njlie:

- Refactored `backend` GraphQL responses to no longer use `code`, `message`, or `success`. If a response is not successful, it will now throw a GraphQL error instead of returning a `success: false` and `code: 4XX/5XX` response.
- Adds a logging plugin to the Apollo Server that will log if a resolver throws a GraphQL Error.
- Updated `frontend` package and local MASE to use the new backend GraphQL response models.

## Context

Fixes #2575.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/graphql/middleware/index.test.ts
+++ b/packages/backend/src/graphql/middleware/index.test.ts
@@ -10,7 +10,11 @@ import { Config } from '../../config/app'
 import { truncateTables } from '../../tests/tableManager'
 import { AssetService } from '../../asset/service'
 import { randomAsset } from '../../tests/asset'
-import { AssetMutationResponse, CreateAssetInput } from '../generated/graphql'
+import {
+  AssetMutationResponse,
+  CreateAssetInput,
+  UpdateAssetInput
+} from '../generated/graphql'
 import { GraphQLError } from 'graphql'
 
 describe('GraphQL Middleware', (): void => {
@@ -39,9 +43,6 @@ describe('GraphQL Middleware', (): void => {
         mutation: gql`
           mutation CreateAsset($input: CreateAssetInput!) {
             createAsset(input: $input) {
-              code
-              success
-              message
               asset {
                 id
                 code
@@ -64,59 +65,96 @@ describe('GraphQL Middleware', (): void => {
       })
   }
 
+  const callUpdateAssetMutation = async (input: UpdateAssetInput) => {
+    return appContainer.apolloClient
+      .mutate({
+        mutation: gql`
+          mutation UpdateAsset($input: UpdateAssetInput!) {
+            updateAsset(input: $input) {
+              asset {
+                id
+                code
+                scale
+                withdrawalThreshold
+              }
+            }
+          }
+        `,
+        variables: {
+          input
+        }
+      })
+      .then((query): AssetMutationResponse => {
+        if (query.data) {
+          return query.data.updateAsset
+        } else {
+          throw new Error('Data was empty')
+        }
+      })
+  }
+
   describe('idempotencyGraphQLMiddleware', (): void => {
+    let createInput: CreateAssetInput
+    let createResponse: AssetMutationResponse
+
+    beforeEach(async (): Promise<void> => {
+      createInput = {
+        ...randomAsset(),
+        idempotencyKey: uuid()
+      }
+
+      createResponse = await callCreateAssetMutation(createInput)
+      assert.ok(createResponse.asset)
+    })
+
     test('returns original response on repeat call with same idempotency key', async (): Promise<void> => {
       const idempotencyKey = uuid()
-      const input: CreateAssetInput = {
-        ...randomAsset(),
+      assert.ok(createResponse.asset)
+      const input: UpdateAssetInput = {
+        id: createResponse.asset.id,
+        withdrawalThreshold: BigInt(10),
         idempotencyKey
       }
 
-      const createAssetSpy = jest.spyOn(assetService, 'create')
+      const updateAssetSpy = jest.spyOn(assetService, 'update')
 
-      const initialResponse = await callCreateAssetMutation(input)
+      const initialResponse = await callUpdateAssetMutation(input)
 
-      expect(createAssetSpy).toHaveBeenCalledTimes(1)
+      expect(updateAssetSpy).toHaveBeenCalledTimes(1)
       assert.ok(initialResponse.asset)
       expect(initialResponse).toEqual({
         __typename: 'AssetMutationResponse',
-        message: 'Created Asset',
-        success: true,
-        code: '200',
         asset: {
           __typename: 'Asset',
           id: initialResponse.asset.id,
-          code: input.code,
-          scale: input.scale,
-          withdrawalThreshold: null
+          code: createInput.code,
+          scale: createInput.scale,
+          withdrawalThreshold: '10'
         }
       })
       await expect(
         assetService.get(initialResponse.asset.id)
       ).resolves.toMatchObject({
         id: initialResponse.asset.id,
-        code: input.code,
-        scale: input.scale,
-        withdrawalThreshold: null
+        code: createInput.code,
+        scale: createInput.scale,
+        withdrawalThreshold: BigInt(10)
       })
 
-      createAssetSpy.mockClear()
+      updateAssetSpy.mockClear()
 
-      const repeatResponse = await callCreateAssetMutation(input)
+      const repeatResponse = await callUpdateAssetMutation(input)
 
-      expect(createAssetSpy).not.toHaveBeenCalled()
+      expect(updateAssetSpy).not.toHaveBeenCalled()
       assert.ok(repeatResponse.asset)
       expect(repeatResponse).toEqual({
         __typename: 'AssetMutationResponse',
-        message: 'Created Asset',
-        success: true,
-        code: '200',
         asset: {
           __typename: 'Asset',
           id: initialResponse.asset.id,
           code: initialResponse.asset.code,
           scale: initialResponse.asset.scale,
-          withdrawalThreshold: null
+          withdrawalThreshold: '10'
         }
       })
       await expect(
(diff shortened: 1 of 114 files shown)


packages/backend/src/graphql/middleware/index.test.ts
```diff
@@ -125,50 +163,7 @@ describe('GraphQL Middleware', (): void => {
         id: initialResponse.asset.id,
         code: initialResponse.asset.code,
         scale: initialResponse.asset.scale,
-        withdrawalThreshold: null
-      })
-    })
-
-    test('does not return original response on repeat call with different idempotency key', async (): Promise<void> => {
```
njlie, line 132: It didn't make sense to me to keep this test. This PR adds GraphQL Errors that are thrown if a request is made to create something that already exists, so idempotency wouldn't affect cases like that.

packages/backend/src/graphql/schema.graphql
```diff
@@ -1204,116 +1192,64 @@ interface Model {
...
 }
 
-type LiquidityMutationResponse implements MutationResponse {
-  code: String!
-  success: Boolean!
-  message: String!
-  error: LiquidityError
+type LiquidityMutationResponse {
+  id: ID!
+  liquidity: UInt64
```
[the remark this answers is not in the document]
BlairCurrey: Actually I see we are already doing `success: Boolean!` in `DeletePeerMutationResponse`. Perhaps that would be appropriate here as well? 

The caller already has the returned `id` and `liquidity` values anyways, since what we are returning is just coming from the input.

packages/backend/src/open_payments/payment/incoming/errors.ts
```diff
@@ -24,6 +26,18 @@ export const errorToCode: {
...
 
+export const errorToCode: {
+  [key in IncomingPaymentError]: string
+} = {
+  [IncomingPaymentError.UnknownWalletAddress]: GraphQLErrorCode.NotFound,
+  [IncomingPaymentError.InvalidAmount]: GraphQLErrorCode.BadUserInput,
+  [IncomingPaymentError.UnknownPayment]: GraphQLErrorCode.NotFound,
+  [IncomingPaymentError.InvalidState]: GraphQLErrorCode.BadUserInput,
+  [IncomingPaymentError.InvalidExpiry]: GraphQLErrorCode.BadUserInput,
+  [IncomingPaymentError.WrongState]: GraphQLErrorCode.BadUserInput,
```
BlairCurrey, line 37: Perhaps Conflict/409 would be a better match? The docs mention using it when the resource is in the wrongs state https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409

```suggestion
  [IncomingPaymentError.WrongState]: GraphQLErrorCode.Conflict,
```

also matches how we map it here: https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/payment/incoming/errors.ts#L23

packages/backend/src/open_payments/payment/outgoing/errors.ts
```diff
@@ -46,6 +51,21 @@ export const errorToCode: {
   [OutgoingPaymentError.InvalidReceiver]: 400
 }
 
+export const errorToCode: {
+  [key in OutgoingPaymentError]: string
+} = {
+  [OutgoingPaymentError.UnknownWalletAddress]: GraphQLErrorCode.NotFound,
+  [OutgoingPaymentError.UnknownPayment]: GraphQLErrorCode.NotFound,
+  [OutgoingPaymentError.UnknownQuote]: GraphQLErrorCode.NotFound,
+  [OutgoingPaymentError.WrongState]: GraphQLErrorCode.BadUserInput,
```
BlairCurrey, line 60: same suggestion as https://github.com/interledger/rafiki/pull/2768/files#r1633467866

```suggestion
  [OutgoingPaymentError.WrongState]: GraphQLErrorCode.Conflict,
```

packages/backend/src/graphql/schema.graphql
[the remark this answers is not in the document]
njlie: I used this approach as a holdover from when I took this over from [Sabine's branch on this](https://github.com/interledger/rafiki/pull/2667/files#diff-fdaa49a1159ec466e5ccb9daa6e086e9188bb521d630840a15f67f1188c9443c). I think returning a `success` boolean makes sense but I'm curious to hear if @sabineschaller has any particular reasoning for this before switching it over.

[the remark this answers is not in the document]
njlie: Based on today's call it doesn't seem like she's all that attached to whatever approaches she used before I took over. I also think the `success` boolean makes more sense so I'll change it to use that as well.

koekiebox (changes requested):
Minor items. Otherwise looks good.

BlairCurrey (dismissed):
Looks good. I tested manually with some of the endpoints on bruno and admin UI and didnt see any issues.

BlairCurrey (approved):

test/integration/lib/integration-server.ts
```diff
@@ -134,7 +134,7 @@ export class WebhookEventHandler {
     })
     const { walletAddress } = response
 
-    if (!response.success) {
+    if (!response.walletAddress) {
```
mkurapov, line 137: can remove this line since we are checking it in the next if

packages/frontend/app/routes/wallet-addresses.$walletAddressId.tsx
```diff
@@ -239,10 +239,11 @@ export async function action({ request }: ActionFunctionArgs) {
     ...result.data
   })
 
-  if (!response?.success) {
+  console.log('response=', response)
```
mkurapov, line 242:
```suggestion
```

mkurapov (commented):
Looks good, thank you for going through this! Just a few comments

packages/backend/src/graphql/resolvers/asset.test.ts
```diff
@@ -751,9 +734,7 @@ describe('Asset Resolvers', (): void => {
           }
         })
 
-      expect(response.success).toBe(false)
-      expect(response.code).toEqual('500')
-      expect(response.message).toEqual('Error trying to delete asset')
+      await expect(gqlQuery).rejects.toThrow(ApolloError)
```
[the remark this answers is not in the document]
njlie: https://www.apollographql.com/docs/apollo-server/data/errors/#throwing-errors
> If a resolver throws a generic error that is not a GraphQLError instance, that error is still thrown with an extensions field that includes a stacktrace and code (specifically [INTERNAL_SERVER_ERROR](https://www.apollographql.com/docs/apollo-server/data/errors/#internal_server_error)), along with any other relevant error details.

Based on that I'd think it should expect a `GraphQLError`, this snippet passes the test:
```
await expect(gqlQuery).rejects.toThrow(new GraphQLError('unexpected', {
  extensions: {
    code: GraphQLErrorCode.InternalServerError
   }
}))

**Review by koekiebox — APPROVED** [review]
LGTM 🚀

**Review by mkurapov — DISMISSED** [review]
Would be good to have the 
`[key in ...Error]: GraphQLErrorCode` and the new test assertions (what you did in the last two commits) updated throughout the rest of the files, but non blocking.

**Review by koekiebox — APPROVED** [review]

File: packages/backend/src/graphql/resolvers/peer.test.ts [hunk 8]
```diff
@@ -212,7 +222,13 @@ describe('Peer Resolvers', (): void => {
             throw new Error('Data was empty')
           }
         })
-      await expect(gqlQuery).rejects.toThrow(ApolloError)
+      await expect(gqlQuery).rejects.toThrow(
+        new GraphQLError('unexpected', {
+          extensions: {
+            code: GraphQLErrorCode.InternalServerError
```
mkurapov, line 228: I think if you change this to be a different enum value (e.g. `GraphQLErrorCode.Conflict`), the test will still pass. Maybe `.toThrow` isn't that smart :/. I think you'd have to check the extensions field in an explicit try-catch

njlie: Yeah...it seems like the resolver won't throw a `GraphQLError` at the top level, but rather it gets caught and then added to a list of GraphQLErrors in a `ApolloError.graphQLErrors` field that ultimately gets thrown at the top level.

packages/backend/src/graphql/resolvers/auto-peering.test.ts
```diff
@@ -244,85 +235,90 @@ describe('Auto Peering Resolvers', (): void => {
...
-            mutation CreateOrUpdatePeerByUrl(
-              $input: CreateOrUpdatePeerByUrlInput!
-            ) {
-              createOrUpdatePeerByUrl(input: $input) {
-                code
-                success
-                message
-                peer {
-                  id
+      try {
```
mkurapov, line 243:
```suggestion
      expect.assertions(2)
      try {
```
to make sure this actually throws and the test does the correct number of assertions (same comment throughout)

mkurapov (commented):
final thing

mkurapov (approved):
There are other spots where we need to verify the number of assertions in tests with try-catches, so I'll approve and put up a PR against this branch to handle those cases.

Thanks for taking care of this!

Issue #2740, opened 2024-05-22
Allow `updateWalletAddress` mutation to take in additional properties


mkurapov:
### Context
Similar to wallet address creation in #2739, we want to let the ASE allow updating additional properties to be associated with the wallet address.

We need to update the `updateWalletAddress` mutation input to take in `additionalProperties`.
These properties are the same as the `createWalletAddress` input: an optional array of objects with `key` & `value` and `visibleInOpenPayments`.

### Tasks
- [x] Update `updateWalletAddress` mutation input to take in additional params
- [x] Update these new fields in the new wallet address table, hard deleting keys that are no longer used
- [x] Update Documentation, including update behavior (overrides entire `additionalFields` on each update)

PR #2769, opened 2024-06-07
feat(backend): update wallet address additional properties

BlairCurrey:

- Updates walletAddressUpdate mutation so that `additionalProperties` can be updated

## Context

fixes https://github.com/interledger/rafiki/issues/2740

## Checklist

- [X] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -180,12 +192,29 @@ async function updateWalletAddress(
...
-    return await walletAddress
-      .$query(deps.knex)
+    const updatedWalletAddress = await walletAddress
+      .$query(trx)
       .patchAndFetch(update)
       .withGraphFetched('asset')
       .throwIfNotFound()
+
+    // Override all existing additional properties if new ones are provided
+    if (additionalProperties && additionalProperties.length) {
```
BlairCurrey, line 202: Does nothing if `additionalProperties` is `[]`. Perhaps this is wrong though and an empty array should delete them? Since it's more consistent with how we're otherwise deleting non-existant keys. And because there is currently no way to delete everything. 

I did it this way originally because I thought it would be too easy to accidentally delete everything, but now I see that this reasoning would also apply to deleting non-existent keys.

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -360,6 +358,190 @@ describe('Wallet Address Resolvers', (): void => {
...
+              }
+            `,
+            variables: {
+              input: updateOptions
+            }
+          })
+          .then((query): UpdateWalletAddressMutationResponse => {
+            if (query.data) {
+              return query.data.updateWalletAddress
+            } else {
```
koekiebox, line 524: Ok, it is nicer that the additional properties has their own section.

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: This is a bit of a tricky one. Here is a suggestion:
- `undefined` would ignore doing anything to the additional properties
- `[]` would delete all the additional properties
- keys that match existing values, but have their new value is empty, gets deleted
- new keys with valid values are inserted
- keys with updated values are updated (non-case-sensitive, not sure if this is already expected to the handled by PSQL behaviour?)

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -134,75 +133,68 @@ describe('Wallet Address Resolvers', (): void => {
...
-                    }
-                    url
-                    publicName
+    test('Can create a wallet address with additional properties', async (): Promise<void> => {
+      input.publicName = 'Bob'
+      input.additionalProperties = [
+        { key: '', value: '', visibleInOpenPayments: false },
+        { key: 'key', value: '', visibleInOpenPayments: false },
+        { key: '', value: 'val', visibleInOpenPayments: false },
+        { key: 'key', value: 'val', visibleInOpenPayments: false },
```
koekiebox, line 142: I just realized that I haven't thought of duplicating key's for additional properties. 
Do we want to:
1. generate an error and reject the create.
2. Only use the value for the last matching keys.

koekiebox (changes requested):
Couple of items, otherwise it looks good.

packages/backend/src/open_payments/wallet_address/service.ts
BlairCurrey: Agree on all counts except I'd rather _effectively_ update existing keys for the reasons stated here: https://github.com/interledger/rafiki/pull/2769#discussion_r1631403514

What do you think?

packages/backend/src/graphql/resolvers/wallet_address.test.ts
BlairCurrey: That is an interesting case.

I actually think we should let them do it because it doesn't break nor complicate anything on our end and I dont want to assume there is no valid use-case for duplicate keys. Perhaps they want to group by key or something where only some of the values are private to open payments: 

```ts
// graphql update wallet address input
const input = {
  additionalProperties: [
    { key: 'myUserArray', value: 'publicUser1', visibleInOpenPayments: 'true' },
    { key: 'myUserArray', value: 'privateUser2', visibleInOpenPayments: 'false' },
    { key: 'myUserArray', value: 'publicUser2', visibleInOpenPayments: 'true' },
    { key: 'nickname', value: 'someNickname', visibleInOpenPayments: 'true' }
  ]
 }
 
 ...

// /GET wallet-address response
const response = {
  additionalProperties: [
    { key: 'myUserArray', value: 'publicUser1', visibleInOpenPayments: 'true' },
    { key: 'myUserArray', value: 'publicUser2', visibleInOpenPayments: 'true' },
    { key: 'nickname', value: 'someNickname', visibleInOpenPayments: 'true' }
  ]
}

// group values by 'myUserArray'
const userArrayValues = response.additionalProperties
  .filter(obj => obj.key === 'myUserArray')
  .map(obj => obj.value);

// users is ["publicUser1", "publicUser2"]
```

This is a capability we would remove by handling duplicate keys. Do they really want to do this? I have no idea - but  I think that's their concern and we shouldn't reduce their power to choose to use it this way, or other un-anticipated ways when it doesn't break nor complicate anything on our end.

packages/backend/src/open_payments/wallet_address/service.ts
mkurapov: Agree with both of the last two messages 👍 @koekiebox @BlairCurrey

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -165,8 +176,9 @@ async function createWalletAddress(
 
 async function updateWalletAddress(
   deps: ServiceDependencies,
-  { id, status, publicName }: UpdateOptions
+  { id, status, publicName, additionalProperties }: UpdateOptions
 ): Promise<WalletAddress | WalletAddressError> {
+  const trx = await WalletAddress.startTransaction()
   try {
     const update: UpdateInput = { publicName }
     const walletAddress = await WalletAddress.query(deps.knex)
```
mkurapov, line 184: not a big deal but should we also use the `trx` here?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -180,12 +192,31 @@ async function updateWalletAddress(
...
+      await WalletAddressAdditionalProperty.query(trx)
+        .where('walletAddressId', id)
+        .delete()
+    }
+    if (additionalProperties?.length) {
+      await WalletAddressAdditionalProperty.query(trx).insert(
+        additionalProperties.map((prop) => ({
+          walletAddressId: id,
+          ...prop
+        }))
```
mkurapov, line 212: just in case I think we should "clean" the values passed into this. Would probably best to have this "cleaning" made into a fn and also used during create in L150-159.

BlairCurrey: Let me know if you have anything else in mind, but I made a fn that takes the existing check in create (that key/value arent empty) and added a step to trim them to get rid of whitespace. Does that cover what you had in mind?

https://github.com/interledger/rafiki/pull/2769/commits/7c21e2db99974ded6c501dc324cd0198e61a1cae

packages/backend/src/open_payments/wallet_address/service.ts
BlairCurrey: I suppose probably so... and more importantly the underlying `IncomingPayment.patch`in `deactivateOpenIncomingPaymentsByWalletAddress` (called a few lines below this).

packages/backend/src/open_payments/wallet_address/service.ts
BlairCurrey: I guess this suggestion kinda veers into the suggestion to de-duplicate https://github.com/interledger/rafiki/pull/2769#discussion_r1635789637

I guess I still prefer to let the caller handle that and not enforce it but I guess if we do this would be the place. Not sure if that's something you had in mind or not.

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -134,75 +133,68 @@ describe('Wallet Address Resolvers', (): void => {
...
-                  message
-                  walletAddress {
-                    id
-                    asset {
-                      code
-                      scale
-                    }
-                    url
-                    publicName
+    test('Can create a wallet address with additional properties', async (): Promise<void> => {
```
mkurapov, line 136: we should check the correct properties were returned in the response

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -165,27 +184,50 @@ async function createWalletAddress(
...
-    return await walletAddress
-      .$query(deps.knex)
+    const updatedWalletAddress = await walletAddress
+      .$query(trx)
       .patchAndFetch(update)
       .withGraphFetched('asset')
       .throwIfNotFound()
+
+    // Override all existing additional properties if new ones are provided
+    if (additionalProperties) {
```
mkurapov, line 210: we should add some tests for these

packages/backend/src/graphql/resolvers/wallet_address.test.ts
BlairCurrey: updated [f73e66d](https://github.com/interledger/rafiki/pull/2769/commits/f73e66d5a8bbf554e965b29a664c90993181caf5)

packages/backend/src/open_payments/wallet_address/service.ts
BlairCurrey: added: [770c0bd](https://github.com/interledger/rafiki/pull/2769/commits/770c0bd99d3f24a94dce38a21977a44ae4221478)

koekiebox (approved):
LGTM! 🎸

mkurapov (approved):
LGTM ✅

PR #2749, opened 2024-05-28
feat(backend)!: update and rename env variables

golobitch:
## Changes proposed in this pull request

- update default values of some env variables
- rename env variable for enabling SPS
- update documentation

## Context

fixes #2684 

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [x] Documentation added
- [x] Make sure that all checks pass
- [ ] Bruno collection updated

PR #2773, opened 2024-06-12
fix: ci

BlairCurrey:

- Adds `--privileged` flag to docker start on tigerbeetle container start
- patches pnpm packages where vulnerabilities were found (cherry-picked from https://github.com/interledger/rafiki/pull/2749) to resolve CI failure: `Node Build / docker-build (linux/arm64, arm64, frontend) (pull_request)` etc

## Context

fixes recent ci failures

Seems to come from an upgrades version of docker in github. more context on this issue and getting to the resolution in this slack thread: https://interledgerfoundation.slack.com/archives/C06KZCKEEKW/p1718211242499809

see:
- https://github.com/tigerbeetle/tigerbeetle/issues/1860
- https://github.com/tigerbeetle/tigerbeetle/pull/1995
## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


golobitch:
LGTM, but will wait also on @mkurapov to do a review.

packages/backend/src/tests/tigerbeetle.ts
```diff
@@ -33,13 +33,16 @@ export async function startTigerbeetleContainer(clusterId?: number): Promise<{
       '--replica-count=1',
       `${TIGERBEETLE_DIR}/${tigerbeetleFile}`
     ])
+    .withPrivilegedMode()
```
mkurapov, line 36: why don't we need to do this in the local playground?

mkurapov (approved):

packages/backend/src/tests/tigerbeetle.ts
BlairCurrey: Good question... My local docker version is the same version that we suspect is causing a problem on github.  Not sure if there would be differences between docker verions on Mac/Ubuntu (or w/e the github environment is).

Issue #2756, opened 2024-06-03
Review TigerBeetle integration based on current TigerBeetle release


koekiebox:

koekiebox:
I will be keeping this PR open,  the next TB `0.15.4` will release this or next week, which will enable us to make use of the `--development` flag. 
In the meantime, I will get the remainder of the PR ready for review.

Issue #2446, opened 2024-02-22
Fix how Tigerbeetle settlement accounts are defined in the local playground


mkurapov:
## Context
When we are making settlement accounts for assets with Tigerbeetle accounting enabled, we end up using the same account `id`s for the underlying tigerbeetle accounts:

https://github.com/interledger/rafiki/blob/8100fa73d2ecc3b15300fbbdfa68cb5b1cc3de17/packages/backend/src/accounting/tigerbeetle/service.ts#L127-L133

`ledger` is just an incrementing smallint in our databases (a property of the asset). Since cloud-nine-backend and happy-life-backend share the same tigerbeetle node, the underlying tigerbeetle accounts end up being the same for the assets, as we end up calling this with id=1,2,3,4 in both rafiki instances (1-4, since we have 4 assets defined).

We should figure out a way to prevent this sharing of the same underlying settlement account.

Some options to fix this, other than using a second tigerbeetle node:
- Add `settlementAccountId` and `liquidityAccountId` to the Asset model

koekiebox:
Myself and @mkurapov / @BlairCurrey had a discussion regarding splitting the underlying ledgers for the two wallets. 
Ironically deciding that HappyLife Bank would make use of the internal Rafiki ledger, and CloudNine making use of the TB ledger implementation. 

The issue was further elevated due to the atomic Asset liquidity/settlement account creation in the PR below: 
https://github.com/interledger/rafiki/pull/2776/files

The change to ensure we make use of both PSQL and TigerBeetle implementations in our local environment:
https://github.com/interledger/rafiki/pull/2776/files#r1682606435

PR #2776, opened 2024-06-15
feat(backend): ip lookup, atomic asset creation, document updates.

koekiebox:

- https://github.com/interledger/rafiki/issues/2756

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/accounting/tigerbeetle/service.ts
```diff
@@ -286,6 +325,12 @@
   })
 }
 
+function bilateralIdentification(source: string, destination: string): bigint {
+  const arr = [source, destination].sort()
+  const md5Hex = createHash('md5').update(`${arr[0]}${arr[1]}`).digest('hex')
```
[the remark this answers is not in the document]
koekiebox: I agree, lets discuss alternatives. 😄

packages/backend/src/accounting/psql/service.ts
```diff
@@ -49,8 +52,15 @@ export function createAccountingService(
   return {
     createLiquidityAccount: (options, accTypeCode, trx) =>
       createLiquidityAccount(deps, options, accTypeCode, trx),
-    createSettlementAccount: (ledger, trx) =>
-      createSettlementAccount(deps, ledger, trx),
+    createSettlementAccount: (ledger, accountId, trx) =>
+      createSettlementAccount(deps, ledger, accountId, trx),
+    createLiquidityAndLinkedSettlementAccount: (options, accTypeCode, trx) =>
+      createLiquidityAndLinkedSettlementAccount(
```
koekiebox, line 58: Added to create both accounts in an atomic manner.

packages/backend/src/accounting/service.ts
```diff
@@ -81,13 +124,19 @@ export interface AccountingService {
   createWithdrawal(withdrawal: Withdrawal): Promise<void | TransferError>
   postWithdrawal(id: string): Promise<void | TransferError>
   voidWithdrawal(id: string): Promise<void | TransferError>
+  getAccountTransfers(
```
koekiebox, line 127: TigerBeetle now supports fetching of transfers. Adding to the accounting service later.

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -4,23 +4,30 @@ import {
...
 
 export interface CreateAccountOptions {
   id: AccountId
   ledger: number
-  code: TigerbeetleAccountCode
+  code: TigerBeetleAccountCode
+  // This flag links the result of this account creation to the result of the next one in the request,
+  // such that they will either succeed or fail together.
+  // The last account in a chain of linked accounts does not have this flag set.
+  linked?: boolean
```
koekiebox, line 18: Linked transfer for atomic account creation in TigerBeetle.

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -4,23 +4,30 @@ import {
...
+  // When set, the account will retain the history of balances at each transfer.
+  history?: boolean
```
koekiebox, line 20: Historic feature for account balances.

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -4,23 +4,30 @@ import {
...
+  userData128: AccountUserData128
```
koekiebox, line 21: Make use of the 16-byte userData to store the bilateral relationship between debtor/creditor.

packages/backend/src/accounting/tigerbeetle/service.ts
```diff
@@ -286,6 +329,12 @@
...
```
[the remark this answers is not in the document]
koekiebox: Looking at alternative hash functions (other than `md5` to make use of the bilateral relationship).

packages/backend/src/accounting/tigerbeetle/transfers.ts
```diff
@@ -154,3 +166,26 @@ export async function createTransfers(
...
+  id: string | number,
+  limit: number = 100_000
+): Promise<GetLedgerTransfersResult> {
+  const account_id = toTigerBeetleId(id)
+  const filter: AccountFilter = {
+    account_id,
+    timestamp_min: 0n,
+    timestamp_max: 0n,
+    limit, //100_000
+    flags: AccountFilterFlags.credits | AccountFilterFlags.debits
```
koekiebox, line 181: We do not currently split between debits / credits. Fetch all for account, we may want to change that in future.

packages/backend/src/index.ts
```diff
@@ -491,6 +493,37 @@ export function initIocContainer(
   return container
 }
 
+export const parseAndLookupAddresses = async (
+  replicaAddresses: string[]
+): Promise<string[]> => {
+  const parsed = []
```
koekiebox, line 499: Get around TigerBeetle not yet supporting DNS lookups.

packages/backend/src/accounting/tigerbeetle/service.ts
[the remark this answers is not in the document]
koekiebox: Replaced with `siphash`. 
Due to:
- Features: Fast and cryptographically secure hash function designed for short inputs and hash-table implementations.
- Security: Provides a high level of security against hash-flooding DoS attacks, with versions like SipHash-2-4 providing strong guarantees.

packages/backend/src/accounting/tigerbeetle/service.ts
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line
+const siphash = require('siphash')
```
koekiebox, line 2: Used require due to the way the module was declared for `siphash`. Suggestions welcome!

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -50,12 +54,24 @@ export async function createAccounts(
...
+): AccountFlags {
+  let returnVal =
+    options.code === TigerBeetleAccountCode.SETTLEMENT
+      ? AccountFlags.credits_must_not_exceed_debits
+      : AccountFlags.debits_must_not_exceed_credits
+  if (options.linked) returnVal |= AccountFlags.linked
+  if (options.history) returnVal |= AccountFlags.history
+  return returnVal
+}
+
```
BlairCurrey, line 68: Just reading this and finding it a bit confusing. Not saying there is anything wrong with it necessarily, and I realize this probably traces back to simply how the tb node client works, but I'm just not sure how this is supposed to work. Can you add a unit test for this - I think that would go a long way in clarifying how this is expected to work.

As I understand it, if we call `flagsBasedOnAccountOptions` with options:

```ts
{
  code: TigerBeetleAccountCode.SETTLEMENT, 
  linked: true, 
  history: true
}
````

Then the `returnValue` is  `13`, but this doesnt make sense to me because the return type is `AccountFlags` and `13` is not one of the values for this `AccountFlags` enum.

Let me show my math.

| Variable            | Value                            |
|---------------------|----------------------------------|
| `AccountFlags.credits_must_not_exceed_debits`      | `4` |
| `AccountFlags.linked`    | `1`                           |
| `AccountFlags.history`   | `8`                          |

So it the function in this case basically resolves to this (you can copy/paste this into a js/node interpreter to verify):

```js
returnVal = 4 //  let returnVal = options.code === TigerBeetleAccountCode.SETTLEMENT ? AccountFlags.credits_must_not_exceed_debits : AccountFlags.debits_must_not_exceed_credits
returnVal |= 1 // if (options.linked) returnVal |= AccountFlags.linked
returnVal |= 8 // if (options.history) returnVal |= AccountFlags.history

returnVal // resolves to 13
```

This just doesnt seem to add up because the return value is supposed to be of type `AccountFlags`

localenv/happy-life-bank/docker-compose.yml
```diff
@@ -53,9 +53,7 @@ services:
       CONNECTOR_PORT: 3002
       OPEN_PAYMENTS_PORT: 80
       DATABASE_URL: postgresql://happy_life_bank_backend:happy_life_bank_backend@shared-database/happy_life_bank_backend
-      USE_TIGERBEETLE: ${USE_TIGERBEETLE-false}
-      TIGERBEETLE_CLUSTER_ID: ${TIGERBEETLE_CLUSTER_ID-0}
-      TIGERBEETLE_REPLICA_ADDRESSES: ${TIGERBEETLE_REPLICA_ADDRESSES-''}
+      USE_TIGERBEETLE: false
```
koekiebox, line 56: Please see: 
https://github.com/interledger/rafiki/issues/2446

packages/backend/src/accounting/tigerbeetle/service.test.ts
```diff
@@ -793,4 +873,58 @@ describe('Tigerbeetle Accounting Service', (): void => {
       })
     })
   })
+
+  test('Test TigerBeetle Account Flags', async (): Promise<void> => {
```
koekiebox, line 877: Added test cases to test each of the possible account flag combinations.

packages/backend/src/accounting/tigerbeetle/accounts.ts
koekiebox: @BlairCurrey that is correct. ✅ I have added test cases with explanations to clarify. The TigerBeetle `flags` field is a unsigned 16bit field. The bitfield is used to flag none, one or many different "feature" flags. The resulting `flags` value will depend on what combination of flags need to be active.
The "opposite" is true (using the `&` operator) for when you want to verify if a certain flag/feature is enabled. Example:
```
const credExcDebitAndHistory = 12
    expect(
      flagsBasedOnAccountOptions({
        id: uuid(),
        code: TigerBeetleAccountCode.SETTLEMENT,
        ledger: 0,
        userData128: 0n,
        history: true
      })
    ).toEqual(credExcDebitAndHistory)

    // Tests to see if correct flags are set.
    expect(
      AccountFlags.credits_must_not_exceed_debits & credExcDebitAndHistory
    ).toBeTruthy()
expect(AccountFlags.linked & credExcDebitAndHistory).toBeFalsy()
```

Please let me know if it all makes sense? 


Basically the Linux file permissions work similar:
image



https://github.com/interledger/rafiki/pull/2776/files#r1686518809

packages/documentation/src/content/docs/concepts/accounting/accounts-and-transfers.md
```diff
@@ -751,12 +764,12 @@ Sender and receiver do not have wallet addresses at the same Rafiki instance.
 | -------------- | -------------- |
 | Peer Liquidity | Wallet Address |
 
-- Example: A wallet address receives 2 USD from an outgoing payment at a peer's Rafiki instance.
+- Example: A wallet address receives `2 USD` from an outgoing payment at a peer's Rafiki instance.
```
mkurapov, line 767: I don't think "from an outgoing payment" is correct here, since SPSP is more like direct/raw ILP access instead

packages/documentation/src/content/docs/concepts/accounting/accounts-and-transfers.md
```diff
@@ -12,8 +12,8 @@ A liquidity account may only hold a positive balance. Rafiki enforces that its t
 
 There is one liquidity account for each of the following resource:
 
-- Asset
-- Peer
+- [Asset](/reference/glossary#asset)
```
mkurapov, line 15: might be good to have this in a separate PR so the writers can also take a glance at it

mkurapov (commented):
Main question is just re: pagination resolver

packages/backend/src/accounting/tigerbeetle/utils.ts
```diff
@@ -1,21 +1,77 @@
...
   return uuidToBigInt(id)
 }
 
 export function uuidToBigInt(id: string): bigint {
   return BigInt(`0x${id.replace(/-/g, '')}`)
 }
+
+export function hexTextToBigInt(hex: string): bigint {
+  return BigInt(`0x${hex}`)
+}
```
BlairCurrey, line 22: This is unused. Remove?

packages/backend/src/accounting/tigerbeetle/utils.ts
```diff
@@ -1,21 +1,77 @@
...
+
+export function fromTigerBeetleId(bi: bigint): string {
```
BlairCurrey, line 24: Why do we need this function? Why cant we keep the tiger beetle id as a big int? I'm guessing its to satisfy some check that we're returning a uuid (in graphql?) 

It looks like this is taking something that is not a UUID and making it look like a UUID. If its just to satsify some downstream validation that assumes uuid I think maybe just loosening that check makes more sense.

BlairCurrey: > It looks like this is taking something that is not a UUID

Guess this understanding was wrong... I said that based off how TB is storing the ID but see we are doing `toTigerBeetleId(uuid())`.

packages/backend/src/accounting/tigerbeetle/transfers.ts
```diff
@@ -68,42 +76,46 @@ export async function createTransfers(
...
-        tbTransfer.pending_id = toTigerbeetleId(transfer.voidId)
+        tbTransfer.pending_id = toTigerBeetleId(transfer.voidId)
       }
     }
-    if (i < transfers.length - 1) {
-      tbTransfer.flags |= TransferFlags.linked
-    }
+
+    if (transfer.userData128)
+      tbTransfer.user_data_128 = toTigerBeetleId(transfer.userData128)
```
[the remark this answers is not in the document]
BlairCurrey: also, do we know that this should be converted to a tigerbeetle id? We know that it needs to be a bigint, but isnt that all we know?

Couldn't this be some json string that gets converted to a big int (string -> buffer -> hex -> bigint)? Complicated, I know, but isn't that the point of this field? arbitrary metadata? 

I remember the general intention to use this for some ID field but that just seems like an idea in our heads at the moment. I wonder if we should change `TransferOptions` to
```ts
interface TransferOptions {
  userData128?: BigInt
  code?:
  }
 ```
 
And then use `toTigerBeetleId` where we want to the `createAccountTransfer` is called.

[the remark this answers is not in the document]
Comment (BlairCurrey) [reply packages/backend/src/accounting/tigerbeetle/transfers.ts:113]: maybe my assumption about this being for generic metadata is wrong though.... its only 128 bytes (same as uuid). I kinda wish we could store an object like { someResourceId: <uuid>, ... } but even that would be too large. I suppose we just have to codify what this field is in the code (and just never change it without quite a bit of difficulty I guess).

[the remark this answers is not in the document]
Comment (BlairCurrey) [reply packages/backend/src/accounting/tigerbeetle/transfers.ts:113]: IDK if my point still stands or not, but I guess uuid -> bigint will take us down to more like `53` bits for the `MAX_SAFE_INTEGER`, if [the method for getting the bits of a bigint described here](https://stackoverflow.com/questions/54758130/how-to-obtain-the-amount-of-bits-of-a-bigint):

```js
> const x = BigInt(Number.MAX_SAFE_INTEGER)
> const i = (x.toString(16).length - 1) * 4
> i + 32 - Math.clz32(Number(x >> BigInt(i)))
53
```

and

```js
> BigInt(Number.MAX_SAFE_INTEGER).toString(2).length
53
```

So I guess we can store something like `{ someResourceId: <uuid>, ... }` if we wanted but it still seems like we could pretty easily try to store some JSON that exceeds the limit.

packages/backend/src/accounting/service.ts
```diff
@@ -57,6 +74,26 @@ export interface Transaction {
...
+}
+
+export interface LedgerTransfer extends BaseTransfer {
+  id: string
+  debitAccount: string
+  creditAccount: string
+  amount: bigint
+  timeout: number
+  timestamp: bigint
+  userData128: bigint
```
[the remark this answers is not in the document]
BlairCurrey: I was feeling similarly but as I understand it the initial intention is to store a UUID (128 bits) on it. Given it only holds 128 bits we couldnt store anything BUT the uuid. If thats the main use case im not sure if calling it additional data would be misleading. Did some more detailed thinking out loud here: https://github.com/interledger/rafiki/pull/2776/files/895999edb6cd7664e6aab7c300aa5bdf044eab66#r1686901609

packages/backend/src/accounting/tigerbeetle/transfers.ts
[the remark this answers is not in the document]
BlairCurrey: It does look like the docs suggest this should probably be an id and not necessarily more arbitrary metadata (hence 128 bits I suppose)

https://docs.tigerbeetle.com/reference/account#user_data_128

packages/backend/src/graphql/schema.graphql
```diff
@@ -91,6 +91,14 @@ type Query {
     filter: PaymentFilter
   ): PaymentConnection!
 
+  "Fetch a page of accounting transfers"
+  accountingTransfers(
```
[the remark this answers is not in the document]
BlairCurrey: > What are your thoughts on this? I'm hesitant to call this query "pagination" because it just takes limit - do we want to ship as is, or move it out into a separate issue and decide on the signature of this resolver separately?

That's my vote personally. Probably good not to call it pagination since its just limiting but nothing is even consuming this query ATM so I dont think we need to worry to much about improving it here.

.github/workflows/node-build.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/workflows/rafiki/env-setup
       - run: pnpm --filter backend build:deps
-      - run: NODE_OPTIONS=--max-old-space-size=4096 pnpm --filter backend test:ci
+      - run: NODE_OPTIONS="--max-old-space-size=4096" pnpm --filter backend test:ci
```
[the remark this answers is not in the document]
koekiebox: No it was, I was playing around with other node options due to a weird behaviour where I had to add `--experimental-vm-modules` additional option. Just thought to add the parentheses to ensure it is future proof. 
Shall I revert it, or are you happy with the parentheses?

packages/backend/src/graphql/schema.graphql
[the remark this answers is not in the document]
koekiebox: Yeah, I agree. Lets see how the API changes in future TigerBeetle releases. It would be cool for now to have the behaviour between the ledgers the same. I am going to resolve this comment for now. 🙇‍♂️

packages/backend/src/accounting/service.ts
```diff
@@ -10,6 +10,19 @@ export enum LiquidityAccountType {
   WEB_MONETIZATION = 'WEB_MONETIZATION'
 }
 
+export enum TransferType {
+  DEPOSIT = 'DEPOSIT',
+  WITHDRAWAL = 'WITHDRAWAL',
+  TRANSFER = 'TRANSFER',
+  DEFAULT = 'DEFAULT'
```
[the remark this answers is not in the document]
koekiebox: Hmm, that was me being silly. Fixed it now.

packages/backend/src/accounting/tigerbeetle/transfers.ts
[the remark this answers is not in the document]
koekiebox: That is correct. I agree with what you said. I have updated the logic to mimic what we are doing in our PSQL ledger, and that is to use it for `transferRef`.

packages/backend/src/index.ts
```diff
@@ -491,6 +493,37 @@ export function initIocContainer(
...
+  for (const addr of replicaAddresses) {
+    const parts = addr.split(':')
+    if (!parts)
+      throw new Error(
+        `Cannot parse replicaAddresses in 'accountingService' startup - value: "${addr}"`
+      )
+    if (parts.length === 1 && !isNaN(Number(parts[0]))) {
```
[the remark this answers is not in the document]
koekiebox: We allow for a port to be used only (no ip address), so I just wanted to keep that behaviour the same.

packages/backend/src/accounting/psql/service.ts
```diff
@@ -251,9 +274,8 @@ export async function createTransfer(
         )
       )
 
-      if (isTransferError(pendingTransfersOrError)) {
```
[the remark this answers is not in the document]
koekiebox: I have reverted back to add the curlies `{/}`

packages/backend/src/accounting/service.ts
```diff
@@ -65,8 +102,14 @@ export interface AccountingService {
   ): Promise<LiquidityAccount>
   createSettlementAccount(
```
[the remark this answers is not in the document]
koekiebox: No, I am removing it.

packages/backend/src/accounting/tigerbeetle/utils.ts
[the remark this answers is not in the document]
koekiebox: Added tests to take care of the number and uuid conversions.

koekiebox: Yeah, for the settlement accounts we store the id as numeric (basically converts to a 1n), but for the liquidity accounts we make use of uuid()
Added test cases to `service.test.ts` to demonstrate this.

packages/documentation/src/content/docs/concepts/accounting/accounts-and-transfers.md
koekiebox: Hi @melissahenderson @tselit , would you mind having a look at the updates I made to accounts and transfers?
Let me know if it is ok reviewing from the link below, or whether you want me to create a separate PR?
https://github.com/interledger/rafiki/pull/2776/files#r1686775804

Thank you!

koekiebox: Preview of the actual changes:
https://deploy-preview-2776--brilliant-pasca-3e80ec.netlify.app/concepts/accounting/accounts-and-transfers/

packages/documentation/src/content/docs/concepts/accounting/accounts-and-transfers.md
```diff
@@ -336,18 +349,18 @@ Rafiki transfers perform double-entry accounting. Every transfer increases both
 
 #### Payments (Same Asset)
 
-##### SPSP / Web Monetization
+##### Simple Payments Setup Protocol (SPSP) / Web Monetization (WM)
```
melissahenderson, line 352: Super small thing, but its Simple Payment, not Simple Payments (plural).

koekiebox: Thank you @melissahenderson ! All done.

melissahenderson:
@koekiebox Most recent doc updates look good to me. Not sure if I should be the one to approve the PR given all the other updates.

packages/backend/src/accounting/psql/service.test.ts
```diff
@@ -108,33 +108,33 @@ describe('Psql Accounting Service', (): void => {
     })
   })
 
-  describe('createSettlementAccount', (): void => {
+  describe('createLiquidityAndLinkedSettlementAccount', (): void => {
     test('creates account', async (): Promise<void> => {
```
mkurapov, line 112: let's have a test to make sure the settlement account was also created

packages/backend/src/accounting/psql/ledger-transfer/model.ts
```diff
@@ -3,12 +3,7 @@ import { Model, QueryContext } from 'objection'
 import { Asset } from '../../../asset/model'
 import { BaseModel } from '../../../shared/baseModel'
 import { LedgerAccount } from '../ledger-account/model'
-
-export enum LedgerTransferState {
-  PENDING = 'PENDING',
-  POSTED = 'POSTED',
-  VOIDED = 'VOIDED'
-}
```
mkurapov, line 11: Let's have the services import from the models, instead of the other way around. To keep the different layers in order. 

Or we re-delcare them in the services, like you've done for `LedgerTransferType` it looks like

packages/backend/src/accounting/psql/service.test.ts
```diff
@@ -181,6 +181,11 @@ describe('Psql Accounting Service', (): void => {
       ])
 
       const amount = 10n
+      const accTransfersEmpty = await accountingService.getAccountTransfers(
```
mkurapov, line 184: I think a separate test for `getAccountTransfers` would be good IMO

packages/backend/src/config/app.ts
```diff
@@ -105,11 +105,12 @@ export const Config = {
...
   streamSecret: Buffer.from(process.env.STREAM_SECRET || '', 'base64'),
-  useTigerbeetle: envBool('USE_TIGERBEETLE', true),
-  tigerbeetleClusterId: envInt('TIGERBEETLE_CLUSTER_ID', 0),
-  tigerbeetleReplicaAddresses: process.env.TIGERBEETLE_REPLICA_ADDRESSES
+  useTigerBeetle: envBool('USE_TIGERBEETLE', true),
+  tigerBeetleClusterId: envInt('TIGERBEETLE_CLUSTER_ID', 0),
+  tigerBeetleReplicaAddresses: process.env.TIGERBEETLE_REPLICA_ADDRESSES
     ? process.env.TIGERBEETLE_REPLICA_ADDRESSES.split(',')
     : ['3004'],
+  tigerBeetleTwoPhaseTimeout: envInt('TIGERBEETLE_TWO_PHASE_TIMEOUT', 5),
```
mkurapov, line 113: Let's include a time unit in here

packages/backend/src/accounting/psql/service.ts
```diff
@@ -13,7 +13,8 @@ import {
   Transaction,
   TransferOptions,
   TransferToCreate,
-  Withdrawal
+  Withdrawal,
+  GetLedgerTransfersResult
```
mkurapov, line 17:
```suggestion
  
```

packages/backend/src/accounting/psql/service.ts
```diff
@@ -30,9 +31,11 @@ import {
   createTransfers,
   CreateTransfersResult,
   postTransfers,
-  voidTransfers
+  voidTransfers,
+  getAccountTransfers
 } from './ledger-transfer'
 import { LedgerTransfer, LedgerTransferType } from './ledger-transfer/model'
+import { tbTransferToLedgerTransfer } from '../tigerbeetle/utils'
```
mkurapov, line 38:
```suggestion
```

packages/backend/src/index.ts
```diff
@@ -491,6 +493,36 @@ export function initIocContainer(
...
+      parsed.push(parts[0])
+      continue
+    }
+
+    if (net.isIP(parts[0]) === 0) {
+      try {
+        await dns.promises.lookup(parts[0], { family: 4 }).then((resp) => {
+          parsed.push(`${resp.address}:${parts[1]}`)
+        })
```
mkurapov, line 515:
```suggestion
        const resp = await dns.promises.lookup(parts[0], { family: 4 })
        parsed.push(`${resp.address}:${parts[1]}`)
```
I'd have to check but I don't remember what happens when doing await with `.then` and no `.catch` handler. Probably ok but this is more straightforward

packages/backend/src/graphql/resolvers/accounting_transfer.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      amount: BigInt(transferAmount),
+      debitAccount: accountDebitLedger[0],
+      debitAccountId: accountDebitLedger[0].id,
+      creditAccount: accountCreditLedger[0],
+      creditAccountId: accountCreditLedger[0].id,
+      ledger: ledger,
+      state: LedgerTransferState.POSTED,
+      transferRef: uuid(),
+      type: LedgerTransferType.DEPOSIT
+    })
```
mkurapov, line 114: I think you can use `createLedgerTransfer` from `packages/backend/src/tests/ledgerTransfer.ts` here

packages/backend/src/graphql/schema.graphql
```diff
@@ -895,6 +903,37 @@ input PaymentFilter {
...
+  credits: [AccountingTransfer!]!
+}
+
+type AccountingTransfer implements Model {
+  "Payment id"
+  id: ID!
+  "Debit account id"
+  debitAccount: ID!
+  "Credit account id"
+  creditAccount: ID!
```
mkurapov, line 917:
```suggestion
  debitAccountId: ID!
  "Credit account id"
  creditAccountId: ID!
```

packages/backend/src/graphql/resolvers/accounting_transfer.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      })
+
+    expect(response.debits).toBeDefined()
+    expect(response.credits).toBeDefined()
+    expect(response.debits).toHaveLength(0)
+    expect(response.credits).toHaveLength(1)
+
+    expect(response.credits[0]).toMatchObject({
+      id: insertedTransfer.id,
+      debitAccount: accountDebitId,
```
mkurapov, line 216: let's check for the creditAccountId in here as well

packages/backend/src/accounting/psql/service.ts
```diff
@@ -82,13 +92,13 @@ export async function createLiquidityAccount(
...
     trx
   )
-
   return account
 }
 
 export async function createSettlementAccount(
   deps: ServiceDependencies,
   ledger: number,
+  assetId: string | number,
```
mkurapov, line 101: do we need assetId here?

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -60,21 +62,42 @@ export async function getAccountTransfers(
...
+    amount: dbModel.amount,
+    creditAccountId: dbModel.creditAccount ? dbModel.creditAccount.id : '',
+    debitAccountId: dbModel.debitAccountId,
+    id: dbModel.id,
+    ledger: dbModel.ledger,
+    timestamp: BigInt(dbModel.createdAt.getTime()),
+    type:
+      dbModel.type === LedgerTransferType.WITHDRAWAL
+        ? TransferType.WITHDRAWAL
+        : TransferType.DEPOSIT,
```
mkurapov, line 93: I _think_ we don't end up passing through the `TransferType.TRANSFER` for normal account to account transfers for the PSQL accounting service. We should, otherwise this will mark "normal" transfers as `TransferType.DEPOSIT`.

mkurapov (commented):
Sorry for another big batch of comments, mostly just small things!

packages/backend/src/accounting/tigerbeetle/accounts.ts
```diff
@@ -4,23 +4,30 @@ import {
...
   ledger: number
-  code: TigerbeetleAccountCode
+  code: TigerBeetleAccountCode
+  // This flag links the result of this account creation to the result of the next one in the request,
+  // such that they will either succeed or fail together.
+  // The last account in a chain of linked accounts does not have this flag set.
+  linked?: boolean
+  // When set, the account will retain the history of balances at each transfer.
+  history?: boolean
+  assetId: AccountUserData128
```
BlairCurrey, line 21: Should we get rid of `assetId` here? Or use it? Not sure what the intention was but it looks like it's unused here.

bruno/collections/Rafiki/Rafiki Admin APIs/Get Accounting Ledger Transfers.bru
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+  
+}
+
+body:graphql:vars {
+  {
+      "id": "{{outgoingPaymentId}}"
+  }
+}
```
BlairCurrey, line 43: getting an error without the limit var

```suggestion
body:graphql:vars {
  {
      "id": "{{outgoingPaymentId}}",
      "limit": 100
  }
}
```
Actually my first though was not to require the limit in the GQL schema and set some default limit on our backend. This is what out paginated resolvers do. However, since we aren't returning any sort of page info back I think that behavior would be unclear so maybe requiring it is the way to go for now - just something to keep in mind if we do implement paging here.

bruno/collections/Rafiki/Rafiki Admin APIs/Get Accounting Ledger Transfers.bru
```diff
@@ -0,0 +1,49 @@
...
+        createdAt
+      }
+    }
+  }
+  
+}
+
+body:graphql:vars {
+  {
+      "id": "{{outgoingPaymentId}}"
```
BlairCurrey, line 41: Im not able to get this to work because it doesnt find the `outgoingPaymentId`. Maybe you can tell me what I'm doing wrong?

I did the Open Payments example flow (which saves the `outgoingPaymentId` env var) and then did this and saw empty results. For good measure I tried sending this request to `{{PeerGraphqlHost}}/graphql` (happy life bank) instead but saw the same result. I went into postgres db and grabbed a `debitAccountId` from one of the `ledgerTransfers` and it worked. Should this be something other than `outgoingPaymentId` or...?

packages/backend/src/accounting/psql/service.test.ts
koekiebox: I agree. Done.

packages/backend/src/accounting/tigerbeetle/accounts.ts
koekiebox: @BlairCurrey The one in the PSQL wasn't used (I removed it now), but we want to use the `assetId` in order to have the same assetId when there is a liquidity/settlement account being created for an asset.
Also, good catch. I did not properly set it on the account. I have now. 
```
    accounts.map((account) => ({
      id: toTigerBeetleId(account.id),
      user_data_32: 0,
      user_data_64: 0n,
      user_data_128: toTigerBeetleId(account.assetId),
      reserved: 0,
      ledger: account.ledger,
      code: account.code,
      // Credit and debit accounts can both send and receive
      // but are restricted by their respective TigerBeetle flags.
      // In Rafiki transfers:
      // - the source account's debits increase
      // - the destination account's credits increase
      flags: flagsBasedOnAccountOptions(account),
      debits_pending: 0n,
      debits_posted: 0n,
      credits_pending: 0n,
      credits_posted: 0n,
      timestamp: 0n
    }))
```

bruno/collections/Rafiki/Rafiki Admin APIs/Get Accounting Ledger Transfers.bru
koekiebox: @BlairCurrey Apologies, I have made the limit optional, then also mentioning the limit in the docs. 

image

bruno/collections/Rafiki/Rafiki Admin APIs/Get Accounting Ledger Transfers.bru
koekiebox: @BlairCurrey Updated it now to be proper and always working. 🙇 

image

BlairCurrey (approved):

packages/backend/src/accounting/psql/ledger-transfer/index.ts
```diff
@@ -60,21 +62,42 @@ export async function getAccountTransfers(
...
+  if (dbModel.type) transferType = TransferType[dbModel.type]
+
+  return {
+    amount: dbModel.amount,
+    creditAccountId: dbModel.creditAccountId,
+    debitAccountId: dbModel.debitAccountId,
+    id: dbModel.id,
+    ledger: dbModel.ledger,
+    timestamp: BigInt(dbModel.createdAt.getTime()),
+    type: transferType,
```
mkurapov, line 93:
```suggestion
    type: dbModel.type ? TransferType[dbModel.type] : TransferType.TRANSFER,
```

packages/backend/src/graphql/resolvers/accounting_transfer.tigerbeetle.test.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * @jest-environment ./packages/backend/jest.tigerbeetle-environment.ts
+ */
+
+import { IocContract } from '@adonisjs/fold'
```
mkurapov, line 5: I think we could probably get rid of this test file and just add tests for `getAccountTransfers` in `/src/accounting/tigerbeetle/service.test.ts`, similar to how you did for the PSQL tests. 
IMO resolver tests shouldn't really care what is used under the hood, just that they return the correct response and possible errors

mkurapov (approved):
Looks good, just final thing re: tests mainly

packages/backend/src/graphql/resolvers/accounting_transfer.tigerbeetle.test.ts
koekiebox: I picked up an issue during the testing that only occurred in the resolver. So I added it due to test coverage and to prevent it from creeping up in the future. If it is ok, I would like to leave it?

BlairCurrey (approved):

packages/backend/src/accounting/tigerbeetle/utils.test.ts
```diff
@@ -0,0 +1,33 @@
...
+      // @ts-expect-error error message will always be present:
+      expect(error.message).toEqual('wrong format of id')
+    }
+  })
+
+  test('Test ', async (): Promise<void> => {
+    expect(uuidToBigInt('1ed4bdd8-adba-4e43-a76c-278ec1ae23cc')).toEqual(
+      40981457350021189385292318948847395788n
+    )
+  })
```
BlairCurrey, line 32: This test could pass incorrectly if `toTigerBeetleId` doesn't throw.

Could use `expect.assertions` to ensure the correct number of assertions were made. It's used around our tests for this purpose.

koekiebox: Thank you @BlairCurrey . Done.

BlairCurrey (approved):

PR #2779, opened 2024-06-21
chore(auth): return objects during GNAP errors

mkurapov:

Updates how errors are displayed when making requests to the AS, to be more in-line with the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#name-error-response). This improves error handling in general as it allows for parsing error codes, and making use of the description.

Before:
Screenshot 2024-06-21 at 20 01 08

After:
Screenshot 2024-06-21 at 20 00 43

This PR also adds some if-checks to the OP post requests scripts in bruno, in order not to prevent any errors from being hidden, since when post requests scripts would fail to parse response body information, the script error would override the actual response error.

## Context

General OP error handling #1905

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/auth/src/shared/gnapErrors.ts
```diff
@@ -57,10 +57,19 @@ export async function gnapServerErrorMiddleware(
         'Received error when handling Open Payments GNAP request'
```
mkurapov, line 57: main changes here: don't throw error, but return status & body. Follows error handling koa docs: https://github.com/koajs/koa/blob/master/docs/error-handling.md

packages/auth/src/shared/gnapErrors.ts
```diff
@@ -72,7 +81,11 @@ export async function gnapServerErrorMiddleware(
         'Received OpenAPI validation error when handling Open Payments GNAP request'
       )
 
-      ctx.throw(finalStatus, err.message)
+      ctx.status = finalStatus
+      ctx.body = {
+        error: { description: err.message }
+      }
```
mkurapov, line 87: this is not a GNAP error per se, just a validation error, for which we just add the description

mkurapov, line 87: this is not a GNAP error per se, just a validation error, for which we just add the description

njlie (approved):
LGTM

PR #2782, opened 2024-06-25
chore(backend, auth): verify try-catches in tests

mkurapov:

- Updating tests cases containing `try-catch` blocks to make sure that we don't bypass the assertions in catch statements (in case there is no actual error in the "try" block). Follow-up to changes in #2768.

Relevant Jest [docs](https://jestjs.io/docs/asynchronous#asyncawait)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/graphql/resolvers/liquidity.test.ts
```diff
@@ -104,6 +104,7 @@ describe('Liquidity Resolvers', (): void => {
     })
 
     test('Returns an error for invalid id', async (): Promise<void> => {
+      let error
```
mkurapov, line 107: in this file, because we had expects in `beforeEach`/`beforeAll` functions, doing `expect.assertions` wasn't intuitive at all (sometimes # of actual assertions would end up being 2 or 3 more than in the test block itself), so I just decided to check for the errors this way

njlie (dismissed):
Thanks for doing this!

njlie (approved):

PR #2783, opened 2024-06-25
chore(backend): estimate ILP quote receive amount, and fail early

### LINKED ISSUE #2770: Fix minimum delivery amounts
## Context

We are able to get "unrealistic" minimum delivery amounts when we are getting quotes .

For example, if we are sending between a RON (scale 2) wallet address to a USD (scale 2) wallet address, where the exchange rate is 1 RON = 0.2 USD, we can set debitAmount to 0.02 RON and get back a quote for 0.01 USD, even though, because of the exchange rate, the outgoing payment will never actually go through, since 0.02 RON * 0.2 = 0.004 USD, which is under a single unit of value of 0.01 USD. This is caused by the way that the ILP pay rounding works.

## Example of behaviour

> So even though you can create the quote (and corresponding outgoing payment) for a debitAmount that will clearly not deliver the minimum unit of the receiving asset, the payment will fail when trying to send.

> I set up a scenario, sending from RON (scale 2) to USD (scale 2), where 1 USD = 20 RON (ie 1 RON = 0.05 USD).
First attempt, you can create a fixed-send 0.02 RON quote that results in 0.01 USD delivery. The outgoing payment can be created from the quote, but it fails when sending.

> Second attempt, 0.19 RON quote results in a 0.01 USD delivery. The outgoing payment can be created, but still fails when sending.

> Final attempt, trying to send 0.20 RON results in a 0.01 USD delivery. The outgoing payment can be created and is successful.
This means, even though the quoting works in all 3 scenarios, the payment will only be completed when the correct delivery amount is reached, in our case, when at least 0.20 RON is sent, which is good. This means there are no ILP pay "issues" or "money making" during actual payment.

https://github.com/interledger/rafiki/assets/15069181/702d7a98-fa85-4577-95e3-ee997aad614d

### Slack threads:
- Initial [thread](https://interledgerfoundation.slack.com/archives/C05EMHTG6FM/p1716803842183209) from Radu
- [Thread](https://interledger.slack.com/archives/CMCUD5B36/p1718891806093109) with example of behaviour 

## Todos
- [ ] When getting a quote for a small amount from the sending wallet, we should error if we cannot fulfill a minimum, single unit  of value of the receiving amount. In other words, depending on the exchange rate, we need to send an amount of sending that is _at least_ a single unit of value in the receiving asset.

mkurapov:

Error during quoting if the ILP quote will not fulfil at least a single unit of the receiving asset. See #2770 for detailed explanation.

## Context

Fixes #2770 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/open_payments/quote/service.test.ts
+++ b/packages/backend/src/open_payments/quote/service.test.ts
@@ -31,6 +31,10 @@ import { PaymentMethodHandlerService } from '../../payment-method/handler/servic
 import { ReceiverService } from '../receiver/service'
 import { createReceiver } from '../../tests/receiver'
 import * as Pay from '@interledger/pay'
+import {
+  PaymentMethodHandlerError,
+  PaymentMethodHandlerErrorCode
+} from '../../payment-method/handler/errors'
 
 describe('QuoteService', (): void => {
   let deps: IocContract<AppServices>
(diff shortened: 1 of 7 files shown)


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -159,6 +163,14 @@ async function createQuote(
     if (isQuoteError(err)) {
       return err
     }
+
+    if (
+      err instanceof PaymentMethodHandlerError &&
+      err.code === PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount
+    ) {
+      return QuoteError.NonPositiveReceiveAmount
+    }
```
mkurapov, line 172: Handle gracefully, avoid returning 500 for an unhandled `PaymentMethodHandlerError`.
This is also prepping for having better error codes in the RS in general.

packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -466,6 +470,32 @@ describe('QuoteService', (): void => {
...
+          })
+        })
+
+      await expect(
+        quoteService.create({
+          walletAddressId: sendingWalletAddress.id,
+          receiver: receiver.incomingPayment!.id,
+          method: 'ilp',
+          debitAmount: {
+            value: 2n,
```
sabineschaller, line 491: Why does that end up being non-positive? Because of slippage?

sabineschaller (dismissed):
Looks good, just one curiosity question

packages/backend/src/open_payments/quote/service.test.ts
mkurapov: We mock the call to `paymentMethodHandlerService.getQuote` so this can be anything

raducristianpopa (approved):

PR #2787, opened 2024-06-26
feat: transaction time histogram metric

### LINKED ISSUE #2735: Track transaction time per outgoing payment
### Context

We want to track how long an Open Payments outgoing payment takes to complete. For now, we will not differentiate between payments at the local rafiki (the incoming payment is at the same rafiki), or a remote rafiki (where the incoming payment is at a different rafiki).

Note: we only care about "completed" outgoing payments, not partially completed, or failed ones.  

### Tasks
- [x] After an outgoing payment is completed, we should increment a new `outgoing_payment_completion_time_ms` metric using "Date.now() - payment.createdAt"

BlairCurrey:

- adds histogram metric to telemetry service and uses for new `outgoing_payment_completion_time_ms` metric 
- adds visualizations to grafana in new "Transaction Time Development" dashboard (will merge into other dashboards when its finalized): https://rafikitelemetry.grafana.net/d/ddpy9febd69s0d/transaction-time-development?orgId=1

image

## Context

- fixes #2735 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -81,13 +81,24 @@ export async function handleSending(
...
+        source: deps.telemetry.getInstanceName()
+      })
+
+    deps.telemetry
+      .getOrCreateHistogramMetric('outgoing_payment_completion_time_ms', {
+        description: 'Time to complete a payment'
+      })
+      .record(Date.now() - payment.createdAt.getTime(), {
+        source: deps.telemetry.getInstanceName()
+      })
```
BlairCurrey, line 100: @mkurapov (and anyone else) do you think these following bucket thresholds would be better?:

```ts
deps.telemetry
      .getOrCreateHistogramMetric('outgoing_payment_completion_time_ms', {
        description: 'Time to complete a payment',
        advice: {
          explicitBucketBoundaries: [
            100,
            200,
            300,
            400,
            500,
            600,
            700,
            800,
            900,
            1000,
            2000,
            5000,
            10000,
            Infinity
          ]
        }
```

The default buckets which it's currently using (as can be seen in the pr body screenshot) are 0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000, +Inf. So the lower part of the range is too low (it will never complete in 0, 5, etc. ms). Bad bucket sizes make the distribution graphs less useful.

However, this would probably complicate things on grafana, as the `outgoing_payment_completion_time_ms` already exists in prometheus with the default buckets so I didn't want to just change it. I think we'd need to either add a new name or somehow delete this existing one in prometheus. 

@beniaminmunteanu wondering if you can speak to whether or not we could easily whipe the existing `outgoing_payment_completion_time_ms` metric in prometheus? I think I'd prefer that if we are going to set different bucket thresholds because this is the name we want to use and I dont want to pollute prometheus with unused metrics. Also, what is the retention policy on metrics (should be set in prometheus config somewhere I think). If we do rename I guess these unused ones will just drop off eventually.

packages/backend/src/telemetry/service.ts
```diff
@@ -85,6 +87,24 @@ class TelemetryServiceImpl implements TelemetryService {
...
   }
 
+  private createHistogram(name: string, options: MetricOptions | undefined) {
+    const histogram = metrics
+      .getMeter(METER_NAME)
+      .createHistogram(name, options)
+    this.histograms.set(name, histogram)
+    return histogram
+  }
+  public getOrCreateHistogramMetric(
```
BlairCurrey, line 97: the analogous one for our counters is named `getOrCreateMetric`. So it implicitly assumed metrics were counters (which was true up until adding this). I didn't touch that one but considered renaming it to `getOrCreateCounterMetric`.

BlairCurrey:
Some notes on the graphs...

- It was difficult to generate even a small amount of data to populate the graphs. Just ran the bruno open payments flow a number of times for the one pictured in the pr body. More data would have been nice to validate the usefulness of the graphs (two distributions and transactions per second in particular).
- heavily inspired by this grafana blogpost. https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/ That can serve as some example for what the distribution graphs might look like with more data. In particular the distributions over time.
- the "Shorter than 500ms" metric might be useful if there is some threshold we want to target for transaction time. otherwise we can probably just get rid of that one. similar information is show in the quartile group.

BlairCurrey:
@mkurapov @JoblersTune please also review the queries in the dashboard if you can https://rafikitelemetry.grafana.net/d/ddpy9febd69s0d/transaction-time-development?orgId=1

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
mkurapov: I think we can leave the default for now TBH, and then fine-tune them with more data

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -81,13 +81,24 @@ export async function handleSending(
...
+    deps.telemetry
+      .getOrCreateMetric('transactions_total', {
+        description: 'Count of funded transactions'
+      })
+      .add(1, {
+        source: deps.telemetry.getInstanceName()
+      })
+
+    deps.telemetry
+      .getOrCreateHistogramMetric('outgoing_payment_completion_time_ms', {
```
mkurapov, line 95: If we go with tracking just the ILP payment, maybe we can rename this to be something ILP related? e.g. `ilp_pay_time_ms`

JoblersTune:
> Some notes on the graphs shared in the pr body:
> 
> * I wasnt really able to generate a large, meaningful amount of data for these graphs. Just ran the bruno open payments flow a number of times. More data would have been nice to validate the usefulness of the graphs (two distributions and transactions per second in particular).
> * heavily inspired by this grafana blogpost. https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/ That can serve as some example for what the distribution graphs might look like with more data. In particular the distributions over time.
> * the "Shorter than 500ms" metric might be useful if there is some specific threshold we care about. Most but not all completed under 500ms locally, which is why I picked it. Otherwise we can probably just get rid of that one. The quartiles kind of show similar info.

Is there a reason you're using `rate(outgoing_payment_completion_time_ms_count[1m]) ` instead of just `rate(transactions_total[1m]) `?

In terms of the interval, I'd probably go with maybe 5m instead, but it's much of a muchness. It just smooths things out more. But it depends if we want to look and see general trends or if we want to look and see anomalies.

packages/backend/src/telemetry/service.ts
JoblersTune: I agree with renaming for clarity.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: Sounds good... I will probably need to change the name anyways since we're changing to time just the ilp pay part. I just dreaded a `outgoing_payment_completion_time_ms` and `outgoing_payment_completion_time_ms_v2` situation but I think at some point they might fall off from being unused (depends on our configuration though I guess).

BlairCurrey:
> > Some notes on the graphs shared in the pr body:
> > 
> > * I wasnt really able to generate a large, meaningful amount of data for these graphs. Just ran the bruno open payments flow a number of times. More data would have been nice to validate the usefulness of the graphs (two distributions and transactions per second in particular).
> > * heavily inspired by this grafana blogpost. https://grafana.com/blog/2020/06/23/how-to-visualize-prometheus-histograms-in-grafana/ That can serve as some example for what the distribution graphs might look like with more data. In particular the distributions over time.
> > * the "Shorter than 500ms" metric might be useful if there is some specific threshold we care about. Most but not all completed under 500ms locally, which is why I picked it. Otherwise we can probably just get rid of that one. The quartiles kind of show similar info.
> 
> Is there a reason you're using `rate(outgoing_payment_completion_time_ms_count[1m]) ` instead of just `rate(transactions_total[1m]) `?
> 
> In terms of the interval, I'd probably go with maybe 5m instead, but it's much of a muchness. It just smooths things out more. But it depends if we want to look and see general trends or if we want to look and see anomalies.
> 
> If we used transactions_total we'd get something more like this image

I agree that  `transactions_total` is probably better. I was focusing on the new metric I created so I hadn't really considered that existing one. `transactions_total` at 5m intervals looks better - will update to that.

packages/backend/src/telemetry/service.ts
```diff
@@ -13,6 +13,7 @@ import { BaseService } from '../shared/baseService'
 export interface TelemetryService {
   shutdown(): void
   getOrCreateMetric(name: string, options?: MetricOptions): Counter
+  getOrCreateHistogramMetric(name: string, options?: MetricOptions): Histogram
```
mkurapov, line 16: This doesn't have to be in this PR of course, but what do you think about changing the `TelemetryService` interface/implementation?

Looking at `getOrCreateMetric` I'm not sure if I like how some telemetry "details" are always needed in the calling services. For example, when we increment the counter, we pass in (we should always pass in) the counter attributes object: 
{
    source: deps.telemetry.getInstanceName()
} 
even though IMO, it shouldn't be the problem of the calling service to do it (even though, of course, it can provide additional attributes to the counter, if it wants).

Ideally, including the addition of the histogram, we should only have three public methods in the service: `shutdown` `incrementCounter` `recordHistogram`.

packages/backend/src/telemetry/service.ts
```diff
@@ -47,6 +48,7 @@ class TelemetryServiceImpl implements TelemetryService {
   private aseRatesService: RatesService
 
   private counters = new Map()
+  private histograms = new Map()
```
mkurapov, line 51: we should add types to the `Map`s

packages/backend/src/telemetry/service.ts
mkurapov: Of course, those are just suggested methods, maybe we could also have `incrementAmountWithPrivacy` or something as well, and the `privacy` obj doesn't need to be exposed outside of `TelemetryService` itself

packages/backend/src/telemetry/service.ts
BlairCurrey: going to hold off on this renaming because this will change with https://github.com/interledger/rafiki/issues/2799

mkurapov (dismissed):

packages/documentation/src/content/docs/telemetry/overview.md
```diff
@@ -70,5 +70,8 @@ Currently collected metrics:
 - `transactions_amount` - Counter metric
   - Description: “Amount sent through the network”.
   - This amount metric increases by the amount sent in each ILP packet.
+- `ilp_pay_time_ms` - Histogram metric
```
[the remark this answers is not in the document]
BlairCurrey: updated

mkurapov (approved):

PR #2788, opened 2024-06-26
feat(auth): inspect access during token introspection

### LINKED ISSUE #835: Check request access during token introspection
- https://github.com/interledger/rafiki/issues/798#issuecomment-1335314019
- https://github.com/interledger/rafiki/issues/798#issuecomment-1335342847

Pass the Open Payments request access type/action in the token introspection request `access` field.
The AS should return `{ active: false }` if the token's grant does not include the specified type/action.
Otherwise, the token introspection response's `access` field should only include the access permitting the request.

https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers#name-token-introspection

njlie:

- Adds `access` parameter to token introspection request
- `auth` service now verifies if provided `access` is sufficiently allowed by grant retrieved during token introspection
- `token-introspection` extends capability of verifying a token has sufficient access for a requested access scope.

## Context

Fixes #835.

## Checklist

- [x] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/access/utils.test.ts
+++ b/packages/auth/src/access/utils.test.ts
@@ -152,7 +152,7 @@ describe('Access utilities', (): void => {
         requestAccessItem,
         toOpenPaymentsAccess(grantAccessItemSuperAction)
       )
-    )
+    ).toBe(true)
   })
 
   test('access comparison fails if grant action items are insufficient', async (): Promise<void> => {
--- a/packages/auth/src/access/utils.test.ts
+++ b/packages/auth/src/access/utils.test.ts
@@ -121,8 +121,7 @@ describe('Access utilities', (): void => {
     const grantAccessItemSuperAction = await Access.query(trx).insertAndFetch({
       grantId: grant.id,
       type: AccessType.OutgoingPayment,
-      actions: [AccessAction.ReadAll
-      ],
+      actions: [AccessAction.ReadAll],
       identifier,
       limits: {
         receiver,
@@ -149,7 +148,10 @@ describe('Access utilities', (): void => {
     }
 
     expect(
-      compareRequestAndGrantAccessItems(requestAccessItem, toOpenPaymentsAccess(grantAccessItemSuperAction))
+      compareRequestAndGrantAccessItems(
+        requestAccessItem,
+        toOpenPaymentsAccess(grantAccessItemSuperAction)
+      )
     )
   })
 
--- a/packages/auth/src/access/utils.test.ts
+++ b/packages/auth/src/access/utils.test.ts
@@ -117,6 +117,42 @@ describe('Access utilities', (): void => {
     ).toBe(true)
   })
 
+  test('Can compare an access item on a grant and an access item from a request with a subaction of the grant', async (): Promise<void> => {
+    const grantAccessItemSuperAction = await Access.query(trx).insertAndFetch({
+      grantId: grant.id,
+      type: AccessType.OutgoingPayment,
+      actions: [AccessAction.ReadAll
+      ],
+      identifier,
+      limits: {
+        receiver,
+        debitAmount: {
+          value: '400',
+          assetCode: 'USD',
+          assetScale: 2
+        }
+      }
+    })
+
+    const requestAccessItem: AccessItem = {
+      type: 'outgoing-payment',
+      actions: ['read'],
+      identifier,
+      limits: {
+        receiver,
+        debitAmount: {
+          value: '400',
+          assetCode: 'USD',
+          assetScale: 2
+        }
+      }
+    }
+
+    expect(
+      compareRequestAndGrantAccessItems(requestAccessItem, toOpenPaymentsAccess(grantAccessItemSuperAction))
+    )
+  })
+
   test('access comparison fails if grant action items are insufficient', async (): Promise<void> => {
     const identifier = `https://example.com/${v4()}`
     const receiver =
--- a/packages/auth/src/access/utils.test.ts
+++ b/packages/auth/src/access/utils.test.ts
@@ -1,0 +1,171 @@
+import { v4 } from 'uuid'
+import { IocContract } from '@adonisjs/fold'
+import { Knex } from 'knex'
+import { faker } from '@faker-js/faker'
+import {
+  AccessItem,
+  AccessType,
+  AccessAction
+} from '@interledger/open-payments'
+
+import { AppServices } from '../app'
+import { Access, toOpenPaymentsAccess } from './model'
+import { Grant, GrantState, StartMethod, FinishMethod } from '../grant/model'
+import { initIocContainer } from '..'
+import { Config } from '../config/app'
+import { createTestApp, TestContainer } from '../tests/app'
+import { truncateTables } from '../tests/tableManager'
+import { generateToken, generateNonce } from '../shared/utils'
+import { compareRequestAndGrantAccessItems } from './utils'
+
+describe('Access utilities', (): void => {
+  let deps: IocContract<AppServices>
+  let appContainer: TestContainer
+  let trx: Knex.Transaction
+  let identifier: string
+  let grant: Grant
+  let grantAccessItem: Access
+
+  const receiver: string =
+    'https://wallet.com/alice/incoming-payments/12341234-1234-1234-1234-123412341234'
+
+  beforeAll(async (): Promise<void> => {
+    deps = initIocContainer(Config)
+    appContainer = await createTestApp(deps)
+  })
+
+  beforeEach(async (): Promise<void> => {
+    identifier = `https://example.com/${v4()}`
+    grant = await Grant.query(trx).insertAndFetch({
+      state: GrantState.Processing,
+      startMethod: [StartMethod.Redirect],
+      continueToken: generateToken(),
+      continueId: v4(),
+      finishMethod: FinishMethod.Redirect,
+      finishUri: 'https://example.com/finish',
+      clientNonce: generateNonce(),
+      client: faker.internet.url({ appendSlash: false })
+    })
+
+    grantAccessItem = await Access.query(trx).insertAndFetch({
+      grantId: grant.id,
+      type: AccessType.OutgoingPayment,
+      actions: [AccessAction.Read, AccessAction.Create, AccessAction.List],
+      identifier,
+      limits: {
+        receiver,
+        debitAmount: {
+          value: '400',
+          assetCode: 'USD',
+          assetScale: 2
+        }
+      }
+    })
+  })
+
+  afterEach(async (): Promise<void> => {
+    await truncateTables(appContainer.knex)
+  })
+
+  afterAll(async (): Promise<void> => {
+    await appContainer.shutdown()
+  })
+
+  test('Can compare an access item on a grant and an access item from a request', async (): Promise<void> => {
+    const requestAccessItem: AccessItem = {
+      type: 'outgoing-payment',
+      actions: ['create', 'read', 'list'],
+      identifier,
+      limits: {
+        receiver,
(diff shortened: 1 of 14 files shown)


packages/token-introspection/src/client/introspection.test.ts
```diff
@@ -98,17 +99,15 @@ describe('introspection', (): void => {
     })
   })
 
-  describe('validateTokenInfo', (): void => {
-    test('returns valid token info', async (): Promise<void> => {
+  describe('validateTokenAccess', (): void => {
```
BlairCurrey, line 102: can you add a case where the validation fails?

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -52,6 +56,19 @@ function contextToRequestLike(ctx: HttpSigContext): RequestLike {
...
+  action: RequestAction,
+  identifier?: string
+): AccessItem {
+  return {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    type: type as any,
+    actions: [action],
+    identifier: identifier ?? undefined
+  }
+}
```
BlairCurrey, line 70: i tried to make this fully typesafe and it got brutal 😆 

Effectively the same, but perhaps a little cleaner?

```suggestion
function toOpenPaymentsAccess(
  type: AccessType,
  action: RequestAction,
  identifier?: string
): AccessItem {
  return {
    type: type,
    actions: [action],
    identifier: identifier ?? undefined
  } as AccessItem
}
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -90,41 +114,32 @@ export function createTokenIntrospectionMiddleware({
...
+        ctx.accessAction = AccessAction.ReadAll
+      } else if (
+        requestAction === AccessAction.List &&
+        (access as Access).actions.includes(AccessAction.ListAll)
+      ) {
+        ctx.accessAction = AccessAction.ListAll
+      } else {
+        ctx.accessAction = requestAction
+      }
+
```
BlairCurrey, line 142: Could we de-duplicate this logic some by returning RequestAction | undefined from `validateTokenAccess` instead of the `AccessItem`?

Looks like the only other place we use the `AccessItem` is to set the grant limits but I think the `tokenInfo` should work fine for that instead, if we wanted.

BlairCurrey: er... maybe not. I didn't fully understand the type of `tokenInfo`. Dont think we can use that for grant limit since we won't know which access item we're dealing with. Guess we could return both access and accessAction from `validateToken` if we cared to.

packages/auth/src/access/utils.ts
```diff
@@ -0,0 +1,47 @@
+import { isDeepStrictEqual } from 'util'
```
BlairCurrey, line 1: how did I not realize this was in the node standard library? Cool.

packages/auth/src/access/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (
+      typeof requestAccessItemValue === 'object' &&
+      !isDeepStrictEqual(
+        requestAccessItemValue,
+        restOfgrantAccessItem[key as keyof typeof restOfgrantAccessItem]
+      )
+    ) {
+      return false
+    } else if (
+      restOfRequestAccessItem[key as keyof typeof restOfRequestAccessItem] !==
```
BlairCurrey, line 39: could use `requestAccessItemValue`, since it was already initialized on L28
```suggestion
      requestAccessItemValue !==
```

BlairCurrey (approved):

packages/auth/src/accessToken/service.ts
```diff
@@ -81,6 +89,21 @@ async function introspect(
...
+        if (
+          !grantAccess.find((grantAccessItem) =>
+            compareRequestAndGrantAccessItems(
+              accessItem,
+              toOpenPaymentsAccess(grantAccessItem)
+            )
+          )
+        ) {
+          return undefined
+        }
```
mkurapov, line 104: to understand this correctly, in this implementation, the `access` will consist of an array with a single item of { `identifier` `access` `type` }. We then compare this object to the existing grantAccessItem. 
And even though `accessItem` will never have `limits`, `compareRequestAndGrantAccessItems` checks if all the other keys are equivalent, correct?

packages/auth/src/accessToken/service.ts
```diff
@@ -42,7 +48,8 @@ export async function createAccessTokenService({
   return {
     getByManagementId: (managementId: string) =>
       getByManagementId(managementId),
-    introspect: (tokenValue: string) => introspect(deps, tokenValue),
+    introspect: (tokenValue: string, access?: AccessItem[]) =>
```
mkurapov, line 51: will/should `access` ever be undefined?

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -81,7 +98,14 @@ export function createTokenIntrospectionMiddleware({
...
         tokenInfo = await tokenIntrospectionClient.introspect({
-          access_token: token
+          access_token: token,
+          access: [
+            toOpenPaymentsAccess(
+              requestType,
+              requestAction,
+              ctx.walletAddressUrl
+            )
+          ]
```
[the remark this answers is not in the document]
njlie: The GNAP Resource Server spec [specifies that this be an array](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers-05#name-token-introspection).

packages/auth/src/accessToken/service.ts
njlie: It's optional in the GNAP spec. It's not required in the OpenAPI spec for the `token-introspection` package, but we certainly _could_ make it required. Payments are pretty sensitive entities, I'd be in favor of making it required.

packages/auth/src/access/utils.ts
njlie: I just discovered it when completing this PR. I knew there surely was a library that could handle this but I wasn't expecting a node-native solution either!

packages/backend/src/open_payments/auth/middleware.ts
njlie: Looks like we can just return a filtered version of `access` with a single item from the introspection endpoint, and have the RS just assume that the lone item in that array is the relevant access.

Then the backend can just check if there is a `Read-All` or `List-All` action in the list if the request for a `Read` or `List` action, respectively.

[From the Resource Server Spec (emphasis mine)](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers-05#name-token-introspection):
> access (array of strings/objects):
REQUIRED. The access rights associated with this access token. This MUST be in the format described in the [Section 8](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-18#section-8) of [[GNAP](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-18)]. _**This array MAY be filtered or otherwise limited for consumption by the identified RS,**_ including being an empty array, indicating that the token has no explicit access rights that can be disclosed to the RS.

packages/auth/src/accessToken/service.ts
njlie: Yeah that's pretty much it.

packages/token-introspection/src/client/introspection.test.ts
njlie: Ended up removing it, actually

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -90,41 +109,29 @@ export function createTokenIntrospectionMiddleware({
...
-        }
-        return access.actions.find((tokenAction: AccessAction) => {
-          if (isActiveTokenInfo(tokenInfo) && tokenAction === requestAction) {
-            ctx.accessAction = requestAction
-            return true
-          }
-          return false
-        })
-      })
+      if (tokenInfo.access.length > 1) {
```
njlie, line 112: Whether or not we need this check depends on if it is decided that the token introspection endpoint will always return ONLY the request's matching access item in the response.

packages/backend/src/open_payments/auth/middleware.ts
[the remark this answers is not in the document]
mkurapov: Got it, makes sense!

packages/backend/src/open_payments/auth/middleware.ts
mkurapov: However, looking at

https://github.com/interledger/rafiki/pull/2788/files#diff-4c4c1504c8978d5fcd7a7bfa8210903c7dada6ec793ac0ad8e895fda2f0a5115R120-R122

since it looks like its "possible" to return multiple accesses, we should keep the original behavior of allowing multiple? Or we make `accessItem` required in the auth introspection route handler.

packages/auth/src/accessToken/service.ts
mkurapov: Edit, I think if we are doing [this](https://github.com/interledger/rafiki/pull/2788#discussion_r1690159011), ie returning the single relevant access item, we should make this required

packages/auth/src/access/utils.test.ts
```diff
@@ -0,0 +1,209 @@
...
+          assetScale: 2
+        }
+      }
+    }
+
+    expect(
+      compareRequestAndGrantAccessItems(
+        requestAccessItem,
+        toOpenPaymentsAccess(grantAccessItemSuperAction)
+      )
```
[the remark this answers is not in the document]
njlie: I think it works, updated just to be sure

packages/backend/src/open_payments/auth/middleware.ts
njlie: @mkurapov Here's what I decided - the introspection endpoint on the auth server will always return whatever access rights were presented in the request if the token is valid, no more, no less.

The backend service can then expect the response to have exactly one item since the middleware only ever presents one access during the introspection request.

packages/auth/src/accessToken/service.ts
```diff
@@ -68,21 +75,40 @@ async function getByManagementId(
...
-  tokenValue: string
-): Promise<Grant | undefined> {
+  tokenValue: string,
+  access?: AccessItem[]
+): Promise<{ grant: Grant; access: Access[] } | undefined> {
   const token = await AccessToken.query(deps.knex)
     .findOne({ value: tokenValue })
     .withGraphFetched('grant.access')
 
+  let foundAccessItem: Access | undefined
```
mkurapov, line 85: Does this need to be defined here or just in the L95 loop?

mkurapov (approved):
Looks good! just a minor comment

packages/backend/src/open_payments/auth/middleware.ts
mkurapov: @njlie makes sense to me 👍

mkurapov:
Feel free to re-request after fixing the test

packages/auth/src/access/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+      restOfgrantAccessItem[key as keyof typeof restOfgrantAccessItem]
+    ) {
+      return false
+    }
+  })
```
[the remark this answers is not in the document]
njlie: I decided to update this to compare the keys explicitly. I added a test in light of [this comment](https://github.com/interledger/rafiki/pull/2788#discussion_r1690592555) to ensure that only the requested access rights get returned, and it wouldn't work without explicit key comparison.

mkurapov (approved):
🚀

PR #2797, opened 2024-07-09
feat: added packet count telemetry metrics

### LINKED ISSUE #2734: Track count of prepare, fulfill, and reject packets
As part of Telemetry V2, we want to add counts, and amount for the number of prepare, fulfill, and reject packets.

These new metrics are:
* packet_count_fulfill
* packet_count_reject
* packet_count_prepare
* packet_amount_fulfill (treat the amount as we do now: we need to convert it to a base currency, see the current `collectTelemetryAmount` function for this)

All of these metrics will be collected at the "receiving" connector, meaning, a count is updated only after it has "reached" the connector that receives that particular type of packet. We should track all packets (we shouldn't, and really, can't, differentiate between packets for quoting and actual transfers).

##### Example 1:
image

After this, the counts would look like:

packet_count_prepare = 2
packet_count_fulfill = 2
packet_amount_fulfill = amount of fulfill 1 + amount of fulfill 2

##### Example 2:

image

After this, the counts would look like:

packet_count_prepare = 2
packet_count_fulfill = 1
packet_count_reject = 1
packet_amount_fulfill = amount of fulfill

JoblersTune:

- Added packet counting metrics for telemetry v2
- Updated the transactions_total metric to `packet_amount_fulfill` and included source metadata with the amount metric collection.
- ~~Moved the collection of all packet data over to the receiving node so that it can log a consistent match of incoming prepares to outgoing responses. This way, since telemetry is optional, we'll get a coherent reporting structure since one node can send coherent results in isolation.~~
- Collecting packet counts and packet amount values on the sending node. The values are now collected from the `handleIlpData` function which is called only on sending nodes, and gives us access to whether packets are unfulfillable so we don't need to count quote packets.
- Moved packet count and amount metrics out of middleware (since metrics should only be collected based on finalised values after all other middleware has run)
- ~~Cleaned up old telemetry code that is no longer needed after changes were made~~

Outside of this PR but related:
- Changes to the telemetry dashboard to be reviewed as well. I've created a test dashboard called TEST packet count which is viewable on Grafana. 

## Context

Working towards expanding telemetry for telemetry v2
- fixes #2734 

TODO:
- add tests
- Update docs
- Consider transaction count being on sending node - planning to track on outgoing and incoming

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


JoblersTune:
Last outstanding task is updating the tests

JoblersTune:
Are there any lifecycle tests for outgoing payments somewhere? I wanted to add in telemetry tests for the outgoing payment count but couldn't see where to put them?

packages/backend/src/accounting/service.ts
JoblersTune, line 1: This was here from the first telemetry implementation and it just got left here. We're not using asset scale and code here anymore so we don't need the LiquidityAccountAsset to expose them.

JoblersTune:
Please don't forget to look at the new dashboard additions as well please: https://rafikitelemetry.grafana.net/d/cdq2hn0w6frpcr/test-packet-count?orgId=1

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -60,6 +61,7 @@ describe('Incoming Payment Service', (): void => {
 
   afterAll(async (): Promise<void> => {
     await appContainer.shutdown()
+    jest.resetAllMocks
```
BlairCurrey, line 64: this function isnt being called (missing `()`) but I'm not sure if this is necessary anyways. We have the [tests configured](https://github.com/interledger/rafiki/blob/main/packages/backend/jest.config.js#L22) to [clear mocks ](https://jestjs.io/docs/configuration#clearmocks-boolean)between all tests already.

packages/documentation/src/content/docs/telemetry/overview.md
```diff
@@ -60,18 +62,31 @@ If an ASE does not provide the necessary exchange rate for a transaction, the te
 
 ## Instrumentation
 
-Rafiki currently has three counter metrics. All data points (counter increases and histogram records) are exported to collection endpoints at a configurable interval (default recommended to 15s).
+Rafiki currently has seven counter metrics. All data points (counter increases and histogram records) are exported to collection endpoints at a configurable interval (default recommended to 15s).
```
BlairCurrey, line 65: This is kinda prone to drift from the truth (count, type of metrics). Feel like we could make our lives easier and avoid potentially misleading people by removing some of the incidental details and rephrasing like:

```suggestion
Data points for all metrics (e.g. counter increases) are exported to collection endpoints at a configurable interval (default recommended to 15s).
```

Currently there are more than 7 metrics (my bad - forgot to include the latest one I added 😬 ) and they arent all counters (`ilp_pay_time_ms` is a histogram for example).

packages/backend/src/payment-method/ilp/connector/core/middleware/ilp-packet.ts
```diff
@@ -234,5 +235,34 @@ export function createIlpPacketMiddleware(
     ctx.assert(!ctx.body, 500, 'response body already set')
     ctx.assert(response.rawReply, 500, 'ilp reply not set')
     ctx.body = response.rawReply
+
+    if (ctx.services.telemetry && Number(prepare.amount)) {
+      ctx.services.telemetry?.incrementCounter('packet_count_prepare', 1, {
```
BlairCurrey, line 240: optional chaining not necessary since we checked `ctx.services.telemetry`
```suggestion
      ctx.services.telemetry.incrementCounter('packet_count_prepare', 1, {
```

BlairCurrey:
Regarding the `Packet count (by source)` viz, should we either:
- use the prepare amount for the total
OR
- sum the fulfill/reject amounts for the total

I think probably just the prepare amount? Currently we are summing fulfill/reject/prepare but the prepare include fulfill and reject as far as I can see. 

image

packages/backend/src/payment-method/ilp/connector/core/middleware/ilp-packet.ts
```diff
@@ -234,5 +235,34 @@ export function createIlpPacketMiddleware(
...
+            value,
+            assetCode: code,
+            assetScale: scale
+          },
+          {
+            description: 'Amount sent through the network',
+            valueType: ValueType.DOUBLE
+          }
+        )
+        ctx.services.telemetry?.incrementCounter('packet_count_fulfill', 1, {
```
BlairCurrey, line 258: optional chaining not necessary since we checked ctx.services.telemetry
```suggestion
        ctx.services.telemetry.incrementCounter('packet_count_fulfill', 1, {
```

packages/backend/src/payment-method/ilp/connector/core/middleware/ilp-packet.ts
```diff
@@ -234,5 +235,34 @@ export function createIlpPacketMiddleware(
...
+          description: 'Count of outgoing fulfill packets'
+        })
+      } else if (response.reject) {
+        ctx.services.telemetry?.incrementCounter('packet_count_reject', 1, {
```
BlairCurrey, line 262: optional chaining not necessary since we checked ctx.services.telemetry
```suggestion
        ctx.services.telemetry.incrementCounter('packet_count_reject', 1, {
```

BlairCurrey:
How will the existing visualizaitons be impacted? It looks like maybe the new `Packet count (by source)` replace the `Transaction Count (by source)` ones but then the `Value Sent Through network` line graphs will be updated to use the new (and `Total Transactions Per Second` would need the same)? I like the new stat visualizations but think the existing line graphs are still useful.

packages/backend/src/open_payments/payment/incoming/service.test.ts
JoblersTune: I did find this necessary otherwise the last system time reset was messing up my tests below and giving some weird NaN values to the date function. But when I included the reset it resolved that. I assumed it as because `afterEach` sets the time back to reality ` jest.useRealTimers()` but it doesn't reset the system time after calling `jest.useFakeTimers()
      jest.setSystemTime(incomingPayment.expiresAt)`?

JoblersTune:
> Regarding the `Packet count (by source)` viz, should we either:
> 
> * use the prepare amount for the total
>   OR
> * sum the fulfill/reject amounts for the total
> 
> I think probably just the prepare amount? Currently we are summing fulfill/reject/prepare but the prepare include fulfill and reject already it looks like. So I think we can just remove the last one here (and related graphs):
> 

I don't have strong feelings about whichever approach. 

I know Max was initially hoping we'd collect this to get a handle on packet loss through the network but this tricky when telemetry is optional.

Given that I assumed we would still be most interested in 1. the total actual number of packets moving through the network (prepare + reject + fulfill) and 2. the relationship between them (we expect prepare = fulfill + reject) and maybe good to see how many rejects we get vs fulfills as well. Of course some of the rejects are quote based as well.  

I guess I'm not exactly sure. 

Maybe we can just graph fulfill and reject and leave the prepare and totals out?

JoblersTune:
> How will the existing visualizaitons be impacted? It looks like maybe the new `Packet count (by source)` replace the `Transaction Count (by source)` ones but then the `Value Sent Through network` line graphs will be updated to use the new (and `Total Transactions Per Second` would need the same)? I like the new stat visualizations but think the existing line graphs are still useful.

Good question. This felt obvious when I started but less so now.
I think we need a call about this.
I'm gonna formulate some ideas and then let's rather discuss.

packages/backend/src/index.ts
```diff
@@ -292,6 +292,9 @@ export function initIocContainer(
       knex: await deps.use('knex'),
       accountingService: await deps.use('accountingService'),
       walletAddressService: await deps.use('walletAddressService'),
+      telemetry: config.enableTelemetry
+        ? await deps.use('telemetry')
```
koekiebox, line 296: Happy to see this. :-).

BlairCurrey:
> I don't have strong feelings about whichever approach.
> 
> I know Max was initially hoping we'd collect this to get a handle on packet loss through the network but this tricky when telemetry is optional.
> 
> Given that I assumed we would still be most interested in 1. the total actual number of packets moving through the network (prepare + reject + fulfill) and 2. the relationship between them (we expect prepare = fulfill + reject) and maybe good to see how many rejects we get vs fulfills as well. Of course some of the rejects are quote based as well.
> 
> I guess I'm not exactly sure.
> 
> Maybe we can just graph fulfill and reject and leave the prepare and totals out?

>  Given that I assumed we would still be most interested in 1. the total actual number of packets moving through the network (prepare + reject + fulfill) and 2.

Might need to take a step back and verify something... is this the correct way to count packets in [ilp-packet.ts](https://github.com/interledger/rafiki/pull/2797/files#diff-51b4d4b2e1a4d5514f9d18db014ac3856ff37d72f1ef7d9edccc30bb1f47c671)? 

- always incrementing preprare
AND
- maybe incrementing fulfill
OR
- maybe incrementing reject

So prepare is actually equal to the total. But based off your comment it sounds like prepare/fulfill/reject should be mutually exclusive. Is that how it should be? I guess that would make sense. My original comment  was taking the logic there at face value.

And just to elaborate on how its currently working, look at the graph. `fulfill` (yellow) `reject` (blue) almost equal the `prepare` (green). The difference is the loss. So that's everything we want. The total (orange) then sums these, which is equivalent to fulfill + reject + loss (the prepare) + fulfill + reject. Hence the double counting and recommendation to remove the Total (orange). But maybe we shouldnt be incrementing the prepare counter for every packet?

image

JoblersTune:
@BlairCurrey I take your point.

> ... based off your comment it sounds like prepare/fulfill/reject should be mutually exclusive.

This came to be because if you look at the issue, the initial suggestion was to collect prepare packets at the receiver node, and fulfill/reject packets at the sending node. But, because telemetry is optional, I pointed out that we might get a distorted view of the situation. Because if only one of the nodes involved has enabled telemetry we could see a whole bunch of prepares but no responses or vice versa. 

This is why I suggested collecting ALL the packet counts on the same node (the receiving node). However, maybe this makes collecting prepare packets redundant... Since we only send a response if we get a prepare, and I'm assuming we ALWAYS send a response when we get a prepare... Is that true? 

In this case, maybe 
1. Collecting prepare packet counts actually adds no value and we should drop it
2. We should collect prepare packet amounts because they are still part of the total number of packets flowing on the network (prepare + fulfill + reject = total)
3. We go back to the idea of collecting the number of prepare packets on the sending node and accept that these metrics (prepares vs. responses) will possibly be misaligned...  

I feel this always goes back to asking what the purpose of these metrics is. 
Do we want to have a number for marketing that says x number of packets were sent over the Interledger network in the last year.
Or do we want to estimate packet loss over the network - which is in a pure sense, impossible to do because telemetry is optional. But maybe having the asymmetric view still adds some other kind of value?

But I still want to make sure we're in alignment that the total number of ILP packets that traverse the Interledger network = prepare + reject + fulfill. Because these are all packets, they are all distinct packet types, and they all travel over the network.

packages/backend/src/payment-method/ilp/connector/core/controllers/client.ts
```diff
@@ -36,6 +36,6 @@ export function createClientController({
       expiresAt
     )
 
-    response.rawReply = await send(axios, outgoing, outgoingPrepare)
+    response.rawReply = await send(axios, outgoing, outgoingPrepare) // This is the sender's response from the receiver, but we don't collect here in case an error is thrown which will be caught by error handling middleware
```
BlairCurrey, line 39: remove comment? not sure how much sense this comment makes outside the context of this PR

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -85,8 +85,8 @@ export async function handleSending(
   const payEndTime = Date.now()
 
   if (deps.telemetry) {
-    deps.telemetry.incrementCounter('transactions_total', 1, {
-      description: 'Count of funded transactions'
+    deps.telemetry.incrementCounter('transactions_count', 1, {
+      description: 'Count of funded outgoing transactions'
```
BlairCurrey, line 89: can you elaborate on why we're renaming `transactions_total` -> `transactions_count` (I remember the intermediate name). Just for consistency with the other incoming transaction count? Just asking since now we'll have transactions_total (unused from this point on) and now this new one that is the same thing.

I guess its all still in the test pool of data so maybe it doesnt matter that much. More specific description definitely makes sense (leave the high level "transaction" more for the display).

packages/backend/src/payment-method/ilp/connector/core/controllers/stream.ts
```diff
@@ -23,9 +23,9 @@ export function createStreamController(): ILPMiddleware {
       return
     }
 
-    const moneyOrReply = streamServer.createReply(request.prepare)
+    const moneyOrReply = streamServer.createReply(request.prepare) // Creates a reply
```
BlairCurrey, line 26: im not sure what context the comment is adding here since its basically just repeating the method name.... add more context or remove?

BlairCurrey:
Is this how we intend to count? Matches my understanding from our previous call but wanted to verify.

```mermaid
sequenceDiagram
    Title: Happy Path
    participant A
    participant B

    note left of A: prepare=0
    note left of A: fulfill=0
    A->>A: ++prepare
    A->>B: Send prepare.
    note right of A: prepare=1
    B->>A: Send fulfill
    note right of A: prepare=1

    A->>A: ++fulfill
    note left of A: prepare=1
    note left of A: fulfill=1
```

BlairCurrey:
And then I guess this would be multi hop:

```mermaid
sequenceDiagram
    Title: Happy Path Multi-Hop
    participant A
    participant B
    participant C

    note left of A: prepare=0
    note left of A: fulfill=0
    A->>A: prepare++  
    A->>B: Send prepare.
    note right of A: prepare=1
    note right of A: fulfill=0

    B->>B: prepare++
    B->>C: Send prepare
    note right of B: prepare=2
    note right of B: fulfill=0

    C->>B: Send fulfill
    note right of B: prepare=2
    note right of B: fulfill=0

    B->>B: fulfill++
    B->>A: Send fulfill
    note right of A: prepare=2
    note right of A: fulfill=1

    A->>A: fulfill++
    note left of A: prepare=2
    note left of A: fulfill=2
```

BlairCurrey:
Max mentioned this for the loss which I think still seems right?

> Number of packets - (2 * (rejected + fulfilled))

Which is just `prepare` - `reject` - `fulfill`. (Number of packets = `prepare` + `fulfill` + `reject`. Expand the terms then add). Makes sense I think - each prepare should have a fulfill or reject.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
JoblersTune: Oops, I thought it was in the original issue, buty it's not. I'll revert it back to `transaction_total`

JoblersTune:
Thanks for the diagrams @BlairCurrey they match my understanding as well.

BlairCurrey (approved):

packages/backend/src/payment-method/ilp/connector/core/README.md
```diff
@@ -1,3 +1,43 @@
...
+- packet_count_prepare: Counts the prepare packets sent, collected before the middleware routes are executed.
+- packet_count_fulfill: Counts the fulfill packets received, collected after receiving a reply from the receiver.
+- packet_count_reject: Counts the reject packets received.
+- packet_amount_fulfill: Records the amount sent in fulfill packets.
+
+These metrics provide valuable insights, including potential packet loss.
+
+### Challenges with the Current Setup
+
+While `handleIlpData` is an effective location for telemetry collection, it has some limitations:
```
mkurapov, line 30: I would prefer not having function names in documentation since it becomes outdated quite quickly, I think we can just remove references to it and keep just the general explanation which you have

JoblersTune: The reason I've included function names is try and avoid any duplicated work when the next person needs to update this or change the metric location. Is there somewhere better for me to use function names and get really specific outside of the README?

packages/backend/src/payment-method/ilp/connector/core/telemetry.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export async function incrementAmount(
+  unfulfillable: boolean,
+  prepareAmount: string,
+  response: IlpResponse,
+  code: string,
+  scale: number,
+  telemetry: TelemetryService
+): Promise<void> {
+  if (!unfulfillable && Number(prepareAmount)) {
```
koekiebox, line 44: The two if statements could be on the same line?

koekiebox (changes requested):
Have a couple of comments.

packages/backend/src/payment-method/ilp/connector/core/telemetry.ts
```diff
@@ -0,0 +1,61 @@
+import { TelemetryService } from '../../../../telemetry/service'
+import { IlpResponse } from './middleware/ilp-packet'
+import { ValueType } from '@opentelemetry/api'
+
+export function incrementPreparePacketCount(
+  unfulfillable: boolean,
+  prepareAmount: string,
```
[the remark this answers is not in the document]
JoblersTune: @koekiebox  The prepare packet has the amount set as a string, so I was just staying true to that. It gives the function a simple signature then since we don't need to check if Number(prepare.amount) before calling this function since it was explicitly designed to abstract all this reasoning into one simple call.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -1161,6 +1161,79 @@ describe('OutgoingPaymentService', (): void => {
...
+      )
+      await expectOutcome(payment, {
+        accountBalance: 0n,
+        amountSent: payment.debitAmount.value,
+        amountDelivered: payment.receiveAmount.value,
+        incomingPaymentReceived: payment.receiveAmount.value,
+        withdrawAmount: 0n
+      })
+
+      expect(incrementTrxCounterSpy).toHaveBeenCalledTimes(1)
```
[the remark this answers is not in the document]
JoblersTune: Like so?

```suggestion
      expect(incrementTrxCounterSpy).toHaveBeenCalledWith('transactions_total', 1, {
        description: 'Count of funded transactions'
      })
```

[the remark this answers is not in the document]
JoblersTune: Done

packages/backend/src/payment-method/ilp/connector/core/telemetry.ts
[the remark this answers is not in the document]
JoblersTune: @mkurapov, do you have any feelings about this and which way you'd like me to have this implemented since Jason is on vacation this week.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -1161,6 +1161,85 @@ describe('OutgoingPaymentService', (): void => {
       return payment
     }
 
+    test('Telemetry Transaction Counter increments for COMPLETED transactions', async (): Promise<void> => {
```
mkurapov, line 1164: thanks for adding this

packages/backend/src/payment-method/ilp/connector/core/README.md
```diff
@@ -1,3 +1,47 @@
...
+
+### Why We Collect on the Sending Side
+
+The first decision in collecting this data was whether to do so on the sender's side or the receiver's side. We opted for the sender’s side to maintain consistency across our metrics, particularly for calculating average transaction amounts and times, which are tied to the outgoing payment flow in the Open Payments (OP) layer. By collecting metrics from the same perspective, we ensure alignment with how other metrics (like transaction completion times) are captured.
+
+We also considered collecting metrics on both the sending and receiving sides, capturing metrics when prepare packets were received by the receiver and when fulfill or reject responses were received by the sender. However, this could lead to unreliable metrics, as telemetry is optional and might not be enabled on all nodes.
+
+### Why We Chose to collect metrics before and after the sendering node's middlware routes
```
mkurapov, line 15:
```suggestion
### Why We Chose to collect metrics before and after the sending node's middleware routes
```

packages/backend/src/payment-method/ilp/connector/core/README.md
```diff
@@ -1,3 +1,47 @@
...
+
+### Challenges with the Current Setup
+
+Sender-Side Limitation: Currently, metrics are only collected on the sender side. This is adequate for now but does not capture data from connecting nodes, which we plan to address in future implementations when multi-hop support is added.
+
+In a scenario where a sender (node A) wants to send money to the receiver (node C), and this transaction takes place via an intermediray (node B).
+
+sender (node A) --> connector (node B) --> receiver (node C)
+
+With the cuirrent metric collection location, we only collect packet count and amount infomation from the sender (at node A) and we miss the packets and amounts forwarded by the connecting node (node B). Ideally we'd like to collect information from the sender and the connector nodes once multihop routing has been implemented.
```
mkurapov, line 36:
```suggestion
With the current metric collection location, we only collect packet count and amount infomation from the sender (at node A) and we miss the packets and amounts forwarded by the connecting node (node B). Ideally we'd like to collect information from the sender and the connector nodes once multihop routing has been implemented.
```

packages/backend/src/payment-method/ilp/connector/core/test/telemetry.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import {
+  incrementPreparePacketCount,
+  incrementFulfillOrRejectPacketCount,
+  incrementAmount
+} from '../telemetry'
+import { IlpResponse } from '../middleware/ilp-packet'
+import { IlpFulfillFactory, IlpRejectFactory } from '../factories'
+import { ValueType } from '@opentelemetry/api'
+
+describe('Connector Core Telemetry', () => {
```
mkurapov, line 15: we should add a few tests for handling non-positive cases (e.g., "... doesn't collect metrics for unfulfillable packets")

packages/documentation/src/content/docs/telemetry/overview.md
```diff
@@ -7,8 +7,10 @@ title: Overview
 The objective of the telemetry feature is to gather metrics and establish an infrastructure for visualizing valuable network insights. The metrics we at the Interledger Foundation collect include:
 
 - The total amount of money transferred via packet data within a specified time frame (daily, weekly, monthly).
-- The number of transactions from outgoing payments that have been at least partially successful.
+- The number of transactions that have been at least partially successful.
+- The number of ILP packets flowing through the network.
 - The average amount of money held within the network per transaction.
+- The avergae time it takes for an outgoing payment to complete.
```
mkurapov, line 13:
```suggestion
- The average time it takes for an outgoing payment to complete
```

mkurapov (commented):
Looking good, minor comments (main one about some tests)

packages/backend/src/payment-method/ilp/connector/core/README.md
JoblersTune: Oops. Clearly got tired.

packages/backend/src/payment-method/ilp/connector/core/test/telemetry.test.ts
JoblersTune: So should I just pass in `unfulfillable` as `true` and check it doesn't call incrementCounter?

JoblersTune: I did my best, have a look and see if the new tests meet this criteria :)

mkurapov:
@JoblersTune just need to fix the build and good to go

JoblersTune:
@melissahenderson 

This PR has been waiting a while, so I nearly forgot there was a doc change here too. 

I'll just paste the text here for your reference and remove the file changes in this PR (telemetry overview page)


```
---
title: Overview
---

## Purpose

The objective of the telemetry feature is to gather metrics and establish an infrastructure for visualizing valuable network insights. The metrics we at the Interledger Foundation collect include:

- The total amount of money transferred via packet data within a specified time frame (daily, weekly, monthly).
- The number of transactions that have been at least partially successful.
- The number of ILP packets flowing through the network.
- The average amount of money held within the network per transaction.
- The average time it takes for an outgoing payment to complete

We aim to track the growth of the network in terms of transaction sizes and the number of transactions processed. Our goal is to use these data for our own insights and to enable [Account Servicing Entities](/reference/glossary#account-servicing-entity) (ASEs) to gain their own insights.

## Privacy and Optionality

Privacy is a paramount concern for us. Rafiki's telemetry feature is designed to provide valuable network insights without violating privacy or aiding malicious ASEs. For more information, please [read the privacy docs](/telemetry/privacy).

The telemetry functionality is currently enabled by default on test (non-livenet) environments, i.e. any environment that is not dealing with real money. When active, it transmits metrics to the "testnet" collector. In the future, those ASEs operating in a production "livenet" environment (real money) will be able to opt-in to sharing their metrics with a "livenet" collector. Regardless of environment, Account Servicing Entities (ASEs) can also opt-out of telemetry completely.

## Architecture

The architecture of the telemetry feature is illustrated below:

![Telemetry architecture](/img/telemetry-architecture.png)

## OpenTelemetry

We have adopted [OpenTelemetry](https://opentelemetry.io/) to ensure compliance with a standardized framework that is compatible with a variety of tool suites. This allows clients to use their preferred tools for data analysis, while Rafiki is instrumented and observable through a standardized metrics format.

## Telemetry ECS Cluster

The Telemetry Replica service is hosted on AWS ECS Fargate and is configured for availability and load balancing of custom ADOT (AWS Distro for Opentelemetry) Collector ECS tasks.

When ASEs opt for telemetry, metrics are sent to our Telemetry Service. To enable ASEs to build their own telemetry solutions, instrumented Rafiki can send data to multiple endpoints. This allows the integration of a local [Otel collector](https://opentelemetry.io/docs/collector/) container that can support custom requirements. Metrics communication is facilitated through [gRPC](https://grpc.io/).

## Otel SDK - Rafiki Instrumentation

The Opentelemetry SDK is integrated into Rafiki to create, collect, and export metrics. The SDK integrates seamlessly with the OTEL Collector.

## Prometheus - AMP

We use Amazon's managed service for Prometheus (AMP) to collect data from the Telemetry cluster.

**Note**: AMP offers limited configuration options and cannot crawl data outside of AWS. This limitation led us to adopt a push model, using prometheusRemoteWrite, instead of a pull model. For future development, we may consider hosting our own Prometheus.

## Grafana - Grafana Cloud

Grafana Cloud is used for data visualization dashboards. It offers multiple tools that extend Prometheus Promql.

**Note**: We initially used Amazon hosted Grafana, but it did not meet our needs for embedding dashboards. Grafana Cloud offers a feature called “Public dashboards”, which allows us to share dashboards. However, embedding may still pose a challenge.

## Exchange Rates

For telemetry purposes, all amounts collected by instrumented Rafiki should be converted to a base currency.

**Privacy Reasoning**: If only two ASEs are peered over a non-USD currency and we collect data in that currency, it would be easy to determine the volumes moved between those two ASEs. To maintain privacy, we convert all amounts to a base currency.

If an ASE does not provide the necessary exchange rate for a transaction, the telemetry solution will still convert the amount to the base currency using external exchange rates. A Lambda function on AWS retrieves and stores these external exchange rates. It is triggered by a daily CloudWatch event and stores the rates in a public S3 Bucket. The S3 Bucket does not have versioning, and the data is overwritten daily to further ensure privacy.

## Instrumentation

Data points for all metrics (e.g. counter increases) are exported to collection endpoints at a configurable interval (default recommended to 15s).

Currently collected metrics:

- `transactions_total` - Counter metric
  - Description: “Count of funded outgoing transactions”
  - This counter metric increases by 1 for each successfully funded outgoing payment resource.
- `packet_count_prepare` - Counter metric
  - Description: “Count of prepare packets that are sent”
  - This counter metric increases by 1 for each prepare packet that is sent.
- `packet_count_fulfill` - Counter metric
  - Description: “Count of fulfill packets”
  - This counter metric increases by 1 for each fulfill packet that is received.
- `packet_count_reject` - Counter metric
  - Description: “Count of reject packets”
  - This counter metric increases by 1 for each reject packet that is received.
- `packet_amount_fulfill` - Counter metric
  - Description: “Amount sent through the network”
  - This amount metric increases by the amount sent in each ILP packet.
- `transaction_fee_amounts` - Counter metric
  - Description: “Fee amount sent through the network”.
  - This fee amount metric increases by the (amount sent minus amount received) for an outgoing payment.
- `ilp_pay_time_ms` - Histogram metric
  - Description: “Time to complete an outgoing ILP payment”
  - This histogram metric records the time taken to make an ILP payment.

**Note**: The current implementation only collects metrics on the SENDING side of a transaction. Metrics for external open-payments transactions RECEIVED by a Rafiki instance in the network are not collected.
```

melissahenderson:
@JoblersTune Thank you! I'll get the updates worked in to our other branch.

mkurapov (approved):

koekiebox (approved):
🚀

PR #2805, opened 2024-07-11
fix: fix incorrect loading of admin ui settings page

### LINKED ISSUE #2793: [BUG] The settings page is not loading for the Admin UI
## Bug Report

**Describe the bug**

The Admin UI settings page is not loading correctly. When you click the Settings menu item you get a 404 error. This is because the browser is getting stuck at the Kratos endpoint and not following the Kratos redirect back to the settings page in the Admin UI with the flow ID appended. You can access the settings page by manually reloading in the browser when you see the 404, or by typing the url directly into the search bar. 

**To Reproduce**

Log in to the Admin UI and navigate to the settings page. The page redirects and the browser gets stuck at the Kratos self service settings endpoint, instead of handling the redirects correctly. 
 

**Expected behavior**

The settings page should redirect to the Kratos self service settings endpoint and be redirected back to the Admin UI settings page with a flow ID, without any interaction from a user (right now this requires a manual reload from the browser).

 - OS: [e.g. Ubuntu]
 - Version [e.g. 20.04LTS]
 - Node version [e.g. 18.16.0]
 - pnpm version [e.g. 8.6.7]

**Screenshots**

image

**Additional context**

It seems like this is a Remix issue, and can be resolved by using `redirectDocument` instead of `redirect`
See https://remix.run/docs/en/main/utils/redirectDocument
It seems to be that Remix handles reloading after redirects differently depending on the path which is perhaps why this bug did not appear for local development work.

JoblersTune:

- This PR addresses a bug in the Admin UI where the settings page fails to load correctly, resulting in a 404 error. The issue arises due to the browser getting stuck at the Kratos endpoint and not following the redirect back to the settings page in the Admin UI with the flow ID appended. This bug is resolved by replacing the redirect function with redirectDocument to ensure the correct handling of document-level redirects. In the settings.tsx page I'm simply swapping `redirect` with `redirectDocument`. 

We are using a nightly build to test if this resolves issues in possible production deployments. 

## Context

- fixes #2793 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/frontend/app/routes/settings.tsx
+++ b/packages/frontend/app/routes/settings.tsx
@@ -1,4 +1,8 @@
-import { json, redirect, type LoaderFunctionArgs } from '@remix-run/node'
+import {
+  json,
+  redirectDocument,
+  type LoaderFunctionArgs
+} from '@remix-run/node'
 import { uuidSchema } from '~/lib/validate.server'
 import {
   isUiNodeInputAttributes,
@@ -39,7 +43,7 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
 
     return { responseData }
   } else {
-    return redirect(
+    return redirectDocument(
       `${variables.kratosBrowserPublicUrl}/self-service/settings/browser`
     )
   }
--- a/packages/frontend/app/routes/auth.login.tsx
+++ b/packages/frontend/app/routes/auth.login.tsx
@@ -1,4 +1,8 @@
-import { json, type LoaderFunctionArgs, redirectDocument, } from '@remix-run/node'
+import {
+  json,
+  type LoaderFunctionArgs,
+  redirectDocument
+} from '@remix-run/node'
 import { uuidSchema } from '~/lib/validate.server'
 import { isUiNodeInputAttributes } from '@ory/integrations/ui'
 import type { UiContainer } from '@ory/client'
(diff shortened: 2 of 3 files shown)


packages/frontend/app/routes/settings.tsx
```diff
@@ -39,7 +43,7 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
 
     return { responseData }
   } else {
-    return redirect(
+    return redirectDocument(
```
mkurapov, line 46: is this the only place this is needed?

mkurapov (approved):

packages/frontend/app/routes/settings.tsx
golobitch: We only found issue with the settings page.

golobitch (dismissed):

packages/frontend/app/routes/settings.tsx
JoblersTune: @mkurapov  is actually a very good question. It made me realise that we need to make improvements on catching direct references to the login page and account recovery page as well! Both were already redirecting to the kratos endpoint when the `login` / `Forgot password?` buttons were clicked but weren't handling direct navigation optimally and should also be using `redirectDocument` if they are navigated to directly instead of just throwing an error.

I'm going to include an additional commit now that covers this as well.

golobitch (approved):

Issue #2799, opened 2024-07-09
Update telemetryService counter collection


mkurapov:
### Context

Originating from [comment](https://github.com/interledger/rafiki/pull/2787#discussion_r1669159473).

We want to simplify the telemetryService interfaces, and abstract away some of the details that should live only within the service. 

For example, in `getOrCreateMetric` when we increment the counter, we (should always) pass in the counter attributes object with `source`: 
```
{
    source: deps.telemetry.getInstanceName()
} 
```

All this can just be abstracted away in a new function like `incrementCounter`, that will end up handling the `counter.add` call itself, always adding `source` (or other attributes we want), for all of the counter metrics.

Additionally, we can also have `incrementCounterWithAmount` that will by default take care of the privacy preserving calls, meaning we don't need to expose the privacy functions outside of the telemetry service itself.

PR #2807, opened 2024-07-12
refactor(backend): telemetry service interface

BlairCurrey:

- abstracts counter/histogram metric recording behind new `incrementCounter`, `incrementCounterWithAmount`, `recordHistogram`.
  -  in addition to doing the counter `.add` method, `incrementCounterWithAmount` handles the amount conversion and privacy obfuscation
- removes/makes private methods no longer used due to above refactor, leaving telemetry service interface with only: `shutdown`,  `incrementCounter`, `incrementCounterWithAmount`, and `recordHistogram` methods

## Context

Fixes: https://github.com/interledger/rafiki/issues/2799

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/telemetry/service.ts
```diff
@@ -6,20 +6,28 @@ import {
...
-  ): Promise<bigint | ConvertError>
+  shutdown(): Promise<void>
+  incrementCounter(
+    name: string,
+    amount: number,
+    attributes?: Record<string, unknown>
+  ): void
+  incrementCounterWithAmount(
+    name: string,
+    convertOptions: Pick<ConvertOptions, 'sourceAmount' | 'sourceAsset'>,
```
JoblersTune, line 23:
```suggestion
    amount: Pick<ConvertOptions, 'sourceAmount' | 'sourceAsset'>,
```

Surely we want to make it clear that this is an amount coming in here?

JoblersTune:
I'm also thinking about the wording `incrementCounterWithAmount`, we're always incrementing a counter by an amount. Maybe we could make it clearer by saying something like `incrementCounterWithValue` or `incrementTransactionValueCounter` or something else.

JoblersTune:
Other than that LGTM, I'll just need to make a few tweaks with the introduction of my PR here as well. But now that I've seen all the changes that shouldn't be too big of a job :)

packages/backend/src/telemetry/service.ts
BlairCurrey: yeah this should have been `amount`. just for extra context, an underlying call (`convertAmount`) uses `convertOptions` which is where this originally came from. the `incrementCounterWithAmount` implementation uses `amount` but i forgot to change this one.

BlairCurrey:
> I'm also thinking about the wording `incrementCounterWithAmount`, we're always incrementing a counter by an amount. Maybe we could make it clearer by saying something like `incrementCounterWithValue` or `incrementTransactionValueCounter` or something else.

I somewhat agree but I'm struggling to come up with something better because `amount` is what we call an object like this, which is what `Amount` in `incrementCounterWithAmount` is referring to:

```ts
interface Amount {
  value: string
  assetCode: string
  assetScale: number
}
```

So I think part of the problem is we just use a generic name for our specific concept of the above `Amount`. But here we are.

This is why I don't really like `incrementCounterWithValue` - because we need to pass in something like `Amount` (with the asset info), not just the string/number/bigint value. In fact perhaps `value` would be a better name for the argument in `incrementCounter` to help distinguish it. I wouldn't want to specify that it's transaction specific either (such as `incrementTransactionValueCounter`), because this method should work with any sort of `Amount` (fees, limits, etc).

But still I don't love it as-is. `incrementCounter` takes an `amount: number` but `incrementCounterWithAmount` takes a this different kind of amount (yuck).

And in this case the amount in `incrementCounterWithAmount` is not actually `Amount`, but `{ sourceAmount: bigint,  sourceAsset: Asset }`. This was driven by the underlying convert call but I think not well-named (`amount.sourceAmount` kinda sucks).

I'm warming up to something more like this. What do you think?

```ts
export interface TelemetryService {
  incrementCounter(
    name: string,
    value: number,
    attributes?: Record<string, unknown>
  ): void
  incrementCounterWithAmount(
    name: string,
    // value could be string or number instead. not sure which makes the most sense.
    amount: { value: bigint, assetCode: string, assetScale, number },
    attributes?: Record<string, unknown>
  ): Promise<void>
 }
```

Or does it still seem unclear because we're always increment with an amount (in the generic sense)? I would argue we need a better name for our amount like `AssetAmount` or something but this introduces a new name for the same thing we have all over the place which I don't like.

packages/backend/src/telemetry/service.ts
```diff
@@ -67,17 +75,16 @@ class TelemetryServiceImpl implements TelemetryService {
...
 
-    deps.collectorUrls.forEach((url) => {
-      const metricExporter = new OTLPMetricExporter({
-        url: url
-      })
-
-      const metricReader = new PeriodicExportingMetricReader({
-        exporter: metricExporter,
-        exportIntervalMillis: deps.exportIntervalMillis ?? 15000
+    this.meterProvider = new MeterProvider({
```
mkurapov, line 78: we can delete the setting of `this.meterProvider` on L74 now

packages/backend/src/tests/telemetry.ts
```diff
@@ -36,22 +31,14 @@ export class MockRatesService implements RatesService {
...
-  public getOrCreateHistogramMetric(
-    _name: string,
-    _options?: MetricOptions | undefined
-  ): Histogram<Attributes> {
-    return mockHistogram
-  }
+
+  incrementCounter(): void {}
+  async incrementCounterWithAmount(): Promise<void> {}
+  recordHistogram(): void {}
```
mkurapov, line 37: we can also get rid of the `getBaseAssetCode` & `getBaseAssetCode` here as well?

mkurapov:
I think `incrementCounterWithAmount` is ok based on the signature, maybe a (long) alternative is `incrementCounterWithTransactionAmount`?

Likewise, we can also have the `incrementCounter` function _not_ take in a value: number, since incrementing counter implies its going up by one? Then we can have a private method just use '1' under the hood when doing the actual metric incrementation. 

Not too big of a deal though.

JoblersTune:
I was also leaning toward `incrementCounterWithTransactionAmount` then. It's better to have something long but clear in my opinion.

BlairCurrey:
@JoblersTune @mkurapov updated to `incrementCounterWithTransactionAmount`. 

We can leave it there if we like, although I still wonder if this is making the function name more opinionated (that the amount is a transaction amount) than the actual function body does (and should). I mean shouldn't this same method be used for other amounts, such as fees (which might actually be right around the corner)? In which case the call would be something like:

```ts
incrementCounterWithTransactionAmount('fees_amount', feeAmount)
```

I think the function should signify this is _our_ concept of amount and avoid signify its for a specific use-case. The `incrementCounterWithAmount` name doesnt signify this currently, but the whole function signature (with arg type) does. Again, I wish we had a less generic term than "amount" for the value/asset pair concept. `AssetAmount` maybe, which would make this `incrementCounterWithAssetAmount`, but maybe that's a larger issue.

> Likewise, we can also have the incrementCounter function not take in a value: number, since incrementing counter implies its going up by one? Then we can have a private method just use '1' under the hood when doing the actual metric incrementation.

Didn't do this one... our `incrementCounter` is calling `counter.add` under the hood so I think we should allow passing in whatever we need to. We are using this currently in `incrementCounterWithTransactionAmount` and could conceivable use it outside the service via `incrementCounter` as well. Also seems incongruent to assume 1 in `incrementCounter` but then not `incrementCounterWithTransactionAmount` IMO. I agree there is some un-intuitiveness but for me this traces back to the opentelemetry api. I don't think "incrementing" (our word choice) implies +1, but I do think a "counter" (their name) does - and obviously there is nothing we can do about that. Perhaps if anything we change our name from `incrementCounter` -> `addToCounter` (and similar).

packages/backend/src/telemetry/service.ts
BlairCurrey: whoops, yes. good catch

packages/backend/src/tests/telemetry.ts
BlairCurrey: yup, removed

JoblersTune:
@BlairCurrey wouldn't we use a different exposed function for fees? Since I assume we don't need to do privacy stuff on fees, only asset conversions. In which case `incrementCounterWithTransactionAmount` and `incrementCounterWitFeesAmount` would arguably add more clarity?

mkurapov:
> Didn't do this one... our incrementCounter is calling counter.add under the hood so I think we should allow passing in whatever we need to. 

Makes sense, I think that's fine. 

@BlairCurrey @JoblersTune for privacy preserving amounts, we can have
`incrementCounterWithPrivacyTransactionAmount` or just add a `preservePrivacy` arg into `incrementCounterWithTransactionAmount `

BlairCurrey:
> @BlairCurrey wouldn't we use a different exposed function for fees? Since I assume we don't need to do privacy stuff on fees, only asset conversions. In which case `incrementCounterWithTransactionAmount` and `incrementCounterWitFeesAmount` would arguably add more clarity?

That's a fair point... I feel like an new arg (like @mkurapov suggested) might work well since everything else in the function would be the same (doesnt really add any complicated logic and will re-use the conversion stuff). But I could also see a new explicit method if the difference ends up being more than that... I guess we can roll with this for now and cross that bridge when we come to it.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -85,21 +85,14 @@ export async function handleSending(
...
-      .getOrCreateHistogramMetric('ilp_pay_time_ms', {
-        description: 'Time to complete an ILP payment'
-      })
-      .record(payDuration, {
-        source: deps.telemetry.getInstanceName()
-      })
+    deps.telemetry.recordHistogram('ilp_pay_time_ms', payDuration, {
+      description: 'Time to complete an ILP payment'
+    })
   }
```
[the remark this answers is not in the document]
BlairCurrey: they arent promises 😄. the underlying add/record sdk methods are synchronous.

`incrementCounterWithTransactionAmount` returns a promise due to the amount conversion.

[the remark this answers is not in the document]
mkurapov: Ah, that's right. LGTM now

mkurapov (approved):

PR #2816, opened 2024-07-18
feat(localenv): add trace collection (with Tempo) in local playground

### LINKED ISSUE #2808: Configure Rafiki to be able to track traces
### Context

Adding [traces](https://opentelemetry.io/docs/concepts/signals/traces/) to Rafiki will give us useful insights into different parts of the code, and how each operation runs, like network requests, PSQL queries, and particular sections of code (e.g. ILP payments). 

This is an example trace from Grafana:

Image

In the above, each block represents "spans" and corresponding nested "spans" that show different kind of attributes like what the operation was, metadata, and how long each operation took.

These spans can be created manually within the code, but also automatically. 

There are a lot of auto-instrumentation packages available for us to use (PSQL auto-instrumentation, GraphQL, Koa etc), which will enable tracing (mostly) out of the box. 

Since we have added a telemetry stack within the local playground in #2800, we can also add an service that ingests traces via the OpenTelemetry APIs. This service is [Grafana Tempo](https://grafana.com/oss/tempo/) , and it will take traces sent from the Rafiki backend, and be able to serve as a data source for the Grafana frontend.

### Tasks
- [ ] Add option to collect traces in Rafiki (under an env flag)
- [ ] Add Grafana Tempo to local playground stack
- [ ] Add basic auto-instrumentation packages (GraphQL, PSQL, HTTP)

mkurapov:

- Adds the ability to collect traces in Rafiki
- Adds trace auto-instrumentation for HTTP requests, our GraphQL Admin API and Postgres.
- Traces are pushed to the open telemetry collector, which proceeds to push traces to Grafana Tempo. These then are able to be visualized in Grafana:

Screenshot 2024-07-18 at 15 55 10

I added a small panel in the example Grafana dashboard that shows some of the longer traces than can be explored.

## Context

Fixes #2808 

## Checklist

- [ ] Tests added/updated
- [x] Documentation added


packages/backend/src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+tracerProvider.register()
+
+registerInstrumentations({
+  tracerProvider,
+  meterProvider,
+  instrumentations: [
+    new UndiciInstrumentation(),
```
mkurapov, line 73: Auto-instruments node `fetch` module. Useful for us to track requests made with the open payments client, since it uses fetch under the hood.

packages/backend/src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    new HttpInstrumentation(),
+    new PgInstrumentation(),
+    new GraphQLInstrumentation({
+      mergeItems: true,
+      ignoreTrivialResolveSpans: true,
+      ignoreResolveSpans: true
+    })
```
mkurapov, line 80: UndiciInstrumentation: Auto-instruments  node `fetch` module. Useful for us to track requests made with the open payments client, since it uses fetch under the hood.

HttpInstrumentation: Auto-instruments http and https modules. Useful to track client requests made with axios and server requests.

PgInstrumentation: auto-instruments all of our postgres queries.

GraphQLInstrumentation: auto-instruments our GraphQL APIs, particularly resolvers. The settings are to reduce the amount of noise we get. Otherwise, we see spans for every single field that we resolve. We can adjust if necessary.

packages/backend/src/telemetry/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+if (Config.enableTelemetryTraces) {
+  for (const url of Config.openTelemetryTraceCollectorUrls) {
+    const traceExporter = new OTLPTraceExporter({
+      url
+    })
+
+    tracerProvider.addSpanProcessor(new BatchSpanProcessor(traceExporter))
+  }
+}
```
mkurapov, line 65: seperate config variables, to differentiate between metrics and traces. (IMO we want metrics on by default, but not traces).

packages/backend/Dockerfile.prod
```diff
@@ -59,4 +59,4 @@ COPY --from=builder /home/rafiki/packages/backend/dist ./packages/backend/dist
 COPY --from=builder /home/rafiki/packages/token-introspection/dist ./packages/token-introspection/dist
 COPY --from=builder /home/rafiki/packages/backend/knexfile.js ./packages/backend/knexfile.js
 
-CMD ["node", "/home/rafiki/packages/backend/dist/index.js"]
+CMD ["node", "-r", "/home/rafiki/packages/backend/dist/telemetry/index.js", "/home/rafiki/packages/backend/dist/index.js"]
```
mkurapov, line 62: Because auto-instrumentation essentially needs to wrap over multiple modules, everything needs to be loaded as the very first thing before app startup:

https://opentelemetry.io/docs/languages/js/getting-started/nodejs/#setup

otherwise, auto-instrumentation will not wok.

mkurapov: We could just turn this CMD call `pnpm start` or something, but maybe its ok like this?

BlairCurrey: just curious, what happens if we dont include the `--require` option? some error? silently fails? Or did it just coincidentally work before?

dont have a very strong opinion on the pnpm start command vs this but leaning towards this. since we may only want to start it this way in this context as opposed to Dockerfile.dev, from host machine, etc. Could always have different start commands in that case I guess though.

mkurapov: It just "fails" silently, and doesn't generate traces. For example, if the pg module gets loaded before this postgres auto instrumentation, it just won't pick up those traces. 

If you uncomnent the debugger in the telemetry file, you'd be able to see those warnings.

packages/backend/src/telemetry/index.ts
BlairCurrey: We are always instrumenting these, regardless of `enableTelemetryTraces`. Is that OK? I mean the important thing is that we're not sending them anywhere, so maybe it's fine. But I wonder if this could have a performance impact or cause any unnecessary concern.

BlairCurrey (approved):
Looks like there are merge conflicts with the telemetry service refactor but generally looks good to me. Spun it up and viewed the traces locally. Just had the one open-ended comment here: https://github.com/interledger/rafiki/pull/2816#discussion_r1686754993

Ping me when ready for re-review

packages/backend/src/telemetry/index.ts
mkurapov: Although there shouldn't be much of a difference at all, it would probably be safer to put this behind the flag. 

I moved the code behind the flags, tested with them enabled & disabled, and everything seems to work. It also works if we make telemetry non optional and try to increment a counter while everything is disabled - it just does a noop.

packages/backend/src/telemetry/index.ts
```diff
@@ -0,0 +1,80 @@
...
+import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
+import {
+  BatchSpanProcessor,
+  NodeTracerProvider
+} from '@opentelemetry/sdk-trace-node'
+import { registerInstrumentations } from '@opentelemetry/instrumentation'
+import { UndiciInstrumentation } from '@opentelemetry/instrumentation-undici'
+
+// debug logger:
+// diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG)
```
njlie, line 23: nit: cleanup

njlie (approved):
LGTM, just some dangling comments to be cleaned up. Was able to get the trace logs in the Grafana dashboard on my machine as well.

packages/backend/src/telemetry/index.ts
mkurapov: I was thinking about this one, but its actually quite useful to keep for seeing whether the telemetry service is working at all, particularly the connection between the backend service and the collector itself. Otherwise need to dig into docs to find this. 

might be a rare exception to keeping comments in

Issue #2798, opened 2024-07-09
[BUG] Admin UI displays liquidity in one asset scale and withdrawal input in a different scale


JoblersTune:
## Bug Report

**Describe the bug**

When you view liquidity information it is displayed in asset scale two but when withdrawing liquidity it converts back to the original asset scale without warning the user.

**Expected behavior**

Either make it clear to the user that the asset scale has changed, or even better, make the asset scale consistent. 

**Screenshots**

![image](https://github.com/interledger/rafiki/assets/77615158/ff41c439-3534-4ce2-98ff-ba9d1a87b33a)
![image](https://github.com/interledger/rafiki/assets/77615158/f75ab702-da59-444c-94b4-aff5ece97b8b)
![image](https://github.com/interledger/rafiki/assets/77615158/fe62c28e-0805-4aff-b3db-7f639c05d796)

JoblersTune:
@Emanuel-Palestino that would be lovely. Thanks for the offer of help!

> The expected result based on your screenshots is 9999999900.60 USD right? So the problem is that it is ambiguous on what scale the user is withdrawing or depositing, correct? 
Yes, exactly!

If you have any more questions please feel free to ask.

JoblersTune:
For some additional context that might be helpful:
- I think this is a problem comes up for withdrawing or depositing liquidity for peers, assets, incoming and outgoing payments, and wallet addresses.
- The bug shows up whenever asset scale is not set to 0.
- The approach may be to use the loaders to get asset details and pass them into the liquidity dialog so that we can at least display 1 = $0.01 for instance.
- Also open to other suggestions if you have a different approach.

Emanuel-Palestino:
I opened a pull request #2817 implementing your approach c:

JoblersTune:
Fantastic @Emanuel-Palestino, thank you. I am out of office but will review on Wednesday when I am properly back at my desk.

PR #2817, opened 2024-07-18
fix(frontend): It is ambiguous on what scale is the withdrawal and deposit input

Emanuel-Palestino:

- Add an asset prop in liquidity dialog component to receive the asset info.
- Get the asset info on the loader of withdraw and deposit components using the `getPeer` api method.
- Get the asset info on the loader of withdraw and deposit components using the `getAssetInfo` api method.

## Context

In the rafiki Admin UI, the liquidity information of peer and asset details are displayed using the current asset scale. When the user wants to deposit or withdraw liquidity, the amount introduced is handle using a scale of 0. So it is ambiguous on what scale the user is withdrawing or depositing.

Fixes: #2798

### DIFF
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -30,7 +30,10 @@ export const LiquidityDialog = ({
     const userInput = e.target.value
     const scaledInput = parseFloat(userInput) * Math.pow(10, asset.scale)
     const integerScaledInput = Math.floor(scaledInput)
-    if (scaledInput !== integerScaledInput) {
+    if (scaledInput < 0) {
+      const error = 'The amount should be a positive value'
+      setErrorMessage(error)
+    } else if (scaledInput !== integerScaledInput) {
       const error = 'The asset scale cannot accomodate this value'
       setErrorMessage(error)
     } else {
@@ -63,18 +66,17 @@ export const LiquidityDialog = ({
                 {title}
               </Dialog.Title>
               <div className='mt-2'>
+                <Input
+                  required
+                  type='number'
+                  name='displayAmount'
+                  label='Amount'
+                  onChange={handleChange}
+                  addOn={asset.code}
+                  step='any'
+                  error={errorMessage}
+                />
                 <Form method='post' replace preventScrollReset>
-                  <Input
-                    required
-                    type='number'
-                    name='displayAmount'
-                    label='Amount'
-                    onChange={handleChange}
-                    addOn={asset.code}
-                    step='any'
-                    min={1 / Math.pow(10, asset.scale)}
-                    error={errorMessage}
-                  />
                   <Input
                     required
                     min={1}
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -1,6 +1,7 @@
 import { Dialog } from '@headlessui/react'
 import { Form } from '@remix-run/react'
-import { ChangeEvent, useState } from 'react'
+import type { ChangeEvent } from 'react'
+import { useState } from 'react'
 import { XIcon } from '~/components/icons'
 import { Button, Input } from '~/components/ui'
 
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -1,5 +1,6 @@
 import { Dialog } from '@headlessui/react'
 import { Form } from '@remix-run/react'
+import { useState } from 'react'
 import { XIcon } from '~/components/icons'
 import { Button, Input } from '~/components/ui'
 
@@ -14,6 +15,8 @@ export const LiquidityDialog = ({
   onClose,
   type
 }: LiquidityDialogProps) => {
+  const [amount, setAmount] = useState(0)
+
   return (
     <Dialog as='div' className='relative z-10' onClose={onClose} open={true}>
       <div className='fixed inset-0 bg-tealish/30 bg-opacity-75 transition-opacity' />
@@ -46,6 +49,10 @@ export const LiquidityDialog = ({
                     name='amount'
                     label='Amount'
                   />
+                  <div className='mt-2'>
+                    <p>Based on the asset:</p>
+                    <p>Amount {amount} = {amount / 100} USD </p>
+                  </div>
                   <div className='flex justify-end py-3'>
                     <Button aria-label={`${type} liquidity`} type='submit'>
                       {type} liquidity
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -4,18 +4,25 @@ import { useState } from 'react'
 import { XIcon } from '~/components/icons'
 import { Button, Input } from '~/components/ui'
 
+type BasicAsset = {
+  code: string
+  scale: number
+}
+
 type LiquidityDialogProps = {
   title: string
   onClose: () => void
   type: 'Deposit' | 'Withdraw'
+  asset: BasicAsset
 }
 
 export const LiquidityDialog = ({
   title,
   onClose,
-  type
+  type,
+  asset
 }: LiquidityDialogProps) => {
-  const [amount, setAmount] = useState(0)
+  const [amount, setAmount] = useState<number>(0)
 
   return (
     <Dialog as='div' className='relative z-10' onClose={onClose} open={true}>
@@ -48,10 +55,11 @@ export const LiquidityDialog = ({
                     type='number'
                     name='amount'
                     label='Amount'
+                    onChange={e => setAmount(Number(e.target.value))}
                   />
                   <div className='mt-2'>
                     <p>Based on the asset:</p>
-                    <p>Amount {amount} = {amount / 100} USD </p>
+                    <p>Amount {amount} = {amount / Math.pow(10, asset.scale)} {asset.code} </p>
                   </div>
                   <div className='flex justify-end py-3'>
                     <Button aria-label={`${type} liquidity`} type='submit'>
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -57,7 +57,7 @@ export const LiquidityDialog = ({
                     label='Amount'
                     onChange={e => setAmount(Number(e.target.value))}
                   />
-                  <div className='mt-2'>
+                  <div className='text-gray-500 text-sm mt-2'>
                     <p>Based on the asset:</p>
                     <p>Amount {amount} = {amount / Math.pow(10, asset.scale)} {asset.code} </p>
                   </div>
--- a/packages/frontend/app/components/LiquidityDialog.tsx
+++ b/packages/frontend/app/components/LiquidityDialog.tsx
@@ -1,6 +1,6 @@
 import { Dialog } from '@headlessui/react'
 import { Form } from '@remix-run/react'
-import { useState } from 'react'
+import { ChangeEvent, useState } from 'react'
 import { XIcon } from '~/components/icons'
 import { Button, Input } from '~/components/ui'
 
@@ -22,7 +22,21 @@ export const LiquidityDialog = ({
   type,
   asset
 }: LiquidityDialogProps) => {
-  const [amount, setAmount] = useState<number>(0)
+  const [actualAmount, setActualAmount] = useState<number>(0)
+  const [errorMessage, setErrorMessage] = useState<string>('')
+
+  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+    const userInput = e.target.value
+    const scaledInput = parseFloat(userInput) * Math.pow(10, asset.scale)
+    const integerScaledInput = Math.floor(scaledInput)
+    if (scaledInput !== integerScaledInput) {
+      const error = 'The asset scale cannot accomodate this value'
+      setErrorMessage(error)
+    } else {
+      setErrorMessage('')
+    }
+    setActualAmount(integerScaledInput)
+  }
 
   return (
     <Dialog as='div' className='relative z-10' onClose={onClose} open={true}>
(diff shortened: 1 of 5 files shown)


packages/frontend/app/components/LiquidityDialog.tsx
```diff
@@ -45,7 +55,12 @@ export const LiquidityDialog = ({
                     type='number'
                     name='amount'
                     label='Amount'
+                    onChange={e => setAmount(Number(e.target.value))}
                   />
+                  <div className='mt-2'>
+                    <p>Based on the asset:</p>
+                    <p>Amount {amount} = {amount / Math.pow(10, asset.scale)} {asset.code} </p>
```
BlairCurrey, line 62: I think this could be simplified a bit. I don't think we need to restate the amount since its in the input nor that it's translating to the current asset scale because of the context.

I think this would be fine:

```suggestion
                    <p>
                      {amount / Math.pow(10, asset.scale)} {asset.code}{' '}
                    </p>
```
image

But perhaps a dummy input field for the asset scale really drives home the point (user can see 0 and 100 => 100, 2 and 100 => 1, etc.):

```tsx
                  <Input
                    required
                    disabled
                    type='number'
                    label='Asset Scale'
                    value={asset.scale}
                  />
```
image

BlairCurrey (commented):
@Emanuel-Palestino thank you for the contribution - this looks great. Functionality is good, just left a couple style suggestions.

JoblersTune:
Thanks Blair. Iwiill add my review on Wednesday as well. Just OOO right now.

Emanuel-Palestino:
Hello @BlairCurrey , I was thinking about your suggestions using a simplified description and the possibility of adding a dummy input. I think the best option is to use a simplified description, I coded it.

I also thought in another approach:
- I can use the `addOn` input prop to display the asset code in the input, so that with this change, the user can clearly know the unit of their deposit or withdrawal. Then, when the user submits the form, the amount can be captured and converted from original asset scale to `0` (which is the scale the backend expects to receive).
  - Screenshot
image

I would like to know your feedback too @JoblersTune c:

JoblersTune:
@Emanuel-Palestino that's a great suggestion to use the `addOn` input prop. I think you're onto something. Then we can abstract all the complexity foir the users entirely. What about something like this?

```
export const LiquidityDialog = ({
  title,
  onClose,
  type,
  asset
}: LiquidityDialogProps) => {
  const [actualAmount, setActualAmount] = useState<number>(0)
  const [errorMessage, setErrorMessage] = useState<string>('')

  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
    const userInput = e.target.value
    const scaledInput = Number(userInput) * Math.pow(10, asset.scale)
    const integerScaledInput = Math.floor(scaledInput)
    if (scaledInput !== integerScaledInput) {
      const error = 'The asset scale cannot accomodate this value'
      setErrorMessage(error)
    } else {
      setErrorMessage('')
    }
    setActualAmount(integerScaledInput)
  }

  return (
    <Dialog as='div' className='relative z-10' onClose={onClose} open={true}>
      <div className='fixed inset-0 bg-tealish/30 bg-opacity-75 transition-opacity' />
      <div className='fixed inset-0 z-10 overflow-y-auto'>
        <div className='flex min-h-full items-center justify-center p-4 text-center'>
          <Dialog.Panel className='relative transform overflow-hidden rounded-lg max-w-full transition-all bg-white px-4 pb-4 pt-5 text-left shadow-xl w-full sm:max-w-lg'>
            <div className='absolute right-0 top-0 pr-4 pt-4'>
              <button
                type='button'
                className='text-gray-400 hover:text-gray-500 focus:outline-none'
                onClick={onClose}
              >
                <span className='sr-only'>Close</span>
                <XIcon className='h-8 w-8' aria-hidden='true' />
              </button>
            </div>
            <div>
              <Dialog.Title
                as='h3'
                className='font-semibold leading-6 text-lg text-center'
              >
                {title}
              </Dialog.Title>
              <div className='mt-2'>
                <Input
                  required
                  type='number'
                  name='displayAmount'
                  label='Amount'
                  onChange={handleChange}
                  addOn={asset.code}
                  step='any'
                  error={errorMessage}
                />
                <Form method='post' replace preventScrollReset>
                  <Input
                    required
                    min={1}
                    type='hidden'
                    name='amount'
                    value={actualAmount}
                  />
                  <div className='flex justify-end py-3'>
                    <Button aria-label={`${type} liquidity`} type='submit'>
                      {type} liquidity
                    </Button>
                  </div>
                </Form>
              </div>
            </div>
          </Dialog.Panel>
        </div>
      </div>
    </Dialog>
  )
}
```
This way with the visible input we can display the value in asset scale 0, the same way we display the liquidity info, but with the hidden field we can submit the value in the relevant asset scale.

What do you think?

packages/frontend/app/routes/assets.$assetId.deposit-liquidity.tsx
```diff
@@ -1,28 +1,45 @@
...
   const navigate = useNavigate()
   const dismissDialog = () => navigate('..', { preventScrollReset: true })
+  const { asset } = useLoaderData<typeof loader>()
 
   return (
     <LiquidityDialog
       onClose={dismissDialog}
       title='Deposit asset liquidity'
       type='Deposit'
+      asset={asset}
```
JoblersTune, line 42:
```suggestion
      asset={{ code: asset.code, scale: asset.scale }}
```

packages/frontend/app/routes/assets.$assetId.withdraw-liquidity.tsx
```diff
@@ -1,28 +1,45 @@
...
   const navigate = useNavigate()
   const dismissDialog = () => navigate('..', { preventScrollReset: true })
+  const { asset } = useLoaderData<typeof loader>()
 
   return (
     <LiquidityDialog
       onClose={dismissDialog}
       title='Withdraw asset liquidity'
       type='Withdraw'
+      asset={asset}
```
JoblersTune, line 42:
```suggestion
      asset={{ code: asset.code, scale: asset.scale }}
```

packages/frontend/app/routes/peers.$peerId.deposit-liquidity.tsx
```diff
@@ -1,28 +1,45 @@
...
   const navigate = useNavigate()
   const dismissDialog = () => navigate('..', { preventScrollReset: true })
+  const { asset } = useLoaderData<typeof loader>()
 
   return (
     <LiquidityDialog
       onClose={dismissDialog}
       title='Deposit peer liquidity'
       type='Deposit'
+      asset={asset}
```
JoblersTune, line 42:
```suggestion
      asset={{ code: asset.code, scale: asset.scale }}
```

packages/frontend/app/routes/peers.$peerId.withdraw-liquidity.tsx
```diff
@@ -1,28 +1,45 @@
...
   const navigate = useNavigate()
   const dismissDialog = () => navigate('..', { preventScrollReset: true })
+  const { asset } = useLoaderData<typeof loader>()
 
   return (
     <LiquidityDialog
       onClose={dismissDialog}
       title='Withdraw peer liquidity'
       type='Withdraw'
+      asset={asset}
```
JoblersTune, line 42:
```suggestion
      asset={{ code: asset.code, scale: asset.scale }}
```

BlairCurrey:
> Hello @BlairCurrey , I was thinking about your suggestions using a simplified description and the possibility of adding a dummy input. I think the best option is to use a simplified description, I coded it.
> 
> I also thought in another approach:
> 
> * I can use the `addOn` input prop to display the asset code in the input, so that with this change, the user can clearly know the unit of their deposit or withdrawal. Then, when the user submits the form, the amount can be captured and converted from original asset scale to `0` (which is the scale the backend expects to receive).
>   
>   * Screenshot
>     image
> 
> I would like to know your feedback too @JoblersTune c:

`addOn` looks great 👍. And converting to 0 on submit as you and Sarah are suggesting looks like a good solution.

Emanuel-Palestino:
@JoblersTune that's a great implementation. I will work on it.

packages/frontend/app/routes/assets.$assetId.deposit-liquidity.tsx
Emanuel-Palestino: I don't think I understand why you suggested this change, can you explain me please? 🙏

JoblersTune: Sure. Since the definition for the BasicAsset is only scale and code this makes it more clear what we are passing in (not the entire asset is used, only code and scale). In other words it has better type matching.

```
type BasicAsset = {
  code: string
  scale: number
}

type LiquidityDialogProps = {
  title: string
  onClose: () => void
  type: 'Deposit' | 'Withdraw'
  asset: BasicAsset
}
```

So having
```
return (
    <LiquidityDialog
      onClose={dismissDialog}
      title='Deposit asset liquidity'
      type='Deposit'
      asset={asset}
    />
)
```

Explicitly matches what we have set in the props definition

Emanuel-Palestino:
I have made the implementation. Waiting for your feedback c:

Additionally, working in the project I found and issue and a suggestion to improve de UX related to assets and peers pages.
- In the asset details page, when a modal is opened, the scroll is setted to the top. This doesn't happen in peers details page.
- I think it's a good idea to make de user input auto focusable, so that the user can type at the same time it opened the modal.

I don't know if I can make this changes in this pull request, or I have to open a new issue.

packages/frontend/app/components/LiquidityDialog.tsx
```diff
@@ -1,19 +1,43 @@
 import { Dialog } from '@headlessui/react'
 import { Form } from '@remix-run/react'
+import { ChangeEvent, useState } from 'react'
```
BlairCurrey, line 3: our prettier check failed here: https://github.com/interledger/rafiki/actions/runs/10097295226/job/27927389684?pr=2817

can either fix with running `pnpm format` or:

```suggestion
import type { ChangeEvent } from 'react'
import { useState } from 'react'
```

BlairCurrey:
> I have made the implementation. Waiting for your feedback c:
> 
> Additionally, working in the project I found and issue and a suggestion to improve de UX related to assets and peers pages.
> 
> * In the asset details page, when a modal is opened, the scroll is setted to the top. This doesn't happen in peers details page.
> * I think it's a good idea to make de user input auto focusable, so that the user can type at the same time it opened the modal.
> 
> I don't know if I can make this changes in this pull request, or I have to open a new issue.

Looks good to me @Emanuel-Palestino. Just a small fix to make prettier happy. https://github.com/interledger/rafiki/pull/2817#discussion_r1691930936

The UX suggestions are good. Lets address that in another issue. Made one here: https://github.com/interledger/rafiki/issues/2825 Feel free to assign yourself if you'd like.

packages/frontend/app/components/LiquidityDialog.tsx
Emanuel-Palestino: Done 👌

packages/frontend/app/components/LiquidityDialog.tsx
```diff
@@ -41,10 +66,21 @@ export const LiquidityDialog = ({
                 <Form method='post' replace preventScrollReset>
```
JoblersTune, line 66:
```suggestion
                <Input
                required
                type='number'
                name='displayAmount'
                label='Amount'
                onChange={handleChange}
                addOn={asset.code}
                step='any'
                min={1 / Math.pow(10, asset.scale)}
                error={errorMessage}
                />
                <Form method='post' replace preventScrollReset>
                  <Input
                    required
                    min={1}
                    type='hidden'
                    name='amount'
                    value={actualAmount}
                  />
                  <div className='flex justify-end py-3'>
                    <Button aria-label={`${type} liquidity`} type='submit'>
                      {type} liquidity
                    </Button>
                  </div>
                </Form>
```

This is a really small (and last) suggestion because everything looks great! But if we move this dummy input field above the form it actually doesn't get sent to the backend when the form is submitted. This way we keep our data collection cleaner on the backend and make sure no-one working on this in the future gets confused about which amount to use for making the GraphQL request or anything like that, we only send what we actually expect to receive on the server-side.

packages/frontend/app/components/LiquidityDialog.tsx
JoblersTune, line 1: This is a really small (and last) suggestion because everything looks great!. But instead of

```
                <Form method='post' replace preventScrollReset>
                  <Input
                    required
                    min={1}
                    type='number'
                    name='amount'
                    name='displayAmount'
                    label='Amount'
                    onChange={handleChange}
                    addOn={asset.code}
                    step='any'
                    min={1 / Math.pow(10, asset.scale)}
                    error={errorMessage}
                  />
                  <Input
                    required
                    min={1}
                    type='hidden'
                    name='amount'
                    value={actualAmount}
                  />
                  <div className='flex justify-end py-3'>
                    <Button aria-label={`${type} liquidity`} type='submit'>
                      {type} liquidity
                    </Button>
                  </div>
                </Form>
```

Can we move the dummy input field above the form?

```
                <Input
                required
                type='number'
                name='displayAmount'
                label='Amount'
                onChange={handleChange}
                addOn={asset.code}
                step='any'
                min={1 / Math.pow(10, asset.scale)}
                error={errorMessage}
                />
                <Form method='post' replace preventScrollReset>
                    <Input
                      required
                      min={1}
                      type='hidden'
                      name='amount'
                      value={actualAmount}
                    />
                    <div className='flex justify-end py-3'>
                      <Button aria-label={`${type} liquidity`} type='submit'>
                        {type} liquidity
                      </Button>
                    </div>
                  </Form>
```

If we move this dummy input field above the form it actually doesn't get sent to the backend when the form is submitted. This way we keep our data collection cleaner on the backend and make sure no-one working on this in the future gets confused about which amount to use for making the GraphQL request or anything like that, we only send what we actually expect to receive on the server-side.

Emanuel-Palestino: Initially, I move the input out of the form, but I notice when the user submits the form it doesn't make the native html validation and it doesn't prevent the user from submitting any amount. The validation is made succesfully until the amount reaches the backend.
I understand that what you mention is also important. What do you think? @JoblersTune

JoblersTune: Hmm, I was hoping the error message would cover that, even more so with your added minimum specification. Can you give me an example of where it is still failing @Emanuel-Palestino?

Emanuel-Palestino: Sure. When I submit a negative value or a very small value, it submits to the backend.

- Negative values, the error message doesn't appear and the validation is made in the backend.
image

- With very small value, the error message appear (although it doesn't prevent to the user from submit). The validation is made in the backend.
image

The message displayed at the top, is made by the backend.

JoblersTune: @Emanuel-Palestino good catch on the negative values, thanks. We do most of our validation on the backend already. It creates a simpler frontend with faster load times and less JS complexity. We want to validate our input on the backend since that is where we'll be using it, so while it's nice to have frontend, reactive and immediate feedback it can lead to slower load times and some redundancy since backend validation is required as well.

So it's fine to get a server-side error for incorrect data entries. However, let's mitigate this by just adding a frontend message for the case where the value is negative as well, to give users a clearer error where we can. 

```
const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
    const userInput = e.target.value
    const scaledInput = parseFloat(userInput) * Math.pow(10, asset.scale)
    const integerScaledInput = Math.floor(scaledInput)
    if (scaledInput < 0) {
      const error = 'The amount should be a positive value'
      setErrorMessage(error)
    } else if (scaledInput !== integerScaledInput) {
      const error = 'The asset scale cannot accomodate this value'
      setErrorMessage(error)
    } else {
      setErrorMessage('')
    }
    setActualAmount(integerScaledInput)
  }
```

And then keep that input value above the form element.

Emanuel-Palestino: Perfect, I'll make the changes c:

JoblersTune (approved):
Thank you so much for your positive and engaged contribution!

PR #2827, opened 2024-07-30
feat(incoming-payment): make it actionable

### LINKED ISSUE #2811: Actionable incoming payments
### Context
[Project Outline](https://docs.google.com/document/d/1PGe1iLleFZCDj3PtV2_jMkq9-BIelt4RuRj4CfC1E2k/edit?usp=sharing)

An ASE should be able to "action" on an incoming payment: it should be able to do certain operations before it decides that an account/wallet address in question can in fact receive money. 

This primarily arose out of the PCH project, as one of their requirements is that they would like to do KYC checks/account status checks before a payment is accepted. A payment should then not proceed if the checks are not successful.

What we will do is similar to the `wallet_address.not_found` webhook: after Rafiki sends the `incoming_payment.created` webhook, we will poll for whether an incoming payment is approved or cancelled by the ASE. 

### Tasks
- [x] Add `approvedAt` and `cancelledAt` timestamps on the incoming payment model
- [x] Add `approveIncomingPayment` mutation, takes in an _object_ Input: `{ id: ID! }` This will set the approvedAt timestamp on the incoming payment. This would only work for payments that are PENDING.
- [x] Add `cancelIncomingPayment` mutation, that takes in an _object_ Input: `{ id: ID! }`. This will set the cancelledAt timestamp on the incoming payment. This would only work for payments that are PENDING.
- [x] Add `POLL_INCOMING_PAYMENT_CREATED_WEBHOOK` (default false) `INCOMING_PAYMENT_CREATED_POLL_TIMEOUT_MS` `INCOMING_PAYMENT_CREATED_POLL_FREQUENCY_MS` flags to Rafiki backend.
- [x] If `POLL_INCOMING_PAYMENT_CREATED_WEBHOOK` enabled, do polling similar to  `wallet_address.not_found` behaviour after the incoming payment created webhook has been sent. Wait for one of those timestamps to be set. If neither are set, return error. (maybe 400 for the Open Payments call? Something reasonable :))

golobitch:

- backend migration that will add `approvedAt` and `cancelledAt` columns to the `incomingPayments` table
- `approveIncomingPayment` mutation that accepts incoming payment id and if this payment is in the database and in PENDING state, it will update it's `approvedAt` field to current time.
- `cancellIncomingPayment`mutation that accepts incoming payment id and if this payment is in the database and in PENDING state it will update it's `cancelledAt` field to current time.

## Context
 fixes #2811 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -135,7 +141,34 @@ async function createIncomingPayment(
...
+    if (response) return response
+    return IncomingPaymentError.ActionNotPerformed
+  }catch(err) {
+    return IncomingPaymentError.ActionNotPerformed
+  }
+
+}
+
+async function getApprovedOrCanceledIncomingPayment(deps: ServiceDependencies, options: GetOptions) {
+  const incomingPayment = await getIncomingPayment(deps, { id: options.id })
```
mkurapov, line 165: I think what would be better instead of using `getIncomingPayment` is writing another query here. I want to avoid using `getIncomingPayment` because `addReceivedAmount` ends up doing the accounting lookup -> but we probably only want to do this once we know that the payment is approved or cancelled.

The query can include a check for where `approvedAt` or `cancelledAt` is not null explicitly (ie `IncomingPayment.query(...).whereNotNull(...)`).

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -278,6 +311,50 @@ async function getWalletAddressPage(
...
+      .findById(id)
+      .forUpdate()
+      .withGraphFetched('[asset, walletAddress]')
+
+    if (!payment) return IncomingPaymentError.UnknownPayment
+    if (payment.state !== IncomingPaymentState.Pending)
+      return IncomingPaymentError.WrongState
+
+    await payment.$query(trx).patch({
+      approvedAt: new Date(Date.now())
```
mkurapov, line 351:
```suggestion
      cancelledAt: new Date()
```

golobitch:
I will also add bruno collection to approve or cancel incoming payment through Admin API. Tomorrow :)

packages/backend/src/open_payments/payment/incoming/service.test.ts
```diff
@@ -62,6 +64,160 @@ describe('Incoming Payment Service', (): void => {
...
+          .patch({ state: IncomingPaymentState.Pending })
+
+        const approvedIncomingPayment = await incomingPaymentService.approve(
+          incomingPayment.id
+        )
+        assert.ok(!isIncomingPaymentError(approvedIncomingPayment))
+        expect(approvedIncomingPayment.id).toBe(incomingPayment.id)
+        expect(approvedIncomingPayment.approvedAt).toBeDefined()
+        expect(!approvedIncomingPayment.cancelledAt).toBeTruthy()
+      })
```
mkurapov, line 192:
```suggestion
        expect(approvedIncomingPayment.cancelledAt).toBeFalsy()
```
or `toBeNull`

bruno/collections/Rafiki/Rafiki Admin APIs/Approve Incoming Payment.bru
```diff
@@ -0,0 +1,34 @@
...
+  url: {{RafikiGraphqlHost}}/graphql
+  body: graphql
+  auth: none
+}
+
+body:graphql {
+  mutation ApproveIncomingPayment($input: ApproveIncomingPaymentInput!) {
+  	approveIncomingPayment(input:$input) {
+      id
+      approvedAt
```
mkurapov, line 17:
```suggestion
      payment
```
since it doesnt look like were exposing `approvedAt` in the schema

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -135,7 +141,35 @@ async function createIncomingPayment(
...
+      request: async () =>
+        getApprovedOrCanceledIncomingPayment(deps, { id: incomingPayment.id }),
+      pollingFrequencyMs: deps.config.incomingPaymentCreatedPollFrequency,
+      timeoutMs: deps.config.incomingPaymentCreatedPollTimeout
+    })
+
+    if (response) return response
+    return IncomingPaymentError.ActionNotPerformed
+  } catch (err) {
+    return IncomingPaymentError.ActionNotPerformed
```
mkurapov, line 160: lets add a log here

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -135,7 +141,35 @@ async function createIncomingPayment(
...
+
+  try {
+    const response = await poll({
+      request: async () =>
+        getApprovedOrCanceledIncomingPayment(deps, { id: incomingPayment.id }),
+      pollingFrequencyMs: deps.config.incomingPaymentCreatedPollFrequency,
+      timeoutMs: deps.config.incomingPaymentCreatedPollTimeout
+    })
+
+    if (response) return response
```
mkurapov, line 157: should we return the payment if it was cancelled?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -135,7 +141,35 @@ async function createIncomingPayment(
...
+    return IncomingPaymentError.ActionNotPerformed
```
mkurapov, line 158: maybe we have one error for not being approved an another for this timeout?

bruno/collections/Rafiki/Rafiki Admin APIs/Approve Incoming Payment.bru
golobitch: Oh yes, this was just my check of that in the schema.

packages/backend/src/open_payments/payment/incoming/service.ts
golobitch: This can be done yes. I did this because end result is basically the same. Action was not performed in this timeframe. I can do `IncomingPaymentError.ActionTimeout` ?

packages/backend/src/open_payments/payment/incoming/service.ts
golobitch: Always to return payment? How will end user now that the IP was cancelled? With other calls?

mkurapov: I think for now lets only return the payment if it was approved, otherwise error. Since it might be confusing returning a payment that money can't be sent to at all.

Maybe in the future we can let the ASE specify what error is shown if a payment is not approved

packages/backend/src/open_payments/payment/incoming/service.ts
mkurapov: We can also keep it like this as long as we have logs for both so we know what happened

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -278,6 +329,50 @@ async function getWalletAddressPage(
...
+    if (payment.state !== IncomingPaymentState.Pending)
+      return IncomingPaymentError.WrongState
+
+    await payment.$query(trx).patch({
+      approvedAt: new Date(Date.now())
+    })
+
+    return await addReceivedAmount(deps, payment)
+  })
+}
```
BlairCurrey, line 352: I can cancel a pending incoming payment and then approve an incoming payment and both timestamps will be set.

How do we want to handle that? I think its probably fine to switch from canceled to approved but it should unset the other one. This applies to cancelIncomingPayment as well. I can approve then cancel a payment and both timestamps are set.

Tangential, and probably not a problem, but I can approve over and over again and it resets the `approvedAt`. I think this might be the best behavior though. We could instead just return the payment without patching (not erroring) but I feel like the way we have it now is more predictable... unless we have some reason not to lose the original `approvedAt` time but I dont think so. (note that this applies to cancelledAt too).

golobitch: Wow, great catch. Don't know how I missed this. Instead of checking timestamps, I would suggest that when IP is approved/cancelled, that we check if opposite field is already set. If it is, then throw error, otherwise, check if it was already approved/cancelled and don't update timestamp

```
// approve IP
incomingPayment := getIncomingPayment()
if incomingPayment.cancelledAt then
    throw Error('IP was already cancelled')
elif incomingPayment.approvedAt then
    return incomingPayment
end

incomingPayment.approvedAt = now()

return incomingPayment
```

Something like this

golobitch: I added additional checks when approve or cancel incoming payment method is called. I also added tests to checks exactly what you wrote above.

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -278,6 +329,62 @@ async function getWalletAddressPage(
...
+      .findById(id)
+      .forUpdate()
+      .withGraphFetched('[asset, walletAddress]')
+
+    if (!payment) return IncomingPaymentError.UnknownPayment
+    if (payment.state !== IncomingPaymentState.Pending)
+      return IncomingPaymentError.WrongState
+
+    if (payment.cancelledAt) {
+      throw new Error('Cannot approve already cancelled incoming payment.')
```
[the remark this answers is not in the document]
golobitch: I better like option 1 than option 2.

mkurapov:
@tadejgolobic some formatting fixes needed

golobitch:
> ~@tadejgolobic some formatting fixes needed~
> 
> ah I see you are still testing

Yeah, I did a push from different computer, where I did not have all the changes ... now it should be everything

mkurapov (approved):

PR #2832, opened 2024-08-01
fix(auth): interact redirect

### LINKED ISSUE #2826: [BUG] Internal Server Error when redirecting to interaction
## Bug Report

**Describe the bug**

After an interactive grant is created and an application tries to redirect the user to the URL, there is an `Internal Server Error`.
This is not happening on a regular basis. After the page is refreshed, the user is correctly redirected to the IdP page.

**To Reproduce**

This happened to me when connecting my wallet from Rafiki Money to the Web Monetization Extension

**Expected behavior**

The user should be redirected directly to the IdP page, without having to refresh a page that showcases an Internal Server Error.

 - OS: [e.g. Ubuntu]
 - Version [e.g. 20.04LTS]
 - Node version [e.g. 18.16.0]
 - pnpm version [e.g. 8.6.7]

**Screenshots**
image

**Additional context**
Logs from Rafiki Auth (rafiki.money)

```
  TypeError: option maxAge is invalid
      at new Cookie (/home/rafiki/node_modules/.pnpm/cookies@0.9.1/node_modules/cookies/index.js:182:11)
      at Cookies.set (/home/rafiki/node_modules/.pnpm/cookies@0.9.1/node_modules/cookies/index.js:117:16)
      at ContextSession.save (/home/rafiki/node_modules/.pnpm/koa-session@6.4.0/node_modules/koa-session/lib/context.js:331:26)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async ContextSession.commit (/home/rafiki/node_modules/.pnpm/koa-session@6.4.0/node_modules/koa-session/lib/context.js:244:5)
      at async session (/home/rafiki/node_modules/.pnpm/koa-session@6.4.0/node_modules/koa-session/index.js:46:9)
      at async cors (/home/rafiki/node_modules/.pnpm/@koa+cors@5.0.0/node_modules/@koa/cors/index.js:109:16)
```

sabineschaller:

- Fixes the `maxAge` is invalid error by changing from hget/hset (saved as redis hashmap) to set/get (string) so that we can use `JSON.parse` and `JSON.stringify`. Saving the `JSON.stringify`ied session preserves the datatypes so that we don't need to [deal with manually re-forming the session](https://github.com/interledger/rafiki/pull/2832#discussion_r1700705829) and can just do `JSON.parse` on read.
- Fixes cookie not expiring in browser by returning `null` from `get` when the session is not found.
- removes unused session middleware in interaction server (`'/grant/:id/:nonce'`, `'/grant/:id/:nonce/:choice'`)

## Context

- fixes #2826 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -361,10 +361,10 @@ export class App {
             },
             async set(key, session) {
               // Add a delay to cookie age to ensure redis record expires after cookie
-              const expireInMs = maxAgeMs + 10 * 1000
+              const expireInSec = maxAgeMs / 1000 + 10
               const op = redis.multi()
               op.set(key, JSON.stringify(session))
-              op.expire(key, expireInMs)
+              op.expire(key, expireInSec)
               await op.exec()
             },
             async destroy(key) {
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -353,24 +353,22 @@ export class App {
           signed: true,
           store: {
             async get(key) {
-              const s = await redis.hgetall(key)
-              const session = {
-                nonce: s.nonce,
-                _expire: Number(s._expire),
-                _maxAge: Number(s._maxAge)
-              }
-              return session
+              const s = await redis.get(key)
+
+              if (!s) return null
+
+              return JSON.parse(s)
             },
             async set(key, session) {
               // Add a delay to cookie age to ensure redis record expires after cookie
               const expireInMs = maxAgeMs + 10 * 1000
               const op = redis.multi()
-              op.hset(key, session)
+              op.set(key, JSON.stringify(session))
               op.expire(key, expireInMs)
               await op.exec()
             },
             async destroy(key) {
-              await redis.hdel(key)
+              await redis.del(key)
             }
           }
         },
@@ -447,17 +445,6 @@ export class App {
 
     koa.use(cors())
     koa.keys = [this.config.cookieKey]
-    koa.use(
-      session(
-        {
-          key: 'sessionId',
-          maxAge: 60 * 1000,
-          signed: true
-        },
-        koa
-      )
-    )
-
     koa.use(router.middleware())
     koa.use(router.routes())
 
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -353,7 +353,13 @@ export class App {
           signed: true,
           store: {
             async get(key) {
-              return await redis.hgetall(key)
+              const s = await redis.hgetall(key)
+              const session = {
+                nonce: s.nonce,
+                _expire: Number(s._expire),
+                _maxAge: Number(s._maxAge)
+              }
+              return session
             },
             async set(key, session) {
               // Add a delay to cookie age to ensure redis record expires after cookie
--- a/packages/auth/src/interaction/routes.ts
+++ b/packages/auth/src/interaction/routes.ts
@@ -181,7 +181,6 @@ async function startInteraction(
 
   const trx = await Interaction.startTransaction()
   try {
-    // TODO: also establish session in redis with short expiry
     await grantService.markPending(interaction.id, trx)
     await trx.commit()
 


packages/auth/src/app.ts
```diff
@@ -353,7 +353,13 @@ export class App {
...
           store: {
             async get(key) {
-              return await redis.hgetall(key)
+              const s = await redis.hgetall(key)
+              const session = {
+                nonce: s.nonce,
+                _expire: Number(s._expire),
+                _maxAge: Number(s._maxAge)
+              }
+              return session
```
BlairCurrey, line 362: this would lose anything from `s` not explicitly set in `session`. not sure we can assume the only properties set on the session in redis will be these. we could do `session = { ...s, _maxAge: Number(s._maxAge), _expire: Number(s._expire)}` but that would have the same problem if the other fields are expected not to be strings (just like we saw with `maxAge`).

I think a more common way of getting/settings is `JSON.stringify(object)` on set and `JSON.parse(string)` on get, which which would take care of converting these to numbers without manually reforming the session object. hashes were suggested to me originally when I implemented this but not sure it's the best way to do it now...

BlairCurrey:
> When the session is retrieved from redis, the _maxAge and _expire parameters are strings instead of numbers. I don't know why that only messed things up every once in a while but once typecasting them back to Numbers, I didn't see the Internal Server Error anymore.

It does make sense that the string would cause the error. the type for session shows `maxAge` should be `number | "session" | undefined` (and unfortunately the return type for `get` is `any`). In my testing it happens in every subsequent grant request (with interaction). So I:
- make a grant request, visit the link. OK
- make another, visit the link. Internal Server Error. Refresh and its OK.
- make another, visit the link. Internal Server Error. Refresh and its OK.
...etc

So at least for the initial one it seems obvious that its working because there is no session being retrieved from redis and parsed incorrectly. And then for subsequent requests there are, so it fails. But not sure why the refresh would work...

raducristianpopa:
> > When the session is retrieved from redis, the _maxAge and _expire parameters are strings instead of numbers. I don't know why that only messed things up every once in a while but once typecasting them back to Numbers, I didn't see the Internal Server Error anymore.
> 
> It does make sense that the string would cause the error. the type for session shows `maxAge` should be `number | "session" | undefined` (and unfortunately the return type for `get` is `any`). In my testing it happens in every subsequent grant request (with interaction). So I:
> 
>     * make a grant request, visit the link. OK
> 
>     * make another, visit the link. Internal Server Error. Refresh and its OK.
> 
>     * make another, visit the link. Internal Server Error. Refresh and its OK.
>       ...etc
> 
> 
> So at least for the initial one it seems obvious that its working because there is no session being retrieved from redis and parsed incorrectly. And then for subsequent requests there are, so it fails. But not sure why the refresh would work...

I ran some tests on this as well. It looks like the same session is used after the first interaction is complete. 

Q:
- Is this the intended behaviour? 
- Whenever we start a new interaction, a new session needs to be created as well or we just have to update the existing one?

Logs: 
```
-------------------
{
  nonce: '53D27796BE7BB9BB',
  _expire: '1722572815903',
  _maxAge: '60000'
}
-------------------
^ logged after accepting the grant (when the user gets redirected to the finish url)
-------------------
{
  nonce: '53D27796BE7BB9BB',
  _expire: '1722572815903',
  _maxAge: '60000'
}
-------------------
^ logged when trying to access the redirect URL for the second grant

  TypeError: option maxAge is invalid
      [redacted]
      
-------------------
{
  nonce: 'D973ED2E15A125A0',
  _expire: '600001722572792015',
  _maxAge: '60000'
}
-------------------
^ logged after refreshing and accepting the second grant (when the user gets redirected to the finish url)
```

What I found really confusing is that when I am getting redirected to the finish URL (`http://localhost:3006/interact/372b14bf-a96d-457a-bc7b-f70b09ddb17c/98F1F12B1BF803C1/finish`) and I start refreshing the page (until the session becomes invalid), I am not seeing the `maxAge` error. The `set` method is called successfully without throwing (the logs from above are located in the `set` method)

sabineschaller:
> In my testing it happens in every subsequent grant request (with interaction).

It didn't happen to me that often. That is what most confused me.

BlairCurrey:
Quick recap after going over this with @raducristianpopa.

While the error is being caused by `maxAge` being parsed from redis incorrectly (`koa-session` expects a number but its a string), there was also a problem of not expiring sessions correctly. That is, if you completed the interaction and keep refreshing, eventually the session would expire but the browser cookie would not be removed. This PR fixes these two things.

To address Radu's question from above:

> Q:
> 
> * Is this the intended behaviour?
> * Whenever we start a new interaction, a new session needs to be created as well or we just have to update the existing one?

Basically yes, this is the intended behavior. If you visit the redirect from the same browser you should have the same cookie which will use the same session. For good measure we confirmed that on rafiki.money different users are generating different sessions.

In addition we did an audit of security related cookie settings and found some things that could be improved and captured it here: https://github.com/interledger/rafiki/issues/2844

packages/auth/src/interaction/routes.ts
```diff
@@ -181,7 +181,6 @@ async function startInteraction(
 
   const trx = await Interaction.startTransaction()
   try {
-    // TODO: also establish session in redis with short expiry
```
BlairCurrey, line 184: confirmed with Nathan this was just about moving sessions into Redis so its no longer a TODO

packages/auth/src/app.ts
```diff
@@ -441,17 +445,6 @@ export class App {
...
-      session(
-        {
-          key: 'sessionId',
-          maxAge: 60 * 1000,
-          signed: true
-        },
-        koa
-      )
-    )
-
```
BlairCurrey, line 454: not used anywhere. looks like a potential copy/paste artifact from the recent(ish) factoring out of interaction routes

njlie (approved):
LGTM

raducristianpopa (approved):

Issue #2815, opened 2024-07-18
Ensure wallet addresses are case insensitive


mkurapov:
## Context
Because wallet addresses (and payment pointers) are in fact URLs, they should be case INsensitive. Currently, we are storing them in the DB simply as unique `string` without doing any cleanup beforehand, meaning, they are technically case-sensitive now.

## Todos
- [ ] Make sure the wallet address URLs we store in `backend` are case-insensitive

s100tist:
Is there anyone working on this? I'd like to try it:)

golobitch:
> Is there anyone working on this yet? I'd like to try it:)

Go for it :)

s100tist:
Hi, I looked at the code and then I thought it would be enough modifying the wallet addres url to lower case before it got registered in the database.

When i tested the wallet address creation in Rafiki Admin it stored the wallet address url in lower case successfully. 
Then I tested a wallet address creation via bruno scenarios and the wallet address was succesfully created as well. However the sucessive events failed; this because the middleware was still case sensitive. 

By this time i left my code letting both the storage and the getWalletAddressUrl using the 'toLowerCase' function. This seems to patch the case-sensitiveness in the storing and in the looking. I don't know if the modified storage would affect any other scenarios. What would your feedback be?

mkurapov:
> When i tested the wallet address creation in Rafiki Admin it stored the wallet address url in lower case successfully.

Even if you had upper case characters in the URL, correct?

> However the sucessive events failed; this because the middleware was still case sensitive.

What exactly did you end up trying? So I can understand correctly

s100tist:
That's correct, it stores the url in lower case even if it has a upper case characters.

In first place i just convert to lowercase the url
![image](https://github.com/user-attachments/assets/d8cda431-e7a5-4635-9420-226f6134ee9a)
And then i added this to the methods where we obtain the url 
![image](https://github.com/user-attachments/assets/d032b46d-3ff9-4fa5-9656-2ea1cbdd631f)

s100tist:
I made the middleware modifications because when i did some tests rafiki looked for the wallet address as it was firstly written it looked for it using the url with the upper case, when they're only stored in lower case

sabineschaller:
@s100tist feel free to open a draft PR such that we can see what you did 🙂

PR #2833, opened 2024-08-01
fix(backend): make the rafiki wallet url case insensitive 

s100tist:
Converting the wallet address to lower case when it get stored
Making the getWalletAddres middleware methods looking for a wallet addres url converted to lower case
- 

## Context
When a wallet address is created, the wallet address gets stored in the database as it was written originally. 
This causes undesired beheaviour such as 'wallet address not found' being raised due to the case sensitiveness.

This PR fixes issue #2815

## Checklist


golobitch:
@s100tist CI prerequisite job is failing due to code format. Please run `pnpm format` and then commit & push changes.

packages/backend/src/open_payments/wallet_address/middleware.ts
```diff
@@ -15,23 +15,24 @@ export async function getWalletAddressUrlFromRequestBody(
   ctx: SignedCollectionContext<CreateBody>,
   next: () => Promise<void>
 ) {
-  ctx.walletAddressUrl = ctx.request.body.walletAddress
+  ctx.walletAddressUrl = ctx.request.body.walletAddress.toLowerCase()
```
mkurapov, line 18: Instead of doing this all over, we can just do `.toLowerCase` in  during creation (`walletAddressService.create`) and when we fetch it by url (`getWalletAddressByUrl`)

s100tist: If I correctly understood it would be better to discard the `packages/backend/src/open_payments/wallet_address/middleware.ts` and `packages/backend/src/graphql/resolvers/wallet_address.ts` changes and better lowercase inside `packages/backend/src/open_payments/wallet_address/service.ts`?

golobitch: @mkurapov maybe going a little bit further, but probably one very neat solution would be to use GraphQL directive for this kind of things?

Our input object would then look like this
```
input CreateWalletAddressInput {
  "Asset of the wallet address"
  assetId: String!
  "Wallet Address URL"
  url: String! @lower
  "Public name associated with the wallet address"
  publicName: String
  idempotencyKey: String
  "Additional properties associated with the [walletAddress]."
  additionalProperties: [AdditionalPropertyInput!]
}
```

Notice that `@lower`? With this kind of directive we would make sure that we always get lower case url in the request.

So the only question would be `getWalletAddressByUrl` right? If I am not mistaken this gets called only when we create receiver right or with other words, it is only getting called from GraphQL not Rest right? In that case, we can also add directive to that field.

Instead of doing this thing manually as close to the database as possible we would have lowercase url since the beginning and it would be the same as if user were inputing url in lower case.

But maybe this is too much for this ticket?

mkurapov: @tadejgolobic 

That's quite a nice solution actually, and we probably have a few opportunities where we could leverage GQL directives more. 

For this case though, I think it makes sense to keep this in the service/domain layer because no matter the interface/API used to create wallet addresses, we always have this baked into the logic before we create the wallet address

s100tist: @tadejgolobic would this change not be retroactive? I mean the already created wallets will not be found. 

Assuming that you mean admin API instead of GraphQL. Yes, the `getWalletAddressByUrl` is geting called from the admin API and the rest (open payments>get method)

packages/backend/src/open_payments/wallet_address/middleware.ts
```diff
@@ -31,7 +31,8 @@ export async function getWalletAddressUrlFromPath(
   ctx: WalletAddressUrlContext,
   next: () => Promise<void>
 ) {
-  ctx.walletAddressUrl = `https://${ctx.request.host}/${ctx.params.walletAddressPath}`
+  ctx.walletAddressUrl =
```
mkurapov, line 34: why did this end up getting updated?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -166,7 +166,7 @@ async function createWalletAddress(
 
     return await WalletAddress.query(deps.knex)
       .insertGraphAndFetch({
-        url: options.url,
+        url: options.url.toLowerCase(),
```
mkurapov, line 169: let's add some tests for these :)

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -421,6 +459,8 @@ describe('Open Payments Wallet Address Service', (): void => {
         }
       )
     )
+
+
```
mkurapov, line 463:
```suggestion
```

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -135,6 +135,44 @@ describe('Open Payments Wallet Address Service', (): void => {
         accountingService.getBalance(walletAddress.id)
       ).resolves.toBeUndefined()
     })
+    
+
+    test.each`
+      url                       | description
+      ${'https://AlIce.me/PEy'} | ${'with a mixed case url'}
```
mkurapov, line 142: we can just turn this to a single test block (no need for `test.each`)

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -135,6 +135,44 @@ describe('Open Payments Wallet Address Service', (): void => {
...
+  `(
+    'Wallet address cannot be created if the url is duplicated',
+    async ({ url }): Promise<void> => {
+      await expect(
+        walletAddressService.create({
+            ...options,
+            url
+          })  
+      ).resolves.toMatchObject(
+        {url:url.toLowerCase()}
```
mkurapov, line 171: This should be also a single test where we call `.create`, once with the first wallet address, and then the second one with different capitalizations, which ends up throwing an error

packages/backend/src/open_payments/wallet_address/service.ts
mkurapov: @s100tist 
in the catch statement, you can check for `err instanceof UniqueViolationError` to catch those cases where you are trying to create a duplicate wallet address

packages/backend/src/open_payments/wallet_address/errors.ts
```diff
@@ -15,13 +16,16 @@ export const errorToCode: {
...
 
 export const errorToMessage: {
   [key in WalletAddressError]: string
 } = {
   [WalletAddressError.InvalidUrl]: 'invalid url',
   [WalletAddressError.UnknownAsset]: 'unknown asset',
-  [WalletAddressError.UnknownWalletAddress]: 'unknown wallet address'
+  [WalletAddressError.UnknownWalletAddress]: 'unknown wallet address',
+  [WalletAddressError.DuplicateWalletAddress]:
+    'wallet address is already registered'
```
mkurapov, line 30: Final thing, let's call this 
```suggestion
    'Duplicate wallet address found with the same url'
```

mkurapov (approved):
🚀

PR #2851, opened 2024-08-08
feat(auth): set session expiry based on interaction expiry env

### LINKED ISSUE #2850: Fix max age for auth session cookie
## Context
When we try to finish a grant interaction, we will end up failing if the interaction took longer that ~1 minute. 

This is because we check that the `nonce` of the finish interaction request matches up to the `nonce` found in the original cookie from the first interaction request, and since the `nonce` is saved in cookie for one minute by default, if a request to finish an interaction is longer than a minute, we will get a "invalid session" error. 

This error doesn't make sense however, in the context of the interaction itself, given that we have a `INTERACTION_EXPIRY_SECONDS` env flag. This session error does not end up respecting the actual value set by this flag, since essentially, any interaction expiry above a minute will just end up in an error regardless.

[Slack thread](https://interledgerfoundation.slack.com/archives/C048W6NDDEZ/p1723046302824609)

## Todos
- [ ] In auth, make the cookie expiration slightly longer than the specified `INTERACTION_EXPIRY_SECONDS

njlie:

- Sets the `maxAge` of a `koa-session`, and by extension, session cookies, to be based off of the value of the interaction expiry environment variable.

## Context

Closes #2850 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -344,7 +344,7 @@ export class App {
     koa.keys = [this.config.cookieKey]
 
     const redis = await this.container.use('redis')
-    const maxAgeMs = 60 * 1000
+    const maxAgeMs = this.config.interactionExpirySeconds * 1000
     koa.use(
       session(
         {


njlie:
Note to self: do NOT pull docker images on really crappy hostel internet

dragosp1011 (approved):

mkurapov (approved):

PR #2857, opened 2024-08-10
feat(backend): add local payment

### LINKED ISSUE #2834: Payments within a Rafiki instance a.k.a. local payments
## Context

Currently, every payment is going through the ILP connector within Rafiki, no matter whether it's local, i.e. from one account to another account within the same Rafiki instance, or whether it is to a peer's account. 

We should be able to improve performance if we account for those local payments directly without going through the connector. Additionally, no slippage is required for these payments since no packets can be lost. 

## Todos
- [x] #2853
- [ ] #2854
- [ ] #2855
- [x] #2856

### LINKED ISSUE #2854: Implement local quoting
### Context
We will be adding a "local" payment method. The payment handler method service needs to support two methods, `getQuote` and `pay`. This issue is about implementing the local quote method in a new "local" payment method handler, which just needs to do the conversion between the sending wallet address and the incoming payment. 

Internally, we should be able to differentiate whether a local quote or local payment happened under the hood. Keep in mind, the `outgoingPayment.id` references (has a FK relationship) to `quote.id` (the outgoing payment id is actually equal to the quote id)

### LINKED ISSUE #2855: Implement local pay method
### Context

To actually make the local payments happen, we will need to implement the `pay` method of the new local payment method handling service.

This method will call the `accountingService.createTransfer` function which will make a transfer between the outgoing payment (the source account) and the incoming payment (destination account) using the amounts from the original quote. This can be a instantly-posted transfer, meaning, we don't need to create a pending transfer with a timeout, we can just make the accounting transfer instantly.

BlairCurrey:

- adds local payments (doesnt go through connector for quote/pay)

## Context

fixes #2834
fixes #2854
fixes #2855

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


BlairCurrey:
# Testing migration

You'll need a way to inspect the db. I just connect to the shell in the docker container:

    docker exec -it rafiki-shared-database-1 bash -c "su postgres -c 'psql cloud_nine_wallet_backend'"

1. Move migration out of migrations directory so that it doesnt run.
2. Start local environment and let migrations run normally.
3. Complete P2P Bruno flow (creates a quote). Inspect the `quotes` table in the db and save the record to verify against later.
4. Stop all containers - do not remove volumes. We want to ensure the existing migrations and data are not lost.
5. Move migration back into the migration directory and spin up the containers: `pnpm localenv:compose:psql up --build`

Inspect `ilpQuoteDetails` table in cloud 9 and see that it was migrated correctly (fields match original quote values):

    SELECT * FROM "ilpQuoteDetails";

Inspect quotes and see the columns no longer exist there:

    SELECT * FROM "quote";

This is already coverd by the unit tests, but if you want to manually test the down:

    docker exec -it rafiki-cloud-nine-backend-1 /bin/sh
    pnpm --filter backend knex migrate:down 20240809171654_add_ilp_quote_details_table.js --env production

packages/backend/migrations/20240809171654_add_ilp_quote_details_table.js
```diff
@@ -0,0 +1,130 @@
...
+        table
+          .decimal('highEstimatedExchangeRateNumerator', 64, 0)
+          .notNullable()
+          .alter()
+        table
+          .decimal('highEstimatedExchangeRateDenominator', 64, 0)
+          .notNullable()
+          .alter()
+      })
+    })
```
BlairCurrey, line 126: this down will fail if local quotes are added before migration is run because they dont have these non null columns. We are setting back to null because that was the original state before the migration. What should we do? 
- leave them as nullable in the down. Doesnt fully restore the original state but not sure we can, short of losing the new quotes that dont fit in the previous model.
-  keep nullable and set to sensible(ish) defaults (derive from exchange rate)?

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -217,12 +217,24 @@ async function pay(
...
+  // Or make better error.
+  // Is there a way to ensure that right here, outgoingPayment.quote is QuoteWithDetails?
+  // I dont think so... this is where the ambiguousness of our DB types surfaces and we have
+  // to do some explicit check like this. Even if we had table inheritance we'd have this problem
+  // so long as an outgoing payment could have either type of quote on them. To truly make this
+  // typesafe you might need to have seperate outgoing payment tables for each type of quote (which
+  // goes way too far)
+  if (!outgoingPayment.quote.ilpQuoteDetails) {
+    throw new Error('Missing ILP quote details')
+  }
```
mkurapov, line 230: I think this is where we do an explicit lookup in `ilpQuoteDetails` table (ie, we don't fetch it in the usual `quote` request), but only fetch it when we need it here. thoughts?

That way, we don't need to mess with types at all

packages/backend/migrations/20240809171654_add_ilp_quote_details_table.js
BlairCurrey: > Let's split this up into multiple migrations, something like:
> 
> * First one will backfill `estimatedExchangeRate`, and mark the field as required
> * second one will create `ilpQuoteDetails`
> * third one will backfill `ilpQuoteDetails`
> 
> Then we deploy the code changes to start reading from `ilpQuoteDetails`
> 
> * last one will drop ilp fields from `quotes`
> 
> this way, we dont run into the risk of losing data

I'm not necessarily against splitting into multiple migrations. I suppose it gives more control over up/down if needed without downside.

In terms of this part:

> - third one will backfill ilpQuoteDetails
>
> Then we deploy the code changes to start reading from ilpQuoteDetails
>
> - last one will drop ilp fields from quotes

Is the idea that we put the drop ilp fields from quotes in another release and communicate as much in patch notes? I mean I dont see a way to ensure integrators arent just upgrading multiple releases and running them all at once in that case.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -89,7 +89,7 @@ async function getOutgoingPayment(
 ): Promise<OutgoingPayment | undefined> {
   const outgoingPayment = await OutgoingPayment.query(deps.knex)
     .get(options)
-    .withGraphFetched('[quote.asset, walletAddress]')
+    .withGraphFetched('[quote.[asset, ilpQuoteDetails], walletAddress]')
```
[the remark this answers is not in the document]
BlairCurrey: That was not the intention. Why would we? 

This was added when factoring out ilpQuoteDetails to maintain current functionality of the previously implicitly ilp quotes. When there is no `ilpQuoteDetails` this will be `null` which seems fine to me.

I guess maybe this relates back to https://github.com/interledger/rafiki/pull/2857#discussion_r1722025824?

packages/backend/src/payment-method/ilp/service.ts
BlairCurrey: Hmm... we might have to parse it either way. Either like this or explicitly looking it up. 

I guess it goes back to where this is coming from. If we were explicitly getting one type or another we could throw an error at the service level when trying to get a payment with an ilp quote and not finding it. But in this case it looks like the worker is just getting the latest payment to process (could be anything). Therefore we're going to just have to check here (or where this is called) one way or another. So I think I'd rather just have the `ilpQuoteDetails` (whether populated or `null`) on the payment by default.

packages/backend/src/payment-method/local/service.ts
```diff
@@ -0,0 +1,157 @@
...
+    throw new PaymentMethodHandlerError('Received error during local quoting', {
+      description: 'No value provided to get quote from',
+      retryable: false
+    })
+  }
+
+  return {
+    receiver: options.receiver,
+    walletAddress: options.walletAddress,
+    // TODO: is this correct for both fixed send/receive? or needs to be flipped?
```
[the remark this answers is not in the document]
BlairCurrey: yeah I think you're right...

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -120,11 +124,7 @@ function getAdjustedAmounts(
   // This is only an approximation of the true amount delivered due to exchange rate variance. Due to connection failures there isn't a reliable way to track that in sync with the amount sent (particularly within ILP payments)
   // eslint-disable-next-line no-case-declarations
   const amountDelivered = BigInt(
-    Math.ceil(
-      Number(alreadySentAmount) *
-        (payment.quote.estimatedExchangeRate ||
-          payment.quote.lowEstimatedExchangeRate.valueOf())
-    )
+    Math.ceil(Number(alreadySentAmount) * payment.quote.estimatedExchangeRate)
```
BlairCurrey, line 127: wont have the low `lowEstimatedExchangeRate` to fall back on anymore (it will be optional) so I ensured `estimatedExchangeRate` is always set in the migration. Set it to be `lowEstimatedExchangeRate` where null, just like we're setting it in the ilp `getQoute` method.

packages/backend/src/payment-method/ilp/service.ts
mkurapov: > So I think I'd rather just have the ilpQuoteDetails (whether populated or null) on the payment by default.

I would say the way that we have this set up is that the outgoing payment itself doesn't know anything about the payment method, it only just needs to have the `payment.quote.method` in order to know what to call when calling
`paymentMethodHandlerService.pay`. The only time we will need to fetch `ilpQuoteDetails` is in this ILP payment service, so we can avoid making additional requests all other times we just need to fetch the outgoing payment data

BlairCurrey: > The only time we will need to fetch ilpQuoteDetails is in this ILP payment service, so we can avoid making additional requests all other times we just need to fetch the outgoing payment data

Makes sense... we can switch to that.

> it only just needs to have the payment.quote.method in order to know what to call when calling
paymentMethodHandlerService.pay

I don't think I'm quite following here. `isLocal` was added to the receiver to determine which method to call. Im using that to call the correct `pay`/`getQuote`. I'm not using the quote payment method for that. In fact I didn't plan on adding a new `local` method at this level as part of not presenting some additional method in the spec (just internal). Guess I'd have to look into this more but not sure if you can clarify in light of that.

packages/backend/migrations/20240809171654_add_ilp_quote_details_table.js
mkurapov: I think this all will go in one release, I was just thinking of how to make the changes safely (or in separate PRs). 

If you want, we can have this:

> Then we deploy the code changes to start reading from ilpQuoteDetails

released the same time as we backfill `ilpQuoteDetails` (and `drop ilp fields from quotes` for that matter)

BlairCurrey: thinking about it more, i dont think we can split the create ilpQuoteDetails table and the migration to backfill it.

if migrations are run through creatye ilpQuoteDetails (but not backill), and then quotes are created in the application with ilpQuoteDetails, then they are backfilled, then the quotes created via the application will try to be inserted again in the ilpQuoteDetails (and it will fail because the quoteId has a unique constraint).

packages/backend/migrations/20240916181643_require_estimated_exchange_rate.js
```diff
@@ -0,0 +1,33 @@
...
+        'lowEstimatedExchangeRateNumerator',
+        'lowEstimatedExchangeRateDenominator'
+      ])
+    })
+    .then(() => {
+      return knex.schema.table('quotes', (table) => {
+        table.decimal('estimatedExchangeRate', 20, 10).notNullable().alter()
+      })
+    })
+}
```
BlairCurrey, line 23: Could these actually be null in prod? They were originally not null and then converted from bigint to decimal where the not null constraint was (I think unintentionally?) lost

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -290,6 +294,22 @@ describe('OutgoingPaymentService', (): void => {
...
+        // call original instead of receiverService.get to avoid infinite loop
+        const receiver = await receiverGet.call(receiverService, url)
+        if (receiver) {
+          // "as any" to circumvent "readonly" check (compile time only) to allow overriding "isLocal" here
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ;(receiver.isLocal as any) = false
+          return receiver
+        }
+        return undefined
+      })
```
BlairCurrey, line 312: There were all sorts of problems with the unit tests (quote/outgoing payemnt) after implementing the logic to pay via local or ilp methods. Before, all methods were implicitly ILP. But now all the existing tests are recognized as local (since there is only 1 rafiki instance in the tests). Took some mocking here to make them all remote by default (and mocking is fine - receiver is not what we're testing here).

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -268,6 +283,70 @@ function calculateFixedSendQuoteAmounts(
   }
 }
 
+// WIP: rework to
+// debitAmount = 500
+// sourceAmount/debitAmountMinusFees = 500 - (500 * 0.02 + 100) = 390
+// receiveAmount = floor(debitAmountMinusFees * estimatedExchangeRate) = floor(390 * 0.91) = 354
```
BlairCurrey, line 289: We wanted to rework `calculateFixedSendQuoteAmounts` to be based off the sending amount as noted here. I started to implement it but was getting some off by 1 errors. messed with how we ceiled/floored a bit with no difference but its probably worth trying again and accounting for the difference (should still work in theory, i think)

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -217,12 +219,27 @@ async function pay(
...
+      throw new PaymentMethodHandlerError(
+        'Could not find required ILP Quote Details',
+        {
+          description: 'ILP Quote Details not found',
+          retryable: false
+        }
+      )
+    }
+  }
+
```
BlairCurrey, line 236: This is how im ensuring we always have the quote details. I also considered simply ensuring the caller passes in a `quote` with `ilpQuoteDetails`, but since the interface is generic (ie same for ILP and LOCAL) I think we'd still need to add a check like this to make typescript happy, so I think adding this and the ilpQuoteDetailsService as a dep is required either way.

packages/backend/src/payment-method/ilp/quote-details/service.ts
```diff
@@ -0,0 +1,32 @@
...
+): Promise<IlpQuoteDetailsService> {
+  const deps = {
+    ...deps_,
+    logger: deps_.logger.child({ service: 'IlpQuoteDetailsService' })
+  }
+  return {
+    getByQuoteId: (quoteId: string) =>
+      getIlpQuoteDetailsByQuoteId(deps, quoteId)
+  }
+}
```
BlairCurrey, line 24: related: https://github.com/interledger/rafiki/pull/2857/files#r1767027492

I created an entirely new service for this table. Initially I though maybe I could just put it in the quote service but it never felt quite right (quote service returning record from not the quotes table, concerned with ILP specific payment method stuff, etc.). Additionally this created a circular dependency when requiring quote service in the payment method - so I think we have to factor out like this either way.

packages/backend/src/payment-method/local/service.ts
BlairCurrey, line 1: lots of work in progress comments here but it is all working fine... only thing is maybe adjusting the final debit amount that gets passed in to just be the `quote.sourceAmount` (or `quote.debitAmountWIthoutFees` or similar, after renaming)

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -350,6 +429,11 @@ async function finalizeQuote(
     : calculateFixedDeliveryQuoteAmounts(deps, quote)
 
   const patchOptions = {
+    sourceAmount: maxReceiveAmountValue
```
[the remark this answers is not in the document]
BlairCurrey: done

bruno/collections/Rafiki/Examples/Admin API -  only locally/Peer-to-Peer Cross Cur Payment (Local, Fixed Send)/Create Outgoing Payment.bru
BlairCurrey, line 1: In regards to all the new bruno requests (not just this specific file), I added them because they were useful for testing but not sure we want all these variants (every combination of local/remote, cross-currency/local, fixed-send/fixed-delivery). I left them for the time being though - maybe just include a basic one (p2p local)? There are unit tests and some integration tests.

packages/backend/src/payment-method/local/service.ts
```diff
@@ -0,0 +1,289 @@
...
+          'Received error during local payment',
+          {
+            description:
+              'Unknown error while trying to create liquidity account',
+            retryable: false
+          }
+        )
+      }
+    }
+  }
```
BlairCurrey, line 246: was not sure if we needed this `incomingPayment.state === IncomingPaymentState.Pending` check here (pulled from connector) - it should probably always be pending right? maybe we check and error if not ... just wasnt quite sure why it made sense in the connector

packages/backend/src/payment-method/local/service.ts
```diff
@@ -0,0 +1,289 @@
...
+  const transferOptions: TransferOptions = {
+    sourceAccount: outgoingPayment,
+    destinationAccount: incomingPayment,
+    sourceAmount: finalDebitAmount,
+    destinationAmount: finalReceiveAmount,
+    transferType: TransferType.TRANSFER,
+    // TODO: no timeout? theoretically possible (perhaps even correct?) but need to
+    // update IncomingPayment state to COMPLETE some other way. trxOrError.post usually
+    // will (via onCredit) but post will return error instead because its already posted.
+    timeout: deps.config.tigerBeetleTwoPhaseTimeout
```
BlairCurrey, line 257: @mkurapov originally you suggested not passing in a timeout which seems like it could make sense, but in that case the incoming payment never gets completed. included the timeout to ensure it does. How important is not setting a timeout? I can eitehr address it here, make a new issue, or remove the todo if its no big deal.

packages/backend/migrations/20240916181643_require_estimated_exchange_rate.js
BlairCurrey: see the TODO comment

@mkurapov can you comment on if dropping the not nullable constraint on `lowEstimatedExchangeRateNumerator` and `lowEstimatedExchangeRateDenominator` were important when the type was changed? im guessing it was inadvertent and we _should_ always be setting it

@raducristianpopa would you be able to verify if any of `lowEstimatedExchangeRateNumerator` or `lowEstimatedExchangeRateDenominator` values are null in testnet?

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -229,10 +239,10 @@ function calculateFixedSendQuoteAmounts(
   quote: Quote,
   maxReceiveAmountValue: bigint
 ): CalculateQuoteAmountsWithFeesResult {
+  // TODO: derive fee from debitAmount and convert that to receiveAmount
```
BlairCurrey, line 242: Took a shot at addressing it and was seeing some off-by-1 errors in the tests. Played with floor/ceiling differently in a few places with no differences. Might be worth tracing that a little further. If not would probably be better to turn the todo into an issue - not sure how critical you think this is @mkurapov

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -350,6 +361,11 @@ async function finalizeQuote(
     : calculateFixedDeliveryQuoteAmounts(deps, quote)
 
   const patchOptions = {
+    debitAmountMinusFees: maxReceiveAmountValue
+      ? // TODO: change calculateFixedSendQuoteAmounts to return the debitAmountMinusFees if using new calculation
```
BlairCurrey, line 365: related https://github.com/interledger/rafiki/pull/2857/files#r1785145020

packages/backend/src/open_payments/quote/service.ts
mkurapov: Yes, I think there is some discrepancy here, or the current behaviour has some incorrect assumptions. Going to think it over some more... this is a tricky one

mkurapov (commented):
Leaving some initial comments, will be back to this!

packages/backend/src/payment-method/ilp/quote-details/service.ts
BlairCurrey: I think I'm coming around to that. 

My immediate reaction was that I'd like to keep the objection calls in the service layers so it can more easily be used across other services and unit tested. And general seperations of concern. Although I recognize the unit test is trivial (just gets by id) and would be implicitly tested by the payment tests. And it's pretty likely we only need the ilp details here. 

Additionally, we are already interfacing with the `ilpQuoteDetails` outside the `ilpQuoteDetail` service in `quoteService.create` (its a bit abstracted away via the objection api but its happening in the `insertGraphAndFetch`). If I wanted to really enforce going through the service layer I guess that would need to change to use some `ilpQuoteDetailsService.create`. And it does generally feel a bit weird to treat it as a thing in-and-of itself (rather than just as _maybe_ part of the quote).

packages/backend/src/payment-method/local/service.ts
BlairCurrey: cleaned up the todo, linking to the single phase transfer issue

mkurapov (commented):
I tested local payment same currency and cross currency payments (without fees) and everything works well for me. Just a few outstanding comments

packages/backend/src/accounting/psql/service.ts
```diff
@@ -264,7 +264,7 @@ export async function createTransfer(
             debitAccount: accountMap[transfer.sourceAccountId],
             creditAccount: accountMap[transfer.destinationAccountId],
             amount: transfer.amount,
-            timeoutMs: BigInt(args.timeout * 1000)
```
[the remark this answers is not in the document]
BlairCurrey: yeah good point, made it required again

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -113,40 +113,50 @@ async function createQuote(
...
+      graph.ilpQuoteDetails = {
+        maxPacketAmount:
+          MAX_INT64 < maxPacketAmount ? MAX_INT64 : maxPacketAmount, // Cap at MAX_INT64 because of postgres type limits.
+        minExchangeRate: quote.additionalFields.minExchangeRate as Pay.Ratio,
+        lowEstimatedExchangeRate: quote.additionalFields
+          .lowEstimatedExchangeRate as Pay.Ratio,
+        highEstimatedExchangeRate: quote.additionalFields
+          .highEstimatedExchangeRate as Pay.PositiveRatio
+      }
+    }
```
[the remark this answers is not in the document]
BlairCurrey: I missed this part. Yeah I think that's right... I was motivated to put it here due to it's relation in the db to quotes but as detailed when retrieving it in `pay`, its really more of a ilp payment method concern. Guess it makes sense to put this part in the `getQuote` method and remove from the quote model as you suggested here: https://github.com/interledger/rafiki/pull/2857#discussion_r1792052018

packages/backend/src/payment-method/local/service.ts
```diff
@@ -0,0 +1,287 @@
...
+        throw new PaymentMethodHandlerError(
+          'Received error during local payment',
+          {
+            description: 'Unknown error while trying to create transfer',
+            retryable: false
+          }
+        )
+    }
+  }
+  await trxOrError.post()
```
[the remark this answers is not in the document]
BlairCurrey: missed this... added.

packages/backend/migrations/20240916181643_require_estimated_exchange_rate.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+
+exports.up = function (knex) {
+  return knex('quotes')
+    .whereNull('estimatedExchangeRate')
```
raducristianpopa, line 8: In which cases the `estimatedExchangeRate` was null?

raducristianpopa (commented):
Migrations look good. Double-checked again with data from the previous test wallet.

packages/backend/src/open_payments/quote/service.ts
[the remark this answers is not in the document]
BlairCurrey: We decided to fully decouple the quote ilp payment method. That is, not dealing with the `ilpQuoteDetails` nor handle any logic based on ILP vs. local quotes in the quote service. The ilp specific stuff (`ilpQuoteDetails`) will be handled in the payment method service. To achieve this we are dropping the FK relationship which allows us to create the `ilpQuoteDetails` in the payment method's `getQuote`.

packages/backend/migrations/20240916181643_require_estimated_exchange_rate.js
BlairCurrey: I dont know of any specific cases where it would be null, but since it's nullable we should handle that possibility.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -429,7 +429,7 @@ async function validateGrantAndAddSpentAmountsToPayment(
     .andWhereNot({
       id: payment.id
     })
-    .withGraphFetched('[quote.asset]')
+    .withGraphFetched('quote.asset')
```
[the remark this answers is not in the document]
BlairCurrey: yeah, the brackets are jsut required if there are multiple joins

packages/backend/src/payment-method/handler/service.ts
```diff
@@ -4,8 +4,11 @@ import { WalletAddress } from '../../open_payments/wallet_address/model'
 import { Receiver } from '../../open_payments/receiver/model'
 import { BaseService } from '../../shared/baseService'
 import { IlpPaymentService } from '../ilp/service'
+import { LocalPaymentService } from '../local/service'
+import { Transaction } from 'objection'
 
 export interface StartQuoteOptions {
+  quoteId?: string
```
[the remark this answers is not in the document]
BlairCurrey: its only used by the ilp implementation (for the `ilpQuoteDetails`). the ilp `getQuote` throws an error if not provided. I figure we dont want to require it for the local `getQuote` since its not used.

It's a little loose type-wise but I don't think there is a better way unless we make more fundamental changes to the payment handler or something.

test/integration/integration.test.ts
```diff
@@ -68,368 +65,588 @@ describe('Integration tests', (): void => {
...
           assetCode: 'USD',
           assetScale: 2,
-          value: value as unknown as bigint
-        },
-        walletAddressUrl:
-          'https://happy-life-bank-test-backend:4100/accounts/pfry'
-      }
+          value: BigInt(quote.receiveAmount.value)
+        })
+        // TODO: fix bug where fixed-send (regardless of local/remote) is not completeing
```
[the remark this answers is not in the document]
BlairCurrey: whoops, thank you... outdated comment

packages/backend/src/payment-method/handler/service.ts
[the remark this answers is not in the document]
mkurapov: Makes sense, we can keep it as optional in that case

packages/backend/src/payment-method/local/service.ts
```diff
@@ -0,0 +1,294 @@
...
+          description: 'Failed to convert debitAmount to receive amount',
+          retryable: false
+        }
+      )
+    }
+    receiveAmountValue = converted
+  } else if (receiveAmount) {
+    receiveAmountValue = receiveAmount.value
+    const converted = await convert({
+      sourceAmount: receiveAmountValue,
```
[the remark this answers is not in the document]
BlairCurrey: I hardcoded in the rates because I'm not sure where to get the sending currency (USD), since there is no `debitAmount` here. Also, I'm not actually sure how to trigger this block as opposed to `receiver.incomingAmount` further down. That's what the fixed send p2p example triggers. Although I think your comment applies to both places so that detail probably doesnt matter.

Anyways, just wanted to validate this. With your calculation patched in and doing a fixed send local payment, I see an outgoing payment with these amounts:

```json
      "receiveAmount": {
        "assetCode": "EUR",
        "assetScale": 2,
        "value": "1000"
      },
      "debitAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1221"
      },
      "sentAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1099"
      },
```

The way it is now, the `receiveAmount` is the same and `debitAmount` and `sentAmount` are 1 higher:

```json
      "receiveAmount": {
        "assetCode": "EUR",
        "assetScale": 2,
        "value": "1000"
      },
      "debitAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1222"
      },
      "sentAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1100"
      },
```

For the same payment over ILP I see the same `1099` `sentAmount` from your revised calculation but a different `debitAmount`. Not sure if slippage would account for this (in which case there is no problem) or something else.

```json
      "receiveAmount": {
        "assetCode": "EUR",
        "assetScale": 2,
        "value": "1000"
      },
      "debitAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1234"
      },
      "sentAmount": {
        "assetCode": "USD",
        "assetScale": 2,
        "value": "1099"
      },
```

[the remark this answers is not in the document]
mkurapov: Also, the estimated exchange rate can be the exchange rate that's received directly from the rates service, instead of having to do `estimatedExchangeRate: Number(receiveAmountValue) / Number(debitAmountValue)`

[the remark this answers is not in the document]
BlairCurrey: 🤦 Of course, the `destinationAsset` passed in to the rates call.

packages/backend/src/rates/util.ts
```diff
@@ -11,9 +11,27 @@ export interface Asset {
   scale: number
 }
 
-export function convert(opts: ConvertOptions): bigint {
+export interface ConvertResults {
+  amount: bigint
+  scaledExchangeRate: number
+}
```
BlairCurrey, line 17: I preferred to return the `scaledExchangeRate` here instead of re-calculating in local `getQuote` because:

1. it should always be the same and duplicating the calculation creates opportunity to diverge. 

2. it was kind of a lot of code duplication and kinda specific stuff I was doing for each conversion case `debitAmount`, `receiver.incomingAmount`, `receiveAmount`.  just felt like this was cleaner.

I mention it though because it did require updating the type many places where we didnt otherwise care about it.

packages/backend/src/rates/service.ts
```diff
@@ -51,22 +53,42 @@ class RatesServiceImpl implements RatesService {
...
-  async convert(
-    opts: Omit<ConvertOptions, 'exchangeRate'>
-  ): Promise<bigint | ConvertError> {
-    const sameCode = opts.sourceAsset.code === opts.destinationAsset.code
-    const sameScale = opts.sourceAsset.scale === opts.destinationAsset.scale
-    if (sameCode && sameScale) return opts.sourceAmount
-    if (sameCode) return convert({ exchangeRate: 1.0, ...opts })
+  async convert(opts: RateConvertOpts): Promise<ConvertResults | ConvertError> {
+    const {
+      reverseDirection = false,
```
BlairCurrey, line 58: convert function is same as before except this `reverseDirection` controls which conversion function is used (`convert` or new `convertReverse`). Alternatively considered making a different `ratesService.convertReverse` - perhaps thats a bit more explicit but I didn't have a strong preference and found this to be simpler.

packages/backend/src/rates/util.ts
```diff
@@ -11,9 +11,27 @@ export interface Asset {
...
+
+export function convertReverse(opts: ConvertOptions): ConvertResults {
+  const scaleDiff = opts.sourceAsset.scale - opts.destinationAsset.scale
+  const scaledExchangeRate = opts.exchangeRate * 10 ** scaleDiff
+
+  return {
+    amount: BigInt(Math.ceil(Number(opts.sourceAmount) / scaledExchangeRate)),
+    scaledExchangeRate
+  }
```
mkurapov, line 36: This looks good and returing `scaledExchangeRate` makes a lot of sense, I'm wondering though whether `convertReverse` might be unclear. 


What do you think about the apporoach of having either `sourceAmount` OR `destinationAmount` provided as an argument to this function?
In our example, USD -> EUR, USD would be the `sourceAsset`, and when we pass in `destinationAmount` in EUR (without providing `sourceAmount`) the function would figure out out the proper direction.

mkurapov (approved):

PR #2862, opened 2024-08-15
feat(backend): track if receiver is local

### LINKED ISSUE #2853: Be able to differentiate local vs remote receiver (in the `Receiver` class, or elsewhere)

BlairCurrey:

- adds `isLocal` boolean to Receiver constructor

## Context

This will be utilized to determine if we should use the local or ilp payment methods. I made this a separate PR since it's not dependent on any of the other changes and should help keep the other PR for that task (a bit) more manageable.

fixes #2853 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -29,8 +29,15 @@ export class Receiver {
...
   public readonly incomingPayment: ReceiverIncomingPayment
+  public readonly isLocal: boolean
 
-  constructor(incomingPayment: OpenPaymentsIncomingPaymentWithPaymentMethod) {
+  constructor(
+    incomingPayment: OpenPaymentsIncomingPaymentWithPaymentMethod,
+    // TODO: lookup incomingPayment where isLocal is used instead of tracking isLocal here?
+    // not sure how simple the check can be. url matches resource server?
+    // exists in local db? exists in local db AND has stream creds (how its checked in receiverService.getReceiver)?
+    isLocal: boolean
```
BlairCurrey, line 39: Going to remove this TODO and keep as-is.

I was trying to see if there was an easier way than passing this flag in everywhere we create a receiver (basically `receiverService.create` and `get` at the moment). However, we do have this information readily available in those places so I guess its not a big deal. I was thinking maybe we could infer it from the resourceServer but this is probably inferior to checking the actual existence of the payment in the local db.

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -147,10 +148,27 @@ async function getReceiver(
...
+      return new Receiver(remoteIncomingPayment, false)
+    }
+  } catch (error) {
+    deps.logger.error(
+      { errorMessage: error instanceof Error && error.message },
+      'Could not get incoming payment'
+    )
   }
+
+  return undefined
```
BlairCurrey, line 171: The real change here in `getReceiver` is that we create the receiver with `isLocal`. The other changes are a little refactor in support of this. Just moves the attempt to get the local incoming payment into this main  `getReceiver` function so I dont have to return more stuff from `getIncomingPayment` (now `getRemoteIncomingPayment`).

sabineschaller (commented):
Generally looks good. Makes code cleaner. Just a question regarding the tests. You updated them by always setting `isLocal = false`. Should we also test a `true` case?

BlairCurrey:
> Generally looks good. Makes code cleaner. Just a question regarding the tests. You updated them by always setting `isLocal = false`. Should we also test a `true` case?

@sabineschaller I changed the receiver model test to true for the sake of this (although it probably doesnt matter much - we're already asserting that its false when we set to false in constructor). 

I dont want to change any of the other tests since everything up until this point was implicitly non-local. I think designating as local might not make sense there and could cause them to fail unexpectedly when we make `isLocal` control which method is used. I'll add tests for the new local payment method and controlling the payment method with `isLocal` when I make these changes in a subsequent PR. Those will be much more substantive.

mkurapov (approved):

Issue #2795, opened 2024-07-05
[BUG] Revoking key didn't actually revoke it


sidvishnoi:
## Bug Report

**Describe the bug**

Revoking key from https://rafiki.money was no longer causing API calls to fail. Following were working fine after key being revoked:
- payments from existing grant going through,
- token rotations too working fine,
- could even revoke a grant and get new grant

**To Reproduce**

This was resolved for me after I changed key-pair (but can occur again). As you can see in screenshots below, jwks.json still has the keys, but UI (and API request in website) don't have any keys.

1. I tested via Web Monetization extension (it's not a bug there)
2. Setup extension, connect wallet.
3. Let some payments go through (go to https://radu.sh to test payments working)
4. Revoke the key from https://rafiki.money/settings/developer-keys
5. API calls work as if key not revoked. Previously, they used to fail with "invalid_client", and "Signature validation error: could not find key in list of client keys" errors.

**Expected behavior**

All API calls relying on key should fail with above errors.

**Desktop (please complete the following information):**

 - Rafiki version: alpha 14
 - Web Monetization extension version: alpha2
 - Wallet address: https://ilp.rafiki.money/sidvishnoi-usd-1

**Screenshots**

![image](https://github.com/interledger/rafiki/assets/8426945/caa2146f-b441-4955-925b-e028e6fbbda0)
![image](https://github.com/interledger/rafiki/assets/8426945/b37e4b23-be21-40f9-b513-1bb79e3d5c01)

golobitch:
I can confirm that rafiki.money is running on version 1.0.0-alpha.14.

I tried reproducing this on rafiki.money, but did not test with web monetization extension. I created few developer keys and checked jwks.json. For me, UI showed exactly the same thing that jwks.json showed.

sidvishnoi:
Are above keys showing up in database for mentioned wallet address? They're not showing up in UI for me (UI part could be a caching issue, I'll check again).

sabineschaller:
I could reproduce that behavior with the extension and a payment pointer from test wallet (rafiki.money). 

I then fired up the Rafiki localenv, 
- checked the keys for a random account (`https://cloud-nine-wallet-backend/accounts/gfranklin`), there was one
- revoked that key using the Admin API endpoint
- fetched the keys for `https://cloud-nine-wallet-backend/accounts/gfranklin`, there was none.

[...]

PR #2863, opened 2024-08-15
feat(backend): unique keys per wallet address

sabineschaller:

This prevents uploading the same key per wallet address multiple times.

## Context

- follow up to #2795 

## Checklist

- [x] Tests added/updated
- [x] Documentation added


packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
```diff
@@ -0,0 +1,29 @@
...
+exports.up = function (knex) {
+  // delete any existing duplicates per wallet address
+  knex.raw(`
+  DELETE FROM walletAddressKeys
+    WHERE ctid NOT IN (
+      SELECT MIN(ctid)
+      FROM walletAddressKeys
+      GROUP BY walletAddressId, kid, x
+    );
+  `)
```
BlairCurrey, line 14: played around with it and it seems to be working.

What's the user experience going to be for the people using the keys being deleted here? I recall the discussion and read the issue but I guess I'm still not quite connecting the dots. For example, is some normal user going to see `invalid signature` or similar from the grant request process and have no idea they need to upload some new key? Or maybe it just fixes the bug in that issue and otherwise people wont be impacted from these duplicates being deleted? 

I see the error for duplicate writes in gql - that makes sense.

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
```diff
@@ -0,0 +1,29 @@
...
+
+  return knex.schema.alterTable('walletAddressKeys', (table) => {
+    table.unique(['walletAddressId', 'kid', 'x'])
+  })
+}
```
BlairCurrey, line 19: noticed this wasn't actually running the DELETE (and when it does we need to quote the camelCase names. Need to do something like:


```suggestion
exports.up = function (knex) {
  // delete any existing duplicates per wallet address
  return knex
    .raw(
      `
    DELETE FROM "walletAddressKeys"
      WHERE ctid NOT IN (
        SELECT MIN(ctid)
        FROM "walletAddressKeys"
        GROUP BY "walletAddressId", kid, x
      );
    `
    )
    .then(() => {
      return knex.schema.alterTable('walletAddressKeys', (table) => {
        table.unique(['walletAddressId', 'kid', 'x'])
      })
    })
}
```

Its kind of annoying but when I want to test migrating data I:
1. move the file out of the migration dir
2. spin up the containers and add whatever data that needs to be migrated (adding a duplicate key in this case)
3. stop the containers (dont remove the volume) and move migration file back in
4. spin back up and check

BlairCurrey:
hmmm what's going on with the history? Just surprised to see a bunch of commits already in main.

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
golobitch: Yes, in PostgreSQL you need to put camel case in quotes. Delete statement was not performed probably because of the missing await, because it is async operation.

BlairCurrey: wish we would have used the snake case mapper when we set the db initially: https://vincit.github.io/objection.js/recipes/snake-case-to-camel-case-conversion.html

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
sabineschaller: The users should not have any issues since they still have another copy of the key. But this makes me think I need to check what happens when a user re-adds a key that they have previously revoked. That should be un-revoked and not throw the duplicate error. And I'll also double check that we actually delete the older versions of the key in the migration and keep the most recent.

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
sabineschaller: I'll add the quotes.

@golobitch all of these have to be async functions?

sabineschaller:
> hmmm what's going on with the history? Just surprised to see a bunch of commits already in main.

@BlairCurrey I messed up merging in main. I haven't touched code in a while so I'm out of practice 😬

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
BlairCurrey: yeah thats a good point... since we're not deleting these and instead setting `revoked` to true I imagine re-adding the previously revoked key will trigger the unique constraint error. I dont think we can just add the `revoked` to the unique constraint because you couldnt revoke it twice.

1. add key, OK
2. revoke key, OK
3. re-add same key, OK
4. revoke same key, NOT OK. already have a revoked key with these unique fields.

BlairCurrey: Not tested, but I think a [partial index](https://www.postgresql.org/docs/current/indexes-partial.html) like this would work: 

```sql
CREATE UNIQUE INDEX walletAddress_keys_revoked_false_idx
ON "walletAddressKeys" (walletAddressId, kid, x)
WHERE revoked = false;
```

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
BlairCurrey: With knex and objection, if you dont await the function the query is not executed. Instead it returns a query builder (so you can modify further). 

That's what happened originally and asserting as much here really made me stop and think how that was even possible (aka, "am I wrong???"). Guess I simply was not familiar with `then()` ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)).

- A SO post asking the same thing: https://stackoverflow.com/questions/65319822/how-does-knex-await-execute-a-database-query
- a little knex query builder prototype that connects all the dots

BlairCurrey: I've found there to be a lot of gotchas with promises in the knex migrations. [Another important thing is returning from the migration](https://github.com/knex/knex/issues/3165#issuecomment-486646428) (which we are doing). We couldn't, for example, await all the queries and not return (something I've done).

I've also seen/tried using `Promise.All` like this but I don't think it would guarantee resolution order because they run concurrently (ie, second might actually modify the db before the first):

```ts
return Promise.All([
  knex.raw(...),
  knex.raw(...)
])
```

Usually I actually use the dreaded promise chaining:

```ts
return knex.raw(...)
 .then(()=>{
   knex.raw(...)
 })
```

Ugly, I know. But clear.

The way it is now seems fine - just wanted to chime in with other general pitfalls I've run into.

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
golobitch: do we really want to unrevoke the key? if we do this, we lose the information that sometime in the past, this key was revoked. I think that we need to have new entry in the database if revoked key was again added

packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js
golobitch: @sabineschaller No, I meant that `knex.raw` needs to be awaited

mkurapov:
@sabineschaller to double check, are we good to review this?

oana-lolea:
Changes:
- Added unrevoke wallet address key query resolver 
- Updated migration script to create a unique index on walletAddressId, kid and x, when the key is not revoked, because we can have the same key revoked multiple times (useful for keeping the key's history)
- Added tests for the service
- Added bruno file for unrevoking

oana-lolea:
Updates based on #3130.

1. Updated migration to: 
- delete all the duplicate active keys of a wallet address, keeping the most recent one.
- unique constraint is now on `(walletAddressId, kid)`, to avoid having duplicate keys.

2. Removed `unrevoke` resolver, because `create` does exactly what unrevoke is supposed to do: create a new row with a key that is not revoked. 

3. Tested that `UniqueConstraint` error is returned when trying to add the same key that is not revoked.

packages/backend/src/open_payments/wallet_address/key/service.test.ts
```diff
@@ -129,11 +146,48 @@ describe('Wallet Address Key Service', (): void => {
...
+      assert.ok(!isWalletAddressKeyError(key))
+
+      expect(unrevokedKey).toMatchObject({
+        jwk: key.jwk,
+        walletAddressId: key.walletAddressId,
+        revoked: false
+      })
+    })
+
+    test('Returns DuplicateKey when the key already exists and it is unrevoked', async (): Promise<void> => {
```
mkurapov, line 180: Looks like you are already handling this in the "'cannot add duplicate key to a wallet address'" test.

packages/backend/src/open_payments/wallet_address/key/service.test.ts
```diff
@@ -129,11 +146,48 @@ describe('Wallet Address Key Service', (): void => {
...
 
       const revokedKey = await walletAddressKeyService.revoke(key.id)
       await expect(walletAddressKeyService.revoke(key.id)).resolves.toEqual(
         revokedKey
       )
     })
   })
+
+  describe('Unrevoke Wallet Address Keys', (): void => {
+    test('Creates a new key unrevoked', async (): Promise<void> => {
```
mkurapov, line 159: I think we can just add this to the "create" describe block

packages/backend/migrations/20241119125038_add_unique_constraint_to_kid.js
```diff
@@ -0,0 +1,37 @@
...
+      // Keep only one active walletAddressKey (the most recent one)
+      `DELETE FROM "walletAddressKeys" w
+        WHERE revoked = false
+        AND "createdAt" <> (
+          SELECT MAX("createdAt")
+          FROM "walletAddressKeys"
+          WHERE revoked = false
+          AND "walletAddressId" = w."walletAddressId"
+          GROUP BY "walletAddressId"
+);
```
mkurapov, line 17: This is a tricky one. One thing we need to add is the `kid` parameter (so we don't delete all non-revoked keys for a wallet address), and also maybe we can come up with a way to not delete by `createdAt` (although very unlikely, it could end up deleting valid, non-duplicated keys which happen to have the same createdAt).

oana-lolea: Yeah, I think that grouping by `kid` is the right way to do it.
```suggestion
      `DELETE FROM "walletAddressKeys" w
        WHERE revoked = false
        AND "createdAt" <> (
          SELECT MAX("createdAt")
          FROM "walletAddressKeys"
          WHERE revoked = false
          AND kid = w.kid
          GROUP BY kid
        );
```
Could we use `TOP` to delete just one key if there are multiple valid keys with the same `createdAt`? I don't see another way to determine which key to choose.

oana-lolea: This is a way to do it, but still based on `createdAt`. If there are multiple keys that have the most recent `createdAt`, we can keep the first one we find and delete the rest. What do you think, @mkurapov ?

```suggestion
      `DELETE FROM "walletAddressKeys" w
       WHERE revoked = false
       AND w."createdAt" = (
          SELECT "createdAt"
          FROM "walletAddressKeys"
          WHERE revoked = false
            AND kid = w.kid
          ORDER BY "createdAt" DESC
          LIMIT 1
       )
       AND w.id <> (
          SELECT id
          FROM "walletAddressKeys"
          WHERE revoked = false
            AND kid = w.kid
          ORDER BY "createdAt" DESC
          LIMIT 1 OFFSET 1
       );
      `
```

mkurapov: @oana-lolea I was looking through the PR history, and I think we can use physical location row `ctid` value:

https://github.com/interledger/rafiki/blob/5e8156950a4806565dbb9b54d67da3cdb9fdf4b3/packages/backend/migrations/20240731084359_add_unique_constraint_to_kid.js#L8-L13

I agree with you on keeping the most recent key so we'd have to adjust the migration it a bit, but that should work.

packages/backend/migrations/20241119125038_add_unique_constraint_to_kid.js
```diff
@@ -0,0 +1,35 @@
...
+  return knex
+    .raw(
+      // Keep only one active walletAddressKey (the most recent row)
+      `DELETE FROM "walletAddressKeys" w
+       WHERE revoked = false
+       AND ctid NOT IN (
+          SELECT MIN(ctid)
+          FROM "walletAddressKeys"
+          WHERE revoked = false
+            AND kid = w.kid
```
mkurapov, line 15: I think we need to group by `walletAddressId` and `kid` here still, otherwise we may delete keys that have the same `kid` but belong to a different wallet address.

Additionally, we should also use `MAX(ctid)` so we can grab the most recent key

packages/backend/migrations/20241119125038_add_unique_constraint_to_kid.js
```diff
@@ -0,0 +1,36 @@
...
+      // Keep only one active walletAddressKey (the most recent row)
+      `DELETE FROM "walletAddressKeys" w
+       WHERE revoked = false
+       AND ctid NOT IN (
+          SELECT MAX(ctid)
+          FROM "walletAddressKeys"
+          WHERE revoked = false
+            AND kid = w.kid
+          GROUP BY kid, "walletAddressId"
+       )`
```
mkurapov, line 17: Do we need 
```suggestion
      `DELETE FROM "walletAddressKeys" w
       AND ctid NOT IN (
          SELECT MAX(ctid)
          FROM "walletAddressKeys"
          WHERE revoked = false
          GROUP BY kid, "walletAddressId"
       )`
```
can you double check if this suggestion has the same behaviour? I _think_ it should

mkurapov (approved):

PR #2881, opened 2024-08-21
feat(backend): improve PSQL balance calculation

### LINKED ISSUE #2879: Improve PSQL account balance calculation
## Context
Currently, we are doing the PSQL balance calculation by fetching all of the transfers from the DB, and doing the calculation in-memory. This is not very efficient with many transfers.

We should move the balance lookup logic in PSQL accounting to a DB call instead of doing it in-memory.

mkurapov:

- For PSQL accounting, instead of calculating balances by fetching all transfers for an account and then doing the balance calculation in memory, do the calculation in PSQL
- Add constraint to ledgerAccounts table to prevent any PENDING transfers being created without an expiryDate (to avoid having "hanging" pending transfers that will never expire)

Running the perf test script with PSQL DB (graph showing transaction counts):

Before:

Screenshot 2024-08-20 at 21 13 00

<br>

After:

Screenshot 2024-08-20 at 21 43 19

## Context

Fixes #2879 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/accounting/psql/balance.ts
```diff
@@ -16,38 +14,45 @@ export async function getAccountBalances(
...
     }
-  }
+  } catch (err) {
+    deps.logger.error(
+      {
+        err,
+        accountId: account.id
+      },
+      'Could not fetch balances for account'
+    )
```
mkurapov, line 54: not doing much with the error here (not expecting it to happen), just logging the accountId we were trying to query

packages/backend/src/tests/ledgerTransfer.ts
```diff
@@ -36,9 +36,13 @@ export const createLedgerTransfer = async (
...
     debitAccountId: debitAccountId,
     amount: amount ?? 10n,
-    expiresAt,
     ledger,
     state: state ?? LedgerTransferState.POSTED,
+    expiresAt:
+      expiresAt ??
+      (state === LedgerTransferState.PENDING
+        ? new Date(Date.now() + 86_400_000)
+        : undefined),
```
mkurapov, line 45: this is only used for tests, we just want to avoid breaking the newly added constraint above

packages/backend/src/accounting/psql/balance.ts
```diff
@@ -16,38 +14,45 @@ export async function getAccountBalances(
   account: LedgerAccount,
   trx?: TransactionOrKnex
 ): Promise<AccountBalance> {
-  const { credits, debits } = await getAccountTransfers(
```
BlairCurrey, line 19: We can tackle this in another issue but can we get rid of `getAccountTransfers` now? I think it's only exposed in a recently added gql resolver. That method still brings the records into code and iterate over all of them (so like half of the total problem you fixed here). So I would still expect performance issues as transfer count grows. 

IDK if we want to:

A) leave it as-is because we dont think the performance issue will matter much. It's possible. I dont think the use-case for the resolver is well defined. I guess its just so people can see each transfer from the admin ui (not currently implemented and not sure what they would do with that info).
B) Remove it
C) Also rework that query. Not exactly sure how much we would leverage this one since its aggregating it a bit differently. Probably just a new similar one.

I guess I would say B or C but im not sure its worth the trouble of refactoring if we dont have a clear use case for consuming it.

packages/backend/migrations/20240820161040_add_ledger_transfer_constraint.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.table('ledgerTransfers', function (table) {
+    table.check(
+      `("state" != 'PENDING') OR ("expiresAt" IS NOT NULL)`,
```
BlairCurrey, line 8: Does the new query error if some record doesnt satisfy the constraint? Just wondering if we can guaruntee all existing records will pass this constraint... migration will fail otherwise. Im also wondering if the error is intelligible if we try to add one that doesnt pass this constraint. Not really sure where that would happen or how it should surface...

mkurapov: Checking rafiki money would be a good test, but I think it's not actually something that should have ever happened, this is how we map transfers to create them:

https://github.com/interledger/rafiki/blob/750f4148618990b3b0d85d1d987b39fc46c1ac19/packages/backend/src/accounting/psql/ledger-transfer/index.ts#L326-L342

i.e. if the timeout is provided, we set PENDING with expiryAt always

packages/backend/src/accounting/psql/balance.ts
mkurapov: > We can tackle this in another issue but can we get rid of getAccountTransfers now? I think it's only exposed in a recently added gql resolver

Yes, I think the gql is worth keeping and since its used more from dev/admin point of view and not something getting called every trx, its like you said, not a big deal, and we can leave as is for now.

Adding pagination (so we grab only a limited set of transactions) would be a good first step in improving it IMO

packages/backend/migrations/20240820161040_add_ledger_transfer_constraint.js
mkurapov: rafiki.money db:

```sh
rafiki_backend=# select count(*) from "ledgerTransfers" where state = 'PENDING' and "expiresAt" is null;
 count 
-------
     0
(1 row)

rafiki_backend=# 
```

packages/backend/src/accounting/psql/balance.ts
```diff
@@ -16,38 +14,45 @@ export async function getAccountBalances(
...
-  let debitsPosted = 0n
-  let debitsPending = 0n
-
-  for (const credit of credits) {
-    if (credit.state === LedgerTransferState.POSTED) {
-      creditsPosted += credit.amount
-    } else if (credit.state === LedgerTransferState.PENDING) {
-      creditsPending += credit.amount
+    if (queryResult?.rows < 1) {
+      throw new Error('No results when fetching balance for account')
```
golobitch, line 33: I was always wondered how do we decide if we should throw error, or just return empty response (in case of a balance, 0)?

BlairCurrey: I think your question still stands but I played around with it because I wasnt sure how it would work and wanted to add some context. When there are no records with the associated id it still aggregates a result, its just 0 for all the balances. So I think like Max mentioned in his other comment, this could probably never happen. But yeah I think it still begs the question, do we need to error or can we just return 0 balances.

BlairCurrey (approved):

packages/backend/src/accounting/psql/balance.ts
mkurapov: This check is if results of the query weren't able to be returned at all - so missing information unexpectedly, a result of a psql issue. So in that case we should break everything IMO

PR #2882, opened 2024-08-22
feat(docs): add header links

BlairCurrey:

- adds header links to docs headings. makes it easier to get link to specific sections by clicking on it (currently have to manually form the url). Style should be unchanged, other than underlining on hover.

https://github.com/user-attachments/assets/b9da2ebe-6065-486d-a65f-098a7599077d

## Context

I wanted to link to a specific section and found it annoying to do so without this.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/documentation/src/styles/rafiki.css
+++ b/packages/documentation/src/styles/rafiki.css
@@ -55,10 +55,11 @@
 }
 
 .sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a {
-  color: var(--sl-color-white);
+  color: inherit;
   text-decoration: none;
+  transition: color 0.2s ease-in-out;
+}
 
-  &:hover {
-    text-decoration: underline;
-  }
+.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a:hover {
+  color: var(--sl-color-accent-high);
 }
--- a/packages/documentation/src/styles/rafiki.css
+++ b/packages/documentation/src/styles/rafiki.css
@@ -53,3 +53,12 @@
 .accounting-table.accounting-table tr td {
   text-align: center;
 }
+
+.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a {
+  color: var(--sl-color-white);
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
--- a/packages/documentation/astro.config.mjs
+++ b/packages/documentation/astro.config.mjs
@@ -5,6 +5,8 @@ import remarkMath from 'remark-math'
 import rehypeMathjax from 'rehype-mathjax'
 import GraphQL from 'astro-graphql-plugin'
 import starlightLinksValidator from 'starlight-links-validator'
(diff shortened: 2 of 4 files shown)


JoblersTune:
@blair, thanks. I'm gonna wait for the doc updates to be in before addressing this because all our links are changing. But it's a good suggestion. So I'll pick it up when our new docs structure is good to go.

JoblersTune:
> @BlairCurrey @JoblersTune Exchange rates, webhook, fees, IdP... these all have their own page in the new doc structure. We might not need this issue, but def leave it open until the new structure is live just in case.

@melissahenderson I believe this is less about any specific item and more about an approach. I think what Blair is suggesting here is that we should make it easy to link directly to any heading on a page, not just the page itself. Once the docs are updated I'm happy to dive in and check it out.

mkurapov:
@melissahenderson @BlairCurrey @JoblersTune 

Should we add this to the new set of docs?

melissahenderson:
Was the issue originally because the page was in mobile view, so the page's table of contents didn't appear on the right side of the screen?

Screenshot 2024-09-13 at 12 38 17 PM

@BlairCurrey Would you have had this issue if the page's table of contents was more apparent?

Screenshot 2024-09-13 at 12 41 44 PM

JoblersTune:
I'm gonna take a proper look at this today

packages/documentation/src/styles/rafiki.css
JoblersTune, line 1: Thanks for this @BlairCurrey. It's great. Can we maybe just edit the styling a bit because otherwise we have an underline on top of our highlight underlines and it looks odd.

How about this instead?

```
/* Heading styles */
.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a {
  color: inherit;
  text-decoration: none;
  transition: color 0.2s ease-in-out;
}

.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a:hover {
  color: var(--sl-color-accent-high);
}
```

BlairCurrey:
> Was the issue originally because the page was in mobile view, so the page's table of contents didn't appear on the right side of the screen?

@melissahenderson I do not recall if I was in mobile view (or just smaller viewport on desktop). 

I didn't think to use the outline. And actually in a smaller viewport I didn't realize where it was. I see how the outline can be used to the same effect. Personally I'm just used to the convention of the headers being links - it feels like a more direct way to get the link when reading/scanning the document.

packages/documentation/src/styles/rafiki.css
BlairCurrey: makes sense.

JoblersTune:
Great, thanks. 

Any idea why the auth tests are failing?

BlairCurrey:
> Great, thanks.
> 
> Any idea why the auth tests are failing?

Something intermittent... tests timed out for some reason. Fine now after re-running.

JoblersTune (approved):

mkurapov (approved):

PR #2884, opened 2024-08-22
feat(backend): add trace to outgoing payment lifecycle

### LINKED ISSUE #2876: Add trace to outgoing payment worker
### Context

We want to be able to narrow down into all of the requests during one of our more important processes in Rafiki:
the outgoing payment lifecycle. Now that we have traces, we can achieve this by manually instrumenting a trace called something like `outgoing_payment_worker`. 

Essentially, we want a trace to cover the `handlePaymentLifecycle` method in `packages/backend/src/open_payments/payment/outgoing/worker.ts`.

We can use `tracerProvider` in the `packages/backend/src/telemetry/index.ts` which exposes `tracerProvider.getTracer(...).startActiveSpan(...)` to be able to manually instrument a span.

The docs for the function can be found [here](https://opentelemetry.io/docs/languages/js/instrumentation/#create-spans).

A consideration: because telemetry is optional, you might need to create the traceProvider outside of `if (Config.enableTelemetryTraces)` conditional

### Tasks
- [ ] Manually instrument a `outgoing_payment_worker` trace

njlie:

- Wraps the `handlePaymentLifecycle` function in a trace that can be consumed by Rafiki's telemetry services.

## Context

Closes #2876.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -74,7 +74,6 @@ services:
       WALLET_ADDRESS_URL: ${CLOUD_NINE_WALLET_ADDRESS_URL:-https://cloud-nine-wallet-backend/.well-known/pay}
       ILP_CONNECTOR_URL: ${CLOUD_NINE_CONNECTOR_URL:-http://cloud-nine-wallet-backend:3002}
       ENABLE_TELEMETRY: true
-      ENABLE_TELEMETRY_TRACES: true
       KEY_ID: 7097F83B-CB84-469E-96C6-2141C72E22C0
     depends_on:
       - shared-database
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -74,6 +74,7 @@ services:
       WALLET_ADDRESS_URL: ${CLOUD_NINE_WALLET_ADDRESS_URL:-https://cloud-nine-wallet-backend/.well-known/pay}
       ILP_CONNECTOR_URL: ${CLOUD_NINE_CONNECTOR_URL:-http://cloud-nine-wallet-backend:3002}
       ENABLE_TELEMETRY: true
+      ENABLE_TELEMETRY_TRACES: true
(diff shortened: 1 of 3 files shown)


njlie:
Regarding this statement from the issue:
> A consideration: because telemetry is optional, you might need to create the traceProvider outside of if (Config.enableTelemetryTraces) conditional

OpenTelemetry handles it [like so](https://opentelemetry.io/docs/languages/js/instrumentation/#initialize-tracing):
> If a TracerProvider is not created, the OpenTelemetry APIs for tracing will use a no-op implementation and fail to generate data.

Handling the outgoing payment lifecycle without the `enableTelemetryTraces` env variable will still run successfully.

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
       WALLET_ADDRESS_URL: ${CLOUD_NINE_WALLET_ADDRESS_URL:-https://cloud-nine-wallet-backend/.well-known/pay}
       ILP_CONNECTOR_URL: ${CLOUD_NINE_CONNECTOR_URL:-http://cloud-nine-wallet-backend:3002}
       ENABLE_TELEMETRY: true
+      ENABLE_TELEMETRY_TRACES: true
```
mkurapov, line 77: I think since we have this in the telemetry docker compose we can keep it turned off here IMO

packages/backend/src/open_payments/payment/outgoing/worker.ts
```diff
@@ -63,16 +64,24 @@ async function handlePaymentLifecycle(
   deps: ServiceDependencies,
   payment: OutgoingPayment
 ): Promise<void> {
-  if (payment.state !== OutgoingPaymentState.Sending) {
-    deps.logger.warn('unexpected payment in lifecycle')
-    return
-  }
+  const tracer = trace.getTracer('outgoing_payment_worker')
```
mkurapov, line 67: actually, let us also include the code where we actually fetch the outgoing payment from the DB (it's happening in `getPendingPayment`)

BlairCurrey (approved):

Issue #2852, opened 2024-08-08
Try to re-request a grant if token rotation fails during remote incoming payment creation


mkurapov:
## Context
Whenever `createReceiver` is called in the Rafiki GraphQL API, and the wallet address is belongs to a remote Rafiki (ie the payment is not a local payment), the underlying code uses the Open Payments client to get an incoming payment grant at the auth server of the receiver's wallet address. Then, once it gets the grant, the auth token in stored to be used for subsequent requests, ie, if createReceiver is called again for the wallet address with the same auth server, the auth token is reused. When the token is expired, a request to rotate the token is made.

However, we sometimes get into an issue when the receiver's auth server does _not_ have the corresponding auth token stored, for example, if the database has been reset. In that case, when the sending Rafiki tries to make a request to rotate the token for the incoming payment grant, we receive an error: https://github.com/interledger/rafiki/blob/4a43281f266bac2439665cf48f2675ee08d43b09/packages/backend/src/open_payments/payment/incoming_remote/service.ts#L123-L136

At this point ^, instead of throwing an error, we should just mark the old grant as "deleted" (ie with a new `deletedAt` field, or by setting expiredAt to "now"), and request a brand new grant at the auth server. Make sure to then update how we fetch the grants going forward, to make sure we always get the most recent grant/accessToken for a particular auth server.

## Todos
- [ ] If token rotation fails when trying to get a grant for remote incoming payment creation, create a new grant instead
- [ ] Update how we fetch grants from the DB to make sure we always use the most recent grant for an auth server

PR #2887, opened 2024-08-22
feat(backend): start of handling grant token rotation more gracefully

mkurapov:

- Adding `getOrCreate` and `delete` methods to `backend` grant service:
    - `getOrCreate` method:
        - soft deletes a grant if token rotation fails
        - uses the open payments client to request grant and store it instead of having it be the responsibility of the caller
    - `delete` method
        - allows callers to soft delete a grant if necessary
    - **these methods are still unused at this point, I will be updating callers to use them in a later PR, so we can break up the changes**
- Adding `deletedAt` field to the grants table to allow soft deletion, as well as dropping the unique constraint on `authServerId` `accessType` and `accessActions` such that we can properly soft delete problematic grants 
       
    
       
## Context

- This is the first part of #2852:
    - We are making the grant service more resilient to edge cases (like being unable to rotate tokens if they are missing from the receivers Rafiki DB)

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/migrations/20240820101148_drop_unique_grants.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('grants', function (table) {
+    table.dropUnique(['authServerId', 'accessType', 'accessActions'])
```
mkurapov, line 7: Dropping the unique constraint on authServerId accessType and accessActions such that we can properly soft delete grants (see migration below)

packages/backend/src/app.ts
```diff
@@ -232,6 +234,8 @@ export interface AppServices {
   incomingPaymentService: Promise<IncomingPaymentService>
   remoteIncomingPaymentService: Promise<RemoteIncomingPaymentService>
   receiverService: Promise<ReceiverService>
+  grantService: Promise<GrantService>
+  authServerService: Promise<AuthServerService>
```
mkurapov, line 238: weren't included before, adding here so we get proper types

packages/backend/src/open_payments/grant/service.ts
```diff
@@ -1,16 +1,26 @@
...
+  isPendingGrant
+} from '@interledger/open-payments'
+import { GrantError } from './errors'
 
 export interface GrantService {
   create(options: CreateOptions): Promise<Grant>
   get(options: GrantOptions): Promise<Grant | undefined>
   update(grant: Grant, options: UpdateOptions): Promise<Grant>
+  getOrCreate(options: GrantOptions): Promise<Grant | GrantError>
+  delete(id: string): Promise<Grant>
```
mkurapov, line 18: still unused

packages/backend/src/open_payments/grant/service.ts
```diff
@@ -89,6 +101,140 @@ async function updateGrant(
...
+  deps: ServiceDependencies,
+  options: GrantOptions
+): Promise<Grant | undefined> {
+  return await Grant.query(deps.knex)
+    .findOne({
+      accessType: options.accessType
+    })
+    .whereNull('deletedAt')
+    .andWhere('authServer.url', options.authServer)
+    .andWhere('accessActions', '@>', options.accessActions) // all options.accessActions are a subset of saved accessActions
```
mkurapov, line 134: e.g. if we have `[AccessActions.Create, AccessActions.Complete, AccessActions.Read]` and we are trying to getOrCreate a grant with `[AccessActions.Create]` we will properly return it, but if one of the elements in the requested list doesnt exist, we will not be able to fetch it

packages/backend/src/app.ts
BlairCurrey: easy to miss this

packages/backend/src/open_payments/grant/service.ts
BlairCurrey: It looks like this differs slightly from the `getGrant` which finds one that has strictly equal `accessActions`.  Its not quite clear what the ramifications of this will be. Can you speak to this difference?

packages/backend/migrations/20240820101148_drop_unique_grants.js
BlairCurrey: seems similar to this case? https://github.com/interledger/rafiki/pull/2863#discussion_r1720065804

So could we use a partial unique index (where `deletedAt` is not null)?

packages/backend/src/open_payments/grant/service.ts
mkurapov: before, if we already had a grant for `[Read, Create]` and then we were trying to getOrCreate a grant with `[Read]`, it would fail to get the existing one, even though the existing grant has the sufficient access requirements

BlairCurrey: I suppose we're getting rid of the `getGrant` here then (in subsequent PR)? looks like its unused but if we keep it for some reason I guess we should update to filter out by `deletedAt` (and maybe this same `accessActions` where clause). But I imagine we can just remove it.

mkurapov: Yes exactly, will be removing the unused methods

packages/backend/migrations/20240820101148_drop_unique_grants.js
mkurapov: I don't think this unique constraint is very useful to begin with, particularly because order matters in `accessActions` array, meaning [Create, Read] is different from [Read, Create]

packages/backend/migrations/20240820101201_add_deleted_at_grants.js
```diff
@@ -0,0 +1,19 @@
+/**
```
njlie, line 1: nit: why not combine this migration with `drop_unique_grants`?

mkurapov: Yup, makes sense, let me do that

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -201,4 +219,400 @@ describe('Grant Service', (): void => {
...
+        accessType: AccessType.IncomingPayment,
+        accessActions: [AccessAction.ReadAll]
+      }
+
+      await expect(
+        getExistingGrant({ knex } as ServiceDependencies, options)
+      ).resolves.toEqual({ ...existingGrant, authServer })
+    })
+
+    test('gets existing grant (requested actions are a subset of saved actions)', async () => {
```
njlie, line 485: Would be necessary to create another test (or modify this one) to ensure that it can retrieve existing grants that have (Read/List)-All rights when (Read/List) is requested?

mkurapov: good point, maybe during create if ReadAll/ListAll is requested, we can also manually add the corresponding Read/List as well

njlie (approved):
LGTM

PR #2888, opened 2024-08-22
refactor(backend): handle grant lookup more gracefully in remote incoming payment service

### LINKED ISSUE #2852: Try to re-request a grant if token rotation fails during remote incoming payment creation

mkurapov:

A new `grantService.getOrCreate` function was created in #2887, that, besides storing grant information, does all of the actual Open Payments grant requests with the OP client (as well as doing token rotation for those grant tokens as necessary).

Previously, the logic for requesting grants and doing token rotation was repeated in the `remoteIncomingPaymentService` as well as the `receiverService`. Both services were doing the same thing: using the OP client to do grant requests and do token rotation. Now, since this logic is consolidated in `grantService.getOrCreate`, we are able to clean up those respective services: 
- `remoteIncomingPaymentService` calls `grantService.getOrCreate` in both of its `create` and `get` methods, which allows us to clean it up considerably
- `receiverService` just has to call `remoteIncomingPaymentService.get/create` to get & create remote incoming payments (which under the hood handles the grant management), which allows us to to clean up _its_ logic

Additionally, in the `remoteIncomingPaymentService`, there is additional logic that is a bit more graceful in its incoming payment requests: if we fail to get or create an incoming payment with an existing grant (by checking for 401 or 403 errors), we delete the old grant, and try a request once more.

Finally, we remove old external grantService methods which are no longer used after these changes.

## Context

Fixes #2852 

## Checklist

- [x] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -57,114 +63,209 @@ async function create(
...
+  }: {
+    createArgs: CreateRemoteIncomingPaymentArgs
+    walletAddress: WalletAddress
+    retryOnTokenError?: boolean
+  }
+): Promise<
+  OpenPaymentsIncomingPaymentWithPaymentMethods | RemoteIncomingPaymentError
+> {
+  const resourceServerUrl =
+    walletAddress.resourceServer ?? new URL(walletAddress.id).origin
```
mkurapov, line 102: We don't really need `new URL(walletAddress.id).origin` at this point, but being defensive

packages/backend/src/open_payments/payment/incoming_remote/service.ts
```diff
@@ -57,114 +63,209 @@ async function create(
...
-    deps.logger.error({ err, walletAddressUrl }, errorMessage)
+
+    if (err instanceof OpenPaymentsClientError) {
+      if ((err.status === 401 || err.status === 403) && retryOnTokenError) {
+        await deps.grantService.delete(grant.id) // force new grant creation
+
+        return createIncomingPayment(deps, {
+          createArgs,
+          walletAddress,
+          retryOnTokenError: false
```
mkurapov, line 141: we could make it such that we retry multiple times, but for our use case, we just want to retry with a new grant once

BlairCurrey (approved):

PR #2909, opened 2024-08-28
feat(backend): create operator kratos identity on backend startup

### LINKED ISSUE #2903: Automatically create operator-level identity in Kratos on startup
When a multi-tenant instance of Rafiki starts up, it should automatically create a user role that represents the operator of the Rafiki instance. We'll call this role `operator`, and represent it in the Kratos Identity object as an admin-only metadata field where the field name is the name of that role, and its value should be `true`. The backend should not accept an identity as an operator unless the value of its field in the metadata is `true`.

njlie:

- Kratos is no longer optional when composing the local environment from pnpm script
  - TODO later: make Kratos not optional at all levels
- On startup, `backend` service generates a Kratos identity with the metadata property `operator: true`, which will be used to determine if the identity is an operator versus a tenant of the Rafiki instance.
- Added `KRATOS_ADMIN_EMAIL` and `KRATOS_ADMIN_URL` environment variables to the `backend` package.

## Context

Fixes #2903.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


golobitch (approved):
LGTM

PR #1812, opened 2023-09-01
chore: build images once

raducristianpopa:
## Changes proposed in this pull request

With these changes, on a clean localenv start-up we will only build the local images once.

Total time spent executing `pnpm localenv:compose up -d`:
- branch `main`: 6.68s user 8.89s system 9% cpu 2:37.82 total
- branch `rp--localenv-docker-images` 4.45s user 5.88s system 6% cpu 2:27.95 total

Despite the fact that all `happy-life-bank` services rely on those of `cloud-nine-wallet`, it does not impact the execution time. The 10s difference might be due to some network delays during image building/pulling.
## Context

- closes #1811 

## Checklist

- [x] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Documentation added
- [ ] Make sure that all checks pass
- [ ] Postman collection updated

PR #2911, opened 2024-08-28
chore: bump seed timeout

BlairCurrey:

- bumps seed timeout from about 4 seconds to about 30 seconds

## Context

Noticed on a slower-running machine that it was timing out. And those of us on M1 macs have noticed it occasionally as well.

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -80,7 +80,7 @@ services:
       - shared-database
       - shared-redis
     healthcheck:
-      test: ["CMD", "wget", "http://localhost:3001/healthz"]
+      test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
       interval: 5s
       retries: 10
       timeout: 3s
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -81,10 +81,10 @@ services:
       - shared-redis
     healthcheck:
       test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
-      start_period: 90s
+      start_period: 60s
       start_interval: 5s
-      interval: 5m
-      retries: 10
+      interval: 30s
+      retries: 1
       timeout: 3s
   cloud-nine-auth:
     hostname: cloud-nine-wallet-auth
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -81,7 +81,7 @@ services:
       - shared-redis
     healthcheck:
       test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
-      start_period: 60s
+      start_period: 90s
       start_interval: 5s
       interval: 5m
       retries: 10
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -81,7 +81,9 @@ services:
       - shared-redis
     healthcheck:
       test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
-      interval: 5s
+      start_period: 60s
+      start_interval: 5s
+      interval: 5m
       retries: 10
       timeout: 3s
   cloud-nine-auth:
--- a/localenv/cloud-nine-wallet/docker-compose.yml
+++ b/localenv/cloud-nine-wallet/docker-compose.yml
@@ -29,7 +29,8 @@ services:
       - ../cloud-nine-wallet/seed.yml:/workspace/seed.yml
       - ../cloud-nine-wallet/private-key.pem:/workspace/private-key.pem
     depends_on:
-      - cloud-nine-backend
+      cloud-nine-backend:
+        condition: service_healthy
   cloud-nine-backend:
     hostname: cloud-nine-wallet-backend
     image: rafiki-backend
@@ -78,6 +79,11 @@ services:
     depends_on:
       - shared-database
       - shared-redis
+    healthcheck:
+      test: ["CMD", "wget", "http://localhost:3001/healthz"]
+      interval: 5s
+      retries: 10
+      timeout: 3s
   cloud-nine-auth:
     hostname: cloud-nine-wallet-auth
     image: rafiki-auth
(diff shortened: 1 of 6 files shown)


packages/auth/src/index.ts
```diff
@@ -325,7 +325,7 @@ const callWithRetry: CallableFunction = async (
   try {
     return await fn()
   } catch (e) {
-    if (depth > 7) {
```
mkurapov, line 328: I think we can remove this one and the backend one this is the configuration for the migration files. 
We are just worried about the ASE in this change, if I understand correctly

BlairCurrey: yes, removed

BlairCurrey:
> @BlairCurrey
> 
> what do you think instead of increasing the retry timeout, we add a healthcheck to the backend service (how you do in the integration tests), and then add a depends_on [service_healthy](https://docs.docker.com/compose/startup-order/) to the mock ASEs? then we for sure know the seeding will wait until the service is ready

I'll try that - sounds better.

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -78,6 +79,13 @@ services:
     depends_on:
       - shared-database
       - shared-redis
+    healthcheck:
+      test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
+      start_period: 60s
+      start_interval: 5s
+      interval: 5m
+      retries: 10
+      timeout: 3s
```
BlairCurrey, line 88: Some things to note here:

First, the healthcheck does _**not**_ stop after it succeeds. I noticed this after seeing lots of logs for this gql operation after startup and successful seeding ([a SO post observing the same thing](https://devops.stackexchange.com/questions/13463/nginx-healthcheck-does-not-stop)).

This is different than I expected and why I specified the start period/interval. This checks more upfront then backs off after 60s (and maybe after the first success?) I think the start period probably needs to be greather than `start_interval` * `retries` (ensures that the start is unhealthy within the `start_period`).

Second, we use `wget` because there is no curl, and the endpoint is [what apollo sever suggests for a healthcheck](https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/). [I initially added](https://github.com/interledger/rafiki/pull/2911/commits/1316e74dcb23d809e20d82925cd26708129a0600#diff-5b481c1a1464d3e282df7102a0cb5882701fc2b2af81037e203812e13acf0176L376-L378) a new `/healthz` path to the koa server which we mount the apollo server to but figured the apollo server itself was a better, more direct check.

mkurapov: Is it a log like this? to check that its working?

```
{"level":30,"time":1726148987820,"pid":42,"hostname":"cloud-nine-wallet-backend","requestId":"53de7860-30f7-46bb-8548-aca479dfb9de","operation":null}

```

mkurapov: So it keeps making the requests even if its the start_period?

localenv/happy-life-bank/docker-compose.yml
```diff
@@ -25,8 +25,10 @@ services:
       - ../happy-life-bank/seed.yml:/workspace/seed.yml
       - ../happy-life-bank/private-key.pem:/workspace/private-key.pem
     depends_on:
-      - cloud-nine-mock-ase
-      - happy-life-backend
+      happy-life-backend:
+        condition: service_healthy
+      cloud-nine-mock-ase:
+        condition: service_started
```
mkurapov, line 31: I wonder why we had the `cloud-nine-mock-ase` dependency originally, do we need it?

raducristianpopa: This was introduced in https://github.com/interledger/rafiki/pull/1812. All `happy-life` services depend on the `cloud-nine` ones to avoid building the same images twice. If we remove the `depends_on`, on a fresh localenv startup all the images will be built twice since they are not found locally.

mkurapov: Got it, thanks @raducristianpopa !

localenv/cloud-nine-wallet/docker-compose.yml
BlairCurrey: yeah thats the log.

> So it keeps making the requests even if it succeeded the first time during the start_period?

and yes it keeps making the requests

30s would be fine, just figured we didn't really care about it after the initial one gating the MASEs (hence made it much longer).

mkurapov: In that case, we can just reduce the # of retries? maybe 30s interval, 5 times?

BlairCurrey: what does that accomplish? after it succeeds it will still send the healtcheck for each interval. lowering the retries just makes the threshold lower for sending out a signal that it failed to start

mkurapov: Is it possible that we don't have any successful starts during the `start_period`, and then we have to wait 5 mins for the next check?

BlairCurrey: Normally no because we're doing 10 retries and checking every 5s (50s + time to make the http request for each) but looking at it again, yes maybe. 

The start interval is 5s and the retries are 10 so within the 60s start period it should fail or succeed if the duration of each healthcheck is not >2s. Which should normally be the case (if its not up its an ECONREFUSED and it fails fast - doesn't hang). Currently the timeout is 3s though, so if it hung for 3s each time I think that could happen.

BlairCurrey: I bumped the start period to 90s. so even if all 10 healthchecks take 8 seconds (interval + timeout) then it will fail within the start period.

mkurapov: Looks like the number of retries are not considered during the start up time ([doc](https://docs.docker.com/reference/dockerfile/#healthcheck)):

> start period provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries.

This means there will be 10 retries after the start up time of 90s, so we just need to determine how long it should take after the startup time to keep trying the check

localenv/happy-life-bank/docker-compose.yml
```diff
@@ -69,6 +71,13 @@ services:
       KEY_ID: 53f2d913-e98a-40b9-b270-372d0547f23d
     depends_on:
       - cloud-nine-backend
+    healthcheck:
+      test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
```
mkurapov, line 75: To confirm, this is just making an empty GQL request > what does this header do?

mkurapov (approved):

localenv/happy-life-bank/docker-compose.yml
BlairCurrey: > To confirm, this is just making an empty GQL request > what does this header do?

Yes, empty gql request. the header is to ensure CSRF protection doesnt block it. Took this request from apollo server docs here which mentions the header: https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/

PR #2937, opened 2024-09-02
feat(backend): add tenant name

golobitch:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added

### DIFF
--- a/packages/backend/src/tenant/service.ts
+++ b/packages/backend/src/tenant/service.ts
@@ -11,11 +11,13 @@ import {
 import { v4 as uuidv4 } from 'uuid'
 import { Pagination, SortOrder } from '../shared/baseModel'
 import { EndpointOptions, TenantEndpointService } from './endpoints/service'
-import { TenantEndpoint } from './endpoints/model'
+import { isTenantEndpointError } from './endpoints/errors'
+import { tr } from '@faker-js/faker'
 
 export interface CreateTenantOptions {
-  idpConsentEndpoint: string
+  name: string
   idpSecret: string
+  idpConsentEndpoint: string
   endpoints: EndpointOptions[]
 }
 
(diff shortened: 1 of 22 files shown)


packages/backend/src/tenant/service.ts
```diff
@@ -85,16 +84,13 @@ async function createTenant(
...
     let tenant: Tenant
     try {
-      // create tenant on backend
-      tenant = await Tenant.query(trx).insert({
-        kratosIdentityId: uuidv4()
-      })
+      const tenantData = {
+        name: options.name,
+        kratosIdentityId: uuidv4(),
+        endpoints: options.endpoints
```
mkurapov, line 90: this allows Objection to automagically create the endpoints in the tenantsEndpoints model, correct?

mkurapov:
Some generic comments:
- Are you thinking we need a separate tenantEndpoints table in order to track changes/deletions? Should the same be done for the tenant endpoints in auth?
- We probably don't need tenantId for grants in `backend`, since I think it is tenant agnostic (the backend grants are the stored grants that the Rafiki instance itself has gotten from Open Payment requests it has made to remote/other Rafiki instances)

packages/backend/src/tenant/service.ts
golobitch: correct

golobitch:
> Some generic comments:
> 
> * Are you thinking we need a separate tenantEndpoints table in order to track changes/deletions? Should the same be done for the tenant endpoints in auth?
> * We probably don't need tenantId for grants in `backend`, since I think it is tenant agnostic (the backend grants are the stored grants that the Rafiki instance itself has gotten from Open Payment requests it has made to remote/other Rafiki instances)

This design is a little bit premature. We should change `tenantEndpoints` to `tenantConfiguration`. Down the line we can have basically all ENVs that are tenant specific there. And that can be then configured from Admin UI. I think that we should have audit trail for data change on the tenant, but audit is completely different feature and I think that we will need to tackle that sooner or later.

TenantId is only mandatory for auth in order to get correct grant for tenant.

PR #2954, opened 2024-09-09
feat(backend): performance and caching poc

koekiebox:

- 

## Checklist

- [ ] Tests added/updated
- [ ] Documentation added


packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -77,9 +79,10 @@ async function getIncomingPayment(
   deps: ServiceDependencies,
   options: GetOptions
 ): Promise<IncomingPayment | undefined> {
-  const incomingPayment = await IncomingPayment.query(deps.knex)
-    .get(options)
-    .withGraphFetched('[asset, walletAddress]')
+  const incomingPayment = await IncomingPayment.query(deps.knex).get(options)
+  await deps.assetService.setOn(incomingPayment)
+  await deps.walletAddressService.setOn(incomingPayment)
+
```
BlairCurrey, line 85: just for documentation's sake (dont necessarily need to do it here), Radu had a good suggestion for how to do this better.

We could override the `withGraphFetched` method on the model instead of manually retrieving/setting asset, wallet etc. everywhere. Would take some parsing of the original arguments or maybe some new arg to specify what to get off the cache.

mkurapov: I think doing explicitly is better, overriding `withGraphFetched` might be a bit too hidden/"magic". Probably long term solution is just doing something like repository pattern and abstracting over ORM/db access, with an ability to add caching in front as well

mkurapov:
- Closing in favour of #2999, #3041

PR #2969, opened 2024-09-11
fix(backend): fix OP resource tests

### LINKED ISSUE #2968: [Multi-Tenant] Fix feature branch tests for Open Payments Resources
A lot of tests were broken by the fact that several existing tables have added foreign key constraints on `tenantId`, fix the tests for the models/services/routes for the Open Payments resources:

- [x] Fix tests on Incoming Payments (`/backend/src/open_payments/payment/incoming`)
- [x] Fix tests on Remote Incoming Payments (`/backend/src/open_payments/payment/incoming_remote`)
- [x] Fix tests on Quotes (`/backend/src/open_payments/quote`)
- [x] Fix tests on Outgoing Payments (`/backend/src/open_payments/payment/outgoing`)
- [x] Fix tests on Wallet Addresses (`/backend/src/open_payments/wallet_address`)
- [x] Fix tests on Combined Payments (`/backend/src/open_payments/payment/combined`)

njlie:

- Adds an `operatorApiSecret` whose inclusion the headers will allow seeding to take place during localenv startup
- Fixes the test environment such that the test app can start up with the Kratos changes
- Fixes the tests for quotes, incoming payments (local & remote), outgoing payments on `backend`

## Context

Fixes #2968.

This is part of a greater effort to bring the multi-tenant feature branch back up to par with the standards of the main branch, as a lot of tests and github actions are failing due to changes to the database that add foreign key constraints to many of the tables.

This PR fixes the tests for the Open Payments resources, as well as sets up some fixes for the test environment that gets set up by many of the tests.

Overall story is captured here: https://github.com/interledger/rafiki/issues/2972

## Checklist

- [x] Tests added/updated


packages/auth/src/index.ts
```diff
@@ -105,42 +105,6 @@ export function initIocContainer(
     }
   )
 
-  container.singleton(
```
njlie, line 108: For some reason, probably due to rebasing, this got duplicated a few times.

packages/backend/migrations/20240827115832_update_resources_with_tenantId.js
```diff
@@ -20,10 +20,6 @@ exports.up = function (knex) {
       table.uuid('tenantId').notNullable()
       table.foreign('tenantId').references('id').inTable('tenants')
     })
-    .table('grants', function (table) {
```
njlie, line 23: Grant's don't need `tenantId` - they're never queried on the backend by that filter, and the tenant it is for can be determined by the auth server URL path. Having the requirement explicitly in this table introduces some backwards compatibility issues.

packages/backend/src/graphql/schema.graphql
```diff
@@ -1071,8 +1071,6 @@ input CreateQuoteInput {
   receiver: String!
   "Unique key to ensure duplicate or retried requests are processed only once. See [idempotence](https://en.wikipedia.org/wiki/Idempotence)"
   idempotencyKey: String
-  "ID of the tenant"
```
njlie, line 1074: `tenantId` got removed from the quote resolver input for backwards compatibility issues and it can be retrieved from the wallet address anyways

packages/backend/src/open_payments/grant/service.test.ts
```diff
@@ -21,10 +21,6 @@ import {
 import { v4 as uuid } from 'uuid'
```
njlie, line 21: The modifications here are all removing the `tenantId` parameter from grant creation calls

packages/backend/src/graphql/schema.graphql
BlairCurrey: > it can be retrieved from the wallet address anyways

I think perhaps this is the better pattern. It's simpler and I don't think we need to verify against the requestor's expectation for what the `tenantId` should be. 

I remember you mentioning that before and not having much of an opinion but I came across it myself when dealing with checking access for the createQuote resolver and felt like passing it in wasn't really providing any value, and it does make the gql input less complicated (IE requiring it only for operators).

packages/backend/src/graphql/resolvers/tenant_endpoints.ts
```diff
@@ -14,45 +9,6 @@ export const mapTenantEndpointTypeToModelEndpointType = {
   [EndpointType.WebhookBaseUrl]: TenantEndpointType.WebhookBaseUrl
 }
 
-export const getTenantEndpoints: TenantResolvers<ApolloContext>['endpoints'] =
```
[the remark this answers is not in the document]
BlairCurrey: looks like it leaves the service `getTenantEndpointsPage` method unused... remove?

BlairCurrey (approved):

PR #2990, opened 2024-09-19
Docs: table updates

brad-dow:
**Current state**

- Wide tables encroach on right side navigation
- Line breaks are disrupting text formatted as code

**Proposed state**

- Remove breaks in tables on text formatted as code
- Add a table wrapper for wide tables (maybe all tables) that adds a horizontal scroll when a table extends past the main content area
- Set fixed height for the table wrapper for both desktop (500px) and mobile (300px)
- Add a sticky table header to ensure column labels stay visible in the case of a longer table

### DIFF
--- a/packages/documentation/src/content/docs/integration/requirements/idp.mdx
+++ b/packages/documentation/src/content/docs/integration/requirements/idp.mdx
@@ -83,38 +83,11 @@ The `result` query parameter will indicate the success or failure of the grant a
 The following table lists examples of the different possible response types on this endpoint.
 
 <div class='wide-table'>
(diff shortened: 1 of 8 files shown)


brad-dow:
I'm going to stop messing with this one for now, but at least it's an idea of some of the changes that can improve the tables.

The thing that is bugging me right now on this one is that if the window is resized to a width of 1280px or higher on a page with the table wrapper, a horizontal scroll is added to the bottom of the page.

brad-dow:
https://github.com/interledger/rafiki/pull/3007 solved this one

PR #2991, opened 2024-09-19
refactor(backend): add NoopTelemetryServiceImpl to make tel optional

### LINKED ISSUE #2980: Make `telemetryService` defined
### Context

We have a lot of cases in the `backend` code where we are checking whether telemetry service is defined or not. However, we should assume that each of the services in the containers are always defined.

In order to still respect the `ENABLE_TELEMETRY` and `ENABLE_TELEMETRY_TRACES` flags, all we can do is just to not register metric/trace providers in the telemetry initialization. 

### Todo
- [ ] Make telemetryService required, update dependents

BlairCurrey:

- makes the telemetry _service_ (not the actual collection) required and removes all the `if(telemetry)` checks
- changes otel collector log config to get rid of a lot of noise that was making logs unreadable

## Context

fixes #2980

## Checklist

- [x] Tests added/updated

### DIFF
--- a/localenv/telemetry/otel-collector-config.yaml
+++ b/localenv/telemetry/otel-collector-config.yaml
@@ -8,6 +8,8 @@ processors:
   batch:
 
 exporters:
+  logging:
+    loglevel: info
   debug:
     verbosity: detailed
   prometheus:
@@ -18,6 +20,9 @@ exporters:
       insecure: true
 
 service:
+  telemetry:
+    logs:
+      level: warn
   pipelines:
     metrics:
       receivers: [otlp]
--- a/packages/backend/src/telemetry/service.test.ts
+++ b/packages/backend/src/telemetry/service.test.ts
@@ -5,7 +5,13 @@ import { Config } from '../config/app'
 import { ConvertError, RatesService } from '../rates/service'
 import { TestContainer, createTestApp } from '../tests/app'
 import { mockCounter, mockHistogram } from '../tests/telemetry'
-import { TelemetryService } from './service'
+import {
+  createTelemetryService,
+  NoopTelemetryServiceImpl,
+  TelemetryService,
+  TelemetryServiceImpl,
+  TelemetryServiceDependencies
+} from './service'
 import { Counter, Histogram } from '@opentelemetry/api'
 import { privacy } from './privacy'
 import { mockRatesApi } from '../tests/rates'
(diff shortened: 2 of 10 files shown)


packages/backend/src/telemetry/service.test.ts
BlairCurrey, line 1: lots of changed lines but its mostly just adding a level of nesting - I didnt change any existing tests. just added the disabled describe block, which ensures the correct telemetry service implementation is used (real one or noop) according to whether telemetry is enabled.

packages/backend/src/telemetry/service.ts
```diff
@@ -32,25 +32,29 @@ export interface TelemetryService {
...
 const METER_NAME = 'Rafiki'
 
 export function createTelemetryService(
   deps: TelemetryServiceDependencies
 ): TelemetryService {
+  if (!deps.enableTelemetry) {
+    return new NoopTelemetryServiceImpl(deps)
+  }
   return new TelemetryServiceImpl(deps)
 }
```
BlairCurrey, line 55: Feel like I need to elaborate on this approach because it might deviate from the expectations in the issue a bit which said:

> In order to still respect the ENABLE_TELEMETRY and ENABLE_TELEMETRY_TRACES flags, all we can do is just to not register metric/trace providers in the telemetry initialization.

What we're doing here is adding a new implementation that just doesnt do anything and using that if telemetry is disabled. 

I can revise this if we don't like it but I think its working well.

I didn't do it as described for a couple reasons: 1) I don't think its possible - we arent registering metric/trace providers on service init. Traces are already registered outside the service so they seem to be a non-factor here. 2) I think even if we did we'd find ourselves in a scenario where we're calling methods like `incrementCounterWithTransactionAmount`, doing all that work (including network call) and then simply do nothing.


Alternatives considerd:
1. in the spirit of the original suggestion, we could simply gate the counter create in `getOrCreateCounter` (and histogram) behind if telemetry is enabled and returned undefined if disabled like:

```ts
  private getOrCreateCounter(name: string, options?: MetricOptions): Counter {
    let counter = this.counters.get(name)
    if (!counter) {
      counter = metrics.getMeter(METER_NAME).createCounter(name, options)
      this.counters.set(name, counter)
    }
    return counter
  }
  ```
  
  Then we just use optional chaining to do nothing if undefined:
  
  ```ts
    const counter = this.getOrCreateCounter(name)
    counter.?add(...)
  ```
  
But again, this has the problem of doing all the work in `incrementCounterWithTransactionAmount` for example only to not actually do anything at the end.

2. Just add an early return to `recordHistogram`/`incrementCounter` if telemetry is disabled. But would the person adding the next similar method remember to do the same?

Comment (mkurapov) [reply packages/backend/src/telemetry/service.ts:55]: >  1) I don't think its possible - we arent registering metric/trace providers on service init

I see what you mean, what happens to the existing metrics/counters/histogramts when we keep the original/existing telemetry service, but with `enableTelemetry` disabled?

Comment (mkurapov) [reply packages/backend/src/telemetry/service.ts:55]: Yes, I'm with you, having a No-op class is clearer, and we don't end up having any side effects (even if they don't really do anything).

File: packages/backend/src/telemetry/service.ts [hunk 2]
```diff
@@ -32,25 +32,29 @@ export interface TelemetryService {
...
 
 const METER_NAME = 'Rafiki'
 
 export function createTelemetryService(
   deps: TelemetryServiceDependencies
 ): TelemetryService {
+  if (!deps.enableTelemetry) {
+    return new NoopTelemetryServiceImpl(deps)
+  }
```
mkurapov, line 53: I _think_ I prefer having this `if (!deps.enableTelemetry)` check in `src/index.ts` instead of here, since it would seem clearer to me to understand how each class/dependency gets registered (kind of like tigerbeetle). Also wouldn't need to pass in `enableTelemetry: boolean`. NABD, just an option.

mkurapov (dismissed):

localenv/telemetry/otel-collector-config.yaml
```diff
@@ -18,6 +20,9 @@ exporters:
       insecure: true
 
 service:
+  telemetry:
+    logs:
+      level: warn
```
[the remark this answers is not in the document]
BlairCurrey: removes a lot of noise from the collector logs. watching the container logs with telemetry enabled is unreadable without it.

packages/backend/src/telemetry/service.ts
BlairCurrey: that makes sense - good to have the switch as high level as possible I think.

Changed to return the noop or real one based on `config. enableTelemetry` in `src/index.ts`. In doing so I realized we don't need to pass in all the deps to the noop one so I removed those as well.

mkurapov (approved):

PR #2999, opened 2024-09-23
feat(localenv): add performance metrics

### LINKED ISSUE #2983: Add additional performance metrics for local playground
### Context

- Let's keep the useful histograms & metrics & timers we added from https://github.com/interledger/rafiki/pull/2954

TODO:
- [ ] add startTimer method to telemetry service
- [ ] copy over metric collection
- [x] copy over dashboard changes from ww branch (including Jason's datasource variable)

BlairCurrey:

- added timer to metric service
- used new timer to power several new graphs tracking durations for most of the main paths in the `create-outgoing-payment.js` k6 script.

Partial view of local dashboard (best observed locally while running performance test):

image

## Context

fixes #2983

currently set to merge into [bc/2980/refactor-telemetry-optionality](https://github.com/interledger/rafiki/tree/bc/2980/refactor-telemetry-optionality) because it depends on those changes.
## Checklist

- [ ] Tests added/updated


packages/backend/src/graphql/resolvers/liquidity.ts
```diff
@@ -440,13 +440,25 @@ export const depositOutgoingPaymentLiquidity: MutationResolvers<ApolloContext>['
...
+    const stopTimerWh = telemetry.startTimer('wh_get_latest_ms', {
+      callName: 'webhookService.getLatestByResourceId'
+    })
     const event = await webhookService.getLatestByResourceId({
       outgoingPaymentId,
       types: [OutgoingPaymentDepositType.PaymentCreated]
     })
+    stopTimerWh()
     if (!event || !isOutgoingPaymentEvent(event)) {
+      stopTimer()
```
mkurapov, line 461: Something we don't necessary have to tackle in this PR, but what can we do when it comes to stopping timers during a happy path vs an error? is there something we can add to make that distinction in the metrics?

packages/backend/src/graphql/resolvers/liquidity.ts
```diff
@@ -455,23 +467,31 @@ export const depositOutgoingPaymentLiquidity: MutationResolvers<ApolloContext>['
...
+    })
     const paymentOrErr = await outgoingPaymentService.fund({
       id: outgoingPaymentId,
       amount: BigInt(event.data.debitAmount.value),
       transferId: event.id
     })
+    stopTimerFund()
+
     if (isFundingError(paymentOrErr)) {
+      stopTimer()
```
mkurapov, line 487: maybe a finally would be nice here to avoid calling `stopTimer()` multiple times

packages/backend/src/open_payments/payment/outgoing/worker.ts
```diff
@@ -79,10 +91,15 @@ async function handlePaymentLifecycle(
     return
   }
 
+  const stopTimer = deps.telemetry.startTimer('handle_sending_ms', {
+    callName: 'handleSending'
```
mkurapov, line 95: what do you think about prefixing `callName` with service name always?

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -102,32 +111,66 @@ async function createQuote(
...
   try {
+    const stopTimerReceiver = deps.telemetry.startTimer(
+      'quote_service_create_resolve_receiver_time_ms',
+      {
+        callName: 'resolveReceiver',
+        description: 'Time to resolve receiver'
+      }
+    )
     const receiver = await resolveReceiver(deps, options)
+    stopTimerReceiver && stopTimerReceiver()
```
mkurapov, line 134:
```suggestion
    stopTimerReceiver
```

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -43,6 +43,13 @@ async function getQuote(
...
 ): Promise<PaymentQuote> {
+  const stopTimerRates = deps.telemetry.startTimer(
+    'ilp_get_quote_rate_time_ms',
+    {
+      callName: 'rates',
+      description: 'Time to get rates'
+    }
+  )
   const rates = await deps.ratesService
     .rates(options.walletAddress.asset.code)
```
mkurapov, line 54: maybe best to just convert to a normal try-catch call?

mkurapov (commented):
A few comments

packages/backend/src/graphql/resolvers/liquidity.ts
BlairCurrey: Yeah we could have a `success: boolean` attribute on the underlying histogram. I was inspired by a pattern that was shown to me, and it had that.

packages/backend/src/payment-method/ilp/service.ts
BlairCurrey: converted from chaining to try/catch/finally

packages/backend/src/open_payments/payment/outgoing/worker.ts
BlairCurrey: messaged you on this with more details. seems like a good idea but if we want to make it a standard prefix I think maybe it needs to be more abstract than service name since not everything is called from outside the service method or is necessarily scoped to a service (gql resolvers,accountService name doesnt properly signify tb/psql, etc.)

packages/backend/src/graphql/resolvers/liquidity.ts
BlairCurrey: Not used currently, but it should be possible now. I updated the closure returned from startTimer to accept additionalAttributes which will be merged on the ones initially provided to `startTimer`. Usage example added to tests (and tests added for `startTimer` in general - noticed we didnt have any).

packages/backend/src/graphql/resolvers/liquidity.ts
BlairCurrey: Not used currently, but it should be possible now. I updated the closure returned from `startTimer` to accept `additionalAttributes` which will be merged on the ones initially provided to `startTimer`. Usage example added to tests (and tests added for `startTimer` in general - noticed we didnt have any).

packages/backend/src/open_payments/payment/outgoing/worker.ts
BlairCurrey: updated to prefix with a pattern like `[AbstractScopeName]:literalFunctionName`. 

for example: 
- `WebhookService:getResourceById`
- `Resolver:depositOutgoingPaymentLiquidity` 

Not every case is as clean as service calls where we have an obvious and unambiguous scope so I'm keeping it a little more abstract to ensure we can standardize.

packages/auth/src/graphql/schema.graphql
```diff
@@ -16,10 +16,7 @@ type Query {
   ): GrantsConnection!
 
   "Fetch a specific grant by its ID."
-  grant(
-    "Unique identifier of the grant."
-    id: ID!
-  ): Grant!
+  grant("Unique identifier of the grant." id: ID!): Grant!
```
BlairCurrey, line 19: some formatting issue made it into main somehow

mkurapov: If you merge in main this should go away

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -43,23 +43,50 @@ async function getQuote(
...
-      throw new PaymentMethodHandlerError('Received error during ILP quoting', {
-        description: 'Could not get rates from service',
-        retryable: false
-      })
+  const stopTimerRates = deps.telemetry.startTimer(
+    'ilp_get_quote_rate_time_ms',
+    {
+      callName: 'RateService:rates',
+      description: 'Time to get rates'
+    }
```
mkurapov, line 51: Should this be directly in the rates service instead? Or instead, called something like `PaymentMethod:ILP:getQuote:getRates`?

mkurapov (commented):
Minor comments

packages/backend/src/payment-method/ilp/service.ts
BlairCurrey: Hmm fair point. 

We have external timers that wrap function calls from the outside like this, and internal timers that start at the beginning of a function and stop at the end. I think in this case we want the external timer because the goal is to measure `getQuote` and it's parts (thats how this metric is graphed). In other words, we want to measure _this_ rates service call, not all of them.

I take your point on the name. I was thinking about it in the context of the graph, where the `PaymentMethod:ILP:getQuote` would be redundant. However, as soon as we want to time this same method somewhere else the name breaks down, so yeah... looks like that convention (scoped by where its called) is good.

mkurapov: > In other words, we want to measure _this_ rates service call, not all of them.

Yes, makes sense. We do have the histogram name `ilp_get_quote_rate_time_ms`, so that is easy to narrow down and trace in the code in any case. 

For the callName, just as long as we can get group metrics from a certain part of the codebase like `PaymentMethod:ILP:*` that's good enough.

mkurapov (approved):

PR #3001, opened 2024-09-24
feat: Allow configurable cookie SameSite header

lengyel-arpad85:
Added a new ENV flag (INTERACTION_COOKIE_SAME_SITE) that is unset by default and allows the ASE to set it to lax, strict, or none if they want.

## Context
Added mainly for control over this value in auth.rafiki.money, in order to be able to have payments work in an iframe, needed in WM Tools

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -13,7 +13,7 @@ import { ApolloServer } from '@apollo/server'
 import { ApolloServerPluginDrainHttpServer } from '@apollo/server/plugin/drainHttpServer'
 import { koaMiddleware } from '@as-integrations/koa'
 
-import { IAppConfig, SameSiteCookieProp } from './config/app'
+import { IAppConfig } from './config/app'
 import { addResolversToSchema } from '@graphql-tools/schema'
 import { GraphQLFileLoader } from '@graphql-tools/graphql-file-loader'
 import { loadSchemaSync } from '@graphql-tools/load'
@@ -345,8 +345,7 @@ export class App {
 
     const redis = await this.container.use('redis')
     const maxAgeMs = this.config.interactionExpirySeconds * 1000
-    const interactionSameSite = this.config
-      .interactionCookieSameSite as SameSiteCookieProp
+    const interactionSameSite = this.config.interactionCookieSameSite
     koa.use(
       session(
         {
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -13,7 +13,7 @@ import { ApolloServer } from '@apollo/server'
 import { ApolloServerPluginDrainHttpServer } from '@apollo/server/plugin/drainHttpServer'
 import { koaMiddleware } from '@as-integrations/koa'
 
-import { IAppConfig } from './config/app'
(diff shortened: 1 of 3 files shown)


BlairCurrey (commented):
This seems OK, although I'm not sure `strict` could possibly work. We aren't defaulting to strict here, but not sure if we should somehow disallow it.

`SameSite=strict` seems to work fine locally but I don't think it will work in production. Not sure of an easy way to test as it would require changing the domain names. Could use another opinion.

My current understanding:

1. make a grant request that requires interaction (`Grant Request Outgoing Payment` in bruno)
2. visit redirect link. auth server creates session and responds with cookie with `sessionId`. Technically the response sends the `Set-Cookie` header with instructions for the client to actually set the cookie, but the server's domain is considered the domain which set it.
3. confirm and auth server successfully receives cookie, evidenced by retrieving the nonce matching the query param in `finishInteraction`. 

I thought this would not work though since the origins are different (localhost:3030) vs localhost:3006)). Judging by the cookie in chrome dev tools it seems like these are considered the same domain (simply `localhost`). 

image

lengyel-arpad85:
Yes, you are correct strict would not work, because that implies same domain (and yes, portnumbers are not considered different domains).
But it's upto the ASE to set this value, and since the possible values include strict as well as a boolean value of 'true' that also is interpreted as strict I don't think we can disallow these (or maybe we can restrict the values to 'lax', 'none' & undefined ... not sure if the session property will allow it)

on another note, I see from your screenshot that Secure is not set ... is that because you were testing on http not https, or do we have to set that implicitly also ?

BlairCurrey:
> on another note, I see from your screenshot that Secure is not set ... is that because you were testing on http not https, or do we have to set that implicitly also ?

Yeah, I am testing on http. I suppose you're anticipating problems with setting `SameSite` to `None`, since it must have the secure flag. If you look at the cookie on testnet its secure, so I think we're good there and we dont need to explicitly set it. 

image

BlairCurrey:
> But it's upto the ASE to set this value, and since the possible values include strict as well as a boolean value of 'true' that also is interpreted as strict I don't think we can disallow these (or maybe we can restrict the values to 'lax', 'none' & undefined ... not sure if the session property will allow it)

I changed the type to `export type SameSiteCookieProp = 'lax' | 'none' | undefined` and not getting any ts errors since its a subset of the valid arg type... I think it makes sense to disallow `strict` since we know it wont work. This set of valid values goes beyond our little env var parsing pattern a bit... I think something like this would work:

```ts
function envEnum<T extends string>(
  name: string,
  allowedValues: T[],
  defaultValue: T | undefined
): T | undefined {
  const envValue = process.env[name]
  if (envValue && allowedValues.includes(envValue as T)) {
    return envValue as T
  }
  return defaultValue
}
```

```ts
  interactionCookieSameSite: envEnum(
    'INTERACTION_COOKIE_SAME_SITE',
    ['lax', 'none'],
    undefined
  ),
  ```
I guess this presumes we don't want to error if they set an invalid value...

lengyel-arpad85:
I updated the code based on the recommendation; went with the envEnum, seems much cleaner to me than the inline solution

BlairCurrey (approved):

Issue #2951, opened 2024-09-05
Docs: Fix markdown tables


melissahenderson:
Most (not all) of the markdown tables look pretty bad. This is especially the case where the first column contains long variable names. 

We need to convert these tables into HTML tables. But first, we need to figure out how we want to style the tables so they're consistent across pages.

brad-dow:
**TEXT FORMATTED AS CODE LINE BREAKS**
I think we can remedy these tables in a way where we don't have to use raw HTML. After doing a little bit of digging on the project, I can see that the main ILF-docs.css file in the core docs-design-system has a line that defines way lines break specifically when the text is formatted as code in a table:

> table code {
  word-break: break-all;
}

There are a few ways to handle this:

- We can modify the docs-design-system CSS to not break up text formatted as code in a table.
- If this is only a Rafiki problem (which I don't think it is), we can alter the rafiki.css to override that style

If we want to pursue the first option, we'll need to update the docs-design-system CSS and then update all of the other docs pages' dependencies in package.json

If we want to handle it within rafiki.css, it's as simple as making updates as seen in this branch: https://github.com/interledger/rafiki/tree/docs-table-updates

brad-dow:
**TABLES THAT EXTEND PAST THE MAIN CONTENT AREA**
In playing around with some markdown tables, I noticed that a table will extend into the right sidebar navigation area. By wrapping the markdown table in a custom div class, we can alter this behavior with a few CSS tweaks. 

To see this functionality, run this branch locally: https://github.com/interledger/rafiki/tree/docs-table-updates

brad-dow:
I've created a PR of my work thus far for @JoblersTune to review and enhance when she gets the chance.

https://github.com/interledger/rafiki/pull/2990

PR #3007, opened 2024-09-30
docs: add custom styles for wider tables

huijing:

- Adds custom styling for tables that overflow to display content better
- Check the following pages:
    - https://deploy-preview-3007--brilliant-pasca-3e80ec.netlify.app/overview/concepts/telemetry/
    - https://deploy-preview-3007--brilliant-pasca-3e80ec.netlify.app/integration/requirements/idp/
    - https://deploy-preview-3007--brilliant-pasca-3e80ec.netlify.app/resources/environment-variables/

## Context

Fixes https://github.com/interledger/rafiki/issues/2951
Supersedes https://github.com/interledger/rafiki/pull/2990

## Checklist

- [ ] Tests added/updated


packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
 }
 
-.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a {
-  color: inherit;
-  text-decoration: none;
-  transition: color 0.2s ease-in-out;
+/* Overflowing table styles */
+.overflow-table {
+  overflow-x: scroll;
+  background:
```
huijing, line 71: This adds a bit of a curved shadow that only shows up when there is something to scroll, to indicate to users that there is something to scroll. Turns off when you scroll to the end.

packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
+    0 0,
+    100%,
+    0 0,
+    100%;
+  background-attachment: local, local, scroll, scroll;
 }
 
-.sl-markdown-content :is(h1, h2, h3, h4, h5, h6) > a:hover {
-  color: var(--sl-color-accent-high);
+@media screen and (min-width: 80rem) {
```
huijing, line 100: Weird hardcode to prevent super long code-lines from not obeying max-width of the content container.

packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
+  .overflow-table {
+    max-width: 755px;
+  }
+}
+
+.overflow-table table code {
```
huijing, line 106: Overrides the DDS word-break: break-all
I suggest keeping these styles here in Rafiki until some other docs site has such tables. For now, only Rafiki's tables have these long code lines in the table.

packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
+  white-space: nowrap;
+}
+
+.overflow-table table tbody tr:nth-child(2n):not(:where(.not-content *)) {
+  background-color: initial;
```
huijing, line 111: The overflow shadow effect only works if the rows do not have background colour.

packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
+}
+
+.overflow-table tbody tr:not(:last-of-type) td {
+  border-block-end: 1px solid var(--sl-color-gray-6);
```
huijing, line 115: Since the background colour was removed, a bottom border helps make the rows look distinct from each other.

packages/documentation/src/styles/rafiki.css
```diff
@@ -54,12 +65,60 @@
...
+}
+
+.wider-160-table.wider-160-table table {
+  width: 160%;
```
huijing, line 119: The tables which have normal text (that breaks normally) in certain columns need to be wider than 100% for the column itself to be wider.
How wide is specific to a table's contents, so these odd helper classes are here.

melissahenderson (approved):
Looks good from my side

JoblersTune:
@huijing how do you feel about an inbetween solution?

```
.overflow-table table code {
  white-space: pre-wrap;
  font-size: 95%;
  max-width: 300px;
  overflow-wrap: break-word;
  word-break: normal;
}

.overflow-table td, .overflow-table th {
  max-width: 300px;
  min-width: 200px;
  word-break: normal;
}
```

This way we get wrap but less of it and also avoid situations where one column gets excessively wide because there is a single value that is very long.

I don't have very strong feelings either way.

But this is definitely an improvement on what we have now. 

@brad-dow what do you think?

JoblersTune:
Thanks for adding in the suggestions @huijing. What do you guys think between the two options @melissahenderson and @brad-dow?

melissahenderson:
They're 100x nicer than they were. I think the writers can make them look even better by breaking some of the tables apart, adjusting the content, etc. Sticky headers are good and while having horizontal and vertical scrollbars isn't ideal, they're necessary for people to know there's more to the table than what they're seeing. 

Also, I think we are going to just have to live with the fact that the Rafiki site isn't going to be great for people on mobile.

What it looks like without touching anything:

IMG_9158

What it looks like using touch to drag things around:

IMG_9159

huijing:
ok, if we're all in agreement, i'm guessing this one can be merged soon?

brad-dow:
This is looking so much better! I only have one very small nit to pick:

In dark mode, the sticky header either loses it's background color or something weird is happening with the opacity on scroll...
image

huijing:
@brad-dow good catch, it's fixed now.

brad-dow:
This looks fantastic, thanks for the fix @huijing!

melissahenderson (approved):

Issue #3012, opened 2024-10-02
Add peer liquidity high and asset liquidity high events


mkurapov:
## Context
Similar to how we have `peer.liquidity_low` and `asset.liquidity_low` webhook events, we should add a `peer.liquidity_high` and `asset.liquidity_high` webhook events. 

Rafiki's A peer liquidity increases whenever payments are being sent _from_ it, which, in other words, means the debt owing to the other peer increases. This information would be useful for the ASE to know as soon as it happens, to better manage their risk, and liquidity in general. 

When it comes to asset liquidity, most of the changes to asset liquidity happen during cross-currency transfers (as it's necessary to convert from one asset account to another). Naturally, asset low and asset high liquidity events will go hand in hand (ie if sending between USD <> EUR, USD will decrease, EUR will increase, Rafiki could end up firing the asset low and asset high events), but this also seems useful to have from a FX risk perspective.

## Todos
- [ ] Add optional `liquidityHighThreshold` input variable to `createPeer` `createAsset` `updatePeer` `updateAsset` graphql mutations, and store this in the `asset` and `peer` tables 
- [ ] Add `peer.liquidity_high` and `asset.liquidity_high` webhook events. These events should get fired in the `onCredit` methods for assets & peers.
- [ ] Create documentation issue for this

PR #3025, opened 2024-10-09
refactor: made webhook event type an enum in GraphQL schema

### LINKED ISSUE #2786: Make Webhook Event types an enum in the Admin API
## Context
We should properly define an `enum` for the different `WebhookEvent` types (`incoming_payment.created`, `outgoing_payment.completed`, etc...) in the GraphQL backend Admin API. This enum would be used in `WebhookEvent.type`, similar to how `IncomingPaymentState` is defined for `IncomingPayment`s.
 
This will provide better clarity in the documentation, and allow consumers of the GraphQL (e.g. MASE) use auto generated TypeScript types for defining the webhooks, instead of having to declare it's own interface.

## Todos
- [ ] Update the GraphQL Admin API schema 
- [ ] Update the consumers of the API (MASE, as well as the frontend)

oana-lolea:

- Added `WebhookEventType` scalar in GraphQL schema
- Added parser to restrict values to be only the allowed types
- Added tests for the scalar methods

## Context

Made `WebhookEvent` type a scalar instead of a string in GraphQL schema. 
Webhook event types format (e.g. `incoming_payment.created`) are accepted. Given a string with the webhook event type, the mapper returns its corresponding `WebhookEventType` enum member, or throws an error if it doesn't exist.
Fixes #2786 
## Checklist

- [ ] Tests added/updated

### DIFF
--- a/localenv/mock-account-servicing-entity/generated/graphql.ts
+++ b/localenv/mock-account-servicing-entity/generated/graphql.ts
@@ -20,6 +20,11 @@ export type Scalars = {
   UInt8: { input: number; output: number; }
   /** The `UInt64` scalar type represents unsigned 64-bit whole numeric values. It is capable of handling values that are larger than the JavaScript `Number` type limit (greater than 2^53). */
   UInt64: { input: bigint; output: bigint; }
+  /**
+   * The 'WebhookEventType' scalar represents the type of an webhook event.
+   * NOTE: The possible options are: 'incoming_payment.created', 'incoming_payment.completed', 'incoming_payment.expired', 'outgoing_payment.created', 'outgoing_payment.completed', 'outgoing_payment.failed', 'wallet_address.not_found', 'wallet_address.web_monetization', 'asset.liquidity_low', 'peer.liquidity_low'.
+   */
+  WebhookEventType: { input: any; output: any; }
 };
 
 export type AccountingTransfer = Model & {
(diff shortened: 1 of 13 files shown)


mkurapov:
@BlairCurrey would be good to get another perspective as well:

what do you think about the issue? I like having a proper enum to be used in the graphql query, but knowing that GQL unfortunately only supports upper snake case values like @oana-lolea mentioned, could it be confusing if people try to use the generated enum value when handing the webhook events in the webhook service, since it won't match up?

BlairCurrey:
> could it be confusing if people try to use the generated enum value when handing the webhook events in the webhook service, since it won't match up?

I think it would be problematic. We couldn't use them for identifying which type of webhook was received (ie `receviedWebhookEvent.type === WebhookEventType.INCOMING_PAYMENT_CREATED`). And documentation-wise it's close but not actually correct which is confusing. 

What about a custom scalar?

```ts
const WebhookEventType = new GraphQLScalarType({
  name: 'WebhookEventType',
  description: 'Webhook event type format (e.g., incoming_payment.created)',
  serialize(value) {
    return value;
  },
  parseValue(value) {
    const allowedTypes = [
      'incoming_payment.created',
      'outgoing_payment.completed',
      // Add other valid event types
    ];
    if (!allowedTypes.includes(value)) {
      throw new Error('Invalid WebhookEventType');
    }
    return value;
  }
});

```

I guess this might not be clear on the auto-generated documentation side though (it will just show `WebhookEventType`). But in terms of behavior of the GQL api it seems like what we want.

mkurapov:
From the meeting:

Let's try to update the `WebhookEventFilter` to take in an array of WebhookEvent scalar values. If it breaks the `frontend` webhooks page out of the box, we can just close the issue with a plan to standardize all of our enums (like #3012) in the future. 

Otherwise, we can keep the scalar.

oana-lolea:
@mkurapov The scalar cannot be used for the filter because it can only be used within the GraphQL schema. We could however use its parser to validate the filter string values.

mkurapov:
@oana-lolea ah, I see. Thanks for taking a look into this. Since we can't use it as part of the filter (which would be useful for automatically validating user/client input), maybe we can just close this issue as not planned, and standardize all our enums separately in the future

Issue #2929, opened 2024-08-30
[Multi-Tenant] Filter OP resources by `tenantId` in Admin API calls


njlie:
The Rafiki Admin API should use the Kratos session token to retrieve and add a `tenantId` to the context. Resolvers should use this context to:

* Only retrieve Open Payments resources that belong to that tenant (unless that tenant is also the instance operator).
* It should prevent requests where provides a valid id is provided for an Open Payments resource, but that resource does not belong to that tenant. Return a `Not Found` response in this case.
* When creating a new resource that requires a tenantId as database field, it should either:
  * Verify that the provided `tenantId` in the input matches the one added to the context, or is from an operator
  * Provided that `tenantId` from the context.
    * Maybe we could even remove `tenantId` as an input from all graphql inputs, and just include it in service requests by pulling it from the context at all times.

This logic should be applied to resolvers for:
* Quotes
* Incoming/Outgoing Payments
* Wallet Addresses

njlie:
Superseded by #3119, #3118, #3117.

PR #3026, opened 2024-10-09
Bc/2929/filter op resources by tenantid

BlairCurrey:
Trying to determine where to put access control. Basic options are:
- in resolver/middleware
- in service, ie an interface like `incomingPaymentService.get(id, tenantId, isOperator)`. Not sure if this is ultimately viable

There are basically 3 classes of access control that I see so far:
- Gets: check `tenantId`s after retrieving if doing in resolver, or filter by if doing in service
- Creates: check the resource-to-be-created's `walletAddress.tenantId` before creating (either in resolver or service)
- Update/deletes: checks the existing resource's tenantId before mutating (either in resolver or service)

This `getIncomingPayment` resolver and the comment shows the issue more clearly. It does access control in the resolver after getting the resource: https://github.com/interledger/rafiki/blob/173b6b23dd7228f44033bf888c2cb1aeec437fa2/packages/backend/src/graphql/resolvers/outgoing_payment.ts#L21-L57

This `cancelIncomingPayment` resolver exemplifies the Create and Update/Delete cases (largely similar, only difference is if we compare given `tenantId` with existing resource or resource-to-be-created's `walletAddress.tenantId`): 

https://github.com/interledger/rafiki/blob/173b6b23dd7228f44033bf888c2cb1aeec437fa2/packages/backend/src/graphql/resolvers/incoming_payment.ts#L139-L181

will fix: #2929

### DIFF
--- a/bruno/collections/Rafiki/Examples/Admin API -  only locally/Peer-to-Peer Payment/Create Outgoing Payment.bru
+++ b/bruno/collections/Rafiki/Examples/Admin API -  only locally/Peer-to-Peer Payment/Create Outgoing Payment.bru
@@ -10,6 +10,10 @@ post {
   auth: none
 }
 
+headers {
+  ~x-operator-secret: tXhxCNRWuVOJs5W/CUgjsc+vBnKj0CVdqSb87EXN64E=
+}
+
 body:graphql {
   mutation CreateOutgoingPayment($input: CreateOutgoingPaymentInput!) {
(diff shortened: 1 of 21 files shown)


mkurapov:
@BlairCurrey we can close this, right?

BlairCurrey:
> @BlairCurrey we can close this, right?

@mkurapov Yeah I wanted to preserve the branch for reference but dont need this draft PR.

PR #3041, opened 2024-10-21
feat(backend): caching on asset and wallet address including other enhancements

### LINKED ISSUE #2981: Cache assets
### Context

During Rafiki WW, one of the performance optimizations we found to be useful is caching assets, given we make so many requests for assets, even though they rarely change. 

We should use redis for caching these.

### Todo
- [x] Remove `public abstract asset: Asset` from `WalletAddressSubresource` model, and instead, add a getter and a setter for the field. (We don't want to use Objection to fetch assets through foreign keys)
- [x] Add redis cache to assetService (via `createRedisDataStore`, and a reasonable key expiry). Delete cache on asset `update`.
- [x] Resolve any requests for assets via the assetService instead of any `.withGraphFetched('asset')` objection calls, use setter on `WalletAddressSubresource` model to add asset to model

### LINKED ISSUE #2982: Cache wallet addresses
### Context

During Rafiki WW, one of the performance optimizations we found to be useful is caching assets, given we make so many requests for wallet addresses, even though they rarely change. 

We should use redis for caching these.

### Todo
- [x] Add redis cache to walletAddressService (via `createRedisDataStore`, and a reasonable key expiry). Delete cache on wallet address `update`.
- [x] Resolve any requests for assets via the assetService instead of any `.withGraphFetched('walletAddress')` objection calls

koekiebox:

- https://github.com/interledger/rafiki/issues/2981
- https://github.com/interledger/rafiki/issues/2982

## Context

Fixes #2981 
Fixes #2982 

## Checklist

- [x] Tests added/updated


packages/backend/src/accounting/psql/service.ts
```diff
@@ -38,7 +38,7 @@ import { TelemetryService } from '../../telemetry/service'
 
 export interface ServiceDependencies extends BaseService {
   knex: TransactionOrKnex
-  telemetry: TelemetryService
+  telemetry?: TelemetryService
```
BlairCurrey, line 41: This should not be optional. I think maybe it ended up this way after resolving a merge conflict or something?

[It used to be optional but we refactored it to always be required (but do nothing if disabled) in this PR](https://github.com/interledger/rafiki/pull/2991).

After making it required we can also change all the timer calls like:

```ts
const stopTimer = deps.telemetry?.startTimer(...)
...
stopTimer && stopTimer()
```

to

```ts
const stopTimer = deps.telemetry.startTimer(...)
...
stopTimer()
```

packages/backend/src/cache/cache.ts
```diff
@@ -0,0 +1,46 @@
+import { CacheDataStore, CacheSupport } from './index'
+
+export interface Cached {
+  expiry: number
+  data: CacheSupport
+}
+
+export function createInMemoryDataStore(keyTtlMs: number): CacheDataStore {
```
BlairCurrey, line 8: We have an existing `createInMemoryDataStore` in `rafiki/packages/backend/src/middleware/cache/data-stores/in-memory.ts`. The difference between that and this seems to what is being cached. Strings vs. `CacheSupport` which is `Asset | WalletAddress | etc.`. Perhaps we can do better than a new `createInMemoryStore` for `CacheSupport` and instead parameterize the type using generics on the existing `createInMemoryStore`. In other words, a function signature like `function createInMemoryDataStore<T>(keyTtlMs: number): CacheDataStore<T>`.

Gets rid of this extra factory and makes the interfaces for the Asset, etc. strongly typed. Whereas currently we're asserting that its an `Asset` etc. on read.

For example, when using this we're doing const `inMem = (await deps.cacheDataStore.get(id)) as Asset`. If we instead initialize a cache per resource like:

```ts
// backend/src/app.ts
    return await createAssetService({
      logger: logger,
      knex: knex,
      accountingService: await deps.use('accountingService'),
      cacheDataStore: createInMemoryDataStore<Asset>(10000)
    })
```

Then instead of `get` returning `Promise<CacheSupport>` it would return `Promise<Asset>`

koekiebox: Resolved.

packages/backend/src/asset/service.ts
```diff
@@ -197,3 +221,13 @@ async function getAssetsPage(
...
+
+async function setAssetOn(
+  deps: ServiceDependencies,
+  obj: ToSetOn
+): Promise<void | Asset> {
+  if (!obj) return
+  const asset = await getAsset(deps, obj.assetId)
+  if (asset) obj.asset = asset
+  return asset
+}
```
BlairCurrey, line 233: This still feels a bit hacky to me. This is returning the asset but I dont see the `setOn` callers actually ever using the returned value. I realize this is to reduce boilerplate everywhere we are setting this but still wondering if there is a better way utilizing the existing `assetService.get(id)`.

Here is one example but it doesnt quite work. The model wallet address expects assetId and asset to be defined because of the non-null assertion in the model.  Still thinking.

```ts
async function getWalletAddressByUrl(
  deps: ServiceDependencies,
  url: string
): Promise<WalletAddress | undefined> {
  let walletAddress = await WalletAddress.query(deps.knex).findOne({ url })
  if (!walletAddress) return
  // await deps.assetService.setOn(walletAddress)
  walletAddress.asset = await deps.assetService.get(walletAddress.assetId)
  return walletAddress || undefined
}
```

packages/backend/src/telemetry/service.ts
```diff
@@ -109,6 +111,14 @@ export class TelemetryServiceImpl implements TelemetryService {
...
   }
 
+  getCounters(): Map<string, Counter> {
+    return this.counters
+  }
+
+  getHistograms(): Map<string, Histogram> {
+    return this.histograms
+  }
+
```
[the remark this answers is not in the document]
koekiebox: Resolved.

BlairCurrey (dismissed):

packages/backend/src/accounting/tigerbeetle/service.ts
```diff
@@ -212,8 +212,15 @@ export async function getAccountBalance(
   deps: ServiceDependencies,
   id: string
 ): Promise<bigint | undefined> {
-  const account = (await getAccounts(deps, [id]))[0]
-  if (account) return calculateBalance(account)
+  const stopTimer = deps.telemetry.startTimer('tb_get_account_balances', {
```
mkurapov, line 215: maybe we can also add one for the TB transfer creation? Might be handy

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
```diff
@@ -11,10 +11,17 @@ export function createIncomingErrorHandlerMiddleware(
...
+      'create_incoming_error_handler_middleware_next',
+      {
+        callName: 'createIncomingErrorHandlerMiddleware:next'
+      }
+    )
     try {
       await next()
       if (!ctx.response.rawReply) {
         ctx.services.logger.error('handler did not return a valid value.')
+        stopTimer()
```
mkurapov, line 24: can use a `finally` here to just have to do `stopTimer()` once

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
```diff
@@ -110,6 +119,7 @@ export function createBalanceMiddleware(): ILPMiddleware {
         } else {
           await trx.void()
         }
+        stopTimer()
```
mkurapov, line 122: I think need to include `stopTimer()` also on L114 and L108

packages/backend/src/config/app.ts
```diff
@@ -127,8 +127,8 @@ export const Config = {
   authServerGrantUrl: envString('AUTH_SERVER_GRANT_URL'),
   authServerIntrospectionUrl: envString('AUTH_SERVER_INTROSPECTION_URL'),
 
-  outgoingPaymentWorkers: envInt('OUTGOING_PAYMENT_WORKERS', 4),
-  outgoingPaymentWorkerIdle: envInt('OUTGOING_PAYMENT_WORKER_IDLE', 200), // milliseconds
+  outgoingPaymentWorkers: envInt('OUTGOING_PAYMENT_WORKERS', 1),
+  outgoingPaymentWorkerIdle: envInt('OUTGOING_PAYMENT_WORKER_IDLE', 10), // milliseconds
```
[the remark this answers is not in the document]
koekiebox: Yeah, having more frequent iterations and less workers did make a positive difference.

packages/backend/src/config/app.ts
```diff
@@ -191,7 +191,8 @@ export const Config = {
   maxOutgoingPaymentRetryAttempts: envInt(
     'MAX_OUTGOING_PAYMENT_RETRY_ATTEMPTS',
     5
-  )
+  ),
+  localCacheDuration: envInt('LOCAL_CACHE_DURATION', 5_000)
```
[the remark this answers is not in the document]
koekiebox: I agree. Making it 15s.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -71,21 +77,26 @@ export interface WalletAddressService {
...
+  setOn(obj: ToSetOn): Promise<void | WalletAddress>
 }
 
 interface ServiceDependencies extends BaseService {
   config: IAppConfig
   knex: TransactionOrKnex
   accountingService: AccountingService
   webhookService: WebhookService
+  assetService: AssetService
+  cacheDataStore: CacheDataStore<WalletAddress>
```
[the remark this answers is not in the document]
koekiebox: It is. Updated for asset also.

packages/backend/src/index.ts
```diff
@@ -303,7 +307,8 @@ export function initIocContainer(
       knex: await deps.use('knex'),
       accountingService: await deps.use('accountingService'),
       walletAddressService: await deps.use('walletAddressService'),
-      config: await deps.use('config')
+      config: await deps.use('config'),
+      assetService: await deps.use('assetService')
```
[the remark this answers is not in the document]
koekiebox: Fixed.

packages/backend/src/asset/service.ts
```diff
@@ -81,6 +93,8 @@ async function createAsset(
       .first()
 
     if (deletedAsset) {
+      await deps.cacheDataStore.delete(deletedAsset.id)
```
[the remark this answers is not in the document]
koekiebox: Done. 👍

packages/backend/src/accounting/tigerbeetle/service.ts
koekiebox: Added. Also added it for PSQL.

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
koekiebox: Done.

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
koekiebox: Reworked it to not have all those `stopTimer()`s.

packages/backend/src/asset/service.ts
```diff
@@ -197,3 +221,13 @@ async function getAssetsPage(
 async function getAll(deps: ServiceDependencies): Promise<Asset[]> {
   return await Asset.query(deps.knex).whereNull('deletedAt')
 }
+
+async function setAssetOn(
```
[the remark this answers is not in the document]
mkurapov: Actually, we don't need to change the model at all, since `asset` is not readonly. I'm mostly coming from the approach that this isn't assetService's responsibility.

We can just do something like 
```ts
incomingPayment.asset = await assetService.get(incomingPayment.assetId) as Asset
```
directly. And if you want to avoid using `as`, we can just have a `assetService.getOrThrow` that returns `Promise<Asset>`.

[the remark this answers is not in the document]
koekiebox: Updated.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -146,9 +154,14 @@ async function getOutgoingPayment(
 ): Promise<OutgoingPayment | undefined> {
   const outgoingPayment = await OutgoingPayment.query(deps.knex)
     .get(options)
-    .withGraphFetched('[quote.asset, walletAddress]')
```
mkurapov, line 149: We can also replace the `withGraphFetched` in the cancelOutgoingPayment function (mostly for consistency)

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -245,9 +262,16 @@ async function getWalletAddress(
   deps: ServiceDependencies,
   id: string
 ): Promise<WalletAddress | undefined> {
-  return await WalletAddress.query(deps.knex)
-    .findById(id)
-    .withGraphFetched('asset')
+  const walletAdd = await deps.walletAddressCache.get(id)
+  if (walletAdd) return walletAdd as WalletAddress
```
mkurapov, line 266:
```suggestion
  if (walletAdd) return walletAdd
```

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -301,9 +325,11 @@ async function getWalletAddressByUrl(
   deps: ServiceDependencies,
   url: string
 ): Promise<WalletAddress | undefined> {
-  const walletAddress = await WalletAddress.query(deps.knex)
-    .findOne({ url: url.toLowerCase() })
-    .withGraphFetched('asset')
+  const walletAddress = await WalletAddress.query(deps.knex).findOne({ url })
```
mkurapov, line 328: Let's keep the `url: url.toLowerCase()` in the `findOne`

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
```diff
@@ -21,96 +21,103 @@ export function createBalanceMiddleware(): ILPMiddleware {
...
+        if (trx) {
+          if (response.fulfill) {
+            await trx.post()
+          } else {
+            await trx.void()
+          }
         }
       }
+    } finally {
+      stopTimer()
```
mkurapov, line 120: unfortunately, I think this timer will encapsulate all of the following middlewares in the timing (because we will be waiting for all the `await next()`s.). We can tackle this in a separate PR, we could probably do something nice in the `compose` function in `src/payment-method/ilp/connector/index.ts` to automatically add telemetry for all middlewares

mkurapov: (same for the error handler middleware)

packages/backend/src/rates/service.ts
```diff
@@ -54,7 +54,7 @@ export function createRatesService(deps: ServiceDependencies): RatesService {
 
 class RatesServiceImpl implements RatesService {
   private axios: AxiosInstance
-  private cachedRates: CacheDataStore
+  private cachedRates: CacheDataStore<string>
```
mkurapov, line 57: ah, we should have just stored the actual object with the generic type from the start (instead of parsing in and out) but we can keep this for now

packages/backend/src/open_payments/payment/outgoing/service.ts
mkurapov: Do we also want to do this for incomingPayments?

localenv/telemetry/grafana/provisioning/dashboards/example.json
```diff
@@ -1592,6 +1643,7 @@
           "sort": "none"
         }
       },
+      "pluginVersion": "11.3.1",
```
mkurapov, line 1646: Not part of your changes, but can you remove the `* 0.5` from L1654? Since we are summing by source we don't need to halve it to account for having both Cloud nine and happy life bank instances running

mkurapov, line 1646: Not part of your changes, but can you remove the `* 0.5` from L1654? Since we are summing by source we don't need to halve it to account for having both Cloud nine and happy life bank instances running

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -54,9 +56,14 @@ async function getQuote(
   deps: ServiceDependencies,
   options: GetOptions
 ): Promise<Quote | undefined> {
-  return Quote.query(deps.knex)
+  const quote = await Quote.query(deps.knex)
     .get(options)
-    .withGraphFetched('[asset, fee, walletAddress]')
+    .withGraphFetched('[fee, walletAddress]')
```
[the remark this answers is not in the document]
koekiebox: Fixed.

localenv/telemetry/grafana/provisioning/dashboards/example.json
```diff
@@ -1493,14 +1542,15 @@
...
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
             "uid": "${PrometheusDS}"
           },
           "editorMode": "code",
-          "expr": "(increase(packet_amount_fulfill_total[30s]) * 0.5 * 0.0001)",
+          "expr": "(increase(packet_amount_fulfill_total[30s]) * 0.0001)",
```
mkurapov, line 1553: We should sum by source as well

packages/backend/src/payment-method/ilp/connector/core/middleware/error-handler.ts
```diff
@@ -11,6 +11,12 @@ export function createIncomingErrorHandlerMiddleware(
   serverAddress: string
 ): ILPMiddleware {
   return async (ctx: ILPContext, next: () => Promise<void>): Promise<void> => {
+    const stopTimer = ctx.services.telemetry.startTimer(
```
mkurapov, line 14: This middleware doesn't do much other than wrap all the other middlewares, so the timing of this will usually just equal the whole packet lifecycle, so as long as we are tracking that (I believe we do) then we don't necessarily need this one here

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -797,3 +801,92 @@ describe('Open Payments Wallet Address Service', (): void => {
...
+          status
+        })
+        assert.ok(!isWalletAddressError(updatedWalletAddress))
+        expect(updatedWalletAddress.isActive).toEqual(expectedIsActive)
+
+        // We expect the [set] to be called again with the new data:
+        expect(spyCacheSet).toHaveBeenCalledTimes(expectedCallCount)
+        expect(spyCacheSet).toHaveBeenCalledWith(
+          walletAddress.id,
+          expect.anything()
```
mkurapov, line 879: we could probably using something like `expect.objectContaining({ id, status })` such that we can verify the few main fields but don't have to go through every single one

localenv/telemetry/grafana/provisioning/dashboards/example.json
mkurapov:
```suggestion
          "expr": "sum by (source) (increase(packet_amount_fulfill_total[30s]) * 0.0001)",
```

packages/backend/src/open_payments/quote/service.ts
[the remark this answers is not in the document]
mkurapov: Still relevant I believe

mkurapov (approved):
🚀

PR #3043, opened 2024-10-22
chore(backend): dont wait 30s when completing/expiring incoming payment

### LINKED ISSUE #2922: [FEATURE REQUEST] Allow the user to change the time between the payment completion and the webhook event sending
## Feature Request

**Describe the feature you'd like to request**

When we make a transfer to Rafiki and the transfer is completed, the ASE has to wait 30 seconds to receive the webhook event call to perform an action.

In the case of the PCH's CNP, we need to perform an action after the transfer completion, the webhook event should be sent from Rafiki almost instantly in order to continues the flow as soon as posible.

**Describe the solution you'd like**

Maybe there should be a environment variable to control the waiting time.

I think this waiting time should be considierated for the actionable payments.

## Todo
- [ ] Remove the 30_000 delay period when completing or expiring incoming payments

mkurapov:

- Removing the 30 second wait time when completing or expiring an incoming payment. 

## Context

- The reason for the incoming payment would wait 30 seconds before "finalizing" it (via `processAt`), was to check for any packets that we're sent but not received, essentially, trying to make sure that we wouldn't "lose" any that should be delivered to the payment. However, in practice, it looks like this delay wouldn't have mattered/worked to begin with. Because we also update the status of the payment when changing `processAt`:
https://github.com/interledger/rafiki/blob/6b971bf4636bbc658530e8447c52866054db2aec/packages/backend/src/open_payments/payment/incoming/model.ts#L154-L164  _and_ we check for the incoming payment status when we received packets: https://github.com/interledger/rafiki/blob/6b971bf4636bbc658530e8447c52866054db2aec/packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts#L60-L80 we wouldn't have been able to receive any packets after the payment was marked as Completed/Expired _anyway_. This is why we can just remove this artificial waiting period, and just complete/expire the payment directly.

## Checklist

Fixes #2922 

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
+++ b/packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
@@ -87,10 +87,6 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
               LiquidityAccountType.INCOMING
             )
           }
-          ctx.services.logger.debug(
-            { incomingPaymentId: incomingPayment.id },
-            'destination account is incoming payment'
-          )
           return incomingPayment
         }
         // Open Payments SPSP fallback account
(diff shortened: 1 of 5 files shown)


packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts
```diff
@@ -87,10 +87,6 @@ export function createAccountMiddleware(serverAddress: string): ILPMiddleware {
               LiquidityAccountType.INCOMING
             )
           }
-          ctx.services.logger.debug(
```
mkurapov, line 90: logs were noisy, especially during quoting step (since we would get 10 of these)

raducristianpopa (approved):

PR #3047, opened 2024-10-27
fix: Admin UI error Auth redirect

### LINKED ISSUE #2840: [BUG]: Admin UI got error when the session expires and the user tries to login again. It enters in an infinite auth loop.
## Bug Report

**Describe the bug**

When the session at the Admin UI expires, appears the welcome page to login again. If the user tries to login again using the welcome page that appears in the interface, it gets a 404 error.
Then if the user click the `Go to homepage` button, it's redirected to root `auth` page, so if the user tries to login, it gets the 404 error over again.

**To Reproduce**

> Important: Don't reload the page at any time.

1. Run the rafiki instance
2. Go to Admin UI and login
2. Using the developer tools of your browser, delete the page cookies. (This is made in order to simulate the session expiration).
3. Now, try to move to another section (wallets, peers, wallet details, etc).
4. Try to login again, using always the buttons that appears in the interface.
5. You will enter in the auth loop.

**Screenshots**

1. Session expired, welcome page appears.
image
2. You try to login again, you got the 404 error.
image
3. If you try to go to homepage and login, you got the same error.
image

I think this is not a correct behavior. The user should be able to login again without the need of reload the page.

DarianM:

Using the `Form` with `redirectDocument(loginUrl);` By using remix's `action` function and `redirectDocument` instead of `href`, the server processes the request, retrieves the necessary information (such as flow parameters from Ory Kratos), and provides the correct redirect response. The form submit allows remix to handle this routing server-side, ensuring the full URL is correctly generated.

The `href` attribute instead tries to directly navigate to the link client-side. If variables.kratosBrowserPublicUrl is undefined or not available at runtime, Remix will interpret it as `undefined`, creating the `localhost:3010/undefined/self-service/login/browser` URL.

## Context
What were you trying to do?
fixes #2840

### DIFF
--- a/packages/frontend/app/routes/auth._index.tsx
+++ b/packages/frontend/app/routes/auth._index.tsx
@@ -11,7 +11,7 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
   return null
 }
 
-export const action = () => {
+export async function action() {
   return redirectDocument(
     `${variables.kratosBrowserPublicUrl}/self-service/login/browser`
   )
--- a/packages/frontend/app/routes/auth._index.tsx
+++ b/packages/frontend/app/routes/auth._index.tsx
@@ -14,8 +14,8 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
 export const action = () => {
   return redirectDocument(
     `${variables.kratosBrowserPublicUrl}/self-service/login/browser`
-  );
-};
+  )
+}
 
 export default function Auth() {
   return (
@@ -39,7 +39,7 @@ export default function Auth() {
             </a>
           </p>
           <div>
-            <Form method="post">
+            <Form method='post'>
               <Button aria-label='logout' type='submit' className='mr-2'>
                 Login
               </Button>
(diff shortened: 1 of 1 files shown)


packages/frontend/app/routes/auth._index.tsx
```diff
@@ -10,8 +11,13 @@ export const loader = async ({ request }: LoaderFunctionArgs) => {
   return null
 }
 
+export const action = () => {
```
mkurapov, line 14: probably not a big deal but should this be `async`?

mkurapov (dismissed):

mkurapov (dismissed):

packages/frontend/app/routes/auth._index.tsx
```diff
@@ -33,9 +39,11 @@ export default function Auth() {
             </a>
           </p>
           <div>
-            <Button aria-label='logout' href={loginUrl} className='mr-2'>
-              Login
-            </Button>
+            <Form method='post'>
+              <Button aria-label='logout' type='submit' className='mr-2'>
```
JoblersTune, line 43:
```suggestion
              <Button aria-label='login' type='submit' className='mr-2'>
```

JoblersTune:
@DarianM thanks for the fix! Would you mind please moving the action function below the default export function to match the styling throughout the `frontend` codebase. We usually stick to loader, then default export, and then action function below that.

mkurapov (approved):

PR #3050, opened 2024-10-29
chore: Fix mock ASE seeding script

### LINKED ISSUE #2945: Make mock ASE seeding script smarter
## Context
Currently, when we start up the mock ASE, we run a seed script. This seed script calls the Admin API to create assets, peers, wallet addresses and add liquidity where necessary. After this, the script creates mock user accounts in memory. These mock accounts are needed to properly receive and send payments between wallet addresses (since without mock accounts, the webhook handlers do not know who to debit/credit).

However, we run into an issue when we spin down the local environment, but do not clear the Rafiki database: the seed script sees that the data has been seeded already, and exits early, without creating mock user accounts. As a result, we cannot process payments, since the webhook handlers would not be able to find the corresponding accounts. 

We should change the seed script, such that it checks if the assets, peers, wallet addresses, etc already exist in Rafiki. If they do, make mock accounts based on that. Otherwise, we should proceed as normal, and create missing resources in Rafiki.

The outcome of this is that if the local playground is spun down without deleting any volumes, spinning the environment back up populates the mock accounts correctly, and payments are able to be processed as usual.

## Todos
- [ ] Make seed script in mock ASE handle existing data

oana-lolea:

- Added checks if assets, peers and wallet addresses already exist.
- If they do, they are returned. Otherwise, they are created as usual.

## Context

When spinning up the local environment and the database volumes were not removed, the mock ASE seeding script tries to create assets, peers and wallet addresses to populate the database, but fails, because the values already exists and the mock user accounts are not created.

By checking if the data already exists by querying by their unique entries, this issue can be fixed. If an entry already exist, the seeding script will use those instead of creating them.

Added #3052 issue for updating the commands in the user docs.

Fixes #2945 
## Checklist

- [ ] Tests added/updated


packages/backend/src/graphql/schema.graphql
```diff
@@ -2,6 +2,14 @@ type Query {
   "Fetch an asset by its ID."
   asset("Unique identifier of the asset." id: String!): Asset
 
+  "Get an asset based on its currency code and scale if it exists."
+  assetByCodeAndScale(
```
mkurapov, line 6: we should add bruno requests for these queries

mkurapov (commented):
Works for me after spinning the env down! Just some minor comments

mkurapov:
@oana-lolea make sure to go through the todo as well

packages/backend/src/graphql/schema.graphql
njlie: An asset code and scale combination is unique, right? Might make sense for the graphql schema to reflect that in some way.

oana-lolea: I thought these queries might be reused in the future, because they are searching by the unique column(s) of the tables. But if not, I can just filter everything in the script and it might be a bit faster than just querying the db that much.

mkurapov: > An asset code and scale combination is unique, right?

Yes, has a unique constraint in DB. 

@njlie what do you mean by "Might make sense for the graphql schema to reflect that in some way."?

njlie: What I mean to say is, from the point of view of someone integrating with the GraphQL API and looking at it for the first time, seeing a resolver like `assetByCodeAndScale` would be an effective way to communicate to that audience that those two fields together is a unique combination

mkurapov: Ah, I see what you mean. As in, the schema itself reflects to the integrator about how assets work. I thought you meant we needed to add a comment or something to the schema. Understand now.

mkurapov: I agree in general, let's keep these resolvers. @oana-lolea I think the outstanding item is just to remove some lines in the resolver test file

mkurapov (dismissed):

mkurapov (approved):

PR #3062, opened 2024-10-31
chore:token invalid throw 401 instead of return public incoming payment

### LINKED ISSUE #2889: Throw error when failing to introspect token instead of returning public incoming payment
### Context
When we do [GET /incoming-payment/:id](https://openpayments.dev/apis/resource-server/operations/get-incoming-payment/) via Open Payments APIs, we return either a "private" incoming payment (that requires an access token in the request) or a "public" incoming payment that returns less details, and that does not require an access token.

Currently, when we receive a request to this route, we always do a token introspection check and a signature verification. If those checks end up throwing an error, we suppress the error (via `bypassError: true` flag), and just return the public incoming payment instead:

https://github.com/interledger/rafiki/blob/b08a157e519b77cdd0de0e1253102c25bc4c3a05/packages/backend/src/app.ts#L573-L580

However, this seems like a wrong behaviour: if the user of the Open Payments client is making a request to get an incoming payment _with_ an access token, then the expected result is receiving the incoming payment with all of the details _or_ an error notifying the client about an invalid token, instead of a public incoming payment. 

If the signature is invalid however (but token is not provided), we can still return the public incoming payment, since the intention of the client is just to get the public incoming payment still.

### Tasks
- [ ] if an `accessToken` is provided with the request to get an incoming payment, and the token is invalid, we should fail with a 401 error instead of returning a public incoming payment

DarianM:

- 

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -577,7 +577,7 @@ export class App {
       createTokenIntrospectionMiddleware({
         requestType: AccessType.IncomingPayment,
         requestAction: RequestAction.Read,
-        bypassError: true
+        canSkipAuthValidation: true
       }),
       authenticatedStatusMiddleware,
       getWalletAddressForSubresource,
--- a/packages/backend/src/open_payments/auth/middleware.ts
+++ b/packages/backend/src/open_payments/auth/middleware.ts
@@ -67,11 +67,11 @@ function toOpenPaymentsAccess(
 export function createTokenIntrospectionMiddleware({
   requestType,
   requestAction,
-  bypassError = false
+  canSkipAuthValidation = false
 }: {
   requestType: AccessType
   requestAction: RequestAction
-  bypassError?: boolean
+  canSkipAuthValidation?: boolean 
 }) {
   return async (
     ctx: WalletAddressUrlContext,
@@ -79,14 +79,20 @@ export function createTokenIntrospectionMiddleware({
   ): Promise<void> => {
     const config = await ctx.container.use('config')
     try {
-      const parts = ctx.request.headers.authorization?.split(' ')
-      if (parts?.length !== 2 || parts[0] !== 'GNAP') {
+      if (canSkipAuthValidation && !ctx.request.headers.authorization) {
+        ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
+        await next()
+        return
+      }
+
+      const authSplit = ctx.request.headers.authorization?.split(' ')
+      if (authSplit?.length !== 2 || authSplit[0] !== 'GNAP') {
         throw new OpenPaymentsServerRouteError(
           401,
           'Missing or invalid authorization header value'
         )
       }
-      const token = parts[1]
+      const token = authSplit[1]
       const tokenIntrospectionClient = await ctx.container.use(
         'tokenIntrospectionClient'
       )
@@ -145,19 +151,16 @@ export function createTokenIntrospectionMiddleware({
               : undefined
         }
       }
+
+      await next()
     } catch (err) {
       if (!(err instanceof OpenPaymentsServerRouteError)) {
         throw err
       }
 
       ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
-
-      if (!bypassError) {
-        throw err
-      }
+      throw err
     }
-
-    await next()
   }
 }
 
--- a/packages/backend/src/open_payments/auth/middleware.test.ts
+++ b/packages/backend/src/open_payments/auth/middleware.test.ts
@@ -87,12 +87,12 @@ describe('Auth Middleware', (): void => {
     await appContainer.shutdown()
   })
 
-  describe('bypassError option', (): void => {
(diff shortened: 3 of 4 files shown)


packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -67,26 +67,32 @@ function toOpenPaymentsAccess(
...
     try {
-      const parts = ctx.request.headers.authorization?.split(' ')
-      if (parts?.length !== 2 || parts[0] !== 'GNAP') {
+      if (canSkipAuthValidation && !ctx.request.headers.authorization) {
+        ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
+        await next()
+        return
+      }
+
+      const authSplit = ctx.request.headers.authorization?.split(' ')
```
mkurapov, line 88:
```suggestion
      const authSplit = ctx.request.headers.authorization.split(' ')
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -145,19 +151,16 @@ export function createTokenIntrospectionMiddleware({
               : undefined
         }
       }
+
+      await next()
```
mkurapov, line 155: Let's keep this in the outside of try-catch. Otherwise, the catch block in this middleware will apply to all of the following middlewares called in `await next()` which might lead to unexpected behaviour, e.g., adding `WWW-Authenticate` when we don't want to.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -145,19 +151,16 @@ export function createTokenIntrospectionMiddleware({
...
       if (!(err instanceof OpenPaymentsServerRouteError)) {
         throw err
       }
 
       ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
-
-      if (!bypassError) {
-        throw err
-      }
+      throw err
```
mkurapov, line 162:
```suggestion
      if (err instanceof OpenPaymentsServerRouteError) {
        ctx.set('WWW-Authenticate', `GNAP as_uri=${config.authServerGrantUrl}`)
      }

      throw err
```

mkurapov: Since we are here, this seemed more straightforward

packages/backend/src/app.ts
```diff
@@ -577,7 +577,7 @@ export class App {
       createTokenIntrospectionMiddleware({
         requestType: AccessType.IncomingPayment,
         requestAction: RequestAction.Read,
-        bypassError: true
+        canSkipAuthValidation: true
       }),
       authenticatedStatusMiddleware,
```
mkurapov, line 582: In `authenticatedStatusMiddleware` we should return early if we know that we ended up skipping auth validation in `createTokenIntrospectionMiddleware` (since it doesn't make sense to do a lot of work for something we already know shouldn't succeed). One way we can do this is to make `client` optional in `HttpSigWithAuthenticatedStatusContext`, and only proceed to call `throwIfSignatureInvalid` if `ctx.client` is defined (setting `ctx.client` is the main function of `createTokenIntrospectionMiddleware`)

mkurapov (commented):
Just need to run `pnpm format` to fix the check. 

We also have a line in `packages/backend/src/open_payments/payment/incoming_remote/service.ts` which we can remove
("// TODO: remove after #2889 is completed").

packages/backend/src/open_payments/auth/middleware.ts
DarianM: `authorization` can still be undefined here, when the `canSkipAuthValidation` is false it won't enter the early return statement and will return 401 because of undefined authorization;
this is the reason i left the `?` there

packages/backend/src/app.ts
DarianM: what do you think about checking for 
`if (!ctx.request.headers.authorization) {
    await next()
    return
  }`
instead of making the `client' optional (string | undefined) for the rest of the code? I also see that `authenticatedStatusMiddleware` is only used for open payments

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -147,14 +152,10 @@ export function createTokenIntrospectionMiddleware({
       }
     } catch (err) {
       if (!(err instanceof OpenPaymentsServerRouteError)) {
```
mkurapov, line 154:
```suggestion
      if (err instanceof OpenPaymentsServerRouteError) {
```

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -166,6 +167,10 @@ export const authenticatedStatusMiddleware = async (
   next: () => Promise<unknown>
 ): Promise<void> => {
   ctx.authenticated = false
+  if (!ctx.request.headers.authorization) {
```
mkurapov, line 170: now that we expect the request to be for sure authorized after this check, we can get rid of the try-catch below altogether

mkurapov (approved):

PR #3091, opened 2024-11-07
docs: fixes #3052

### LINKED ISSUE #3052: Update local environment setup commands
After [#3050](https://github.com/interledger/rafiki/pull/3050) is merged, there is no need to clear volumes using `--volumes` anymore when spinning down the project.

The localenv command should be updated to just `pnpm localenv:compose down` and the issue regarding this bug should be removed from "Known issues".

melissahenderson:
Created to fix #3052, which is in response to #3050.

@oana-lolea I wasn't sure if every reference to `--volumes` should come out or only specific ones. Please take a look and let me know if I need to change anything!

### DIFF
--- a/localenv/README.md
+++ b/localenv/README.md
@@ -202,6 +202,9 @@ For more ways to connect debuggers, see the Node docs for debugging: https://nod
 # tear down
 pnpm localenv:compose down
 
+# tear down and delete database volumes
+pnpm localenv:compose down --volumes
+
 # tear down, delete database volumes and remove images
 pnpm localenv:compose down --volumes --rmi all
 ```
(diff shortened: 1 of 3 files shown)


oana-lolea:
@melissahenderson I don't think the command should be removed entirely, because it may still be used. It should remain in the list of commonly used commands and removed only when it is used as a fix for the bug and as a setup command.

packages/documentation/src/content/docs/integration/playground/overview.mdx
```diff
@@ -220,7 +217,6 @@ The following are the most commonly used commands:
 | pnpm localenv:compose up                       | Start (with TigerBeetle)                                          |
 | pnpm localenv:compose up -d                    | Start (with TigerBeetle) detached                                 |
 | pnpm localenv:compose down                     | Down (with TigerBeetle)                                           |
-| pnpm localenv:compose down --volumes           | Down and kill volumes (with TigerBeetle)                          |
```
oana-lolea, line 223: We should keep this, it may be still used.

packages/documentation/src/content/docs/integration/playground/overview.mdx
```diff
@@ -203,9 +203,6 @@ To shut down your local instance, run the following commands:
 # tear down
 pnpm localenv:compose down
 
-# tear down and delete database volumes
-pnpm localenv:compose down --volumes
```
oana-lolea, line 207: We should keep it as an option.

localenv/README.md
```diff
@@ -202,9 +202,6 @@ For more ways to connect debuggers, see the Node docs for debugging: https://nod
 # tear down
 pnpm localenv:compose down
 
-# tear down and delete database volumes
-pnpm localenv:compose down --volumes
-
```
oana-lolea, line 207: We should keep it as an option.

oana-lolea (approved):

PR #3024, opened 2024-10-09
docs: added missing responses to id-provider requests

oana-lolea:
## Changes proposed in this pull request

- Added missing response codes to id-provider specs

## Context

After reviewing all the OpenAPI specs, I have found two missing response statuses in the `id-provider.yaml`:

- `500` code at the request that starts the user interaction
- `404` code at the request that finishes the user interaction

Fixes #2958 
## Checklist

- [ ] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Make sure that all checks pass
- [ ] Bruno collection updated (if necessary)
- [ ] Documentation issue created with `user-docs` label (if necessary)
- [ ] OpenAPI specs updated (if necessary)

PR #3094, opened 2024-11-08
feat(auth): use gnap error middleware on idp api

### LINKED ISSUE #3029: Properly return errors in Auth Interaction server
## Context
Currently, in the auth server routes (Open Payments & some of the front channel routes), we are throwing `GNAPServerRouteError`s, and properly catching them in `gnapServerErrorMiddleware` in order to return a well formatted GNAP error object that is [defined](https://datatracker.ietf.org/doc/html/rfc9635#section-3.6) in the GNAP spec. However, we are not doing the same for `/grant/:id/:nonce/:choice` `/grant/:id/:nonce`, the two routes exposed in the "Interaction Server".

Relevant OpenAPI [docs](https://swagger.io/docs/specification/v3_0/describing-responses/#_top)

## Todos
- [ ] Add `gnapServerErrorMiddleware` handler in the "Interaction Server" routes, add tests accordingly
- [ ] Update the IDP spec to display the different possible error objects for each route. A good idea might be to define all the possible GNAP errors we throw as components/schemas, and use them in each of the routes (see OpenAPI docs above)

njlie:

- Adds the `gnapServerErrorMiddleware` to the routes on the Identity Provider API on the auth server.

## Context

Closes #3029.

This ended up being a one-line change, as the IDP responses were added already in the spec from #850 and #3024. The code already throws these errors and had tests added as of #2400. There was one potential response missing from the spec, however.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/auth/src/openapi/specs/id-provider.yaml
+++ b/packages/auth/src/openapi/specs/id-provider.yaml
@@ -39,8 +39,16 @@ paths:
               description: Interaction id
         '401':
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '500':
           description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
       operationId: get-interact
       parameters:
         - schema:
@@ -91,9 +99,17 @@ paths:
               description: Client finish endpoint
         '401':
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '404':
           description: Not Found
-      description: "This endpoint is called by the identity provider to end the user interaction and redirect the user to the client's finish URL."
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
+      description: "To finish the user interaction for grant approval, this endpoint redirects the user to the client's finish url."
       parameters:
         - schema:
             type: string
@@ -138,8 +154,16 @@ paths:
                     type: string
         '401':
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '404':
           description: Not Found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
       operationId: get-grant
       description: |
         This endpoint is called by the identity provider to get the grant details associated with the `interactId` on the front-channel. The identity provider will display the details to the user to either accept or deny.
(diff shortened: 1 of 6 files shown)


oana-lolea (approved):

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -39,6 +39,8 @@ paths:
               description: Interaction id
         '401':
           description: Unauthorized
+        '500':
+          description: Internal Server Error
```
mkurapov, line 43: Would it be good to provide the actual GNAP response error objects in the IDP spec?

mkurapov (dismissed):
We can make my comment into a separate issue, if you'd like

packages/auth/src/openapi/specs/id-provider.yaml
njlie: I'll add them since it's a pretty bare PR as it is right now

oana-lolea (approved):

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -39,6 +39,16 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
+        '500':
+          description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
```
njlie, line 51: I figured it might make sense to include this in the auth server spec as well for Open Payments, so I captured it in an issue [here](https://github.com/interledger/open-payments/issues/518).

oana-lolea (dismissed):

packages/auth/src/openapi/specs/id-provider.yaml
mkurapov: Makes sense! if you want, we could probably merge the open payments issue first, bump the client, and import the yaml files in here so we can reference them as `./auth-server.yaml#/components/schemas/...`, what do you think? Unless we want to decouple the IDP spec from the auth server spec?

njlie: Let's bump the OP client for this :+1:

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -165,11 +193,23 @@ paths:
           description: Accepted
         '400':
           description: Not Found
```
mkurapov, line 195:
```suggestion
          description: Bad Request
```

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -165,11 +193,23 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/error-invalid-interaction'
         '401':
```
mkurapov, line 200: I think also we have the `user_denied` `invalid_request` possible in this route

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -39,6 +39,16 @@ paths:
               description: Interaction id
         '401':
```
mkurapov, line 40: Should this be 400?

mkurapov: We also have `request_denied` in this flow possible

njlie: > We also have `request_denied` in this flow possible

We shoud have `request_denied` in https://github.com/interledger/rafiki/pull/3094/files#diff-196f8d522cc447a08b51f0cf830f0c27ebc33e85c7cd936d96d1d6d053a1a754R46-R51

> Should this be 400?

The conditional for throwing this response can be satisfied by not finding the interaction based on the provided nonce + id, not being in the pending state, or the associated grant being revoked. I felt as though these cases sufficiently fell under "incorrect credentials".

Looking at it now, I think we could break up the conditional into 400 (or 404) for the case where it can't find the interaction because the id or the nonce don't match up, and 403 for when the grant is revoked or the interaction isn't in a pending state

mkurapov (approved):

PR #3095, opened 2024-11-12
fix: fixed error where no assets were found when opening create peer page

### LINKED ISSUE #3046: [BUG] An unknown error is displayed when the user try to create a peer with no assets created
## Bug Report

**Describe the bug**

Using a clean installation of Rafiki (no wallets, no peers, no assets). The user get an unknown error (internal server error) when it tries to create a new peer with no peers in Rafiki.

**To Reproduce**

1. Use a clean Rafiki instance.
2. Go to `peers` page
2. Click on 'create peer'
3. You got the error

**Expected behavior**

Rafiki should say to the user that an asset should be created before to create a peer.
Or maybe would be better to allow the user to create a new peer and automatically create a new asset.

Rafiki version: [v1.0.0-alpha.20](https://github.com/interledger/rafiki/releases/tag/v1.0.0-alpha.20)

**Screenshots**

oana-lolea:

- Fixed error when trying to create an peer with no existing assets.
- Added a dialog when opening the create peer page that can route the user to create asset page.

## Context

Fixes #3046 
When trying to create a peer with no existing assets, a 'cannot read properties of undefined' error occurred, because there was a mapper on the assets list. I fixed this by updating `loadAssets` to return `[]` in this case.

However, the user might start filling in the form and when he gets to the assets select field (which is the last one) there will be nothing. To prevent this, I added a dialog that warns the user that an asset should be created first and can go to create assets page.

No assets dialog

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/frontend/app/components/RedirectDialog.tsx
+++ b/packages/frontend/app/components/RedirectDialog.tsx
@@ -1,0 +1,107 @@
+import { Dialog, Transition } from '@headlessui/react'
+import { useNavigate } from '@remix-run/react'
+import { Button } from './ui'
+import { Fragment } from 'react/jsx-runtime'
+import { forwardRef, useImperativeHandle, useState } from 'react'
+
+export type RedirectDialogRef = {
+  display: () => void
+}
+
+type RedirectDialogProps = {
+  title: string
+  message: string
+  redirectPath: string
+  redirectButtonText: string
+  closeButtonText?: string
+}
+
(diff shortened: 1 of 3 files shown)


njlie (approved):

mkurapov (commented):
Looks good, just minor comment about copy

njlie (approved):

PR #3101, opened 2024-11-17
feat(auth): return GNAP 404 error if token cannot be rotated

### LINKED ISSUE #3100: Authorization Server should throw a `GNAPServerRouteError` if invalid token is used during rotation
The following conditional in the [access token routes](https://github.com/interledger/rafiki/blob/main/packages/auth/src/accessToken/routes.ts) file does not throw a `GNAPServerError` like the rest of the errors in that route logic.

https://github.com/interledger/rafiki/blob/be426fad6457d9928f3151cedf1401bf6b1cf65a/packages/auth/src/accessToken/routes.ts#L155-L157

It should throw a `404` http code and a `invalid_rotation` GNAP error code.

golobitch:

- Instead of returning 400 return 404 if token cannot be rotated

## Context

- fixes #3100

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/auth/src/accessToken/routes.ts
+++ b/packages/auth/src/accessToken/routes.ts
@@ -164,7 +164,7 @@ async function rotateToken(
     const errorMessage =
       error instanceof Error ? error.message : 'Could not rotate token'
     throw new GNAPServerRouteError(
-      400,
+      404,
       GNAPErrorCode.InvalidRotation,
       errorMessage
     )
--- a/packages/auth/src/accessToken/routes.test.ts
+++ b/packages/auth/src/accessToken/routes.test.ts
@@ -484,12 +484,12 @@ describe('Access Token Routes', (): void => {
       )
     })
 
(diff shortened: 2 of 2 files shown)


njlie (approved):
Thanks for picking this up. For context, this is to match the spec update here: https://github.com/interledger/open-payments/issues/518

packages/auth/src/accessToken/routes.ts
```diff
@@ -164,7 +164,7 @@ async function rotateToken(
     const errorMessage =
       error instanceof Error ? error.message : 'Could not rotate token'
     throw new GNAPServerRouteError(
-      400,
+      404,
```
mkurapov, line 167: Do we throw in the other service methods that may lead to this 404? or do we just want this error message for the `if (!newToken)` case?

mkurapov: CC @njlie

mkurapov (approved):

PR #3133, opened 2024-11-26
feat(auth): tenants table v1

### LINKED ISSUE #3113: [Multi-Tenant] `auth` tenants table
The authorization server will need to maintain its own tenants table, so that it may present clients with the correct identity server during interactions.

The auth tenants table should implement the following schema:

```
tenants {
  id: string
  idpConsentUrl: string (URL)
  idpSecret: string
}
```

Note that the `id` is not generated by the `auth` database, but is instead provided during the creation process and should correspond directly to the id of a tenant on the `backend`.

- [ ] Create `auth` tenant table migration
- [ ] Create `model.ts` file for tenant model

njlie:

- Add tenants table in migration for backend database
- Add ObjectionJS model.ts file to model tenants table schema

## Context

Closes #3113.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/auth/migrations/20241125233415_create_tenants_table.js
+++ b/packages/auth/migrations/20241125233415_create_tenants_table.js
@@ -10,6 +10,7 @@ exports.up = function (knex) {
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
+    table.timestamp('deletedAt')
   })
 }
 
--- a/packages/auth/migrations/20241125233415_create_tenants_table.js
+++ b/packages/auth/migrations/20241125233415_create_tenants_table.js
@@ -1,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.createTable('tenants', function (table) {
+    table.uuid('id').notNullable().primary()
+    table.string('idpConsentUrl').notNullable()
+    table.string('idpSecret').notNullable()
+  })
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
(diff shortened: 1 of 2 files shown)


packages/auth/src/tenant/model.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseModel } from '../shared/baseModel'
+
+export class Tenant extends BaseModel {
```
golobitch, line 3: Do we also need tenantSettings on auth ?

njlie: I don't think so. The auth server just needs to be able to map identity providers to grants - I don't think anything else is becoming tenanted.

golobitch (approved):

sanducb (approved):

mkurapov (approved):

packages/auth/migrations/20241125233415_create_tenants_table.js
```diff
@@ -0,0 +1,22 @@
...
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.createTable('tenants', function (table) {
+    table.uuid('id').notNullable().primary()
+    table.string('idpConsentUrl').notNullable()
+    table.string('idpSecret').notNullable()
+
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
+    table.timestamp('updatedAt').defaultTo(knex.fn.now())
```
mkurapov, line 12: Are these generated in this table, or passed over from backend?
Should we also include `deletedAt`?

mkurapov (changes requested):

packages/auth/migrations/20241125233415_create_tenants_table.js
golobitch: I would say that yes, this createdAt and updatedAt needs to be generated in this table since it is different entity. Bot somehow, still connected with the one in backend. with createdAt we get info when was the entity created in that database, not when it was created in the backed database. What do you think ?

mkurapov (approved):

PR #3140, opened 2024-11-27
feat(backend): tenants service (#3123)

### LINKED ISSUE #3123: [Multi-Tenant] `backend` tenant service
The backend should provide a tenant service that other services can use to retrieve tenant information, given an id.

Pseudocode for Tenant service:
```
CreateOptions {
  publicName: string
  email: string
  apiSecret: string
  idpConsentUrl: string
  idpSecret: string
}

TenantService {
  create: (options: CreateOptions) => Tenant
  get: (id: string) => Tenant
  update(id: string, input: { ...tenantFields }
  delete: (id: string) => boolean
}
```

- [ ] Create TenantService and add it to IoC container
- [ ] `create` function
  -  `create` function should call create tenant endpoint on auth server admin api
- [ ] `get` function
- [ ] `update` function
- [ ] `delete` function
  - `delete` function should call delete tenant endpoint on auth server admin api
  - use soft deletion

### Prerequisites
* https://github.com/interledger/rafiki/issues/3103
* https://github.com/interledger/rafiki/issues/3113
* https://github.com/interledger/rafiki/issues/3122

njlie:

- Adds a `TenantService` to the backend IoC container
- Adds `updatedAt` and `createdAt` in the backend tenants table.

## Context

Closes #3123.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/migrations/20241125224212_create_tenants_table.js
+++ b/packages/backend/migrations/20241125224212_create_tenants_table.js
@@ -5,10 +5,10 @@
 exports.up = function (knex) {
   return knex.schema.createTable('tenants', function (table) {
     table.uuid('id').notNullable().primary()
-    table.string('email').notNullable()
+    table.string('email')
     table.string('apiSecret').notNullable()
-    table.string('idpConsentUrl').notNullable()
-    table.string('idpSecret').notNullable()
+    table.string('idpConsentUrl')
+    table.string('idpSecret')
     table.string('publicName')
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
--- a/packages/backend/migrations/20241125224212_create_tenants_table.js
+++ b/packages/backend/migrations/20241125224212_create_tenants_table.js
@@ -7,6 +7,8 @@ exports.up = function (knex) {
     table.uuid('id').notNullable().primary()
     table.string('email').notNullable()
     table.string('apiSecret').notNullable()
+    table.string('idpConsentUrl').notNullable()
+    table.string('idpSecret').notNullable()
     table.string('publicName')
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
--- a/packages/backend/migrations/20241125224212_create_tenants_table.js
+++ b/packages/backend/migrations/20241125224212_create_tenants_table.js
@@ -8,6 +8,9 @@ exports.up = function (knex) {
     table.string('email').notNullable()
     table.string('publicName')
     table.string('apiSecret')
+
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
+    table.timestamp('updatedAt').defaultTo(knex.fn.now())
   })
 }
 
--- a/packages/backend/migrations/20241125224212_create_tenants_table.js
+++ b/packages/backend/migrations/20241125224212_create_tenants_table.js
@@ -6,11 +6,12 @@ exports.up = function (knex) {
   return knex.schema.createTable('tenants', function (table) {
     table.uuid('id').notNullable().primary()
     table.string('email').notNullable()
+    table.string('apiSecret').notNullable()
     table.string('publicName')
-    table.string('apiSecret')
 
     table.timestamp('createdAt').defaultTo(knex.fn.now())
     table.timestamp('updatedAt').defaultTo(knex.fn.now())
+    table.timestamp('deletedAt')
   })
 }
 
--- a/packages/backend/src/tenants/model.ts
+++ b/packages/backend/src/tenants/model.ts
@@ -22,11 +22,3 @@ export class Tenant extends BaseModel {
     }
   }
 }
-
-export type TenantWithIdpConfig = Pick<
-  Tenant,
-  'id' | 'email' | 'apiSecret' | 'publicName' | 'createdAt' | 'updatedAt'
-> & {
-  idpConsentUrl: string
-  idpSecret: string
-}
--- a/packages/backend/src/tenants/model.ts
+++ b/packages/backend/src/tenants/model.ts
@@ -1,4 +1,5 @@
 import { BaseModel } from '../shared/baseModel'
+import { Pojo } from 'objection'
 
 export class Tenant extends BaseModel {
   public static get tableName(): string {
@@ -8,6 +9,16 @@ export class Tenant extends BaseModel {
   public email!: string
   public apiSecret!: string
   public publicName?: string
+
+  public deletedAt?: Date
+
+  $formatJson(json: Pojo): Pojo {
+    json = super.$formatJson(json)
+    return {
+      ...json,
+      deletedAt: json.deletedAt.toISOString()
+    }
+  }
 }
 
 export type TenantWithIdpConfig = Pick<
--- a/packages/backend/src/tenants/model.ts
+++ b/packages/backend/src/tenants/model.ts
@@ -9,3 +9,11 @@ export class Tenant extends BaseModel {
   public apiSecret!: string
   public publicName?: string
 }
+
+export type TenantWithIdpConfig = Pick<
+  Tenant,
+  'id' | 'email' | 'apiSecret' | 'publicName' | 'createdAt' | 'updatedAt'
+> & {
+  idpConsentUrl: string
+  idpSecret: string
+}
--- a/packages/backend/src/tenants/model.ts
+++ b/packages/backend/src/tenants/model.ts
@@ -8,6 +8,8 @@ export class Tenant extends BaseModel {
 
   public email!: string
   public apiSecret!: string
+  public idpConsentUrl!: string
+  public idpSecret!: string
   public publicName?: string
 
   public deletedAt?: Date
--- a/packages/backend/src/tenants/service.test.ts
+++ b/packages/backend/src/tenants/service.test.ts
@@ -1,0 +1,366 @@
+import assert from 'assert'
+import { faker } from '@faker-js/faker'
+import { IocContract } from '@adonisjs/fold'
+import nock from 'nock'
+import { AppServices } from '../app'
+import { initIocContainer } from '..'
+import { createTestApp, TestContainer } from '../tests/app'
+import { TenantService } from './service'
+import { Config, IAppConfig } from '../config/app'
+import { truncateTables } from '../tests/tableManager'
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
+import { Tenant } from './model'
+
+describe('Tenant Service', (): void => {
+  let deps: IocContract<AppServices>
+  let appContainer: TestContainer
+  let tenantService: TenantService
+  let config: IAppConfig
+  let apolloClient: ApolloClient<NormalizedCacheObject>
+
+  beforeAll(async (): Promise<void> => {
+    deps = initIocContainer(Config)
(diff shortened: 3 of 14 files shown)


packages/backend/migrations/20241125224212_create_tenants_table.js
```diff
@@ -8,6 +8,9 @@ exports.up = function (knex) {
     table.string('email').notNullable()
     table.string('publicName')
     table.string('apiSecret')
+
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
+    table.timestamp('updatedAt').defaultTo(knex.fn.now())
```
golobitch, line 13: Maybe also deletedAt ? So we would leverage soft delete. What do you think?

packages/backend/src/tenants/service.test.ts
```diff
@@ -0,0 +1,366 @@
...
+                idpConsentUrl: createOptions.idpConsentUrl,
+                idpSecret: createOptions.idpSecret
+              }
+            }
+          }
+        })
+      const apolloSpy = jest.spyOn(apolloClient, 'query')
+      const tenant = await tenantService.get(createdTenant.id)
+      assert.ok(tenant)
+      expect(tenant.id).toEqual(createdTenant.id)
```
golobitch, line 70: Maybe just 
```
expect(tenant).toEqual(createdTenant)
```
? Don't know if this works :D

packages/backend/src/tenants/service.test.ts
```diff
@@ -0,0 +1,366 @@
...
+              idpConsentUrl: createOptions.idpConsentUrl
+            }
+          }
+        })
+      )
+
+      scope.done()
+    })
+
+    test('tenant creation rolls back if auth tenant create fails', async (): Promise<void> => {
```
golobitch, line 161: Yes, this one is important! Nice one!

packages/backend/src/tenants/service.ts
```diff
@@ -0,0 +1,202 @@
...
+    const variables = {
+      input: {
+        id: tenant.id,
+        idpSecret,
+        idpConsentUrl
+      }
+    }
+
+    // TODO: add type to this in https://github.com/interledger/rafiki/issues/3125
+    await deps.apolloClient.mutate({ mutation, variables })
```
golobitch, line 113: Does mutate method throw error in case of 4xx / 5xx errors ?

golobitch (commented):
Generally looks ok. Me personally, I would like to see method getPage to be implemented, that would be used for getting list of tenants

packages/backend/src/tenants/service.test.ts
njlie: It breaks because the `updatedAt` for one will always be different from the other, sadly.

packages/backend/migrations/20241125224212_create_tenants_table.js
njlie: Makes sense to me

packages/backend/src/tenants/service.ts
njlie: It does - a manifestation of this is [having to catch `400` responses inside a try/catch block in the tests](https://github.com/interledger/rafiki/pull/3140/files#diff-38f750d470d3a544f415e5642aed59c1fa4f3d899445deea7c84958eaa1af779R107-R123).

packages/backend/src/tenants/service.test.ts
golobitch: Why would updatedAt field be different ? You are creating a tenant and getting back data, and then you are querying the exact same entity ... so it should be the same one? Or am I missing something here?

njlie: Nevermind that, I conflated that result with a different test. This will work, but `createdTenant` just needs to be augmented with the `idpSecret` and `idpConsentUrl`.

packages/backend/src/tenants/model.ts
```diff
@@ -8,4 +9,22 @@ export class Tenant extends BaseModel {
...
+    }
+  }
+}
+
+export type TenantWithIdpConfig = Pick<
+  Tenant,
+  'id' | 'email' | 'apiSecret' | 'publicName' | 'createdAt' | 'updatedAt'
+> & {
+  idpConsentUrl: string
+  idpSecret: string
```
mkurapov, line 29: Do we need `idpConsentUrl` and `idpSecret` in backend?

test/integration/testenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -31,6 +31,8 @@ services:
       PRIVATE_KEY_FILE: /workspace/private-key.pem
       AUTH_SERVER_INTROSPECTION_URL: http://cloud-nine-wallet-test-auth:3107
       AUTH_SERVER_GRANT_URL: http://cloud-nine-wallet-test-auth:3106
+      AUTH_ADMIN_API_URL: 'http://cloud-nine-wallet-test-auth:3003/graphql'
+      AUTH_ADMIN_API_SECRET: 'test-secret'
```
mkurapov, line 35: Does this need to correspond to auth's `ADMIN_API_SECRET`? Or will this be a separate API altogether?

golobitch: @mkurapov I would say that on both of your comments the answer is affirmitive. We need this in backend, so that with one call, we add tenant data on both backend, and auth service.

packages/backend/src/tenants/model.ts
mkurapov: > We need this in backend, so that with one call, we add tenant data on both backend, and auth service.

I'm guessing you are referring to this comment @golobitch?

Also, I see, we are not storing the `idpConsentUrl` and `idpSecret` in the tenants table, (`TenantWithIdpConfig` is just a type). At first, _not_ storing the data makes sense to me (we shouldn't care about IDP stuff in `backend`). However, how would that work for doing pagination over tenants? If we don't store these, then we would need to call `getTenant` in the auth service for every tenant result?

packages/backend/src/tenants/service.ts
BlairCurrey: My understanding is this will be a new gql server for service-to-service communication so I guess maybe we can simply return a 400 from that server like this expects, but I just wanted to note that our current resolvers dont behave this way.

For errors we do something like `throw new GraphQLError('tenant not found', { extensions: { code: 'NOT_FOUND' } })`. I tested an existing one of these on Bruno and see the gql server replies with a 200 and error details in the body like:

```ts
{
  "errors": [
    {
      "message": "No grant",
      "locations": [
        {
          "line": 2,
          "column": 5
        }
      ],
      "path": [
        "grant"
      ],
      "extensions": {
        "code": "NOT_FOUND"
      }
    }
  ],
  "data": null
}
```

Im not necessarily against doing it a different way, if that seems better. I just wanted to note this. 

Also I think the test would incorrectly pass if we updated the mock to something like the above response because the expects in the catch would never be called. We safeguard against this sort of incorrect pass elsewhere by doing `expect.assertions(3)` (or whatever number).

njlie: Thanks for pointing that out. I updated the nocks to return proper error structure. Mercifully, the client still parses that as an error and throws, but it's a better test to have the correct response from the API.

packages/backend/src/tenants/model.ts
golobitch: Ah yes, @njlie, @mkurapov has a point here :)

njlie: Perhaps we should treat the backend's `idpConsentUrl` and `idpSecret` as some sort of cache-like thing. As in for pagination, we return the backend's entries for each tenant row, but in the query for an individual tenant, make the call to the authorization server. Whenever a tenant is mutated such that the IDP fields are changed, it would update the columns in both backend and auth.

packages/backend/src/tenants/service.ts
BlairCurrey: > Mercifully, the client still parses that as an error and throws

Good to know, was curious about that

packages/backend/src/tenants/model.ts
mkurapov: > but in the query for an individual tenant, make the call to the authorization server

The fields should basically always be the same, so I think calling to the AS isn't needed. The updates shouldn't happen frequently, and if they do, they'll be stored in the backend tenants table anyway as you describe

packages/backend/src/tenants/service.ts
```diff
@@ -0,0 +1,209 @@
...
+async function deleteTenant(
+  deps: ServiceDependencies,
+  id: string
+): Promise<void> {
+  const trx = await deps.knex.transaction()
+
+  try {
+    await Tenant.query(trx).patchAndFetchById(id, {
+      deletedAt: new Date(Date.now())
+    })
```
mkurapov, line 193: maybe we can pass through the `deletedAt` to the mutation, so both tables in backend and auth have the same value

packages/backend/src/tenants/service.ts
```diff
@@ -0,0 +1,209 @@
...
+    const mutation = gql`
+      mutation DeleteAuthTenantMutation($input: DeleteTenantInput!) {
+        deleteTenant(input: $input) {
+          sucess
```
sanducb, line 197: I assume this must be `success`

njlie: It's a basic response, but this could be changed in the PR for the auth graphql API if it is necessary.

koekiebox (approved):
LGTM. 🟢

packages/backend/src/tenants/service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+async function deleteTenant(
+  deps: ServiceDependencies,
+  id: string
+): Promise<void> {
+  const trx = await deps.knex.transaction()
+
+  await deps.tenantCache.delete(id)
+  try {
+    const deletedAt = new Date(Date.now())
```
BlairCurrey, line 185: this would be the same

```suggestion
    const deletedAt = new Date()
```

packages/backend/migrations/20241125224212_create_tenants_table.js
```diff
@@ -6,8 +6,14 @@ exports.up = function (knex) {
   return knex.schema.createTable('tenants', function (table) {
     table.uuid('id').notNullable().primary()
     table.string('email').notNullable()
```
mkurapov, line 8: lets have `email` nullable, based on our slack convo?

packages/backend/migrations/20241125224212_create_tenants_table.js
```diff
@@ -6,8 +6,14 @@ exports.up = function (knex) {
...
+    table.string('apiSecret').notNullable()
+    table.string('idpConsentUrl').notNullable()
+    table.string('idpSecret').notNullable()
```
mkurapov, line 11: I think `idpConsentUrl` and `idpSecret` can be null, integrators may not actually be integrated with an IDP anyway. (echoing @tadejgolobic on our Tues call)

BlairCurrey (approved):

mkurapov (approved):

Issue #3125, opened 2024-11-21
[Mulit-Tenant] `auth` Tenant Admin Graphql API


njlie:
Add the following to `auth` Admin GraphQL API:

```
type Mutation {
  createTenant(input: CreateTenantInput): CreateTenantMutationResponse!
  updateTenant(input: UpdateTenantInput): UpdateTenantMutationResponse!
  deleteTenant(input: DeleteTenantInput): DeleteTenantMutationResponse!
}
```

- [ ] `createTenant` mutation
- [ ] `updateTenant` mutation
- [ ] `deleteTenant` mutation

### Prerequisites
* https://github.com/interledger/rafiki/issues/3122

mkurapov:
What do we think about having this as an event we publish from backend to be consumed by auth and stored in the tenant TB?

njlie:
> What do we think about having this as an event we publish from backend to be consumed by auth and stored in the tenant TB?

What kind of event did you have in mind? Pretty much any communication between these services will suffice.

mkurapov:
I was thinking something like redis pub sub/streams as a message queue, but for V1 we can get away without using new patterns/technologies and just use a new, separate GraphQL API

BlairCurrey:
Per our discussion on slack we decided to host this api on a new http server because:

1. we dont want to expose it on the graphql admin api
2. a new graphql api for this is a bit heavy-handed
3. a new http server would be a bit simpler than something like GRPC. No extra dependencies, codegen, etc.

PR #3148, opened 2024-12-03
feat: auth service-to-service api

BlairCurrey:

- Adds a new `ServiceAPIServer` to `auth`
- Adds new tenant CRUD routes to `ServiceAPIServer`
- Adds new `authServiceAPIClient` to `backend` which is used by tenant service.

## Context

fixes: #3125 
Opened https://github.com/interledger/helm-charts/issues/47 to follow up on the env vars in the helm charts

## Checklist

- [x] Tests added/updated


packages/backend/src/auth-service-client/client.ts
```diff
@@ -0,0 +1,88 @@
...
+      } catch (error) {
+        throw new AuthServiceClientError(
+          `Failed to parse JSON response from ${path}`,
+          response.status
+        )
+      }
+    }
+
+    return (await response.text()) as T
+  }
```
BlairCurrey, line 69: I opted to use the native `fetch` here although there were some behaviors I wasnt quite sure about. I tested with `axios` and mirrored how that works:

1. not OK response (400, 500) etc. throws
2. 204 returns `undefined`
3. Success without body (200, 201, etc.) returns the response.text (just `''` if not set)
4. Interface accepts a generic which asserts the return type, although technically it could still return a `string` or `undefined` in the cases described above.

packages/auth/src/tenant/routes.ts
```diff
@@ -0,0 +1,138 @@
...
+async function createTenant(
+  deps: ServiceDependencies,
+  ctx: CreateContext
+): Promise<void> {
+  const { body } = ctx.request
+
+  await deps.tenantService.create(body)
+
+  ctx.status = 204
+}
```
BlairCurrey, line 83: @mkurapov influenced by the pattern we've been talking about w/ performance. Not 201 with the resource returned. Originally I went with 201 but that really should have a body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

thus the [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) for this and the update and delete.

packages/auth/src/tenant/routes.ts
```diff
@@ -0,0 +1,138 @@
...
+  ctx.body = toTenantResponse(tenant)
+}
+
+function toTenantResponse(tenant: Tenant): TenantResponse {
+  return {
+    id: tenant.id,
+    idpConsentUrl: tenant.idpConsentUrl,
+    idpSecret: tenant.idpSecret
+  }
+}
```
BlairCurrey, line 138: to filter out the created, deleted, update timestamps basically. figured we'd keep those in the return from the service layer and filter out here.

packages/backend/src/auth-service-client/client.ts
```diff
@@ -0,0 +1,88 @@
...
+        body: JSON.stringify(data)
+      }),
+    update: (id: string, data: Partial<Omit<Tenant, 'id'>>) =>
+      this.request(`/tenant/${id}`, {
+        method: 'PATCH',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(data)
+      }),
+    delete: (id: string) => this.request(`/tenant/${id}`, { method: 'DELETE' })
+  }
```
BlairCurrey, line 87: could probably move these methods and the `Tenant` interface out of this file if we want but opted to just put it here since it's really the only thing we need for the forseeable future.

packages/backend/src/auth-service-client/client.ts
```diff
@@ -0,0 +1,88 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    public details?: any
+  ) {
+    super(message)
+    this.status = status
+    this.details = details
+  }
+}
+
+export class AuthServiceClient {
```
BlairCurrey, line 20: Almost feels like this should be a different package outside of `backend` but this is the only consumer at this point so I felt like this keeps it simpler.

BlairCurrey: Intention is to add this as a dep to the tenant service, when thats merged in.

BlairCurrey:
Not able to add an integration test yet because the backend doesnt have any endpoints that uses the backend tenant service. However, I did do some manual integration tests by adding this to the `start` function the `backend/src/index.ts` and found everything to be working as expected:

```ts
  // Test the tenant stuff. Comment out client methods and inspect db as needed to confirm.
  const authServiceClient = container.use('authServiceClient')
  const tenant = await authServiceClient.tenant.get(config.operatorTenantId)
  console.log({ tenant }) // OK

  const id = '438fa74a-fa7d-4317-9ced-dde32ece1788'

  await authServiceClient.tenant.create({
    // OK
    id,
    idpConsentUrl: 'http://localhost:3031/mock-idp2/',
    idpSecret: '3pEcn2kkCclbOHQiGNEwhJ0rucATZhrA807HTm2rNXE='
  })

  await authServiceClient.tenant.update(id, {
    // OK
    idpConsentUrl: 'http://localhost:3031/mock-idp3/'
  })

  const deletedAt = new Date('2024-12-20T18:16:02.019Z')
  await authServiceClient.tenant.delete(id, deletedAt) // OK
  ```

**Comment by njlie** [comment]
> Not able to add an integration test yet because the backend doesnt have any endpoints that uses the backend tenant service. However, I did do some manual integration tests by adding this to the `start` function the `backend/src/index.ts` and found everything to be working as expected:
> 
> ```ts
>   // Test the tenant stuff. Comment out client methods and inspect db as needed to confirm.
>   const authServiceClient = container.use('authServiceClient')
>   const tenant = await authServiceClient.tenant.get(config.operatorTenantId)
>   console.log({ tenant }) // OK
> 
>   const id = '438fa74a-fa7d-4317-9ced-dde32ece1788'
> 
>   await authServiceClient.tenant.create({
>     // OK
>     id,
>     idpConsentUrl: 'http://localhost:3031/mock-idp2/',
>     idpSecret: '3pEcn2kkCclbOHQiGNEwhJ0rucATZhrA807HTm2rNXE='
>   })
> 
>   await authServiceClient.tenant.update(id, {
>     // OK
>     idpConsentUrl: 'http://localhost:3031/mock-idp3/'
>   })
> 
>   const deletedAt = new Date('2024-12-20T18:16:02.019Z')
>   await authServiceClient.tenant.delete(id, deletedAt) // OK
> ```

We can add the proper test for this in https://github.com/interledger/rafiki/issues/3124

**Review by njlie — APPROVED** [review]
LGTM

File: packages/auth/src/config/app.ts [hunk 6]
```diff
@@ -43,6 +43,7 @@ export const Config = {
   authPort: envInt('AUTH_PORT', 3006),
   interactionPort: envInt('INTERACTION_PORT', 3009),
   introspectionPort: envInt('INTROSPECTION_PORT', 3007),
+  serviceAPIPort: envInt('SERVICE_API_PORT', 3011),
```
mkurapov, line 46: what do you think about calling this Tenant service instead? 

Are you approaching this from this service being used for something else, for example?

packages/backend/src/auth-service-client/client.ts
```diff
@@ -0,0 +1,90 @@
...
+        )
+      }
+    }
+
+    return (await response.text()) as T
+  }
+
+  public tenant = {
+    get: (id: string) =>
+      this.request<Tenant>(`/tenant/${id}`, { method: 'GET' }),
```
mkurapov, line 70: do we need the headers object in here as well? 
Can also just have it in the `request` fn

packages/auth/src/shared/utils.ts
```diff
@@ -104,3 +104,8 @@ export async function verifyApiSignature(
 
   return verifyApiSignatureDigest(signature as string, ctx.request, config)
 }
+
+// Intended for Date strings like "2024-12-05T15:10:09.545Z" (e.g., from new Date().toISOString())
+export function isValidDateString(date: string): boolean {
```
mkurapov, line 109: small, but would be good to write a few tests for this

packages/backend/src/app.ts
```diff
@@ -265,6 +266,7 @@ export interface AppServices {
   paymentMethodHandlerService: Promise<PaymentMethodHandlerService>
   ilpPaymentService: Promise<IlpPaymentService>
   localPaymentService: Promise<LocalPaymentService>
+  authServiceClient: AuthServiceClient
```
mkurapov, line 269:
```suggestion
  authServiceClient: Promise<AuthServiceClient>
```

packages/auth/src/shared/utils.ts
BlairCurrey: got em already here https://github.com/interledger/rafiki/pull/3148/files#diff-04126f4f8bb03998a1c86d2551e9442d11fab8403c4d7aff2031088f2dc36092R149

packages/auth/src/config/app.ts
BlairCurrey: > Are you approaching this from this service being used for something else, for example?

Yes, that was the idea. Figured it can serve the purpose of any service-to-service communication. No specific things in mind but if we did want to expose some other resource I dont imagine we'd want a new api just for that.

packages/backend/src/app.ts
BlairCurrey: Technically I dont think I need this? Although I realize everything else is a promise but its driven by the factory function we pass into the singleton. I have:

```ts
  container.singleton('authServiceClient', () => {
    return new AuthServiceClient(config.authServiceApiUrl)
  })
```

I didn't make that `async` since it didn't need to be, hence the `AuthServiceClient` and not a promise fort he type here. I do see that we return a promise elsewhere where it doesnt seem like we need to. Recent in memory cache for example.

```ts
  container.singleton('tenantCache', async () => {
    return createInMemoryDataStore(config.localCacheDuration)
  })
```

So either I'm missing something or we've just done it this way because the other ones looked like this?

packages/backend/src/auth-service-client/client.ts
BlairCurrey: The second arg in request? Yeah I think we need this. Its more than just the headers. `request` just wraps `fetch` and fetch needs the the `method` at the very least since its used for gets, posts, delete, etc.

mkurapov: As in, I think we need `{ 'Content-Type': 'application/json' }` also for the GET method.

My suggestion is to just have `{ 'Content-Type': 'application/json' }` once on L29

BlairCurrey: i added it in `request` as suggested. Actually I dont think we need it for the `GET` (just for specifying format of body no?) but it shouldn't hurt in any case.

Issue #3114, opened 2024-11-20
[Multi-Tenant] Tenanted Wallet Addresses


njlie:
Wallet Addresses should become tenanted as per the Multi-Tenancy design, and will be logically separated in the `backend` database by a `tenantId` row that serves as a foreign key on the tenants table.

- [x] Add `tenantId` row to `walletAddresses` table that is required and maps to an id on the `tenants` table.
- [x] Add `tenantId` to database queries on wallet addresses
- [x] Acquire `tenantId` from request headers on wallet address GraphQL requests and pass them into wallet address service logic.
- [x] Include `tenantId` in base URL of `authServer` and `resourceServer` fields of the current wallet address response.
  - See https://github.com/interledger/rafiki/pull/2986 for existing implementation on old feature branch.
  - eg: `{ ..., resourceServer: OPEN_PAYMENTS_URL/tenantId/ }`

### Prerequisites
* https://github.com/interledger/rafiki/issues/2928

### Pull Request
https://github.com/interledger/rafiki/pull/3152

golobitch:
> Add tenantId to queries on wallet addresses

Note: this means to add tenantId to SQL queries and not GraphQL queries. Just for the clarification :)

PR #3152, opened 2024-12-05
feat(backend): tenant support for wallet address (#3114)

koekiebox:

- See: https://github.com/interledger/rafiki/issues/3114

## Checklist

- [x] Tests added/updated


packages/backend/src/open_payments/wallet_address/model.ts
```diff
@@ -53,6 +62,9 @@ export class WalletAddress
   public readonly assetId!: string
   public asset!: Asset
 
+  public readonly tenantId!: string
+  public tenant!: Tenant
```
mkurapov, line 66: Do we actually need the `tenant!: Tenant` property on the model? I'm not sure we'd end up using it

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -176,13 +181,19 @@ async function createWalletAddress(
     const walletAddress = await WalletAddress.query(
       deps.knex
     ).insertGraphAndFetch({
+      tenantId: options.tenantId,
```
mkurapov, line 184: I think we would need the gets and update to also take in tenantId in the query, right?

koekiebox: I don't think so, since you cannot move a wallet between tenants? Unless that is something we want? @mkurapov

packages/backend/src/open_payments/wallet_address/model.ts
koekiebox: Originally I thought we would need it, but since we have `authServer` and `resourceServer`, we are good. Will update.

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: Updated.

packages/backend/src/open_payments/wallet_address/model.ts
koekiebox: We don't need it @mkurapov , I removed it.

packages/backend/src/app.ts
```diff
@@ -393,7 +394,21 @@ export class App {
     )
 
     let tenantApiSignatureResult: TenantApiSignatureResult
-    if (this.config.env !== 'test') {
+    if (this.config.env === 'test') {
+      const tenantService = await this.container.use('tenantService')
+      let tenant = await tenantService.get(this.config.operatorTenantId)
+      if (!tenant) {
+        tenant = await Tenant.query().insertAndFetch({
```
koekiebox, line 401: I really don't like this. Do you guys have an alternative suggesetion? 
The "re-insert" is necessary due to the tenant table being cleared, then the `TenantedApolloContext` does not have a tenant.

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -52,11 +52,11 @@ export const getWalletAddresses: QueryResolvers<ApolloContext>['walletAddresses'
     }
   }
 
-export const getWalletAddress: QueryResolvers<ApolloContext>['walletAddress'] =
+export const getWalletAddress: QueryResolvers<TenantedApolloContext>['walletAddress'] =
   async (parent, args, ctx): Promise<ResolversTypes['WalletAddress']> => {
     const walletAddressService = await ctx.container.use('walletAddressService')
     const walletAddress = await walletAddressService.get(args.id)
-    if (!walletAddress) {
+    if (!walletAddress || walletAddress.tenantId !== ctx.tenant.id) {
```
koekiebox, line 59: We want to ensure we are not being 'hacked', the tenant can only access their own wallet addresses.

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -69,18 +69,20 @@ export const getWalletAddress: QueryResolvers<ApolloContext>['walletAddress'] =
...
+export const getWalletAddressByUrl: QueryResolvers<TenantedApolloContext>['walletAddressByUrl'] =
   async (
     parent,
     args,
     ctx
   ): Promise<ResolversTypes['WalletAddress'] | null> => {
     const walletAddressService = await ctx.container.use('walletAddressService')
     const walletAddress = await walletAddressService.getByUrl(args.url)
-    return walletAddress ? walletAddressToGraphql(walletAddress) : null
+    return walletAddress && ctx.tenant.id === walletAddress.tenantId
```
koekiebox, line 80: We want to ensure we are not being 'hacked', the tenant can only access their own wallet addresses.

packages/backend/migrations/20241203112902_add_tenant_to_wallet_address.js
```diff
@@ -0,0 +1,25 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return Promise.all([
+    knex.schema.alterTable('walletAddresses', function (table) {
+      table.uuid('tenantId').index().notNullable()
+      //table.foreign(['tenantId']).references('tenants.id')
+    })
+  ])
```
BlairCurrey, line 11: What is the intention with the commented out line here? Waiting on something? I imagine this would be a foreign key and we could uncomment this.

packages/backend/src/app.ts
BlairCurrey: Can you elaborate on why we need to do this? 

Also, dont we need to insert some initial operator for non test environments? Like when initiallty deploying these changes? Im fuzzy on that but thought we were either going to seed it with some default values which need to be updated, or inserting somewhere like here.

If that is the case, I'm just wondering if we can piggy off that here. Insert some default operator which can be updated by the integrator or in the tests.

BlairCurrey: > The "re-insert" is necessary due to the tenant table being cleared

Oh, are we talking about how `truncateTables` clears out everything for the tests? You could add the tenant table to the ignore list inside that function.

packages/backend/src/graphql/resolvers/wallet_address.ts
BlairCurrey: Jason what are you thoughts on where we do the access control (ie does the tenantId match) for the different cases? Updating, getting, etc.

In this PR we are mixing whether the tenant id gets passed into the service layer or not here. The update passes it in and uses it in the `where` to find the resource to update, and the `get` does not.  Looking at the implementation and usage in this PR I think its very simple and like that. I just wonder what happens with the worker where I think the pattern breaks down.

For the worker, we wont have a `tenantId`. So the `walletAddressService.update({tenantId ... })` (or similar) wont work. `OutgoingPayment` update would run into this, for example. 

Nathan and I have had some discussions about where the access control happens. There were a few ideas I recall:

- always doing access control in a middleware that runs before the resolver. getting a wallet address? lookup wallet address and check `tenantId`. updating outgoing payment? lookup and check `tenantId`. This does add some extra calls but centralizes it and keeps the services totally `tenantId` agnostic. Do we really want to care about the tenant everywhere in our application or just at the entry? Does have some complexity in terms of getting the tenant Id from the request as I recall.
- make separate functions with shared logic for updating a resource and doing access control or updating a resource and not doing access control.

BlairCurrey: I had a draft PR for a previous version of multi-tenancy with a helper function in the different services (quote, wallet address, etc) for checking if a tenant could access. Just to elaborate on option #1 above a bit.

https://github.com/interledger/rafiki/pull/3026/files#diff-1695f39120835125efdde9389ae4e400acc244ab1a89105ab531631f1606dff0

BlairCurrey: Do we need an "is operator" check here?

For things like the create, update, etc. the operator needs to specify a tenant id. But I dont think thats the case for the get. Imagine they are on the page listing the wallet addresses. When they click one of them, the tenantId for that resource isnt getting set, and then this resolver fires.

packages/backend/src/app.ts
njlie: I would also suggest with this approach to have a `truncateTables` call in the `afterAll` block that doesn't ignore it, so it gets cleaned up eventually

packages/backend/src/graphql/resolvers/wallet_address.ts
njlie: I think with operations like this where a single item is affected, the tenant id on the context should be considered within the permission of the requester - the middleware should take care if the tenant id is the "right" one, so it could be passed into the service layer.

I would say `isOperator` is more relevant to pagination requests, where it would be necessary to know if it needs to be filtered, or if the request is to `createTenant` which should be operator-only.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -340,9 +344,15 @@ async function getWalletAddressPage(
   pagination?: Pagination,
   sortOrder?: SortOrder
 ): Promise<WalletAddress[]> {
-  return await WalletAddress.query(deps.knex)
-    .getPage(pagination, sortOrder)
-    .withGraphFetched('asset')
+  const addresses = await WalletAddress.query(deps.knex).getPage(
```
njlie, line 347: Would chaining `.withGraphFetched('asset')` work here as well?

packages/backend/src/app.ts
koekiebox: Ok, I agree. So to confirm: 
1. The `20241205153035_seed_operator_tenant.js` will create the tenant as part of migration script;
2. The `truncateTables` will be updated to include `tenant` as one of the ignored tables

koekiebox: Updated.

packages/backend/src/graphql/resolvers/wallet_address.ts
koekiebox: Ok, I have updated the logic to not have the `tenantId` be mandatory on the service layer. 

### In the GraphQL, we ensure that we have flexibility for providing an alternative `tenantId` (for when `isOperator == true`) than what is on the signature, as well as ensuring they are correct when fetching.

Example snippet that performs the validation on the resolver:
```node
// The blow is done when fetching the walletAddress:
if (ctx) {
    const tenantId = tenantIdToProceed(
      ctx.isOperator,
      ctx.tenant.id,
      walletAddress.tenantId
    )
    if (!tenantId) {
      const err = WalletAddressError.InvalidTenantIdNotAllowed
      throw new GraphQLError(errorToMessage[err], {
        extensions: {
          code: errorToCode[err]
        }
      })
    }
  }

// Then in operations like create and update walletAddress:
 const tenantId = tenantIdToProceed(
      ctx.isOperator,
      ctx.tenant.id,
      args.input.tenantId
    )
    if (!tenantId) {
      const err = WalletAddressError.InvalidTenantIdNotAllowed
      throw new GraphQLError(errorToMessage[err], {
        extensions: {
          code: errorToCode[err]
        }
      })
    }


```

> The function for determining the correct `tenantId` to use:

```node
/**
 * The tenantId to use will be determined as follows:
 * - When an operator and the {tenantId} is present, return {tenantId}
 * - When an operator and {tenantId} is not present, return {signatureTenantId}
 * - When NOT an operator and {tenantId} is present, but does not match {signatureTenantId}, return {undefined}
 * - Otherwise return {signatureTenantId}
 *
 * @param isOperator is operator
 * @param signatureTenantId the signature tenantId
 * @param tenantId the intended tenantId
 */
export function tenantIdToProceed(
  isOperator: boolean,
  signatureTenantId: string,
  tenantId?: string
): string | undefined {
  if (isOperator && tenantId) return tenantId
  else if (isOperator) return signatureTenantId
  return tenantId && tenantId !== signatureTenantId
    ? undefined
    : signatureTenantId
}

```


Let me know what you think? @BlairCurrey and @njlie

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: It would, but then it won't be using the cached `asset`.

packages/backend/migrations/20241203112902_add_tenant_to_wallet_address.js
koekiebox: Yip, missed that one, thanks @BlairCurrey fixed.

packages/backend/src/graphql/resolvers/wallet_address.ts
mkurapov: @koekiebox instead of `args.input.tenantId`, we would be getting the tenantId from the header, right?

mkurapov: When the operator is trying to paginate across all resources (e.g. all wallet addresses, for example), what should the behaviour be? we don't pass in the tenantId at all?

packages/backend/src/app.ts
```diff
@@ -393,7 +393,11 @@ export class App {
     )
 
     let tenantApiSignatureResult: TenantApiSignatureResult
-    if (this.config.env !== 'test') {
+    if (this.config.env === 'test') {
+      const tenantService = await this.container.use('tenantService')
+      const tenant = await tenantService.get(this.config.operatorTenantId)
+      tenantApiSignatureResult = { tenant, isOperator: true }
```
mkurapov, line 399: ah, we basically got to the same conclusion :)

I ended up setting a `tenant-id` header in the apollo client during tests: https://github.com/interledger/rafiki/pull/3206/files#diff-5b481c1a1464d3e282df7102a0cb5882701fc2b2af81037e203812e13acf0176R414-R433

Both work, I'm trying to think whether we would actually want to be able to pass in a different tenant id during tests...

packages/backend/src/graphql/resolvers/wallet_address.ts
BlairCurrey: > When the operator is trying to paginate across all resources (e.g. all wallet addresses, for example), what should the behaviour be? we don't pass in the tenantId at all?

I think we still need to pass in the tenant id.

How I see it:

- The user sets api credentials (`tenantId` and `apiSecret`)
- The user visits `/wallet-addresses` and the apollo client makes a request with the `tenant-id` header and signing using the apiSecret
- The backend detects it's the operator by tenant id and sets it on the context
- Bypass as necessary for the operator

mkurapov: > Bypass as necessary for the operator

I assume when we get the page of wallet addresses in the backend, we just fetch all wallet addresses without providing tenantId

BlairCurrey: Hmm, dont we have to pass it in to paginate correctly? Can't filter after we get the page. Not sure I see an alternative.

mkurapov: For the operator tenant that just wants to get all wallet addresses, I think we wouldn't provide it. Otherwise, for a non-operator tenant, we would for sure.

packages/backend/src/app.ts
njlie: Might there be tests to verify that requests are properly permissioned between an operator making a request versus a tenant?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -39,16 +39,15 @@ export type WalletAddressAdditionalPropertyInput = Pick<
 >
 
 export interface CreateOptions extends Options {
+  tenantId?: string
```
njlie, line 42: Where might this be called where `tenantId` is unavailable to be passed into the service call?

packages/backend/src/graphql/resolvers/wallet_address.ts
koekiebox: Yes, that is what the below function does. 
The `tenantId` is optional in the GraphQL, which means the `signatureTenantId` would be preferred, then for `isOperator`, the caller is able to provide a different (other than signature) `tenantId`.


```node
/**
 * The tenantId to use will be determined as follows:
 * - When an operator and the {tenantId} is present, return {tenantId}
 * - When an operator and {tenantId} is not present, return {signatureTenantId}
 * - When NOT an operator and {tenantId} is present, but does not match {signatureTenantId}, return {undefined}
 * - Otherwise return {signatureTenantId}
 *
 * @param isOperator is operator
 * @param signatureTenantId the signature tenantId
 * @param tenantId the intended tenantId
 */
export function tenantIdToProceed(
  isOperator: boolean,
  signatureTenantId: string,
  tenantId?: string
): string | undefined {
  if (isOperator && tenantId) return tenantId
  else if (isOperator) return signatureTenantId
  return tenantId && tenantId !== signatureTenantId
    ? undefined
    : signatureTenantId
}
```

packages/backend/src/app.ts
koekiebox: Good point @njlie , I will add those in. I added a new error called: `WalletAddressError.InvalidTenantIdNotAllowed`:

```node
    const tenantId = tenantIdToProceed(
      ctx.isOperator,
      ctx.tenant.id,
      args.input.tenantId
    )
    if (!tenantId) {
      const err = WalletAddressError.InvalidTenantIdNotAllowed
      throw new GraphQLError(errorToMessage[err], {
        extensions: {
          code: errorToCode[err]
        }
      })
    }
```

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -97,8 +98,10 @@ export const createWalletAddress: MutationResolvers<ApolloContext>['createWallet
         addProps.push(toAdd)
       })
 
+    const tenantId = tenantIdToUseAndValidate(ctx, args.input.tenantId)
```
koekiebox, line 101: @mkurapov 
Unfortunately the `TenantedApolloContext` is an interface. I added this utility function `tenantIdToUseAndValidate` to make the `tenantId` retrieval and validation easier:
```
export function tenantIdToUseAndValidate(
  ctx: TenantedApolloContext,
  tenantId?: string
): string {
  const returnVal = tenantIdToProceed(ctx.isOperator, ctx.tenant.id, tenantId)
  if (!returnVal) {
    const err = WalletAddressError.InvalidTenantIdNotAllowed
    throw new GraphQLError(errorToMessage[err], {
      extensions: {
        code: errorToCode[err]
      }
    })
  }
  return returnVal
}
```

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: By not forcing the tenant-id, the `operatorTenantId` will be used from the config. 
Would it be better for us to just force the `tenantId` @njlie ? 

```nodejs
const tenantId = options.tenantId
      ? options.tenantId
      : deps.config.operatorTenantId
```

packages/backend/src/middleware/tenant/index.ts
```diff
@@ -0,0 +1,58 @@
...
+type Request = () => Promise<unknown>
+
+interface TenantValidateMiddlewareArgs {
+  deps: { ctx: ApolloContext }
+  tenantIdInput: string | undefined
+  onFailValidation: Request
+  next: Request
+}
+
+export async function validateTenantMiddleware(
```
koekiebox, line 12: Validate for any `TenantedApolloContext` and `tenantIdInput` is present.

koekiebox: @mkurapov @njlie @BlairCurrey 
This is a bit of a hack, where the `forTenantId` is sitting on the `TenantedApolloContext`. Will figure that one out.

But the rest basically does what we talked about, without having to update any additional code on each of the resolvers. 

As long as the resolver type is `TenantedApolloContext`, the `tenantId` validations are performed, and it is safe to always use `forTenantId`.

What do you guys think?

packages/backend/src/app.ts
```diff
@@ -395,15 +398,20 @@ export class App {
     )
 
     let tenantApiSignatureResult: TenantApiSignatureResult
-    if (this.config.env !== 'test') {
+    if (this.config.env === 'test') {
+      const tenantService = await this.container.use('tenantService')
```
BlairCurrey, line 402: noticed this is returning an `any` type. Need to add `tenantService: Promise<TenantService>` to the `AppServices` interface in this file.

packages/backend/src/middleware/tenant/index.ts
BlairCurrey: I think this is a bit simpler:

```ts
export async function setForTenantIdMiddleware(
  args: SetForTenantIdMiddlewareArgs
): ReturnType<Request> {
  const {
    deps: { ctx },
    tenantIdInput,
    onFail,
    next
  } = args
  if (!('tenant' in ctx && 'isOperator' in ctx)) return onFail()
  if (!ctx.isOperator && tenantIdInput) return onFail()

  ctx.forTenantId = tenantIdInput ?? ctx.tenant.id
  return next()
}
```

Differences:

- Errors when we dont have the tenant nor operator. Correct me if I'm wrong on this but we never want to allow an request without the tenant and if they are an operator. Hence `tenant` and `isOperator` are required properties on the `TenantedApolloContext`
- Some renaming to convey this is just getting the id, not so much validating

BlairCurrey: > This is a bit of a hack, where the forTenantId is sitting on the TenantedApolloContext. Will figure that one out.

For the types I think you can remove it from the initial `TenantedApolloContext` and make a new interface like:

```ts
interface ForTenantIdContext extends TenantedApolloContext {
  forTenantId: string
}
```

And use in the middleware like:

```ts
export function setForTenantIdGraphQLMutationMiddleware(): {
  Mutation: IMiddleware<TenantedApolloContext, ForTenantIdContext>
} {
```

Then update the context type passed into the resolvers: 


```ts
export const getWalletAddress: QueryResolvers<ForTenantIdContext>['walletAddress'] =
```

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -166,15 +170,20 @@ export const triggerWalletAddressEvents: MutationResolvers<ApolloContext>['trigg
...
-  createdAt: new Date(+walletAddress.createdAt).toISOString(),
-  status: walletAddress.isActive
-    ? WalletAddressStatus.Active
-    : WalletAddressStatus.Inactive
-})
+export function walletAddressToGraphql(
+  walletAddress: WalletAddress,
+  ctx?: TenantedApolloContext
+): SchemaWalletAddress {
+  if (ctx) tenantIdToUseAndValidate(ctx, walletAddress.tenantId)
```
mkurapov, line 177: Do we still need this here now that we have the middleware?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -173,9 +174,14 @@ async function createWalletAddress(
...
       : undefined
 
+    const tenantId = options.tenantId
+      ? options.tenantId
+      : deps.config.operatorTenantId
+
     const walletAddress = await WalletAddress.query(
       deps.knex
     ).insertGraphAndFetch({
+      tenantId,
```
mkurapov, line 184: just to not forget: should check in this fn that the asset is belongs to the correct tenant

koekiebox: Done.

packages/backend/migrations/20241203112902_add_tenant_to_wallet_address.js
```diff
@@ -0,0 +1,25 @@
+/**
```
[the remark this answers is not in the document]
koekiebox: Done.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -168,6 +169,12 @@ async function createWalletAddress(
   }
 
   try {
+    const tenantId = options.tenantId
+      ? options.tenantId
+      : deps.config.operatorTenantId
+    const asset = await deps.assetService.get(options.assetId, tenantId)
```
mkurapov, line 175: I don't think the service needs to know/use `config.operatorTenantId`, it'll get it from the calling service if needs to

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -306,13 +313,63 @@ describe('Wallet Address Resolvers', (): void => {
...
     })
+
+    test('bad input data when not allowed to perform cross tenant create', async (): Promise<void> => {
+      const badInputData = {
+        tenantId: 'ae4950b6-3e1b-4e50-ad24-25c065bdd3a9',
+        assetId: input.assetId,
+        url: input.url
+      }
+      try {
+        await appContainer.apolloClient
```
mkurapov, line 324:
```suggestion
      try {
        expect.assertions(2)
        await appContainer.apolloClient
```

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: I think this was just an old commit.

packages/backend/src/tests/quote.ts
```diff
@@ -168,6 +168,7 @@ export async function createQuote(
   const withGraphFetchedArray = [
     'asset',
     'walletAddress',
+    'walletAddress.tenant',
```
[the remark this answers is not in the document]
koekiebox: Fixed.

packages/backend/src/app.ts
```diff
@@ -448,11 +454,12 @@ export class App {
...
     koa.use(
       koaMiddleware(this.apolloServer, {
-        context: async (): Promise<TenantedApolloContext> => {
+        context: async (): Promise<ForTenantIdContext> => {
           return {
             ...tenantApiSignatureResult,
             container: this.container,
-            logger: await this.container.use('logger')
+            logger: await this.container.use('logger'),
+            forTenantId: this.config.operatorTenantId
```
mkurapov, line 462: Including the operator id by default feels less safe, we should either let it be undefined (and just declare this as the `TenantedApolloContext`, or have `ForTenantIdContext` let `tenantId` be undefined)

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -45,18 +47,27 @@ export const getWalletAddresses: QueryResolvers<ApolloContext>['walletAddresses'
...
-      edges: walletAddresses.map((walletAddress: WalletAddress) => ({
-        cursor: walletAddress.id,
-        node: walletAddressToGraphql(walletAddress)
-      }))
+      edges: walletAddresses
+        .filter(
+          (wa: WalletAddress) =>
+            tenantIdToProceed(ctx.isOperator, ctx.tenant.id, wa.tenantId) !=
+            undefined
+        )
```
mkurapov, line 55: I'm not sure if this works from the client perspective, since we could get into a situation where we set the limit to be e.g. 20 and then because of the filter, end up filtering out a few results. Then we have a situation where we got 15 results, but `hasNextPage` is still true.

mkurapov: Maybe we can have the service methods be able to take in a list of tenantIds? That way we can pass in both the operator tenant id and the (on behalf id if necessary)

koekiebox: @mkurapov, I agree with your first statement. Then the second, I was thinking the opposite. We make the `tenantId` an optional argument for the `WalletAddressesConnection`.
This enables us to do the following:
- When the `args.input.tenantId` is set, we do the usual `ctx.forTenantId` - This allows for specific tenant paging when the operator is permitted
- When the caller `isOperator` and `args.input.tenantId` is not provided, "default" paging is performed where `tenantId` is undefined and not part of the selection criteria
- When the caller `!isOperator`, then we always use: `ctx.forTenantId` - this enforces that the caller is only able to see their own tenant data

Again, we can determine most of this from the resolver. So that we only need to update `getPage` to have an optional parameter for `tenantId`.

packages/backend/src/shared/baseModel.ts
```diff
@@ -66,13 +69,18 @@ class PaginationQueryBuilder<M extends Model, R = M[]> extends QueryBuilder<
...
+      : ''
+
     /**
      * Forward pagination
      */
     if (typeof pagination?.after === 'string') {
       const comparisonOperator = sortOrder === SortOrder.Asc ? '>' : '<'
       return this.whereRaw(
-        `("${tableName}"."createdAt", "${tableName}"."id") ${comparisonOperator} (select "${tableName}"."createdAt" :: TIMESTAMP, "${tableName}"."id" from ?? where "${tableName}"."id" = ?)`,
+        `("${tableName}"."createdAt", "${tableName}"."id") ${comparisonOperator} (select "${tableName}"."createdAt" :: TIMESTAMP, "${tableName}"."id" from ?? where "${tableName}"."id" = ?${tenantFilterClause})`,
```
BlairCurrey, line 83: I think maybe its better not to filter by `tenantId` in the `baseModel.` Especially since we are making it optional here. 

Take a look at the filter in `webhookService.getPage`. 

https://github.com/interledger/rafiki/blob/main/packages/backend/src/webhook/service.ts#L247-L260

It applies filters in the service. Much easier than forming the string manually in the base model. And we can keep all the logic of if it needs to include tenants out of here which makes the already complicated `getPage` less complicated. It will return the correct page for the same filters.

If we keep it here we also need to apply it to the backward pagination. Also is `tenantId` safe to directly insert into the string or does that create an opportunity for sql injection? Does it come directly from the client or are we retrieving it from the db (and is it it safely formed when its created)? The second arg in `whereRaw` are the variables to sanitize and insert: `[this.modelClass().tableName, pagination.after]`. Not sure if `tenantId` should conditionally be included there as well.

I dont really see any advantage to putting it in the base model especially when you consider we want it to be optional. Its just much easier to handle in the caller I think.

BlairCurrey: So in the service it would be something like

```ts
  ...

  const query = WalletAddress.query(deps.knex)

  if (tenantId) {
    query.where('tenantId', tenantId)
  }

  const page = await query.getPage(pagination, sortOrder)
  
  ...
```

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -35,21 +35,27 @@ import {
 import { getPageTests } from './page.test'
```
njlie, line 35: Should there be tests to check if operators can perform wallet address mutations for other tenants? I recall debating whether or not if operators creating wallet addresses for tenants was a case we felt was worth handling, but maybe for the other operations?

packages/backend/migrations/20250117112902_add_tenant_to_wallet_address.js
```diff
@@ -0,0 +1,32 @@
...
+      return knex.raw(
+        `UPDATE "walletAddresses" SET "tenantId" = (SELECT id from "tenants" LIMIT 1)`
+      )
+    })
+    .then(() => {
+      return knex.schema.alterTable('walletAddresses', (table) => {
+        table.uuid('tenantId').notNullable().alter()
+      })
+    })
+}
```
BlairCurrey, line 20: @mkurapov Should we edit this migration for the other resources? It just seems important that we ensure it defaults to the correct tenant and running them at once aids in that. Frankly I even wonder about adding this to the one that seeds the initial operator for that reason, or using an env var instead of the LIMIT 1. Either option reduces assumptions about the state of things when the migrations are run.

For development purposes I see why we'd end up with multiple migrations per resource (because thats how we've split the work) but for deployment purposes I see that more as one change.

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -52,11 +58,14 @@ export const getWalletAddresses: QueryResolvers<ApolloContext>['walletAddresses'
...
 
-export const getWalletAddress: QueryResolvers<ApolloContext>['walletAddress'] =
+export const getWalletAddress: QueryResolvers<TenantedApolloContext>['walletAddress'] =
   async (parent, args, ctx): Promise<ResolversTypes['WalletAddress']> => {
     const walletAddressService = await ctx.container.use('walletAddressService')
     const walletAddress = await walletAddressService.get(args.id)
-    if (!walletAddress) {
+    if (
+      !walletAddress ||
+      !tenantIdToProceed(ctx.isOperator, ctx.tenant.id, walletAddress.tenantId)
```
BlairCurrey, line 67: We shouldn't need this `tenantIdToProceed`. We should be able to just use `forTenantId` set by the middleware. The idea being the middleware picks the correct tenant id but beyond that the service does the access control.

packages/backend/src/graphql/schema.graphql
```diff
@@ -803,6 +805,9 @@ type WalletAddress implements Model {
 
   "Additional properties associated with the wallet address."
   additionalProperties: [AdditionalProperty]
+
+  "Tenant ID of the wallet address."
+  tenantId: String
```
BlairCurrey, line 810: I dont think we want to return the tenantId on the wallet address. unless im missing some use case?

packages/backend/migrations/20250117112902_add_tenant_to_wallet_address.js
mkurapov: My assumption here is that as long as we have the "seed tenant operator" migration first, we should be ok. Then we get:
1. seed operator tenant
2. add tenantId to wallet address and backfill it
3. add tenantId to incoming payments 
4. ...

given that we run the migrations at startup in order, in the backfill migrations, we can only ever fetch the single operator tenant in these migrations. 

The only problem I see here is that we run migrations manually, run the  "seed tenant operator" and create a new tenant without running the other migrations - unlikely to happen. So we either, leave it alone (add a sort by oldest, or use the operator tenant id directly

packages/backend/src/graphql/schema.graphql
```diff
@@ -1226,6 +1231,8 @@ type CreateReceiverResponse {
 }
 
 input CreateWalletAddressInput {
+  "Unique identifier of the tenant associated with the wallet address. This cannot be changed. Optional, if not provided, the tenantId will be obtained from the signature."
+  tenantId: String
```
mkurapov, line 1235:
```suggestion
  tenantId: ID
```

packages/backend/src/graphql/schema.graphql
mkurapov: I think this can be used by the tenant operator during pagination, since then it'll be able to see which tenant this wallet address corresponds to. 

If we don't want to add this in the frontend, we can leave it out

packages/backend/src/tests/walletAddress.ts
```diff
@@ -31,7 +32,10 @@ export async function createWalletAddress(
   const walletAddressService = await deps.use('walletAddressService')
   const walletAddressOrError = (await walletAddressService.create({
     ...options,
-    assetId: options.assetId || (await createAsset(deps)).id,
+    assetId:
+      options.assetId ||
+      (await createAsset(deps, undefined, options.tenantId)).id,
+    tenantId: options.tenantId || (await createTenant(deps)).id,
```
mkurapov, line 38: If I'm not mistaken, if we don't pass in `options.tenantId`, we will end up creating an asset and a wallet address with different `tenantId`s

packages/backend/src/graphql/resolvers/wallet_address.ts
mkurapov: You're suggesting passing the tenantId into the `await walletAddressService.get(args.id)`, correct?

(Also, currently, the `setForTenantIdGraphQLMutationMiddleware` only runs for mutations. So we would pass in  `ctx.tenant.id` instead of `forTenantId`)

packages/backend/src/middleware/tenant/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  const {
+    deps: { context },
+    tenantIdInput,
+    onFailValidation,
+    next
+  } = args
+  if (!tenantIdInput) {
+    ;(context as ForTenantIdContext).forTenantId = context.tenant.id
+    return next()
+  }
```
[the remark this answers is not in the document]
koekiebox: Agree. Fixed.

packages/backend/src/graphql/resolvers/wallet_address.ts
koekiebox: @BlairCurrey, to add to what @mkurapov described, it also ensures that:
- We are avoiding sending the tenantId to the service layer
- Making use of a single index instead of both `tenantId` and `walletId` for the lookup
- We agreed to rather return `UnknownWalletAddress` for the queries (keeping the functionality consistent to what it was before), where middleware would always throw `BadUserInput` with message: `Assignment to the specified tenant is not permitted`

packages/backend/src/graphql/schema.graphql
koekiebox: Agreed. Resolving this one.

packages/backend/src/graphql/resolvers/wallet_address.ts
mkurapov: @koekiebox based on our original call, I was also thinking we _would_ pass in `tenantId` in the `walletAddressService.get(id, tenantId?)` function. Since we'll likely need it in other service methods such as incoming payment creation (where we will pass in tenantId in the `walletAddressService.get`)

packages/backend/src/graphql/schema.graphql
koekiebox: Fixed.

packages/backend/src/graphql/resolvers/wallet_address.ts
koekiebox: Ah, got it @mkurapov , I see you also did that with the asset. Updating ♻️ .

packages/backend/src/tests/walletAddress.ts
koekiebox: Correct.

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -23,19 +23,25 @@ import {
...
     args,
     ctx
   ): Promise<ResolversTypes['WalletAddressesConnection']> => {
     const walletAddressService = await ctx.container.use('walletAddressService')
-    const { sortOrder, ...pagination } = args
+    const { tenantId, sortOrder, ...pagination } = args
     const order = sortOrder === 'ASC' ? SortOrder.Asc : SortOrder.Desc
+
+    const tenantForLookup =
+      tenantId && ctx.isOperator ? tenantId : ctx.tenant.id
```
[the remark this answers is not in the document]
koekiebox: That is correct.

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -35,21 +35,27 @@ import {
...
   let walletAddressService: WalletAddressService
+  let assetService: AssetService
 
   beforeAll(async (): Promise<void> => {
-    deps = await initIocContainer({
+    deps = initIocContainer({
       ...Config,
-      localCacheDuration: 0
+      localCacheDuration: 0,
+      adminApiSecret: '123' //to force not being an operator.
```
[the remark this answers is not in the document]
koekiebox:
```node
if (tenant) {
          tenantApiSignatureResult = {
            tenant,
            isOperator: tenant.apiSecret === this.config.adminApiSecret
          }
```

@mkurapov we make use of the operator check for the queries, so it would still be relevant (in order to switch the isOperator to false).

image

packages/backend/src/graphql/resolvers/wallet_address.test.ts
koekiebox: @njlie By default. the `beforeAll` will ensure that the `isOperator` is `false`. 
I added the following tests to evaluate the above:
- `bad input data when not allowed to perform cross tenant create`
- `bad input data when not allowed to perform cross tenant update`
- `Can get page of wallet addresses with tenantId param`
- `Can get a wallet address` updated to ensure the NotFound error is thrown
Which I think is extensive enough, due to the same `tenantIdToProceed` mechanism being used.

I am going to resolve this for now, but if you think we need more, please re-open and ping me.

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: No, it has been updated already.

packages/backend/src/open_payments/wallet_address/routes.ts
```diff
@@ -60,8 +61,8 @@ export async function getWalletAddress(
     )
 
   ctx.body = walletAddress.toOpenPaymentsType({
-    authServer: deps.config.authServerGrantUrl,
-    resourceServer: deps.config.openPaymentsUrl
+    authServer: `${ensureTrailingSlash(deps.config.authServerGrantUrl)}${walletAddress.tenantId}`,
```
mkurapov, line 64: can we avoid the need to have `urlWithoutTenantId` if we don't include the `tenantId` in the `authServer` response for now? We can just add it in once the AS is ready. Might be less work overall

packages/backend/src/graphql/schema.graphql
```diff
@@ -53,6 +53,8 @@ type Query {
   walletAddress(
     "Unique identifier of the wallet address."
     id: String!
+    "Unique identifier of the tenant associated with the wallet address. Optional, if not provided, the tenantId will be obtained from the signature."
+    tenantId: String
```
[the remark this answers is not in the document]
koekiebox: We do need it for when you want to fetch a wallet as an operator for a tenant that is not "your own".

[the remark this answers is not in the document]
mkurapov: Can we pass in undefined tenantId into the get call if it's an operator request?

packages/backend/src/open_payments/wallet_address/routes.ts
koekiebox: Ah, I did it the other way around, so that we only need to look for the `urlWithoutTenantId` usages? Can we leave it like this? 😬

packages/backend/src/graphql/resolvers/wallet_address.test.ts
```diff
@@ -718,6 +841,56 @@ describe('Wallet Address Resolvers', (): void => {
             }
           ]
         })
+
+        // Attempt to switch tenant:
+        try {
```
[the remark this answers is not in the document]
koekiebox: Done.

packages/backend/src/graphql/schema.graphql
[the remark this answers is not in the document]
koekiebox: We can now. :-) Yes.

packages/backend/src/graphql/resolvers/wallet_address.ts
```diff
@@ -69,18 +91,21 @@ export const getWalletAddress: QueryResolvers<ApolloContext>['walletAddress'] =
...
+export const getWalletAddressByUrl: QueryResolvers<TenantedApolloContext>['walletAddressByUrl'] =
   async (
     parent,
     args,
     ctx
   ): Promise<ResolversTypes['WalletAddress'] | null> => {
     const walletAddressService = await ctx.container.use('walletAddressService')
     const walletAddress = await walletAddressService.getByUrl(args.url)
-    return walletAddress ? walletAddressToGraphql(walletAddress) : null
+    return walletAddress &&
```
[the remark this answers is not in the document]
koekiebox: Done, also added the negative tests for it.

packages/backend/src/graphql/schema.graphql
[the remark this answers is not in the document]
mkurapov: > Can we pass in undefined tenantId into the get call if it's an operator request?

Sorry, I mean without needing `tenantId` in the schema (since it wouldn't ever differ from the tenantId in the signature). We can have logic in the resolvers (like the pagination) and pass in `undefined` for the tenantId in the service layer `walletAddressService.get`.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -338,11 +359,21 @@ async function getWalletAddressByUrl(
...
+  tenantId?: string
 ): Promise<WalletAddress[]> {
-  return await WalletAddress.query(deps.knex)
-    .getPage(pagination, sortOrder)
-    .withGraphFetched('asset')
+  const query = WalletAddress.query(deps.knex)
+
+  if (tenantId && tenantId.length > 0) {
+    query.where({ tenantId })
+  }
```
mkurapov, line 369:
```suggestion
  if (tenantId) {
    query.where({ tenantId })
  }
```

test/integration/lib/test-actions/index.ts
```diff
@@ -4,6 +4,7 @@ import { parseCookies } from '../utils'
 import { WalletAddress, PendingGrant } from '@interledger/open-payments'
 import { AdminActions, createAdminActions } from './admin'
 import { OpenPaymentsActions, createOpenPaymentsActions } from './open-payments'
+import { urlWithoutTenantId } from '../../../../packages/backend/src/shared/utils'
```
mkurapov, line 7: I'm surprised we can reach across to the other package like this. I think we should just re-declare/copy the function somewhere in this project just to make sure it's ok.

packages/backend/src/tests/walletAddress.ts
mkurapov: I don't think that could/should happen realistically. Wallet address tenantId should match tenantId for the asset.

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -168,6 +170,12 @@ async function createWalletAddress(
   }
 
   try {
+    const tenantId = options.tenantId
+      ? options.tenantId
+      : deps.config.operatorTenantId
+    const asset = await deps.assetService.get(options.assetId, tenantId)
```
mkurapov, line 176:
```suggestion
    const asset = await deps.assetService.get(options.assetId, options.tenantId)
```
I think this should be sufficient?

packages/backend/src/graphql/resolvers/wallet_address.ts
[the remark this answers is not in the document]
koekiebox: Fixed it now. to allow for it being defined and undefined. 
```node
const { tenantId, sortOrder, ...pagination } = args
    const order = sortOrder === 'ASC' ? SortOrder.Asc : SortOrder.Desc

    const walletAddresses = await walletAddressService.getPage(
      pagination,
      order,
      ctx.isOperator ? tenantId : ctx.tenant.id
    )
```

packages/backend/src/open_payments/wallet_address/service.ts
koekiebox: I am blind as a bat. I thought I removed that already! 🥹

test/integration/lib/test-actions/index.ts
koekiebox: I thought it was pretty neat. :-) 
Fixed it now.

packages/backend/src/open_payments/wallet_address/errors.ts
```diff
@@ -27,5 +29,7 @@ export const errorToMessage: {
   [WalletAddressError.UnknownAsset]: 'unknown asset',
   [WalletAddressError.UnknownWalletAddress]: 'unknown wallet address',
   [WalletAddressError.DuplicateWalletAddress]:
-    'Duplicate wallet address found with the same url'
+    'Duplicate wallet address found with the same url',
+  [WalletAddressError.InvalidTenantIdNotAllowed]:
+    'Assignment to the specified tenant is not permitted.'
```
mkurapov, line 34: Looks like this is not used anymore

packages/backend/migrations/20250117112902_add_tenant_to_wallet_address.js
BlairCurrey: yeah Im not recommending any changes here. just something to think about.

mkurapov (approved):

BlairCurrey (approved):

PR #3156, opened 2024-12-05
3149/mk/seed operator tenant

mkurapov:

- 

## Checklist

- [ ] Tests added/updated


packages/auth/src/config/app.ts
```diff
@@ -78,7 +78,8 @@ export const Config = {
     process.env.REDIS_TLS_CA_FILE_PATH,
     process.env.REDIS_TLS_KEY_FILE_PATH,
     process.env.REDIS_TLS_CERT_FILE_PATH
-  )
+  ),
+  operatorTenantId: envString('OPERATOR_TENANT_ID')
```
mkurapov, line 82: required (will throw if not provided)

packages/backend/migrations/20241205153035_seed_operator_tenant.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.raw(`
+        INSERT INTO "tenants" ("id", "apiSecret") 
+        VALUES ('${process.env['OPERATOR_TENANT_ID']}', '${process.env['OPERATOR_TENANT_SECRET']}')
```
mkurapov, line 8: technically, we don't need to store the `OPERATOR_TENANT_SECRET` in the DB if we have it in the process, thoughts?

BlairCurrey: How does this behave when running manually? I guess it errors unless you've set them in the process. Not sure there is much to do about this. Maybe throw an error to make it more obvious?

sanducb: Operator email must be inserted as well since it is required.

mkurapov: @sanducb Nathan [updated](https://github.com/interledger/rafiki/pull/3140/commits/3625938cdb5fa3ce8d8ab2b06da0e68fd30357b4) his PR to make some fields including email optional

sanducb:
> One thing I'm concerned about is an integrator entering a mismatched OPERATOR_TENANT_ID across `backend` and `auth`. It would require a manual rollback of migrations to fix this.
> 
> To alleviate this, I was even thinking we don't store the operator tenant in the DB, only in memory. Unfortunately, this is not possible in the way that we have the `tenantId` FK on each of our resources (unless resources _without_ a tenantId are assumed to be under the operator tenant).
> 
> CC @njlie

Should we treat `OPERATOR_TENANT_ID` as a different kind of value from the usual tenant ids? Would it make sense to have the operator tenant id as a known hardcoded value 

> One thing I'm concerned about is an integrator entering a mismatched OPERATOR_TENANT_ID across `backend` and `auth`. It would require a manual rollback of migrations to fix this.
> 
> To alleviate this, I was even thinking we don't store the operator tenant in the DB, only in memory. Unfortunately, this is not possible in the way that we have the `tenantId` FK on each of our resources (unless resources _without_ a tenantId are assumed to be under the operator tenant).
> 
> CC @njlie

I think another solution would be to programatically derive an operator tenant id from an already known "seed"/secret by both `backend` and `auth` servers such that it's always the same no matter 

> One thing I'm concerned about is an integrator entering a mismatched OPERATOR_TENANT_ID across `backend` and `auth`. It would require a manual rollback of migrations to fix this.
> 
> To alleviate this, I was even thinking we don't store the operator tenant in the DB, only in memory. Unfortunately, this is not possible in the way that we have the `tenantId` FK on each of our resources (unless resources _without_ a tenantId are assumed to be under the operator tenant).
> 
> CC @njlie 

I think a solution could be having a sort of "handshake" between backend and auth servers in order to agree upon matching `OPERATOR_TENANT_ID` value. This of course would introduce overhead and would introduce tight coupling between `auth` and `backend` servers.

koekiebox (approved):

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -78,6 +78,8 @@ services:
       ILP_CONNECTOR_URL: ${CLOUD_NINE_CONNECTOR_URL:-http://cloud-nine-wallet-backend:3002}
       ENABLE_TELEMETRY: true
       KEY_ID: 7097F83B-CB84-469E-96C6-2141C72E22C0
+      OPERATOR_TENANT_ID: 438fa74a-fa7d-4317-9ced-dde32ece1787
+      OPERATOR_TENANT_SECRET: 4RRtVErbQqdI9E6o6vOOKuLEbchVvOBvFIXzLWX9GWQ=
```
njlie, line 82: Per our call today, the `ADMIN_API_SECRET` could be converted into this env variable, or be used instead of adding a new env variable

njlie (approved):

PR #3164, opened 2024-12-06
feat(backend): tenant signature validation for admin api

### LINKED ISSUE #2928: [Multi-Tenant] Use http-signatures to determine tenant identity during requests
During tenanted Admin GraphQL requests, the server should leverage the existing `http-signature` signing scheme to determine if the requester is a particular tenant. This should be achieved through the following:

* GraphQL requests should be augmented with a `tenantId` in the header.
* A middleware on the Admin GraphQL server should use this header to retrieve the [`apiSecret` on the associated entry in the tenants table](https://github.com/interledger/rafiki/issues/2932).
* This middleware should then construct a signature with the `apiSecret` it retrieved and either accept or reject the request based on if the signature is valid.
* If the request is accepted, the subsequent business logic should use the `tenantId` provided in the header as part of its input.

We might want to also allow signatures generated using the `ADMIN_API_SECRET` environment variable to be used as valid authentication, as way for operators to identify themselves.

- [ ] Update Admin API middleware to expect `tenantId` in header during request
- [ ] Update Admin API middleware to call signature verification method that constructs signature from a tenant's `apiSecret`, throw if invalid

### Prerequisites:
* #3103

njlie:

- Adds a utility to determine if a http signature was signed by a tenant, the operator, or neither and augment the context of the request to reflect that.

## Context

Fixes #2928.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -399,7 +399,7 @@ export class App {
 
     if (this.config.env !== 'test') {
       async (ctx: TenantedHttpSigContext, next: Koa.Next): Promise<void> => {
-        if (!verifyTenantOrOperatorApiSignature(ctx, this.config)) {
+        if (!(await verifyTenantOrOperatorApiSignature(ctx, this.config))) {
           ctx.throw(401, 'Unauthorized')
         }
         return next()
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -219,7 +219,7 @@ type ContextType<T> = T extends (
 const WALLET_ADDRESS_PATH = '/:walletAddressPath+'
 
 export interface TenantedApolloContext extends ApolloContext {
-  tenant?: Tenant
+  tenant: Tenant
   isOperator: boolean
 }
 
@@ -392,19 +392,17 @@ export class App {
       }
     )
 
-    let tenantApiSignatureResult: TenantApiSignatureResult = {
-      tenant: undefined,
-      isOperator: false
-    }
+    let tenantApiSignatureResult: TenantApiSignatureResult
     if (this.config.env !== 'test') {
       koa.use(async (ctx, next: Koa.Next): Promise<void> => {
         const result = await getTenantFromApiSignature(ctx, this.config)
         if (!result) {
           ctx.throw(401, 'Unauthorized')
-        }
-        tenantApiSignatureResult = {
-          tenant: result?.tenant,
-          isOperator: result?.isOperator ? true : false
+        } else {
+          tenantApiSignatureResult = {
+            tenant: result.tenant,
+            isOperator: result.isOperator ? true : false
+          }
         }
         return next()
       })
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -397,22 +397,13 @@ export class App {
       }
     )
 
-<<<<<<< HEAD
     if (this.config.env !== 'test') {
-      koa.use(async (ctx, next: Koa.Next): Promise<void> => {
-        if (!(await verifyApiSignature(ctx, this.config))) {
+      async (ctx: TenantedHttpSigContext, next: Koa.Next): Promise<void> => {
+        if (!verifyTenantOrOperatorApiSignature(ctx, this.config)) {
           ctx.throw(401, 'Unauthorized')
-=======
-    if (this.config.adminApiSecret) {
-      koa.use(
-        async (ctx: TenantedHttpSigContext, next: Koa.Next): Promise<void> => {
-          if (!verifyTenantOrOperatorApiSignature(ctx, this.config)) {
-            ctx.throw(401, 'Unauthorized')
-          }
-          return next()
->>>>>>> 3805b100 (feat(backend): tenant signature validation for admin api)
         }
-      )
+        return next()
+      }
     }
 
     koa.use(
--- a/packages/backend/src/app.ts
+++ b/packages/backend/src/app.ts
@@ -101,7 +101,10 @@ import { LocalPaymentService } from './payment-method/local/service'
 import { GrantService } from './open_payments/grant/service'
 import { AuthServerService } from './open_payments/authServer/service'
 import { Tenant } from './tenants/model'
-import { verifyTenantOrOperatorApiSignature } from './shared/utils'
+import {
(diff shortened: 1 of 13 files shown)


packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +172,64 @@ async function canApiSignatureBeProcessed(
...
+  ctx.tenant = undefined
+  ctx.isOperator = false
+  const { headers } = ctx.request
+  const signature = headers['signature']
+  if (!signature) {
+    return false
+  }
+
+  const tenantService = await ctx.container.use('tenantService')
+  const tenantId = headers['tenantid']
```
mkurapov, line 196:
```suggestion
  const tenantId = headers['tenant-id']
```
nit, but maybe kebab case is better for http headers?

packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +172,64 @@ async function canApiSignatureBeProcessed(
...
+  const tenant = tenantId ? await tenantService.get(tenantId) : undefined
```
mkurapov, line 197: should we return false if we can't find the tenant?

packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +172,64 @@ async function canApiSignatureBeProcessed(
...
+      signature as string,
+      ctx.request,
+      config,
+      config.adminApiSecret as string
+    )
+  ) {
+    ctx.tenant = tenant
+    ctx.isOperator = true
+    return true
+  }
```
mkurapov, line 228: (related to my [PR](https://github.com/interledger/rafiki/pull/3156/files#diff-b3971e30f50aa3c5a4843015630a6aafd3ea2f1846a23763dc7189d7ce80c462))
If we think that `apiSecret` is required in the DB schema, and we are seeding the operator tenant, we could actually avoid this extra check, and just determine whether its the operator by doing `tenant.apiSecret = config.adminApiSecret` after verifyingApiSignatureDigest

mkurapov (commented):
To clarify, is the idea that we can run Rafiki without an operator tenant?

packages/backend/src/app.ts
```diff
@@ -384,12 +398,14 @@ export class App {
     )
 
     if (this.config.adminApiSecret) {
```
mkurapov, line 400: We can remove this check if this secret is required, right?

packages/backend/src/shared/utils.ts
njlie: The idea here is to account for non-tenanted requests (e.g. createTenant, listTenants, other pagination requests). But this is moot since our call where the tenant id is to be expected in any request.

packages/backend/src/shared/utils.test.ts
```diff
@@ -258,4 +269,182 @@ describe('utils', (): void => {
...
+          },
+          url: '/graphql'
+        },
+        {},
+        appContainer.container
+      )
+
+      ctx.request.body = requestBody
+
+      const result = await verifyTenantOrOperatorApiSignature(ctx, config)
```
mkurapov, line 444: Might be good to check that the `tenantService.get` was in fact called

packages/backend/src/shared/utils.ts
```diff
@@ -184,5 +232,10 @@ export async function verifyApiSignature(
   if (!(await canApiSignatureBeProcessed(signature as string, ctx, config)))
     return false
 
-  return verifyApiSignatureDigest(signature as string, ctx.request, config)
+  return verifyApiSignatureDigest(
+    signature as string,
+    ctx.request,
+    config,
+    config.adminApiSecret as string
```
mkurapov, line 239:
```suggestion
    config.adminApiSecret
```
Since it should already be typed

mkurapov: nit: but we can pass in either whole of `config` or `adminApiSecret` + `adminApiSignatureVersion` explicitly

packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +172,52 @@ async function canApiSignatureBeProcessed(
...
+
+  const tenantService = await ctx.container.use('tenantService')
+  const tenantId = headers['tenant-id']
+  const tenant = tenantId ? await tenantService.get(tenantId) : undefined
+
+  if (!tenant) return false
+
+  if (!(await canApiSignatureBeProcessed(signature, ctx, config))) return false
+
+  // First, try validating with the tenant api secret
```
mkurapov, line 203:
```suggestion

```

packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +172,52 @@ async function canApiSignatureBeProcessed(
   return true
 }
 
+/*
+  Verifies http signatures by first attempting to replicate it with a secret
+  associated with a tenant id in the headers, then with the configured admin secret.
```
mkurapov, line 177: Would be good to update this comment

packages/backend/src/app.ts
```diff
@@ -144,6 +145,19 @@ export type HttpSigContext = AppContext & {
...
+type TenantedHttpSigRequest = Omit<HttpSigContext['request'], 'headers'> & {
+  headers: TenantedHttpSigHeaders
+}
+
+export type TenantedHttpSigContext = HttpSigContext & {
+  headers: TenantedHttpSigHeaders
+  request: TenantedHttpSigRequest
+  tenant?: Tenant
+  isOperator: boolean
+}
```
mkurapov, line 159: These types seem to be more suited for the Open Payments server. The main thing we need to do here is to add the the tenant object to the GraphQL context (`ApolloContext` & lines 409 -> 418), such that we can get the tenant information in the resolvers

packages/backend/src/shared/utils.ts
```diff
@@ -171,6 +173,53 @@ async function canApiSignatureBeProcessed(
   return true
 }
 
+export interface TenantApiSignatureResult {
+  tenant?: Tenant
```
mkurapov, line 177:
```suggestion
  tenant: Tenant
```
(same for `TenantedApolloContext`)

mkurapov (approved):

sanducb (approved):

Issue #3139, opened 2024-11-27
docs: split manage liquidity doc into multiple pages (title edit)


melissahenderson:
After the team spoke with Max yesterday, we decided to make a few changes to the Rafiki IA.

Brad is working on a peering integration doc, so some items A.) can't be complete until #2967 is done and/or B.) will be taken care of as part of 2967.

- [x] Place the peering integration doc within the Integration > Requirements section.
- [x] Create a new section in Integration, called Deployment, and move the "Docker Compose" and "Helm and Kubernetes" pages into the section.
- [x] Move Integration > Services so that it's a subsection of Deployment (e.g., Integration > Deployment > Services > auth service).
- [x] The content in the "Manage peering relationships" page should become part of Brad's peering integration page. The "Manage" page should go away.
- [ ] Split the "Manage Liquidity" page in Administration into multiple pages. Have a section under Administration called either "Liquidity" or "Manage Liquidity". Within that section, have separate pages for Asset liquidity, Peer liquidity, Payment liquidity, and Two-phase withdrawals.
- [x] Update the sidenav's TOC as needed.

brad-dow:
The items checked in the description are taken care of in this PR: https://github.com/interledger/rafiki/pull/3166

This issue basically turns into "split 'Manage Liquidity' into multiple pages"

Issue #3090, opened 2024-11-07
docs: standardize graphql api examples


melissahenderson:
Before working on this issue, wait for:
* https://github.com/interledger/rafiki/pull/3102 to be merged
* https://github.com/interledger/rafiki/issues/3139 to be merged
* Melissa to speak with Sarah about light/dark theme and code block border changes

Example of what we want will eventually be available at `/admin/liquidity/asset-liquidity#deposit-asset-liquidity`, pending 3139. 

In general:
* Shorten header to be more direct, don’t include mutation name, don’t include code style
* If any code block is too long, use wrap (` ```graphql wrap `)

Create three tabs.

**Operation tab**
* Add the mutation to a code block
* Specify `graphql` as the language

**Arguments tab**
* Add the JSON request to a code block
* Specify `json` as the language
* Delete the variable/description table
* Beneath the code block, add _For more information about this mutation's input object, see `NameOfInputObjectInput`._ Link the input object to its page in the API docs.

**Response tab**
* Add the JSON response to a code block
* Specify `json` as the language

melissahenderson:
@brad-dow I was able to merge 3139. The stuff described above is now visible in the liquidity pages. Eg https://rafiki.dev/admin/liquidity/asset-liquidity/#deposit-asset-liquidity

PR #3166, opened 2024-12-06
docs: peering integration

brad-dow:
Initial draft (repurposed the manging peering relationships admin page into a technical integration guide).

Completed a few tasks from https://github.com/interledger/rafiki/issues/3139 as well, including the following:

- Place the peering integration doc within the Integration > Requirements section.
- Create a new section in Integration, called Deployment, and move the "Docker Compose" and "Helm and Kubernetes" pages into the section.
- Move Integration > Services so that it's a subsection of Deployment (e.g., Integration > Deployment > Services > auth service).
- The content in the "Manage peering relationships" page should become part of Brad's peering integration page. The "Manage" page should go away.
- Update sidenav's TOCs as needed

Fixed all of the links that the page moving and TOC updating broke


packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
+---
+title: Peering integration guide
```
melissahenderson, line 2: The other requirements pages are simply titled "Assets" or "Wallet addresses". Do you think we should follow the same format here, or maybe the other page titles should be updated?

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+Whether you are using the Backend Admin API or the Rafiki Admin application, the underlying configurations and requirements remain the same. Choose the interface that best suits your individual workflow.
+:::
+
+## Prerequisites
+
+Before adding a peer, you and the account servicing entity you intend to peer with must both:
+
+- Run an implementation of an [Interledger connector](/integration/deployment/services/backend-service#interledger-connector) (ideally Rafiki).
+- Agree on an [asset](/overview/concepts/accounting#assets) for the peering relationship. You can set up multiple peering relationships with the same peer based on different assets. At least one asset shared by you and your peer must be added to your Rafiki instance prior to setting up the peering relationship. For more information, refer to [Assets](/integration/requirements/assets/).
+- Agree on a `maxPacketAmount` value, which specifies how many packets a payment is split into.
```
melissahenderson, line 23: I know this came from the original doc. The maxPacketAmount variable further down in the page is listed as optional. So, maybe it's not a prerequisite? Or maybe its more of a best practice to set an amount. I'm also not sure if there's a default.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+- Communicate your respective static <LinkOut href="https://interledger.org/developers/rfcs/ilp-addresses/">Interledger (ILP) addresses</LinkOut>.
```
melissahenderson, line 24: Is this something that get assigned as part of Rafiki set up? If so, where would the ASE using this page find the address?

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+- Communicate a connection endpoint for the other peer to send packets to.
+- Exchange `auth` tokens for the connection endpoint.
```
melissahenderson, line 26: Again, I know this was all from the original doc, but I think it's assuming the ASE already knows about connection endpoints and packets and auth tokens. I'm not sure if that's an accurate assumption. Maybe this is something to revisit after the changes are merged.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+- Agree on a settlement mechanism (a means of paying one another for the successful forwarding and delivery of packets), which is outside the scope of Interledger and Rafiki.
+
+Optionally, you may also deposit an initial liquidity for your peer.
+
+:::note
+Peering is not required unless you want to participate in transactions with another ASE running Rafiki. For foundational peering concepts, refer to [Peers](/overview/concepts/interledger/#peers).
```
melissahenderson, line 32: I don't think it's a requirement that the other ASE use Rafiki, just that they are on the Interledger network somehow (with Rafiki being the easiest way to achieve that). Also, the title for this page in the sidenav is Peers, and here we're referring the reader to another page called Peers. That might be confusing.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+:::
+
+## Setting up peering in Rafiki
+
+The basic workflow of setting up a peering relationship starts with adding the agreed upon asset, then adding a peer, and then creating a wallet address from where you can send and receive payments.
```
melissahenderson, line 37: General comment: In places where you say "a peer", maybe change them to "your peer" (where appropriate).

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+{
+  "input": {
+	"staticIlpAddress": "g.othergreatwallet",
+	"name": "The Other Great Wallet"
+	"http": {
+  		"incoming": {"authTokens": ["mytoken"]},
+  		"outgoing": {"endpoint": "ilp.othergreatwallet.com", "authToken": "theirtoken"}
+	  },
+	"assetId": "INSERT_ASSET_ID",
+	"initialLiquidity": <optionally, an intial amount of liquity to provision. Liquidity can also be deposited via the `DepositPeerLiquidity`>
```
melissahenderson, line 86: Small thing: Either change end to "deposited via DepositPeerLiquidity" or "deposited via the DepositPeerLiquidity mutation"

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+</CodeBlock>
+
+<div class="overflow-table">
+
+| Variable                   | Description                                                                                                                 | Required |
+| -------------------------- | --------------------------------------------------------------------------------------------------------------------------- | -------- |
+| `assetID`                  | The ID of the asset that you and your peer will use to ultimately settle your net obligations outside of Interledger.       | Y        |
+| `staticILPaddress`         | Your peer’s static ILP address (the example uses `g.othergreatwallet`)                                                      | Y        |
+| `name`                     | Your peer’s name (the example uses “The Other Great Wallet”)                                                                | Y        |
+| `http.incoming.authTokens` | The token your peer will use to present and connect to and send packets to your Rafiki instance.                            | Y        |
```
melissahenderson, line 100: Can you check that the descriptions for incoming.authTokens and outgoing.authtoken are correct? The info on the [Backend service page](https://rafiki.dev/integration/services/backend-service/#interledger-connector) makes me question what's here. Also, is incoming plural (authTokens) and outgoing singular? Should outgoing be authToken?

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+| `http.outgoing.endpoint`   | Your peer’s connector endpoint. By default it is on local port 0.0.0.0:3002                                                 | Y        |
+| `http.outgoing.authtoken`  | The token that you will use to present to your peer and connect to and send packets to their Rafiki instance.               | Y        |
+| `initialLiquidity`         | Initial amount of liquidity to deposit for peer. Liquidity can also be deposited using the `DepositPeerLiquidity` mutation. | N        |
```
melissahenderson, line 103: Ah, here it says mutation. Can you add "the" to the first sentence? ...deposit for the peer.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+        "name": "The Other Great Wallet"
+      }
+    }
+  }
+}
+```
+
+</CodeBlock>
+
+### Create wallet addresses
```
melissahenderson, line 132: Let's talk about how to handle this section. It reads as if the ASE needs to create a wallet address for their peer, and I don't believe that's right. 

Also it says "refer to Assets" and links to Wallet Addresses :)

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+After adding a peer, we need to create a wallet address. This wallet address is where payments are sent from and received. To learn how to create a new wallet address, refer to [Assets](/integration/requirements/wallet-addresses/).
+
+## Managing peers
+
```
melissahenderson, line 137: Let's also talk about this section. It makes it sound like the ASE should expect to perform ongoing maintenance, but in an ideal world, I don't believe there should be much for the ASE to worry about other than liquidity. I have a few more thoughts but it's a lot to type out.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,287 @@
...
+</CodeBlock>
+
+<div class="overflow-table">
+
+| Variable                   | Description                                                                                                   | Required |
+| -------------------------- | ------------------------------------------------------------------------------------------------------------- | -------- |
+| `id`                       | The ID of the peer you wish to update.                                                                        | Y        |
+| `maxPacketAmount`          | Maximum packet amount that the peer accepts.                                                                  | N        |
+| `liquidityThreshold`       | A webhook event will notify the Account Servicing Entity if peer liquidity falls below this new value.        | N        |
```
hajjimo, line 195: While the example only uses `maxPacketAmount` and `liquidityThreshold` and identifies them as two possible optional fields, I wonder if we should add all the possible fields to the table so that the reader doesn't assume those are the only two.  Alternatively, we could remove the table.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: Changed to "Peers"

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: It's not required and it doesn't need to be discussed with a peer. Reworded this based of off feedback from Max and moved to the next (new) section "Optional settings)

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: ILP addresses are self-assigned and stored local to Rafiki as an environment variable. Updated wording here to reflect this. Worth considering adding additional information somewhere else on this page or maybe somewhere else completely.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: Agreed. This type of information would be beneficial to have here or somewhere else (maybe with the ILP address stuff). Let's revisit this one after the peering guide is merged in.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: Fixed

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: This is correct. HttpIncoming accepts an array of auth tokens while HttpOutgoing accepts a single token.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: Removing "adding wallet address" for now.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: I reworded the introductory paragraph to deemphasize the frequency of editing peers. Hopefully this change captures the essence of "you shouldn't have to edit these peers once they are set up properly". I also moved the contents of "Edit a peer" and "Delete a peer" into details elements to lessen the screen real estate impact that made it feel like they were a significant part of the workflow.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
brad-dow: I'm not dismissing this feedback, but I think that the reader could look at the API reference to see that these are not the only possible fields. In https://github.com/interledger/rafiki/issues/3090, we are going to update these code examples to link to a mutation's input object to hopefully help bridge that gap. Our code examples shouldn't have to account for all fields available.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,289 @@
...
+- Communicate a connection endpoint for the other peer to send packets to.
+- Exchange `auth` tokens for the connection endpoint.
+- Agree on a settlement mechanism (a means of paying one another for the successful forwarding and delivery of packets), which is outside the scope of Interledger and Rafiki.
+
+### Optional settings
+
+- Deposit an `initialLiquidity` for your peer. Liquidity can also be deposited later using the `depositPeerLiquidity` mutation.
+- Define a `maxPacketAmount` value, which specifies the maximum packet size you are willing to accept from the peer. Your peer's `maxPacketAmount` value does not need to match, as this value is independently set by each ASE. If omitted, payments will not be broken into smaller packets.
+
+:::note
```
melissahenderson, line 33: Just a suggestion... move this note up higher. I know it might look a bit busy having a note panel and tip panel so close to each other.

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,289 @@
...
+Peering is not required unless you want to participate in transactions with another ASE on the Interledger network. For foundational peering concepts, refer to the Peers section of [Interledger Concepts](/overview/concepts/interledger/#peers).
+:::
+
+## Setting up peering in Rafiki
```
melissahenderson, line 37: Suggest "Set up" instead of "Setting up"

packages/documentation/src/content/docs/integration/requirements/peers.mdx
```diff
@@ -0,0 +1,289 @@
...
+        "name": "The Other Great Wallet"
+      }
+    }
+  }
+}
+```
+
+</CodeBlock>
+
+## Managing peers
```
melissahenderson, line 134: Suggest "Manage peers" instead of "Managing"

melissahenderson (approved):
I left a few very minor suggestions, but am approving this PR.

melissahenderson (dismissed):

melissahenderson (approved):

Issue #3141, opened 2024-11-28
Check quote expiry in outgoing payment worker


mkurapov:
## Context
Currently, we check the `expiresAt` of the quote during outgoing payment creation. If it is valid, we create the outgoing payment, store it in the DB.

Once the payment is funded, the outgoing payments worker starts processing the payment, whether that is local or over ILP. If the payment fails, it will stay in the SENDING state and the worker will keep retrying it (up to some maximum number of times).

In the worker, however, we do not check the expiry of the quote. The quote could expire between the time when we created the payment, and when it is being processed (or retried) in the worker. We should prevent processing an outgoing payment for an expired quote. (good spot would be in `lifecycle.ts > handleSending`)

## Todos
- [ ] Prevent outgoing payment being processed for expired quotes (mark it as FAILED). Make sure the `outgoing_payment.failed` webhook is correctly sent.

CollinsMunene:
Hi @mkurapov , would it be correct to state that, a quote is marked as expired if the expiresAt time of the quote is less than or equal to the time the outgoing payment was created?

mkurapov:
@CollinsMunene yes :+1: 

and to confirm, outgoing payment should _not_ be retried.

CollinsMunene:
Okay, so here is the implementation attached,
The payment is only retried if the payment state is "SENDING", so by throwing the PaymentFailed event, it changes the payment state to FAILED as well as sends the outgoing_payment.failed webhook.

Kindly confirm if this is okay.
The implementation:
<img width="752" alt="implementation" src="https://github.com/user-attachments/assets/64d207d2-20f7-49d2-901f-e3f74cda2a7f">

The Webhook:
<img width="339" alt="payment_state" src="https://github.com/user-attachments/assets/aa9b370a-41ef-4701-aa3d-e44101597a9c">

The outgoing payment body after:
<img width="576" alt="webhook" src="https://github.com/user-attachments/assets/63e933b3-e007-4bda-8da0-7aa210652660">

CollinsMunene:
Hi @mkurapov, I also noticed on the webhooks, can we also add an error description. For example, when the outgoing_payment.failed is raised when the quote expiry time has reached. The webhook does not really say why it has failed. It would be nice to see a reason like "quote has expired"

or even in any other instance of failure, makes it easy to debug.

mkurapov:
Hi @CollinsMunene, 

> The webhook does not really say why it has failed. It would be nice to see a reason like "quote has expired"

[...]

PR #3173, opened 2024-12-11
feat(backend): Check quote expiry in outgoing payment worker (#3141)

CollinsMunene:
Check quote expiry in outgoing payment worker (#3141)

## Context
Currently, we check the expiresAt of the quote during outgoing payment creation. If it is valid, we create the outgoing payment, store it in the DB.

Once the payment is funded, the outgoing payments worker starts processing the payment, whether that is local or over ILP. If the payment fails, it will stay in the SENDING state and the worker will keep retrying it (up to some maximum number of times).

In the worker, however, we do not check the expiry of the quote. The quote could expire between the time when we created the payment, and when it is being processed (or retried) in the worker. We should prevent processing an outgoing payment for an expired quote. (good spot would be in lifecycle.ts > handleSending)

Also, so as to enable better error handling and debugging, we should add a new lifecycle error for showing quote has expired.

## Checklist

- [ ] Related tasks linked at issue #3141
- [ ] Added check if payment quote has expired under lifecycle.handleSending
- [ ] Added QuoteExpired under LifecycleError enum


packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -17,6 +17,11 @@ export async function handleSending(
 ): Promise<void> {
   if (!payment.quote) throw LifecycleError.MissingQuote
 
+  // Check if the current time is greater than or equal to when the Quote should be expiring
```
mkurapov, line 20: Let's add a test for this case

mkurapov (commented):
Thank you for the PR @CollinsMunene !

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
CollinsMunene: Hi Max, I have been looking to add the test here at (service.test.ts) potentially something similar to ('processNext' test) or ('fund').
Is this correct?

CollinsMunene: Pull Request made

mkurapov (commented):
Looks good @CollinsMunene ! I think just need to run `pnpm format` in the root of the repo

mkurapov (approved):
Thank you!

mkurapov:
Merging @CollinsMunene!

PR #3213, opened 2025-01-14
feat(fronted): tenanted admin api credentials

### LINKED ISSUE #3108: [Multi-Tenant] Tenanted Admin UI Requests
The Admin UI should use the new identity management scheme for tenanted requests on the admin graphql api. The UI could present an input to collect the api secret that signatures should be signed with and add them to the browser session, if the UI does not detect that they have been set already.

- [x] Input to collect tenant id & api secret for signature generation if unset
- [x] Sign all requests from the UI to Admin API using api secret
- [x] Augment all requests from UI to Admin API with tenantId header

### Prerequisites:
* https://github.com/interledger/rafiki/issues/2928

BlairCurrey:

- updates frontend to use api credentials (tenant id, api secret) to make request to admin api and handles 401 errors on bad/missing credentials

## Context

fixes #3108 

## Checklist

- [x] Tests added/updated


localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -84,7 +85,7 @@ services:
       - shared-database
       - shared-redis
     healthcheck:
-      test: ["CMD", "wget", "--header=apollo-require-preflight: true", "http://localhost:3001/graphql?query=%7B__typename%7D", "-O", "/dev/null"]
+      test: ["CMD", "wget", "--spider", "http://localhost:3001/healthz"]
```
BlairCurrey, line 88: duplicated change already made on main. Alternatively, could merge main into the base branch and I can merge the base branch in here.

BlairCurrey: probably _should_ merge the main branch in but wasn't sure if someone else was normally handling that

packages/backend/src/app.ts
```diff
@@ -386,7 +386,9 @@ export class App {
         },
         next: Koa.Next
       ): Promise<void> => {
-        if (ctx.path !== '/graphql') {
+        if (ctx.path === '/healthz') {
+          ctx.status = 200
+        } else if (ctx.path !== '/graphql') {
```
BlairCurrey, line 391: see https://github.com/interledger/rafiki/pull/3213/files#r1916805016

packages/frontend/app/components/Sidebar.tsx
```diff
@@ -91,7 +96,10 @@ export const Sidebar: FC<SidebarProps> = ({ logoutUrl, authEnabled }) => {
                               isActive
                                 ? 'bg-mercury'
                                 : 'text-tealish/70 hover:bg-mercury/70',
-                              'flex p-2 font-medium rounded-md'
+                              'flex p-2 font-medium rounded-md',
+                              !hasApiCredentials &&
+                                name !== 'Home' &&
+                                'text-gray-400 pointer-events-none cursor-not-allowed'
```
BlairCurrey, line 102: disables nav links (Assets, Wallet Addresses, etc.) in nav menu when api credentials arent set. This keeps user on page w/ the form to submit the credentials and avoids an inevitable error (which you can see if manually forming the url).

packages/frontend/app/lib/apollo.server.ts
```diff
@@ -26,50 +23,66 @@ BigInt.prototype.toJSON = function (this: bigint) {
...
+  })
+}
 
 const httpLink = createHttpLink({
   uri: process.env.GRAPHQL_URL
 })
 
-if (!global.__apolloClient) {
-  global.__apolloClient = new ApolloClient({
+export async function getApolloClient(request: Request) {
```
BlairCurrey, line 73: Before this change, we re-used 1 server side apollo client for all requests.

This is no longer tenable because we need to support the concept of different requestors. Hence, a `getApolloClient` that is formed for each request with the correct `apiSecret` and `tenantId`.

I did wonder if this was a weird, not performant thing to do but actually I found it is preferred to avoid side effects between clients (someone else's credentials, cache issues, etc.). Some supporting evidence for this:

- pretty direct instruction to create a new client epr request when there is user specific data (like the tenantId, apiSecret): https://github.com/apollographql/apollo-client/issues/9520#issuecomment-1067397043
- official apollo post about how to use remix. setup is a bit different but a new client is created in handleRequest (which is called for all page loads) https://www.apollographql.com/blog/how-to-use-apollo-client-with-remix

packages/frontend/app/lib/apollo.server.ts
```diff
@@ -26,50 +23,56 @@ BigInt.prototype.toJSON = function (this: bigint) {
...
     cache: new InMemoryCache({}),
-    link: ApolloLink.from([authLink, httpLink]),
-    defaultOptions: {
-      query: {
-        fetchPolicy: 'no-cache'
-      },
-      mutate: {
-        fetchPolicy: 'no-cache'
-      }
-    }
```
BlairCurrey, line 70: no longer needed since clients arent shared between requests and we arent worried about accidentally sharing cache

packages/frontend/app/lib/session.server.ts
```diff
@@ -0,0 +1,17 @@
...
+const SESSION_NAME = '__session'
+
+const { getSession, commitSession, destroySession } =
+  createCookieSessionStorage({
+    cookie: {
+      name: '__session',
+      httpOnly: true,
+      path: '/',
+      sameSite: 'lax',
+      secure: process.env.NODE_ENV === 'production',
```
BlairCurrey, line 12: @mkurapov Is there anything to test here? I recall some shenanigans with the snackbar message cookie and some scenario. Mobile perhaps? Didn't notice any actual problems in testing.

packages/frontend/app/components/Sidebar.tsx
BlairCurrey: if there is more visual feedback we can give on the disabled state im all ears. With these changes it looks the same, they just arent clickable.  See the pic here showing it. 

Normally I'd make it a lighter gray but lighter gray already represents non-active tabs. Perhaps make the text black on all of them (active or not) but keep the dark background to represent active. 

Strikethrough, italics, etc. all seemd kinda meh too and a different color doesnt really seem to communicate disabled... could completely hide them but felt like showing what content there is still made more sense.

image

packages/frontend/app/lib/api/asset.server.ts
```diff
@@ -27,9 +27,10 @@ import type {
   WithdrawAssetLiquidity,
   WithdrawAssetLiquidityVariables
 } from '~/generated/graphql'
-import { apolloClient } from '../apollo.server'
+import { getApolloClient } from '../apollo.server'
 
-export const getAssetInfo = async (args: QueryAssetArgs) => {
+export const getAssetInfo = async (request: Request, args: QueryAssetArgs) => {
+  const apolloClient = await getApolloClient(request)
```
BlairCurrey, line 33: example of changs to apollo client to support requestor specific api signature and tenant id header as described here: https://github.com/interledger/rafiki/pull/3213/files#r1916826388

packages/frontend/app/root.tsx
```diff
@@ -184,6 +208,15 @@ export function ErrorBoundary() {
...
     errorMessage = error.message
+
+    if (
+      isApolloError(error) &&
+      error.networkError &&
+      'statusCode' in error.networkError &&
+      error.networkError.statusCode === 401
+    ) {
+      errorMessage = 'Unauthorized. Please set your API credentials.'
+    }
```
BlairCurrey, line 219: The behavior on access a route with missing or bad api credentials is handled here. Clicking the navigation is disabled but if you manually go to /assets for example you'll see our regular error screen with this unauthorized message and prompt to visit the homepage, where the api credentials form lives.

image

BlairCurrey: Parsing this is kinda of a PITA but I think this is the correct way. It's an apollo error but we are actually throwing the error in koa so it ends up as a network error (which could be one of many types, hence `'statusCode' in error.networkError`).

AFAIK this should be reliable but the fact that `networkError` could be one of many types does make you wonder. I also considered simply doing 
 `if(error.message.includes('401'))`. This works but I figured that's likely flakier.

localenv/mock-account-servicing-entity/app/entry.server.tsx
```diff
@@ -39,6 +39,11 @@ if (!global.__seeded) {
   })
     .then(() => {
       global.__seeded = true
+      setTimeout(() => {
+        console.log('Server started with the following environment variables:')
+        console.log(`OPERATOR_TENANT_ID: ${process.env.OPERATOR_TENANT_ID}`)
+        console.log(`SIGNATURE_SECRET: ${process.env.SIGNATURE_SECRET}`)
+      }, 2000)
```
BlairCurrey, line 46: You have to submit these in the admin ui to use it now. so im logging them here to streamline that (dont have to hunt them down in the docker-compose). Could not for the life of me figure out how to get them to log after the seed script without the timeout.

packages/frontend/app/lib/apollo.server.ts
```diff
@@ -26,50 +23,52 @@ BigInt.prototype.toJSON = function (this: bigint) {
...
 
-const authLink = setContext((request, { headers }) => {
-  if (!process.env.SIGNATURE_SECRET || !process.env.SIGNATURE_VERSION)
-    return { headers }
-  const timestamp = Math.round(new Date().getTime() / 1000)
-  const version = process.env.SIGNATURE_VERSION
+async function createAuthLink(request: Request) {
+  return setContext(async (gqlRequest, { headers }) => {
+    const timestamp = Math.round(new Date().getTime() / 1000)
+    const version = 1
```
BlairCurrey, line 29: Wasn't entirely sure what to do about this version. I think we either have to hardcode it like this or make it a configurable at the tenant level. This used to be hardcoded as an env var along with the signature but of course the signature is now set per tenant. 

Im not entirely clear on what the version does and when you'd update it. But I imagine it would be updated alongside a new secret and so it wouldnt make sense to configure this via environment variable while the signature is set per tenant. @njlie do you think hardcoding the version like this makes sense or is there a better way?

packages/frontend/app/routes/_index.tsx
```diff
@@ -23,6 +93,16 @@ export default function Index() {
             In this web application, you&apos;ll be able to manage peering
             relationships, assets, and wallet addresses, among other settings.
           </p>
+
+          <div className='space-y-4'>
+            <p className='text-gray-600'>
+              To get started, please configure your API credentials
+            </p>
+            <div className='max-w-md mx-auto'>
+              <ApiCredentialsForm hasCredentials={hasCredentials} />
```
BlairCurrey, line 102: New form to submit credentials. 

I included it on the homepage because thats naturally where you land and you cant really see anything else without submitting these.

Form validates uuid client side. Saves to cookie based session on submit which you can also delete. Here are the views for credentials set in storage and not set.

image

image

packages/frontend/app/root.tsx
BlairCurrey: actually this isnt right.... not correctly detecting it and showing the message I intended. reverting back to checking the error message.

packages/frontend/app/root.tsx
```diff
@@ -183,7 +206,9 @@ export function ErrorBoundary() {
 
   let errorMessage = 'Unknown error'
   if (error instanceof Error) {
-    errorMessage = error.message
+    errorMessage = error.message.includes('401')
+      ? 'Unauthorized. Please set your API credentials.'
+      : error.message
```
BlairCurrey, line 211: On 401 (missing or bad credentials), shows our error page with this message and prompt to return to homepage where credentials can be set. 
image

packages/frontend/app/components/Sidebar.tsx
oana-lolea: To me the api credentials form is similar to a login form, so I believe hiding the contents from the sidebar is clearer for the user that he cannot navigate anywhere until they add their credentials.

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -166,9 +167,7 @@ services:
       GRAPHQL_URL: http://cloud-nine-wallet-backend:3001/graphql
       OPEN_PAYMENTS_URL: https://cloud-nine-wallet-backend/
       ENABLE_INSECURE_MESSAGE_COOKIE: true
-      AUTH_ENABLED: false      
-      SIGNATURE_VERSION: 1
-      SIGNATURE_SECRET: iyIgCprjb9uL8wFckR+pLEkJWMB7FJhgkvqhTQR/964=
```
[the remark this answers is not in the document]
BlairCurrey: Is that appropriate logic to have in the production image?

I was worried about having that functionality when using multiple tenants. For a single tenant it's good - it just works the way it always has. But if you have multiple tenants and set this then everyone is that tenant. That made it seem potentially risky to have so removing it was defensive. But thinking from the other side, it does streamline things not only in local development but for the single-tenant use-case which may be the most common one.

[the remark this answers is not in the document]
BlairCurrey: Thinking about it another way, these env vars are required. So all existing deployments have them set.

If an existing deployment adds a new tenant, now how do we know which signature to use? How do we know not to use this one? I think the signature is just no longer a decision we can make at deployment.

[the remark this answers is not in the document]
BlairCurrey: Im not against streamlining the localenv more. Currently it just logs the credentials and you copy-paste them. I did think a bit about alternatives.

I think its probably possible to simply trigger the action from the MASE with the env vars we are using there. THe `action` is a function which runs on the remix server at `POST /`. So I think in theory we could expose the port in docker and add a `fetch` call in the MASE to set them on startup. It does seem like a more appropriate thing to do from our development fixture than the frontend app itself.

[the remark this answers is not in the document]
mkurapov: > So all existing deployments have them set.

The SIGNATURE_SECRET and SIGNATURE_VERSION are optional in `frontend`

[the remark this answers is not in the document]
mkurapov: > So I think in theory we could expose the port in docker and add a fetch call in the MASE to set them on startup. It does seem like a more appropriate thing to do from our development fixture than the frontend app itself.

I think that makes sense in terms of the developer environment. Doesn't have to be in this PR.

packages/frontend/app/components/Sidebar.tsx
BlairCurrey: Updated, thanks for the feedback.

localenv/cloud-nine-wallet/docker-compose.yml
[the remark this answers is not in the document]
BlairCurrey: > The SIGNATURE_SECRET and SIGNATURE_VERSION are optional in frontend I believe

I think it would still be an issue for anyone going from single tenant to multi-tenant. This happens over the course of using the application. IDK if we can securely determine if we should use these or not - seems dangerous to fallback to making every request to this single tenant (likely the operator).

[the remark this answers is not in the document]
BlairCurrey: I'm thinking more about allowing them to set SIGNATURE_SECRET and then using that if we can, otherwise requiring one from the session. I think the question is how do we know if the session one is required or if we can use the env var? I think maybe it's just if there is only 1 tenant in the backend or not?

I think we'd need an extra api call before making the gql request, and it may be difficult to use the gql api for it since we are assuming all requests will have the tenant details on it and we wont have those, because those details are what we're trying to figure out.

```ts
export const loader = async ({ request }: LoaderFunctionArgs) => {
   // ...
   
  const shouldUseSessionSecret = await checkApiIfMultitenanted() // boolean

  const assets = await listAssets(request, shouldUseSessionSecret, {
    ...pagination.data
  })
  
  // ... 
}
```

I guess this saves some intermittent input from operator (ie when there is no cookie with session) but how important is that? Dont really like this and I dont see another way.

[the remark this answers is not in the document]
BlairCurrey: Ah shoot... setting from the MASE doesn't work because its a cookie... the browser wont have it.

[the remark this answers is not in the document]
BlairCurrey: I changed the MASE to log a link which will prefill the form. So you can click the link and press save. This is something integrators could leverage as well when distributing credentials. They can just send such a link.

In theory I think we should be able to automatically submit it in a `useEffect` hook but I wasn't able to get that working.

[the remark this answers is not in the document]
BlairCurrey: nvm, got the auto submit working. MASE logs a link like http://localhost:3010/?tenantId=438fa74a-fa7d-4317-9ced-dde32ece1787&apiSecret=iyIgCprjb9uL8wFckR%2BpLEkJWMB7FJhgkvqhTQR%2F964%3D which you can use to set without any additional input.

mkurapov:
https://github.com/user-attachments/assets/fd052946-4e99-47e1-be85-7b6e2f44777b

Getting odd behaviour where 401 on the request is happening when navigating to the same page in succession

BlairCurrey:
> Screen.Recording.2025-01-20.at.13.43.51.mov 
> Getting odd behaviour where 401 on the request is happening when navigating to the same page in succession

Im seeing this on `main` too. Would you mind verifying you see the same? Looks like its an existing bug to me.

mkurapov:
@BlairCurrey yep, seeing on main as well. Looking into it

mkurapov:
@BlairCurrey Ah, I think I got it. If you do it within one second of clicking, it looks like the signature validation fails because of the `canApiSignatureBeProcessed` replay attack protection in `backend`

mkurapov (approved):

packages/frontend/app/lib/apollo.server.ts
BlairCurrey: decided to keep the env var after talking w/ Nathan. The backend has this env var and frontend should match it, so it needs to stay configurable.

localenv/mock-account-servicing-entity/app/entry.server.tsx
```diff
@@ -30,6 +30,15 @@ async function callWithRetry(fn: () => any, depth = 0): Promise<void> {
 }
 
 if (!global.__seeded) {
+  const tenantId = process.env.OPERATOR_TENANT_ID
+  const apiSecret = process.env.SIGNATURE_SECRET
```
njlie, line 34: It doesn't seem like `SIGNATURE_SECRET` is set in the environment anymore.

BlairCurrey: It was removed from the frontend - is that what you're referring to? The docker compose sets them for the mock ase. When I spin up rafiki it doesnt error on the lines below for the secret, and it spits out a link with the `apiSecret`.

njlie: Ah, got it. My mistake.

njlie (approved):

PR #3224, opened 2025-01-20
build(deps): docs dependency bumps

JoblersTune:

- Routine dependency updates and small changes for the updated Astro 5 bump.

## Checklist

- [ ] Tests added/updated


JoblersTune:
This upgrade is causing issues because the astro-graphql-plugin dependency requires an update first.

So this is pending until that happens.

PR #3236, opened 2025-01-22
chore: make signature verification & generation use ms for timestamp

### LINKED ISSUE #3228: Make signature use milliseconds for timestamp
### Context

When verifying the signature in the Admin API requests, we also do a check for replay attacks by storing the signature in redis, and checking if a subsequent requests has the same signature.
Given we use unix timestamp (in seconds) as part of the signature, this means that if we have the same request (with the same payload), happen within a second, we will fail the signature verification.

Normally this is ok, but in some cases this may lead to weird behaviour, particularly in the admin API if we navigate to the same page within a second:
https://github.com/interledger/rafiki/pull/3213#issuecomment-2602337500

To fix this, for signature generation, we can make the timestamp be milliseconds (instead of seconds) in the signature verification and generation. 

### Todo
- [ ] Make signature verification and generation use a timestamp with milliseconds instead of seconds. Make the change in all of the spots: backend, auth, frontend, and mock ASE

mkurapov:

- For signature generation and verification, use timestamp to the nearest millisecond instead of seconds to prevent issues for sending the same request (like we would often do in the admin UI, see links in issue)
- Start signing apollo client requests in integration tests

## Context

Fixes #3228 

## Checklist

- [x] Tests added/updated


packages/backend/src/shared/utils.test.ts
```diff
@@ -251,6 +248,13 @@ describe('utils', (): void => {
       )
       ctx.request.body = requestBody
 
+      await expect(
+        verifyApiSignature(ctx, {
+          ...Config,
+          adminApiSecret: 'test-secret'
+        })
+      ).resolves.toBe(true)
```
mkurapov, line 256: make the test agnostic to the actual implementation of cache (just call the signature verification twice directly)

BlairCurrey (commented):
timestamp in integration test's apollo client needs to be updated too?

packages/backend/src/tests/quote.ts
```diff
@@ -165,11 +165,7 @@ export async function createQuote(
     maxPacketAmount: BigInt('9223372036854775807')
   })
 
-  const withGraphFetchedArray = [
-    'asset',
-    'walletAddress',
-    'walletAddress.asset'
-  ]
```
mkurapov, line 172: was causing a console.warn

Screenshot 2025-01-23 at 13 03 49

mkurapov:
@BlairCurrey 

It wasn't signing previously, so I added it

.trivyignore
```diff
@@ -1 +1 @@
-CVE-2024-21538 exp:2024-12-31
\ No newline at end of file
+CVE-2024-21538 exp:2025-03-01
```
mkurapov, line 1: to make trivy scanner pass again, this will be resolved by the earlier of the underlying issue to be merged in node, or upgrading to node 22

.trivyignore
```diff
@@ -1 +1 @@
-CVE-2024-21538 exp:2024-12-31
\ No newline at end of file
+CVE-2024-21538
```
mkurapov, line 1: to make trivy scanner pass again, this will be resolved when the underlying issue gets fixed in node 20 (or we end up updating to node 22)

BlairCurrey (approved):

PR #3248, opened 2025-01-28
feat(backend): check grant receiver to match quote receiver

### LINKED ISSUE #3112: Enforce `receiver` field in outgoing payment grant creation
## Context
Currently, the Open Payments AS spec allows to specify a `receiver` field as part of the outgoing payment limits for grant creation, which specifies for which incoming payment outgoing payments can be created under. 
Even though this field is spec'ed out in Open Payments, we do not enforce it in Rafiki (`receiver` field is ignored).

## Todos
- [ ] If an outgoing payment is trying to be created under a grant that has a different `receiver` (incoming payment) than the one the outgoing payment is paying into, we should fail with a 403 (forbidden) error

cozminu:
Enforce receiver field in outgoing payment grant creation

## Context
On before creation of an outgoing payment, check if grant.limit.receiver exists and if it has the same value as the quote.receiver
Provide further details about how the feature should be tested/reviewed if necessary.
- fixes #3112 

## Checklist
- [ ] Tests added/updated


packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -265,10 +268,24 @@ async function createOutgoingPayment(
 
     if (isQuoteError(quoteOrError)) {
       return quoteErrorToOutgoingPaymentError[quoteOrError]
+    } else {
+      quote = quoteOrError
     }
     quoteId = quoteOrError.id
```
golobitch, line 274:
```suggestion
    if (isQuoteError(quoteOrError)) {
      return quoteErrorToOutgoingPaymentError[quoteOrError]
    }
    
    quote = quoteOrError
    quoteId = quote.id
```

golobitch: just for the readability

localenv/README.md
```diff
@@ -289,6 +289,6 @@ Keep-Alive: timeout=5
 
 #### TigerBeetle container exits with code 137
 
-This is a known [issue](https://docs.tigerbeetle.com/getting-started/with-docker-compose/#exited-with-code-137) when running TigerBeetle in Docker: the container exits without logs and simply shows error code 137. To fix this, increase the Docker memory limit.
+This is a known [issue](https://docs.tigerbeetle.com/operating/docker#exited-with-code-137) when running TigerBeetle in Docker: the container exits without logs and simply shows error code 137. To fix this, increase the Docker memory limit.
```
mkurapov, line 292: Thanks for updating this!

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -265,10 +268,24 @@ async function createOutgoingPayment(
...
   } else {
     quoteId = options.quoteId
+    if (doExtraChecks) {
+      quote = await deps.quoteService.get({ id: quoteId })
```
mkurapov, line 278: Given that we already fetch the quote from the database for both scenarios on L342 (`.withGraphFetched('quote')`), we can do the check after we've created the payment (within the DB transaction).

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -265,10 +268,24 @@ async function createOutgoingPayment(
...
+    }
+  }
+
+  if (doExtraChecks) {
+    if (!quote) {
+      return OutgoingPaymentError.UnknownQuote
+    }
+    if (quote.receiver !== options.grant?.limits?.receiver) {
+      return OutgoingPaymentError.InvalidReceiver
```
mkurapov, line 287: In the context of Open Payments, I think we should return a 403 error here (`InsufficientGrant`), since the request itself (and the receiver) is valid, but the client making this outgoing payment request does not have the correct access.

(I'll update the issue to reflect this)

packages/backend/src/open_payments/payment/outgoing/service.ts
cozminu: good point

localenv/README.md
cozminu: no problem

packages/backend/src/open_payments/payment/outgoing/service.ts
cozminu: got it

packages/backend/src/open_payments/payment/outgoing/service.ts
cozminu: Looks like we already check this. I wrote 2 tests instead.

cozminu: Screenshot 2025-01-31 at 17 25 19
receiver is already checked here

mkurapov (approved):

PR #3249, opened 2025-01-28
docs: Updated TRUST_PROXY as a conditionally required variable

hajjimo:
- [ ] Moved TRUST_PROXY variable from optional to conditionally required with an additional note that if Rafiki runs behind a firewall, then the default value must be updated to `true`
- [ ] Added note in docker-compose guide to ensure the TRUST_PROXY variable is set to true


packages/documentation/src/partials/backend-variables.mdx
```diff
@@ -27,6 +27,7 @@ import { LinkOut } from '@interledger/docs-design-system'
 | Variable        | Helm value name         | Default     | Description                                                                                                                                                               |
 | --------------- | ----------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `INSTANCE_NAME` | `backend.instance.name` | _undefined_ | Your Rafiki instance's name used to communicate for auto-peering and/or [telemetry](/overview/concepts/telemetry). Required when auto-peering and/or telemetry is enabled |
+| `TRUST_PROXY`   | `backend.trustProxy`    | `false`     | Must be set to `true` when running Rafiki behind a proxy. When `true`, the `X-Forwarded-Proto` header is used to determine if connections are secure.                     |
```
mkurapov, line 30: Also need it for the auth variables

mkurapov: Should we add the `wider-column` for this table as well?

mkurapov (approved):

PR #3171, opened 2024-12-10
feat(backend): tenanted quotes and outgoing payments

sanducb:
## Changes proposed in this pull request

-  The purpose of this PR is to make quotes and outgoing payments tenanted as per multi-tenant design

## Context

fixes #3118 #3119

## Checklist

- [x] Related issues linked using `fixes #number`
- [x] Tests added/updated
- [x] Make sure that all checks pass
- [x] Bruno collection updated (if necessary)
- [x] Documentation issue created with `user-docs` label (if necessary)
- [x] OpenAPI specs updated (if necessary)

PR #3271, opened 2025-02-05
feat(backend): tenanted incoming payments

### LINKED ISSUE #3117: [Multi-Tenant] Tenanted Incoming Payments
Incoming Payments should become tenanted as per the Multi-Tenancy design, and will be logically separated in the `backend` database by a `tenantId` row that serves as a foreign key on the tenants table.

- [ ] Add `tenantId` row to `incomingPayments` table that is required and maps to an id on the `tenants` table.
- [ ] Add `tenantId` to database queries on incoming payments
- [ ] Acquire `tenantId` from request headers on routes involving incoming payments and pass them into incoming payment service logic.
- [ ] Include `tenantId` in path of incoming payment URL. e.g. `http://exampe-rafiki/:tenantId/incoming-payments/:id`
  - See https://github.com/interledger/rafiki/pull/2986 for existing implementation on old feature branch.
- [ ] Additionally, we should be able to support fetching an incoming payment with and _without_ tenant id:
    - for GraphQL Admin API requests to get (and update) an incoming payment, `tenantId` will be provided in those requests, and it _should_ be used when getting (or updating) the incoming payments.
    - However, because we are able to make payments across tenanted resources (ie. an outgoing payment on tenant A can pay into an incoming payment on tenant B), we should also have the functionality to to look up the incoming payment just by the Id, tenant-agnostic

### Prerequisites
* https://github.com/interledger/rafiki/issues/2928
* https://github.com/interledger/rafiki/issues/3114

oana-lolea:

- Added optional tenantId for get, mandatory for post and delete
- Updated unit and integration tests
- Updated bruno scripts

## Context

Fixes #3117

## Checklist

- [x] Tests added/updated


packages/backend/src/graphql/schema.graphql
```diff
@@ -1247,6 +1255,8 @@ input UpdateIncomingPaymentInput {
   id: ID!
   "The new metadata object to save for the incoming payment. It will overwrite any existing metadata."
   metadata: JSONObject!
+  "The tenant UUID associated with the incoming payment. If not provided, it will be obtained from the signature."
+  tenantId: String
```
mkurapov, line 1259: Given that this is for an existing resource, we don't need to have this on the Input. We will use the signature tenant id to check if they can update the payment.

packages/backend/src/graphql/schema.graphql
```diff
@@ -372,11 +372,15 @@ type AssetEdge {
...
   id: ID!
+  "The tenant UUID associated with the incoming payment."
+  tenantId: String!
 }
 
 input CancelIncomingPaymentInput {
   "Unique identifier of the incoming payment to be canceled. Note: incoming payment must be PENDING."
   id: ID!
+  "The tenant UUID associated with the incoming payment."
+  tenantId: String!
```
mkurapov, line 383: Same as comment below, we don't need to specify tenantId in the mutations when doing operations for an existing resource so we can remove it from `ApproveIncomingPaymentInput` and `CancelIncomingPaymentInput`.

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -97,13 +99,14 @@ async function createLocalIncomingPayment(
...
-  const { expiresAt, incomingAmount, metadata } = args
+  const { expiresAt, incomingAmount, metadata, tenantId } = args
 
   const incomingPaymentOrError = await deps.incomingPaymentService.create({
     walletAddressId: walletAddress.id,
     expiresAt,
     incomingAmount,
-    metadata
+    metadata,
+    tenantId: tenantId ?? Config.operatorTenantId
```
mkurapov, line 109: Instead of using `Config.operatorTenantId` we should just throw if `tenantId` is undefined, since its the responsibility of the caller of the function (in this case, the Admin API resolver) to provide tenantId if its a local payment.

This means we should add `tenantId` to the `createReceiver` mutation

packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -11,19 +11,20 @@ import {
...
-export const getIncomingPayment: QueryResolvers<ApolloContext>['incomingPayment'] =
+export const getIncomingPayment: QueryResolvers<TenantedApolloContext>['incomingPayment'] =
   async (parent, args, ctx): Promise<ResolversTypes['IncomingPayment']> => {
     const incomingPaymentService = await ctx.container.use(
       'incomingPaymentService'
     )
     const payment = await incomingPaymentService.get({
-      id: args.id
+      id: args.id,
+      tenantId: ctx.tenant.id
```
mkurapov, line 27:
```suggestion
      tenantId: ctx.isOperator ? undefined : ctx.tenant.id
```
since an operator can request incoming payments for other tenants

packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -79,7 +82,7 @@ export const getWalletAddressIncomingPayments: WalletAddressResolvers<ApolloCont
       })
     }
   }
-export const createIncomingPayment: MutationResolvers<ApolloContext>['createIncomingPayment'] =
+export const createIncomingPayment: MutationResolvers<TenantedApolloContext>['createIncomingPayment'] =
```
mkurapov, line 85:
```suggestion
export const createIncomingPayment: MutationResolvers<ForTenantIdContext>['createIncomingPayment'] =
```

packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -94,7 +97,8 @@ export const createIncomingPayment: MutationResolvers<ApolloContext>['createInco
         ? undefined
         : new Date(args.input.expiresAt),
       incomingAmount: args.input.incomingAmount,
-      metadata: args.input.metadata
+      metadata: args.input.metadata,
+      tenantId: ctx.tenant.id
```
mkurapov, line 101:
```suggestion
      tenantId: ctx.forTenantId
```
This allows us the operator to create incoming payments for a specific tenant they've specified in the `tenantId` mutation input

packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -56,14 +57,16 @@ export const getWalletAddressIncomingPayments: WalletAddressResolvers<ApolloCont
     const incomingPayments = await incomingPaymentService.getWalletAddressPage({
       walletAddressId: parent.id,
       pagination,
-      sortOrder: order
+      sortOrder: order,
+      tenantId: ctx.tenant.id
```
mkurapov, line 61:
```suggestion
      tenantId: ctx.isOperator ? undefined : ctx.tenant.id
```
so the operator can view incoming payments for other tenants

test/integration/lib/utils.ts
```diff
@@ -87,3 +87,25 @@ export function urlWithoutTenantId(url: string): string {
...
   return validate(id) && version(id) === 4
 }
+
+/**
+ * A helper function to add tenantId to incoming payment id.
+ * @param tenantId tenant that must be added
+ * @param incomingPaymentId the id of the incoming payment
+ * @returns the updated incoming payment
+ */
+export function addTenantToIncomingPaymentId(
```
oana-lolea, line 97: Added this function as a workaround for `createReceiver` method, because its id does not contain the tenant.

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -129,7 +140,8 @@ async function createIncomingPayment(
     client,
     expiresAt,
     incomingAmount,
-    metadata
+    metadata,
+    tenantId
```
mkurapov, line 144: We should add the `tenantId` in `walletAddressService.get` so we make sure its the same tenant for the wallet address and the incoming payment

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -75,9 +86,9 @@ export async function createIncomingPaymentService(
   return {
     get: (options) => getIncomingPayment(deps, options),
     create: (options, trx) => createIncomingPayment(deps, options, trx),
-    approve: (id) => approveIncomingPayment(deps, id),
-    cancel: (id) => cancelIncomingPayment(deps, id),
-    complete: (id) => completeIncomingPayment(deps, id),
+    approve: (id, tenantId) => approveIncomingPayment(deps, id, tenantId),
+    cancel: (id, tenantId) => cancelIncomingPayment(deps, id, tenantId),
+    complete: (id, tenantId) => completeIncomingPayment(deps, id, tenantId),
     getWalletAddressPage: (options) => getWalletAddressPage(deps, options),
```
mkurapov, line 92: We also need to update the `getWalletAddressPage` method so it adds tenantId to the query if provided:

```
  if (tenantId) {
    query.where({ tenantId })
  }
```


In the incoming payment routes, we should also make sure we are passing the tenantId from `ctx.params` into `getWalletAddressPage`

packages/backend/src/open_payments/receiver/service.ts
```diff
@@ -97,13 +98,20 @@ async function createLocalIncomingPayment(
   args: CreateReceiverArgs,
   walletAddress: WalletAddress
 ): Promise<OpenPaymentsIncomingPaymentWithPaymentMethods | ReceiverError> {
-  const { expiresAt, incomingAmount, metadata } = args
+  const { expiresAt, incomingAmount, metadata, tenantId } = args
+
+  if (!tenantId) {
```
mkurapov, line 103: since `tenantId` is required we can remove this check right?

packages/backend/src/graphql/resolvers/receiver.ts
```diff
@@ -32,17 +32,23 @@ export const getReceiver: QueryResolvers<ApolloContext>['receiver'] = async (
...
 
-export const createReceiver: MutationResolvers<ApolloContext>['createReceiver'] =
+export const createReceiver: MutationResolvers<ForTenantIdContext>['createReceiver'] =
   async (_, args, ctx): Promise<ResolversTypes['CreateReceiverResponse']> => {
     const receiverService = await ctx.container.use('receiverService')
 
+    const tenantId = ctx.forTenantId
+    if (!tenantId) {
+      throw new Error('Tenant id is required to create a receiver')
+    }
```
mkurapov, line 42: we can just use `ctx.tenant.id` here (since the mutation doesn't take in `tenantId`)

test/integration/lib/utils.ts
mkurapov: Wouldn't we have tenantId in all scenarios? (Since you've updated the incoming payments model to always return `tenantId` in the `url`)

bruno/collections/Rafiki/Examples/Open Payments Without Quote/Create Incoming Payment.bru
```diff
@@ -5,7 +5,7 @@ meta {
 }
 
 post {
-  url: {{receiverOpenPaymentsHost}}/incoming-payments
+  url: {{receiverOpenPaymentsHost}}/{{receiverTenantId}}/incoming-payments
```
njlie, line 8: You can remove the explicit addition of the `receiverTenantId` (or `senderTenantId`) in other requests now that #3187 & #3152 are merged into the feature branch.

I made a [similar comment](https://github.com/interledger/rafiki/pull/3171#discussion_r1943360155) on #3171 for reference.

test/integration/lib/utils.ts
oana-lolea: The incoming payment doesn't include a `tenantId` because the wallet address we get by URL also does not have the `tenantId`

packages/backend/src/tests/outgoingPayment.ts
```diff
@@ -42,7 +43,9 @@ export async function createOutgoingPayment(
     const streamCredentials = streamServer.generateCredentials()
 
     const incomingPayment = await createIncomingPayment(deps, {
-      walletAddressId: options.walletAddressId
+      walletAddressId: options.walletAddressId,
+      // TODO: replace
```
njlie, line 47: Does this still need to be replaced?

njlie (approved):
LGTM aside from Max's remaining comments

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -70,7 +70,8 @@ async function getIncomingPaymentPublic(
 ) {
```
mkurapov, line 70: In the `listIncomingPayments` function, we need to make sure to pass in `ctx.params.tenantId` in the `deps.incomingPaymentService.getWalletAddressPage`.

(something like [this](https://github.com/interledger/rafiki/pull/3171/commits/799026a362166aba563031e9862bb2caa41c9940#diff-789c9d4089f6ed3faac07bb57227080d576bb0d9f24d8eb9dafac9498a5efbf5R153-R159))

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -399,10 +420,13 @@ async function getWalletAddressPage(
...
 async function approveIncomingPayment(
   deps: ServiceDependencies,
-  id: string
+  id: string,
+  tenantId: string
 ): Promise<IncomingPayment | IncomingPaymentError> {
   return deps.knex.transaction(async (trx) => {
-    const payment = await IncomingPayment.query(trx).findById(id).forUpdate()
+    const payment = await IncomingPayment.query(trx)
+      .get({ id, tenantId })
```
sanducb, line 428: Just out of curiosity, does this have the same behavior as `findOne`? I think that `get` will require having `tenantId` in `GetOptions` if I'm not mistaken.

packages/backend/src/open_payments/wallet_address/model.ts
```diff
@@ -191,13 +191,15 @@ export interface GetOptions {
   id: string
   client?: string
   walletAddressId?: string
+  tenantId?: string
```
sanducb, line 194: I think this is only needed for using `get` here, so we could get rid of it by using some other function instead.

sanducb (commented):
I only left a couple of comments. LGTM as well aside from Max's comments.

packages/backend/src/open_payments/payment/incoming/service.ts
oana-lolea: It doesn't actually, because `get` does not use `tenantId` in the query. I'll replace it with `findOne`

packages/backend/src/tests/outgoingPayment.ts
oana-lolea: Yeah, I'll use the `tenantId` of the wallet address instead

mkurapov (approved):

sanducb (approved):

PR #3277, opened 2025-02-06
build: containers will now run as the non-root rafiki user

bosbaber:
- Modified both the dev and prod Dockerfiles of all the services to run as the non-root Rafiki user.
- To achieve this we have to chown the packages back to rafiki after each copy to ensure the ownership is correct. Otherwise the build process create the symlinks it wants to make.

## Context
As part of me trying to get into understanding the build pipeline, I noticed that the docker containers currently run as the root user which can be considered a security issue.

- For dev builds we have to switch between Rafiki and root user during the build process to ensure everything links correctly
- For prod builds, only the run-container will now run as Rafiki user, but with read-only access to the packages folder

Just to make it clear, after the change this is the status:
```sh
$ docker exec rafiki-happy-life-backend-1 touch /etc/danger-zone
touch: /etc/danger-zone: Permission denied
$ docker exec rafiki-happy-life-backend-1 whoami
rafiki
```

## Checklist
- [x] Tests added/updated

### DIFF
--- a/packages/auth/Dockerfile.dev
+++ b/packages/auth/Dockerfile.dev
@@ -1,24 +1,37 @@
 FROM node:20-alpine3.20
 
+RUN adduser -D rafiki
 WORKDIR /home/rafiki
 
-RUN corepack enable
+# Install Corepack and pnpm as the Rafiki user
+USER rafiki
+RUN mkdir -p /home/rafiki/.local/bin
+ENV PATH="/home/rafiki/.local/bin:$PATH"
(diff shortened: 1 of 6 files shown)


golobitch:
Adding also @mkurapov as a reviewer. But to me, this looks like the correct approach. However I would say that you need to create rafiki user only in the runtime container, not in any before.

mkurapov (approved):

golobitch (approved):

PR #3281, opened 2025-02-09
feat(tenantSettings): initial implementation

golobitch:

- 

## Checklist

- [x] Tests added/updated


packages/backend/src/graphql/schema.graphql
```diff
@@ -241,6 +241,10 @@ type Mutation {
     input: CreateWalletAddressKeyInput!
   ): CreateWalletAddressKeyMutationResponse
 
+  createTenantSettings(
+    input: [CreateTenantSettingsInput!]!
```
mkurapov, line 245:
```suggestion
    input: CreateTenantSettingsInput!
```
and then just have the input have a `settings` list (like you have for the CreateTenantSettingsMutationResponse). Good for consistency and allows us to add additional fields to the input as well

packages/backend/migrations/20250117205655_create_tenant_settings_table.js
```diff
@@ -0,0 +1,24 @@
...
+    table.string('value').notNullable()
+
+    table
+      .uuid('tenantId')
+      .notNullable()
+      .references('tenants.id')
+      .index()
+      .onDelete('CASCADE')
+
+    table.timestamp('createdAt').index().defaultTo(knex.fn.now())
```
mkurapov, line 14: do we need the index still?

packages/backend/src/tenants/settings/service.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (options.key) {
+    obj.key = options.key
+  }
+  await TenantSetting.query(extra?.trx ?? deps.knex)
+    .findOne(obj)
+    .patch({
+      deletedAt: new Date()
+    })
```
mkurapov, line 97: Do we need to filter for those with deletedAt = null?

packages/backend/src/tenants/settings/service.ts
```diff
@@ -0,0 +1,139 @@
...
+
+async function createTenantSetting(
+  deps: ServiceDependencies,
+  options: CreateOptions,
+  extra?: ExtraOptions
+) {
+  const dataToInsert = options.setting.map((s) => ({
+    tenantId: options.tenantId,
+    ...s
+  }))
```
mkurapov, line 122: Do we want to check that each key of the setting is something legit? ie the key is in [EXCHANGE_RATES_URL, WEBHOOK_URL, WEBHOOK_TIMEOUT, WEBHOOK_MAX_RETRY]?

packages/backend/src/tenants/service.ts
```diff
@@ -88,8 +91,17 @@ async function createTenant(
...
+      {
+        tenantId: tenant.id,
+        setting: TenantSetting.default()
+      },
+      { trx }
+    )
+
     await trx.commit()
 
+    tenant.settings = settings
```
mkurapov, line 104: Is this to set it on the cache? If so, we would also need to change the update method. 

Or we just have a separate cache for tenant settings (or no cache at all initially)

mkurapov (commented):
Initial comments

mkurapov:
- https://github.com/interledger/rafiki/pull/3245 seems like a subset of this, do you want to get reviews/merge that one first?

packages/backend/src/tenants/settings/service.ts
sanducb: I think an enum would do justice

packages/backend/src/tenants/settings/service.ts
```diff
@@ -0,0 +1,139 @@
...
+    ...deps_,
+    logger: deps_.logger.child({ service: 'TenantSettingService ' })
+  }
+
+  return {
+    get: (options: GetOptions) => getTenantSettings(deps, options),
+    create: (options: CreateOptions, extra?: ExtraOptions) =>
+      createTenantSetting(deps, options, extra),
+    update: (options: UpdateOptions) => updateTenantSetting(deps, options),
+    delete: (options: GetOptions) => deleteTenantSetting(deps, options),
```
sanducb, line 64: `extra` seems to be missing here

packages/backend/src/tenants/settings/service.ts
golobitch: At first, I was a little hesitant to put that in enum. Because if we use enum, then we are somewhat restricted by what we can save in database...we can save only what is defined in enum, however, we can leave this as raw string and we can save whatever we want. But maybe, enum is a better option here. I will update it.

packages/backend/src/tenants/service.ts
golobitch: Ah this is some leftover from experimenting. Nice catch. Will remove this. cache can probably be separate for tenant and for settings.

packages/backend/src/tenants/settings/service.ts
golobitch: Nice one!

packages/backend/src/tenants/settings/service.test.ts
```diff
@@ -0,0 +1,277 @@
...
+        tenantId: tenantSetting[0].tenantId,
+        key: createOptions.setting[0].key
+      })
+
+      const dbTenantSetting = await TenantSetting.query().findById(
+        tenantSetting[0].id
+      )
+      expect(dbTenantSetting?.deletedAt).toBeDefined()
+      expect(dbTenantSetting?.deletedAt?.getTime()).toBeLessThanOrEqual(
+        new Date(Date.now()).getTime()
```
mkurapov, line 206:
```suggestion
        Date.now()
```

packages/backend/src/tenants/settings/service.test.ts
```diff
@@ -0,0 +1,277 @@
...
+    })
+  })
+
+  describe('get', () => {
+    let tenantSetting: TenantSetting[]
+
+    function createTenantSetting() {
+      const options: CreateOptions = {
+        tenantId: tenant.id,
+        setting: [randomSetting()]
```
mkurapov, line 103: Double-checking here, since we are checking for valid keys in the create, would this end up inserting anything?

packages/backend/src/graphql/resolvers/tenant_settings.ts
```diff
@@ -0,0 +1,122 @@
...
+    } catch (err) {
+      throw new GraphQLError('failed to delete tenant setting', {
+        extensions: {
+          code: GraphQLErrorCode.InternalServerError
+        }
+      })
+    }
+  }
+
+// export const updateTenantSetting: MutationResolvers<TenantedApolloContext>['updateTenantSetting'] =
```
mkurapov, line 94: can get rid of this one

packages/backend/src/graphql/schema.graphql
```diff
@@ -241,6 +241,14 @@ type Mutation {
     input: CreateWalletAddressKeyInput!
   ): CreateWalletAddressKeyMutationResponse
 
+  createTenantSettings(
+    input: CreateTenantSettingsInput!
+  ): CreateTenantSettingsMutationResponse
+
+  deleteTenantSettings(
```
mkurapov, line 248: this makes me think, do we want to have this mutation in general? since if a tenant gets rid of their `EXCHANGE_RATES_URL` for example, it would be bad 😅

packages/backend/src/tenants/service.ts
```diff
@@ -152,6 +163,13 @@ async function deleteTenant(
   await deps.tenantCache.delete(id)
   try {
     const deletedAt = new Date()
```
mkurapov, line 165: we can just pass in this `deletedAt` into `tenantSettingService.delete`

packages/backend/src/tenants/settings/model.test.ts
```diff
@@ -0,0 +1,12 @@
+import { TenantSetting } from './model'
+
+describe('TenantSetting Model', (): void => {
+  describe('dfeault', () => {
```
njlie, line 4: nit: typo

packages/backend/src/graphql/resolvers/tenant_settings.ts
golobitch: We wont have support for updating tenant settings?

packages/backend/src/graphql/schema.graphql
golobitch: I think that maybe we do not need it. I don't think that we have some string based config that can be optional for example. Should I remove this mutation?

packages/backend/src/graphql/resolvers/tenant_settings.ts
mkurapov: I see - you are going to have it in a separate PR?
I see the `updateTenantSetting` mutation in the backend graphql schema, so we should either uncomment this or remove the `updateTenantSetting` mutation from this PR

packages/backend/src/graphql/schema.graphql
mkurapov: Yes, I think we can remove it for now

packages/backend/src/graphql/resolvers/tenant_settings.ts
golobitch: Yeah, let's do it in separate PR so that we can unblock other tasks.

mkurapov (approved):
Going to approve, just removing the delete mutation and update mutation can go in separate PR

mkurapov (approved):

sanducb (approved):

Issue #3256, opened 2025-01-30
[Multi-Tenant] Allow viewing deleted tenants


mkurapov:
## Context
Currently, if a tenant is deleted, it is not possible to fetch it using the `tenant` query. 
However, this doesn't play nicely with the Admin UI: you are unable to navigate to the tenant details screen from the tenant list page.

If you are the operator tenant, we should allow fetching deleted tenants, and being able to view details.

## Todos
- [x] Allow the `tenant` Admin API resolver to return deleted tenants _if an operator is making the request_.
- [x] Allow clicking into tenant details in the Admin UI for deleted tenants. The fields should not be editable.

## PR:
https://github.com/interledger/rafiki/pull/3299

mkurapov:
- Blocked by #2915

PR #3289, opened 2025-02-11
feat(frontend): add operator-contextual dropdowns for tenant selection

### LINKED ISSUE #3273: [Multi-Tenant] Add tenant selection when creating resources in Admin UI
### Context

As an operator, you should be able to create assets, peers and wallet addresses on behalf of other tenants. We need to reflect this in the UI, and have the tenantId be selectable when creating those resources (via dropdowns). 

For non-operator tenants, we can hide the UI elements to select different tenants (since in the backend, the resolvers will use the tenantId from the signature/apollo context). 

### Todo
- [ ] For the create asset page, _if you are an operator_, we need to provide a dropdown that lists the tenants (with name and id). An operator can then select a particular tenant to create the asset for. This will pass in the `tenantId` in the `createAsset` mutation. The mutation input will need to be updated if it hasn't been done so.
- [ ] For the create wallet address page, _if you are an operator_, we need to update the dropdown of assets to provide information about the tenant as well. (alternatively, maybe two dropdowns, where you select the tenant first, then select the asset).
- [ ] For the create peers page, _if you are an operator_, we will need to update the asset dropdown similar to the create wallet address page, such that we can select assets across different tenants.

njlie:

- Adds tenant dropdown menus if the user is an operator for the following `frontend` pages:
  - Asset creation
  - Wallet Address creation
  - Peer creation
  - These resources will be created with the selected tenant, or with the selected asset associated with a given tenant.
- Does not display this option if the user is a tenant.

## Context

Fixes #3273.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -675,7 +675,7 @@ type Asset implements Model {
   ): FeesConnection
   "The date and time when the asset was created."
   createdAt: String!
-  tenantId: String
+  tenantId: ID!
 }
 
 enum SortOrder {
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -394,6 +394,8 @@ input CreateAssetInput {
   liquidityThreshold: UInt64
   "Unique key to ensure duplicate or retried requests are processed only once. For more information, refer to [idempotency](https://rafiki.dev/apis/graphql/admin-api-overview/#idempotency)."
   idempotencyKey: String
+  "Unique identifier of the tenant associated with the asset. This cannot be changed. Optional, if not provided, the tenantId will be obtained from the signature."
+  tenantId: ID
 }
 
 input UpdateAssetInput {
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -673,6 +673,7 @@ type Asset implements Model {
   ): FeesConnection
   "The date and time when the asset was created."
   createdAt: String!
+  tenantId: String
 }
 
 enum SortOrder {
--- a/packages/frontend/app/lib/api/asset.server.ts
+++ b/packages/frontend/app/lib/api/asset.server.ts
@@ -123,6 +123,7 @@ export const listAssets = async (request: Request, args: QueryAssetsArgs) => {
               scale
               withdrawalThreshold
               createdAt
+              tenantId
             }
           }
           pageInfo {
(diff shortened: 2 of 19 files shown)


packages/frontend/app/routes/assets.create.tsx
```diff
@@ -63,6 +80,22 @@ export default function CreateAssetPage() {
                     label='Withdrawal Threshold'
                     error={response?.errors.fieldErrors.withdrawalThreshold}
                   />
+                  {tenants && (
+                    <Dropdown
+                      options={tenants.map((tenant) => ({
+                        label: tenant.node.id,
+                        value: `${tenant.node.id} ${tenant.node.publicName ? `(${tenant.node.publicName})` : ''}`
```
BlairCurrey, line 87: Good call.... came here to suggest using the name for the select after just seeing the id in the UI, but I understand the default operator just doesnt have one.

packages/frontend/app/lib/api/asset.server.ts
```diff
@@ -139,8 +140,12 @@ export const listAssets = async (request: Request, args: QueryAssetsArgs) => {
...
 }
 
-export const createAsset = async (request: Request, args: CreateAssetInput) => {
-  const apolloClient = await getApolloClient(request)
+export const createAsset = async (
+  request: Request,
+  args: CreateAssetInput,
+  tenantId?: string
+) => {
+  const apolloClient = await getApolloClient(request, tenantId)
```
BlairCurrey, line 148: I think something is missing here.

The `tenantId` is never passed in `createAsset` (just called once) nor `getApolloClient` (only passed in here but its `undefined`). And there is no `tenantId` on the `CreateAssetInput`. The backend mutation uses the header, so it's just always going to be the operator.

You can validate this by changing the `defaultValue` on the tenant dropdown to any bogus value and submitting and see that it works (should fail because its not a real tenant id).

I think the intention was probably to pass the `tenantId` in here and override the header but I think previously we agreed we didn't want to do that to ensure the header always identifies the requestor. 

I think we can remove the header overriding, put the `tenantId` on the gql input, and use the `ctx.forTenantId` provided by the [setForTenantIdGraphQLMutationMiddleware](https://github.com/interledger/rafiki/blob/2893/multi-tenancy-v1/packages/backend/src/graphql/middleware/index.ts#L55) (looks like mutation context type for assets needs to be updated to `ForTenantIdContext`).

packages/backend/src/graphql/schema.graphql
```diff
@@ -669,6 +671,7 @@ type Asset implements Model {
   ): FeesConnection
   "The date and time when the asset was created."
   createdAt: String!
+  tenantId: String
```
mkurapov, line 674:
```suggestion
  tenantId: ID!
```

packages/frontend/app/lib/api/tenant.server.ts
```diff
@@ -0,0 +1,84 @@
+import { gql } from '@apollo/client'
```
mkurapov, line 1: If you merge update `2893/multi-tenancy-v1` locally, you should see changes for this file from #3256

BlairCurrey (approved):

PR #3314, opened 2025-02-21
fix(backend): allow admin api to query all receivers and move checks

### LINKED ISSUE #3253: Allow querying completed and expired receivers (incoming payments)
## Context
In our backend GraphQL Admin API, we have a `receiver` [query](https://rafiki.dev/apis/graphql/backend/queries/#receiver). It allows fetching a "local" incoming payment or a remote incoming payment, which is fetched at a different Rafiki instance using Open Payments.

Currently, the `receiver` query returns a payment _only_ if it hasn't been completed yet. This is not the intended behaviour. 

### Expected behaviour
We should be able to fetch a receiver (local or remote incoming payment), even if it has already been completed. 

The logic that needs to be changed is in the `Receiver` class, as we do not allow constructing a new `Receiver` instance from a payment that has been completed, expired, or has an empty `methods` array. We need to move those checks from inside the constructor directly to the places where they are needed (like during quote creation or outgoing payment processing).

## Todos
- [ ] Allow creating `Receiver` instances even if the payment is completed, expired, or has an empty methods array
- [ ] Move those checks to the correct places where they are needed
- [ ] Make sure the `receiver` Admin API query returns incoming payments even if they are completed or expired

cozminu:

- moved validation checks (completed, expired and methods availability) to createOutgoingPayment, handleSending, resolveReceiver via `Resolver.isInvalidExpiredOrCompleted`

## Context

fixes #3253 

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/receiver/model.ts
+++ b/packages/backend/src/open_payments/receiver/model.ts
@@ -35,21 +35,9 @@ export class Receiver {
     incomingPayment: OpenPaymentsIncomingPaymentWithPaymentMethod,
     isLocal: boolean
   ) {
-    if (incomingPayment.completed) {
-      throw new Error('Cannot create receiver from completed incoming payment')
-    }
-
     const expiresAt = incomingPayment.expiresAt
       ? new Date(incomingPayment.expiresAt)
       : undefined
-
-    if (expiresAt && expiresAt.getTime() <= Date.now()) {
-      throw new Error('Cannot create receiver from expired incoming payment')
-    }
-
-    if (!incomingPayment.methods.length) {
-      throw new Error('Missing payment method(s) on incoming payment')
-    }
 
     const incomingAmount = incomingPayment.incomingAmount
       ? parseAmount(incomingPayment.incomingAmount)
@@ -119,5 +107,31 @@ export class Receiver {
       sharedSecret: this.sharedSecret,
       requestCounter: Counter.from(0) as Counter
     }
+  }
+
+  public static isInvalidExpiredOrCompleted(
+    receiver: Receiver | undefined
+  ): receiver is undefined {
+    if (!receiver) {
+      return true
+    }
+    const incomingPayment = receiver.incomingPayment
+    if (incomingPayment.completed) {
+      // throw new Error('Cannot create receiver from completed incoming payment')
+      return true
+    }
+    if (
+      incomingPayment.expiresAt &&
+      incomingPayment.expiresAt.getTime() <= Date.now()
+    ) {
+      // throw new Error('Cannot create receiver from expired incoming payment')
+      return true
+    }
+    if (!incomingPayment.methods.length) {
+      // throw new Error('Missing payment method(s) on incoming payment')
+      return true
+    }
+
+    return false
   }
 }
--- a/packages/backend/src/open_payments/receiver/model.ts
+++ b/packages/backend/src/open_payments/receiver/model.ts
@@ -111,25 +111,25 @@ export class Receiver {
 
   public static isActive(receiver: Receiver | undefined): receiver is Receiver {
     if (!receiver) {
-      return true
+      return false
     }
     const incomingPayment = receiver.incomingPayment
     if (incomingPayment.completed) {
       // throw new Error('Cannot create receiver from completed incoming payment')
-      return true
+      return false
     }
     if (
       incomingPayment.expiresAt &&
       incomingPayment.expiresAt.getTime() <= Date.now()
     ) {
       // throw new Error('Cannot create receiver from expired incoming payment')
-      return true
+      return false
     }
     if (!incomingPayment.methods.length) {
       // throw new Error('Missing payment method(s) on incoming payment')
-      return true
+      return false
     }
 
-    return false
+    return true
   }
 }
--- a/packages/backend/src/open_payments/receiver/model.ts
+++ b/packages/backend/src/open_payments/receiver/model.ts
@@ -45,7 +45,7 @@ export class Receiver {
     const receivedAmount = parseAmount(incomingPayment.receivedAmount)
 
     // TODO: handle multiple payment methods
-    const ilpMethod = incomingPayment.methods.find(
+    const ilpMethod = incomingPayment.methods?.find(
       (method) => method.type === 'ilp'
     )
     if (!ilpMethod) {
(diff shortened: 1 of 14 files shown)


cozminu:
tests tbd after review

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -120,4 +108,30 @@ export class Receiver {
       requestCounter: Counter.from(0) as Counter
     }
   }
+
+  public static isInvalidExpiredOrCompleted(
```
BlairCurrey, line 112: Does this need to be a static method? Or can we just make it a regular method, and use `this` to refer to the receiver instead of passing it in? This allows us to not import the `Receiver` model across our different services and leverages the receiver instance we already have. 

Then instead of handling `undefined` in this method we check where this is called like:

So this:

```ts
      if (!receiver || receiver.isInvalidExpiredOrCompleted()) {
        throw OutgoingPaymentError.InvalidQuote
      }
```
Instead of

```ts
      if (Receiver.isInvalidExpiredOrCompleted(receiver)) {
        throw OutgoingPaymentError.InvalidQuote
      }
```

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -120,4 +108,30 @@ export class Receiver {
...
+      incomingPayment.expiresAt &&
+      incomingPayment.expiresAt.getTime() <= Date.now()
+    ) {
+      // throw new Error('Cannot create receiver from expired incoming payment')
+      return true
+    }
+    if (!incomingPayment.methods.length) {
+      // throw new Error('Missing payment method(s) on incoming payment')
+      return true
+    }
```
BlairCurrey, line 133: We can probably remove this payment method check. Its not going to do anything because the `Receiver` constructor will error if it's empty. It looks for ilp method then throws if it cant find it.

I am considering if it should be returned to the constructor but it seems kind of redundant in the first place. With or without a check for this, if method is empty, it will error as described here (missing ilp payment method). I guess the `method.length` check was just slightly more informative. I think just remove it here.

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -120,4 +108,30 @@ export class Receiver {
       requestCounter: Counter.from(0) as Counter
     }
   }
+
+  public static isInvalidExpiredOrCompleted(
+    receiver: Receiver | undefined
+  ): receiver is undefined {
```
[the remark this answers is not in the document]
cozminu: first time around I did it, typescript response was boolean
```
public static isActive(receiver: Receiver | undefined): boolean {
```
the issue was that I had to double check for type:
Screenshot 2025-02-25 at 10 38 03

but if we do the non-static method, then we won't have this problem

packages/backend/src/open_payments/receiver/model.ts
cozminu: moved to constructor for now because it makes more sense since there's already a check there and ILP method is hard-coupled to the Receiver via `ilpAddress` property.
Checked if this will affect querying the receiver via Admin API and it doesn't.

packages/backend/src/open_payments/receiver/model.ts
cozminu: It does not, I chose static for simplicity only. Changed to non-static method.

packages/backend/src/open_payments/receiver/model.ts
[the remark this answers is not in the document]
BlairCurrey: That makes sense, I see how its working now.

packages/backend/src/payment-method/ilp/stream-credentials/service.ts
```diff
@@ -33,9 +33,6 @@ function getStreamCredentials(
   deps: ServiceDependencies,
   payment: IncomingPayment
 ): IlpStreamCredentials | undefined {
-  if (payment.isExpiredOrComplete()) {
```
njlie, line 36: `IncomingPayment.isExpiredOrComplete()` isn't used anywhere else in the code with this deletion - perhaps it can be cleaned up?

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -120,4 +108,28 @@ export class Receiver {
...
     }
   }
+
+  public static isActive(receiver: Receiver | undefined): receiver is Receiver {
+    if (!receiver) {
+      return false
+    }
+    const incomingPayment = receiver.incomingPayment
+    if (incomingPayment.completed) {
+      // throw new Error('Cannot create receiver from completed incoming payment')
```
njlie, line 118: nit: clean up comments

packages/backend/src/open_payments/receiver/service.test.ts
```diff
@@ -251,7 +251,7 @@ describe('Receiver Service', (): void => {
         expect(remoteIncomingPaymentServiceGetSpy).toHaveBeenCalledTimes(1)
       })
 
-      test('returns undefined if error getting receiver from remote incoming payment', async () => {
+      test('gets receiver from completed remote incoming payment', async () => {
         const mockedIncomingPayment = mockIncomingPaymentWithPaymentMethods({
           completed: true // cannot get receiver with a completed incoming payment
```
mkurapov, line 256:
```suggestion
          completed: true
```

packages/backend/src/payment-method/ilp/stream-credentials/service.ts
mkurapov: Given this, we should update the `receiverService.getLocalIncomingPayment` function to not error if we could not get stream credentials for the payment.

mkurapov (approved):

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -35,29 +35,17 @@ export class Receiver {
...
       ? parseAmount(incomingPayment.incomingAmount)
       : undefined
     const receivedAmount = parseAmount(incomingPayment.receivedAmount)
 
     // TODO: handle multiple payment methods
-    const ilpMethod = incomingPayment.methods.find(
+    const ilpMethod = incomingPayment.methods?.find(
       (method) => method.type === 'ilp'
     )
     if (!ilpMethod) {
```
mkurapov, line 51: Maybe we should be a bit more permissive here, and _not_ throw if there are no payment methods provided. (we can keep the check for invalid ilp address - if payment method exists).  What do you think?

njlie (approved):

Issue #3319, opened 2025-02-24
[BUG] Database Deadlocks


BlairCurrey:
## Bug Report

**Describe the bug**

Our database deadlocks in places where we open a new connection before an open transaction is resolved while at max connections.

This, basically: https://github.com/Vincit/objection.js/issues/1137#issuecomment-561149456

In our case, it may look something like:

```ts
await Quote.transaction(async (trx) => {
   await receiverService.get(id)
   await Quote.query(trx).insert({...})
})
```

We open a transaction then try to open a new connection from another db call not using the transaction (in `receiverService.get`). This results in the db call in `receiverService.get` (or similar) waiting to get a connection, which it never gets because the current connections never resolve (because its waiting on `receiverService.get`).

**To Reproduce**

- Set the knex connection settings to a min/max of 1.
- Change performance test to do 1 vu and 1 iteration 
- Run test and see it never completes

That's the easiest way but it doesnt just happen under the contrived scenario of 1 min/max connection. It happens at any number of VUs >= max connections.

**Expected behavior**

Performance tests should pass with 1vu/1iteration with max connection size of 1.

**Resolution**

Note that our current system makes some queries inside transaction blocks without using the transaction. We should probably evaluate if this behavior is correct on a case-by-case basis.

But in general, we either need to move the queries not using the transaction outside of the scope of the transaction (before or after), or we need to pass the transaction into the query. We shouldnt include anything more than necessary in the transaction to reduce transaction lock time and contention.

Issue #3226, opened 2025-01-20
Local payments performance script


mkurapov:
### Context
Currently, we have a performance test script that goes through the steps of making a payment (create incoming payment, create quote, create outgoing payment) between two different Rafikis. With the introduction of local payments in #2834, we also want to be able to measure the performance of "local" payments that happen within the same Rafiki instance. 

For this issue, let's create a script that is very similar to the existing `rafiki/test/performance/scripts/create-outgoing-payments.js` script, except instead of calling `createReceiver` mutation, we call `createIncomingPayment` instead. This will mean we will have all the resources (incoming payment, quote, and outgoing payment) be created on the same, cloud-nine-wallet, Rafiki instance.

### Todo
- [ ] Create k6 script in `rafiki/test/performance/scripts/` that sets up a local payment in the same Rafiki

PR #3320, opened 2025-02-24
fix(backend): deadlocks

BlairCurrey:

- fixes deadlocks by either using transaction after opening one, or moving calls before/after the commit
  - changes quote creation to not create then patch. which required some small changes to the expiryDate (base it off `Date.now()` before creating instead of `quote.createdAt`)
- adds create local outgoing payment k6 script
- adds caching to fees

## Context

fixes https://github.com/interledger/rafiki/issues/3319 #3226 

## Checklist

- [ ] Tests added/updated


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -145,89 +156,88 @@ async function createQuote(
...
+    const quote = await deps.paymentMethodHandlerService.getQuote(
+      paymentMethod,
+      {
+        quoteId,
+        walletAddress,
+        receiver,
+        receiveAmount: options.receiveAmount,
+        debitAmount: options.debitAmount
+      }
+    )
```
BlairCurrey, line 193: Something still needs to be done here one way or another.

1. return ilp quote details from get quote and save in transaction with quote insert. This assumes the ilp connector db calls do not need to use the transaction (they are not using it on main). This should be more performant as it greatly reduces the duration of the transaction. However it does leak the ilp payment method concern into quotes.

2. Do this in the transaction. Pass the transaction into the connector and use for all the db calls.

Originally we called this inside the transaction. The transaction was passed in but only used to save the `IlpQuoteDetails`. It was not used by the connector. This `getQuote` is a heavy operation (for ilp payments) because of all the db/network calls so it would be nice not to include it in the transaction provided its safe.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -347,13 +357,36 @@ function calculateFixedSendQuoteAmounts(
   }
 }
 
+function calculateExpiry(
+  deps: ServiceDependencies,
+  quote: UnfinalizedQuote,
+  receiver: Receiver
+): Date {
+  const quoteExpiry = new Date(Date.now() + deps.config.quoteLifespan)
```
BlairCurrey, line 365: creates an expiry of now, instead of `createdAt` (since we are setting ahead of creation). This was to allow us to simply create the quote and not create then update it.

packages/backend/src/open_payments/quote/service.ts
BlairCurrey: Another idea for **1.** is a new `saveAdditionalDetails(trx: Knex.Transction, quote: Quote)` method on the  `paymentMethodHandlerService` that inserts into `IlpQuoteDetails` for the ilp paymnet method, and does nothing for local payments. Keeps the ilp quote details encapsulated in the ilp payment method handler.

BlairCurrey: >is a good approach: Do you think if we start a new transaction before we call this, and then pass in an optional trx into the .getQuote method, it be a perf issue?

First, if we do this we need to use this transaction everywhere inside `getQuote` (including connector). Currently, and on main, this is not done inside the transaction. This would be required to prevent the same deadlocking issue at max connections. 

I think that should determine the direction here. If everything in `getQuote` needs to start using the transaction we should pass it into `getQuote` and pass into the connector and use it. Otherwise we shouldn't call `getQuote` within the lifecycle of the transaction because it unnecessarily elongates the transaction. The extra work would be the ~11 (?) network calls between rafikis in the connector to get the quote. At that point the question is how to ensure the `ilpQuoteDetails` are still saved inside the transaction. Directly call the IlpQuoteDetails insert here or
 some new `paymentMethodHandlerService. saveAdditionalDetails(trx: Knex.Transction, quote: Quote)` to expose it.

BlairCurrey: As we just discussed in our call, I think we're fine with the current behavior (not inside trx) but will still need to ensure the `IlpQuoteDetails` is happening inside the transaction. I think the new `saveAdditionalDetails` method will work fine for that.

packages/backend/src/open_payments/quote/model.ts
```diff
@@ -22,7 +22,7 @@ export class Quote extends WalletAddressSubresource {
 
   public estimatedExchangeRate!: number
   public feeId?: string
-  public fee?: Fee
+  public fee?: Fee | null
```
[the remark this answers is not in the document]
BlairCurrey: Had to do with how it was populated in different scenarios. If it was joined but there was no fee vs not joined and maybe some other scenarios. Was causing problems in the centralized getPage tests or similar as I recall.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -236,222 +243,242 @@ async function createOutgoingPayment(
...
+        } else if (err instanceof knex.KnexTimeoutError) {
+          deps.logger.error(
+            { grant: grantId },
+            'Could not create outgoing payment: grant locked'
+          )
+        }
+        throw err
+      } finally {
+        stopTimerOP()
+        span.end()
```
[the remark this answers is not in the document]
BlairCurrey: It appears so, yes.

Given [this example from the docs](https://opentelemetry.io/docs/languages/js/instrumentation/):

```ts
export function rollTheDice(rolls: number, min: number, max: number) {
  // Create a span. A span must be closed.
  return tracer.startActiveSpan('rollTheDice', (span: Span) => {
    const result: number[] = [];
    for (let i = 0; i < rolls; i++) {
      result.push(rollOnce(min, max));
    }
    // Be sure to end the span!
    span.end();
    return result;
  });
}
```

And some testing. I removed the `span.end()` and no longer saw this `outgoingPaymentService.createOutgoingPayment` span.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -437,18 +452,14 @@ async function finalizeQuote(
...
+  return patchOptions
 }
 
 async function getWalletAddressPage(
   deps: ServiceDependencies,
   options: ListOptions
 ): Promise<Quote[]> {
-  const quotes = await Quote.query(deps.knex)
-    .list(options)
-    .withGraphFetched('fee')
```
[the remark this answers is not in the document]
BlairCurrey: That, and if I recall correctly there were some inconsistencies with the centralized getQuote, getPage tests. Which expect the `get` methods and `create` methods to return the same object. As I recall the fee is no longer returned from the create. Or it's null instead of undefined.

mkurapov (approved):

packages/backend/src/open_payments/quote/service.ts
BlairCurrey: After _further_ discussion we came to the conclusion that we dont need with ilp quote details and the quote. If there is a failure between the details insert and the quote insert it will not create any sort of consistency issues and will just exist in our db as something that happened during ilp quoting (although not utilized).

mkurapov (approved):

PR #3325, opened 2025-03-01
feat(wallet-address)!: possibility to specify wallet address range

### LINKED ISSUE #3278: [Multi-Tenant] Wallet address prefixes
## Context
Each tenant should have a setting for which wallet address prefix they are allowed to use. For example, in the context of JOPACC, an example would be:

Bank 1 (Tenant 1):
Prefix/range: `https://jopacc.com/arabbank/`
`https://jopacc.com/arabbank/<user>`

Bank 2 (Tenant 2):
Prefix/range: `https://jopacc.com/abcbank/`
`https://jopacc.com/abcbank/<user>`

If a tenant tries to create a wallet address with the incorrect "prefix" or "range", then we should throw an error. 
We should store this prefix/range in the tenant settings.

During wallet address creation, we should check whether the non-operator tenant is trying to create a wallet address with the correct URL. Be wary of slashes.

- Currently blocked by #3127 

## Todos
- [ ] Allow configuring a wallet address prefix or range via tenant settings
- [ ] Check the wallet address prefix during wallet address creation

golobitch:
- feat(graphql): schema update for specifying settings when creating tenant
- feat(tenant): add possibility to specify initial tenant settings as an operator
- test(tenant): create with settings
- feat(wallet-address): rename url to address
- fix(tenant-settings): duplicate key for tenant
- feat(wallet-address): replace url field with address field
- chore(backend): format

- allow operator to specify wallet address range for tenant
- BREAKING CHANGE: replace url field in mutation with address because it will be possible to specify whole url or just a last part of it

## Context

- fixes #3278

## Checklist

- [x] Tests added/updated

### DIFF
--- a/localenv/mock-account-servicing-entity/app/lib/requesters.ts
+++ b/localenv/mock-account-servicing-entity/app/lib/requesters.ts
@@ -88,7 +88,7 @@ export async function createWalletAddress(
       createWalletAddress(input: $input) {
         walletAddress {
           id
-          url
+          address
           publicName
         }
       }
--- a/localenv/mock-account-servicing-entity/app/lib/requesters.ts
+++ b/localenv/mock-account-servicing-entity/app/lib/requesters.ts
@@ -96,7 +96,7 @@ export async function createWalletAddress(
   `
   const createWalletAddressInput: CreateWalletAddressInput = {
     assetId,
-    url: accountUrl,
+    address: accountUrl,
     publicName: accountName,
     additionalProperties: []
   }
--- a/packages/backend/src/graphql/resolvers/tenant_settings.ts
+++ b/packages/backend/src/graphql/resolvers/tenant_settings.ts
@@ -21,7 +21,7 @@ export const getTenantSettings: TenantResolvers<TenantedApolloContext>['settings
       tenantId: parent.id
     })
 
-    return tenantSettings.map((x) => tenantSettingsToGraphql(x))
+    return tenantSettingsToGraphql(tenantSettings)
   }
 
 export const createTenantSettings: MutationResolvers<TenantedApolloContext>['createTenantSettings'] =
@@ -38,13 +38,22 @@ export const createTenantSettings: MutationResolvers<TenantedApolloContext>['cre
     })
 
     return {
-      settings: tenantSettings.map((x) => tenantSettingsToGraphql(x))
+      settings: tenantSettingsToGraphql(tenantSettings)
     }
   }
 
-export const tenantSettingsToGraphql = (
+const tenantSettingToGraphql = (
   tenantSetting: TenantSetting
 ): SchemaTenantSetting => ({
   key: tenantSetting.key,
   value: tenantSetting.value
 })
+
+export const tenantSettingsToGraphql = (
+  tenantSettings?: TenantSetting[]
+): SchemaTenantSetting[] => {
+  if (!tenantSettings) {
+    return []
+  }
+  return tenantSettings.map((x) => tenantSettingToGraphql(x))
+}
--- a/packages/backend/src/graphql/resolvers/tenant_settings.ts
+++ b/packages/backend/src/graphql/resolvers/tenant_settings.ts
@@ -1,21 +1,14 @@
 import { TenantedApolloContext } from '../../app'
-import { Pagination } from '../../shared/baseModel'
-import { getPageInfo } from '../../shared/pagination'
 import { TenantSetting } from '../../tenants/settings/model'
 import {
   ResolversTypes,
-  SortOrder,
   TenantResolvers,
   TenantSetting as SchemaTenantSetting,
   MutationResolvers
 } from '../generated/graphql'
 
 export const getTenantSettings: TenantResolvers<TenantedApolloContext>['settings'] =
-  async (
-    parent,
-    args,
-    ctx
-  ): Promise<ResolversTypes['TenantSettingsConnection']> => {
+  async (parent, args, ctx): Promise<ResolversTypes['TenantSetting'][]> => {
     if (!parent.id) {
       throw new Error('missing tenant id')
     }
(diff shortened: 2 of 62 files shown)


packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -165,7 +171,58 @@ async function createWalletAddress(
...
+
+  const found = (await deps.tenantSettingService.get({
+    tenantId: options.tenantId,
+    key: TenantSettingKeys.WALLET_ADDRESS_URL.name
+  })) as TenantSetting[]
+
+  if (!found || found.length === 0) {
+    if (!options.isOperator) {
+      return WalletAddressError.WalletAddressSettingNotFound
+    }
```
mkurapov, line 184: We should add a test for this

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -165,7 +171,58 @@ async function createWalletAddress(
...
+      if (relativePath.startsWith('/')) {
+        relativePath = relativePath.substring(1)
+      }
+      finalWalletAddressUrl = tenantBaseUrl + relativePath
+    } catch (err) {
+      return WalletAddressError.InvalidUrl
+    }
+  }
+
+  if (!isValidWalletAddressUrl(finalWalletAddressUrl)) {
```
mkurapov, line 225: Small suggestion: we should place this in a separate fn

localenv/mock-account-servicing-entity/app/lib/requesters.ts
```diff
@@ -96,7 +96,7 @@ export async function createWalletAddress(
   `
   const createWalletAddressInput: CreateWalletAddressInput = {
```
mkurapov, line 97: Need to update L91 as well to use `address`

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -165,7 +171,58 @@ async function createWalletAddress(
...
+    const walletAddressUrl = new URL(options.address)
+    if (!walletAddressUrl.href.startsWith(tenantWalletAddressUrl.href)) {
+      return WalletAddressError.InvalidUrl
+    }
+    finalWalletAddressUrl = walletAddressUrl.toString()
+  } else {
+    // in case that client provided just the path / wallet address name, construct the address using the wallet address url from tenant setting
+    try {
+      let relativePath = options.address
+      if (relativePath.startsWith('/')) {
```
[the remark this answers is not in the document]
golobitch: Huh, good point. What do you propose we do in that case? should we just leave it as is? or trim all forward slashes?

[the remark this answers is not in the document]
mkurapov: Thinking about this, I think we can just leave it as is. It's technically a valid path name, so as long as we're handling the typical case of someone entering `/alice` or something, we should be OK>

packages/backend/src/tenants/service.ts
```diff
@@ -102,13 +105,18 @@ async function createTenant(
...
-      { trx }
-    )
+    const createInitialTenantSettingsOptions = {
+      tenantId: tenant.id,
+      setting: TenantSetting.default()
+    }
+    if (settings) {
+      createInitialTenantSettingsOptions.setting =
+        createInitialTenantSettingsOptions.setting.concat(settings)
+    }
```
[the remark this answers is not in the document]
golobitch: Yes it is true. Setting can be at this point twice in the array. However, then in create method on tenant settings, there is upsert. 

This part is the most important one that handles inserts into the database:

```
return TenantSetting.query(extra?.trx ?? deps.knex)
    .insert(dataToUpsert)
    .onConflict(['tenantId', 'key'])
    .merge()
    .returning('*')
```

* **.onConflict(['tenantId', 'key'])** tells the dataase what to consider as the unique constraint for conflicts. So in our case it is a combination of `tenantId` and `key`.
* **.merge()** triggers when a conflict is found. So instead of throwing an error, this command tells the database to update aka merge the existing row with the new data provided. This is what make it an upsert (update or insert) operation 

I think that this should be ok. What do you think? What did you have in mind by "use an object to do this calculation" ?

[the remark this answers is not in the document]
mkurapov: We can leave it as is.

The upsert makes sense, and this does what we want (as long as we concatenate the arrays in this exact order).

> use an object to do this calculation
To be on the safe side & to avoid duplicates in the final array, I was just thinking we just use an object + spread operator to gather the key/values, something like `const finalSettings = { ...defaultSettings, ...customSettings }`. And then just convert to an array before inserting.

packages/backend/src/tenants/settings/service.ts
```diff
@@ -118,20 +118,22 @@ async function createTenantSetting(
...
     return []
   }
 
-  return TenantSetting.query(extra?.trx ?? deps.knex).insertAndFetch(
-    dataToInsert
-  )
+  return TenantSetting.query(extra?.trx ?? deps.knex)
+    .insert(dataToUpsert)
+    .onConflict(['tenantId', 'key'])
+    .merge()
```
mkurapov, line 135: we should add a test for upsert behaviour

packages/backend/src/tenants/settings/service.test.ts
```diff
@@ -276,21 +270,4 @@ describe('TenantSetting Service', (): void => {
       expect(dbTenantData.filter((x) => !x.deletedAt)).toHaveLength(0)
     })
   })
-
-  describe('pagination', (): void => {
```
mkurapov, line 280: did you mean to delete?

golobitch: Yes, because it doesn't make sense at this moment, because you have only few keys that you can add ... and we are also doing upsert ... so ... not a lot of data there :)

mkurapov: I think we should keep the test because the `getTenantSettingPageForTenant` service method is still used by the tenant settings graphql resolver

tadejgolobic: oh, maybe we can remove that resolver then :D

packages/backend/src/graphql/resolvers/tenant_settings.ts
```diff
@@ -24,27 +17,11 @@ export const getTenantSettings: TenantResolvers<TenantedApolloContext>['settings
       'tenantSettingService'
     )
 
-    const { sortOrder, ...pagination } = args
-    const order = sortOrder === 'ASC' ? SortOrder.Asc : SortOrder.Desc
+    const tenantSettings = (await tenantSettingsService.get({
+      tenantId: parent.id
+    })) as TenantSetting[]
```
mkurapov, line 22:
```suggestion
    const tenantSettings = await tenantSettingsService.get({
      tenantId: parent.id
    })
```

packages/backend/src/graphql/schema.graphql
```diff
@@ -1590,32 +1590,7 @@ type Tenant implements Model {
...
-  "A list of edges representing tenant settings and cursors for pagination."
-  edges: [TenantSettingEdge!]!
-}
-
-type TenantSettingEdge {
-  "A tenant setting node in the list."
-  node: TenantSetting!
-  "A cursor for paginating through the tenants."
-  cursor: String!
+  settings: [TenantSetting!]
```
mkurapov, line 1593:
```suggestion
  settings: [TenantSetting!]!
```

packages/backend/src/graphql/resolvers/tenant_settings.ts
golobitch: ha yes, good catch. I need to check if Bogdan changed the return type and rebase to our MT base branch. After that it should be ok

packages/backend/src/graphql/schema.graphql
golobitch: I am just thinking what I didn't do this, because I remember that initially I did exactly like this, but somehow I cannot remember why :/ But sure, let's try again :)

mkurapov (approved):

PR #3327, opened 2025-03-04
feat(backend)!: tenanted rates endpoints

### LINKED ISSUE #3120: [Multi-Tenant] Tenanted Rates endpoints
As quotes and outgoing payments are now tenanted, tenants should now be responsible for setting their own rates and hosting them at their own endpoint.

In the `backend` business logic, the existing rates service gets rates from an environment variable for a URL. The Rates service should instead be updated to require a `tenantId` when acquiring rates, so that it pulls from the tenant's `ratesEndpoint` when calculating rates.

- [ ] Update the `RatesService.rates()` method to retrieve rates based on the url from a provided tenant id.
- [ ] Update the `RatesService` to cache rates on a per-tenant basis.

### Prerequisites
* https://github.com/interledger/rafiki/issues/3103

sanducb:
- rates endpoints are tenanted and use specified URLs in tenant settings if available, otherwise operator URL is used.
- conversion functions of rates service now support an optional `tenantId` as parameter

## Context

fixes #3120 

Tenanted peers is needed for getting `tenantId` inside the connector.

## Checklist

- [x] Tests added/updated


packages/backend/src/payment-method/ilp/connector/core/telemetry.ts
```diff
@@ -50,6 +50,7 @@ export async function incrementAmount(
         assetCode: code,
         assetScale: scale
       },
+      undefined,
```
njlie, line 53: If the [sourceAccount](https://github.com/interledger/rafiki/blob/2b428f57761dcd614f926214bff9832164e7d804/packages/backend/src/payment-method/ilp/connector/core/rafiki.ts#L157) in the caller for `incrementAmount` has a known `staticIlpAddress`, it might be possible to acquire a tenantId for this telemetry call from the peer with that address (once tenanted peers are added).

packages/backend/src/asset/service.ts
```diff
@@ -99,13 +108,30 @@ async function createAsset(
...
+      .select('*')
+
+    const sameCodeAssets = assets.find((asset) => asset.code === code)
+    if (!sameCodeAssets && assets.length > 0) {
+      const exchangeUrlSetting = await deps.tenantSettingService.get({
+        tenantId,
+        key: TenantSettingKeys.EXCHANGE_RATES_URL.name
+      })
+
+      const tenantExchangeRatesUrl = Array.isArray(exchangeUrlSetting)
```
mkurapov, line 122: I wonder why the type returned is `TenantSetting | TenantSetting[]` I think it should always be `TenantSetting[]`, unless I'm missing something

CC @golobitch

packages/backend/src/rates/service.ts
```diff
@@ -24,20 +27,26 @@ export type RateConvertDestinationOpts = Omit<
 >
 
 export interface RatesService {
-  rates(baseAssetCode: string): Promise<Rates>
+  rates(baseAssetCode: string, tenantId?: string): Promise<Rates>
```
mkurapov, line 30: Should `tenantId` be required? Since in the quote requests and the connector balance middleware, we will always have tenantId, right?

packages/backend/src/rates/service.ts
```diff
@@ -99,68 +110,92 @@ class RatesServiceImpl implements RatesService {
...
+    if (this.inProgressRequests[ratesCacheKey] === undefined) {
+      this.inProgressRequests[ratesCacheKey] = this.fetchNewRates(
+        baseAssetCode,
+        tenantId
       )
-
-      throw new Error(errorMessage)
-    } finally {
-      delete this.inProgressRequests[baseAssetCode]
+        .then(async (rates) => {
```
mkurapov, line 162: Do we need to use `.then` here? vs keeping the await + catch as usual

test/integration/integration.test.ts
```diff
@@ -71,6 +71,19 @@ describe('Integration tests', (): void => {
...
+        const c9Settings = {
+          settings: [
+            { key: 'EXCHANGE_RATES_URL', value: c9.config.seed.rates_url }
+          ]
+        }
+        const c9TenantSettings = await testActions.admin.createTenantSettings(
+          c9Settings,
+          'sending'
+        )
+        assert(c9TenantSettings)
```
mkurapov, line 86: Do we still need this one? vs just using the operator one also for the sender?

packages/backend/src/payment-method/local/service.ts
```diff
@@ -117,17 +121,20 @@ async function getQuote(
...
+          code: receiveAmount.assetCode,
+          scale: receiveAmount.assetScale
+        }
       },
-      destinationAsset: {
-        code: receiveAmount.assetCode,
-        scale: receiveAmount.assetScale
-      }
-    })
+      options.walletAddress.tenantId
```
mkurapov, line 136: There's an edge case here: the sending wallet's tenant could have only configured a single asset (and not having any rates URL set as a result), but the receiver's tenant wallet is in a different asset code (who might also be only supporting a single asset). 

I'm not sure there is a straightforward "coding" solution here but a business decision one: _someone_ would need to know in this case what the correct FX is, maybe it is the operator.

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -75,7 +75,7 @@ services:
       API_SECRET: iyIgCprjb9uL8wFckR+pLEkJWMB7FJhgkvqhTQR/964=
       OPEN_PAYMENTS_URL: ${CLOUD_NINE_OPEN_PAYMENTS_URL:-http://cloud-nine-wallet-backend}
       WEBHOOK_URL: http://cloud-nine-wallet/webhooks
-      EXCHANGE_RATES_URL: http://cloud-nine-wallet/rates
+      EXCHANGE_RATES_URL: http://cloud-nine-wallet/prices
```
mkurapov, line 78: Should this be changed?

sanducb: Good catch! I copy-pasted the url from the `values.yaml` which has `/prices` when adding the env var back for default URL. I'll replace `/prices` with `/rates`.

sanducb: Maybe we should also change the `values.yaml` file to use `/rates` instead.

test/integration/integration.test.ts
sanducb: We don't need it. The only reason I left it here is to check and display both scenarios i.e. having an exchange rates URL defined in the tenant settings and using operator exchange rates URL.

packages/backend/src/rates/service.ts
sanducb: I put it optional because if not provided, operator `tenantId` will be used. I thought that it is less error prone to have operator `tenantId` used as default directly in the rates service, so that we don't have to change other calls to the rates service to include this parameter unless we want to specify a different `tenantId` than the operator's.

packages/backend/src/asset/service.ts
golobitch: hm,...good question. I don't remember why is there TenantSetting as a return type. @sanducb I think that you can change the return type of that to always return array.

test/integration/testenv/happy-life-bank/seed.yml
```diff
@@ -51,6 +51,7 @@ accounts:
     id: 60257507-3191-4507-9d77-9071fd6b3c30
     initialBalance: 150000
     assetCode: MXN
+rates_url: 'http://host.docker.internal:8889/rates'
```
mkurapov, line 54: I think this (and the one for cloud-nine) can go too, right? 
(along with `rates_url` in the `SeedInstance` interface of the mase-lib)

packages/backend/src/payment-method/ilp/connector/core/middleware/balance.ts
```diff
@@ -40,11 +40,18 @@ export function createBalanceMiddleware(): ILPMiddleware {
...
     const sourceAmount = BigInt(amount)
-    const destinationAmountOrError = await services.rates.convertSource({
-      sourceAmount,
-      sourceAsset: accounts.incoming.asset,
-      destinationAsset: accounts.outgoing.asset
-    })
+    const token = accounts.outgoing.http?.outgoing.authToken
+    const peer = token
+      ? await services.peers.getByIncomingToken(token)
+      : undefined
```
mkurapov, line 46: Each of the `accounts` here are either a Peer, Incoming Payment, or Outgoing Payment. Each of them should already have a tenantId on them, so if we add `tenantId` as a property of the `ConnectorAccount` interface, you should be able to use it directly.

packages/backend/src/payment-method/ilp/connector/core/rafiki.ts
```diff
@@ -193,6 +193,14 @@ export class Rafiki<T = any> {
...
 
     const { code, scale } = sourceAccount.asset
+    const ilpAddress = sourceAccount.staticIlpAddress
+    const tenantId = ilpAddress
+      ? (
+          await this.publicServer.context.services.peers.getByDestinationAddress(
+            ilpAddress
+          )
+        )?.tenantId
+      : undefined
```
mkurapov, line 203: same comment as in the balance middleware, can leverage tenantId directly here

packages/documentation/src/content/docs/integration/requirements/webhook-events.mdx
```diff
@@ -262,8 +262,8 @@ The incoming payment will either complete or expire.
 <summary>Expand for event sequences</summary>
 {' '}
 
-<Badge text='Single-phase transfer' variant='tip' size='medium' /> An incoming payment
-of $10 was completed.
+<Badge text='Single-phase transfer' variant='tip' size='medium' /> An incoming
```
mkurapov, line 265: Is this from main?

sanducb: I introduced this change (and a few others) when running `pnpm format`. I had to do it because the workflow was failing at the formatting check. In other words, the unformatted code comes from the main multi-tenancy branch if that is what you were asking.

mkurapov (approved):

PR #3328, opened 2025-03-04
feat: handle updating operator api secret

### LINKED ISSUE #3276: Handle Operator API secret Updates
Currently we do not handle operator secret updates correctly. The operator is identified by the api secret from the config (coming from environment variable). As a result:
 
- If we update the operators api secret use it for subsequent api requests, we get a 401 instead of success.
- If we updated the operators api secret and use the original secret on subsequent requests, it works instead of 401ing.

We concluded that we do not want to support updates at run time due to the coordination required with the integration server operator to update on their end as well, and the infrequency of the need to rotate.

TODO:
- [x] Disallow operator secret updates via admin api.
- [x] Allow updating secret via environment variable by updating operator's secret database field to the config value on application start
- [ ] Document this behavior and note that it requires coordination with the integration sever operator, who needs to update the secret as well

BlairCurrey:

- Disallows operator from updating their api secret via admin api (implemented at resolver level, as opposed to service)
- Adds method to update operator api secret from config (if it has changed)
- Changes tenant create form to submit sections independently, similar to our other create forms
- Changes `truncateTables` interface to remove possibility of truncating wrong schema

## Context

fixes #3276

## Checklist

- [x] Tests added/updated


BlairCurrey:
Intended to add some docs explaining how to update operator api secret and how it would require coordination with the integration server owner (would need to update in tandem). However it seems like there is nothing for multi-tenancy. Brought this up on this in our slack workspace here: https://interledgerfoundation.slack.com/archives/C06KZCKEEKW/p1741114127552369

packages/backend/src/tenants/service.ts
```diff
@@ -186,3 +191,23 @@ async function deleteTenant(
...
+
+  if (!tenant) {
+    return TenantError.TenantNotFound
+  }
+  if (tenant.apiSecret !== adminApiSecret) {
+    await Tenant.query(deps.knex)
+      .patch({ apiSecret: adminApiSecret })
+      .where('id', operatorTenantId)
+  }
+}
```
BlairCurrey, line 213: New method to update the secret from config. I made a new method rather than using the update because:

- I wanted to be opinionated about where this is coming from definitely the config, not just some value being passed in which is intended to be from the config.
- I didnt want to update unless its changed. didnt want to add that logic to the update just for this. Perhaps we could just always write it but figured this was safer.

BlairCurrey: I was just noticing that this exemplifies the value of the repo layer... notice we use the cache in the service `get` method but we cant use that here (could if we were getting from the repo). I suppose if thats important I can use/update the cache here as well.

packages/backend/src/index.ts
```diff
@@ -768,6 +769,11 @@ export const start = async (
 
   Model.knex(knex)
 
+  // Update Operator Tenant from config
+  const tenantService = await container.use('tenantService')
+  const error = await tenantService.updateOperatorApiSecretFromConfig()
+  if (error) throw error
+
```
BlairCurrey, line 776: This is happening in the `start` method after we run migrations and start the admin server, etc.

Is this the correct behavior? It will error if the operator id from the env var is not found (tenant service returns an error), or some unknown error throws. The not found error could be from a bad environment variable input.

This will show an error log and exit the process. I debated catching and logging at warning level because strictly speaking the app will still work. Operator tenant id isnt (at this point) used for anything else in the app (just used in the initial migration and here). So it can work without this but it could be in an unexpected state (still with the old api secret). You wouldnt know unless dutifully watching the logs.

BlairCurrey: Thinking about it more, we throw an error when parsing the config if `OPERATOR_TENANT_ID` is not set. So I think its probably correct to throw here if we know its bad.

packages/backend/src/graphql/resolvers/tenant.test.ts
```diff
@@ -85,7 +86,7 @@ describe('Tenant Resolvers', (): void => {
   })
 
   afterEach(async (): Promise<void> => {
-    await truncateTables(appContainer.knex, true)
```
BlairCurrey, line 88: This is an example of the problem that lead to changing the `truncateTables` interface.

packages/backend/src/tests/tableManager.ts
```diff
@@ -9,10 +11,15 @@ export async function truncateTable(
 }
 
 export async function truncateTables(
-  knex: Knex,
-  truncateTenants = false,
-  dbSchema?: string
+  deps: IocContract<AppServices>,
+  options?: { truncateTenants?: boolean }
```
BlairCurrey, line 15: Tightens up the `truncateTables` interface. Most of the changed files in this PR are just changing how `truncateTables` is called.


When we added the options to truncate `tenants` and provide a schema we created an opportunity to truncate tenants from the _wrong_ schema.

For example, here, where we truncated with no `dbSchema` provided: https://github.com/interledger/rafiki/pull/3328/files#r1982206044

This coincidentally wasnt causing problems as most tests just werent interacting with the tenants. However, with these changes we attempt to fetch the tenant every time the app starts (so every test via `createTestApp`) which yielded many `TenantNotFound` errors because another test had dropped them.

This removes the (not completely obvious) responsibility from the caller to pass in dbSchema when truncating tenants. It just uses the config directly to truncate from the same `dbSchema` that the app is using.

packages/backend/src/tenants/service.ts
```diff
@@ -186,3 +191,23 @@ async function deleteTenant(
...
+
+async function updateOperatorApiSecretFromConfig(
+  deps: ServiceDependencies
+): Promise<undefined | TenantError> {
+  const { adminApiSecret, operatorTenantId } = deps.config
+
+  const tenant = await Tenant.query(deps.knex)
+    .findById(operatorTenantId)
+    .whereNull('deletedAt')
+    .first()
```
mkurapov, line 203:
```suggestion
```

packages/backend/src/tenants/service.ts
mkurapov: Makes sense to have another method. As far as caching goes, I think we should update it if we end up updating the secret.

packages/backend/src/tenants/service.test.ts
```diff
@@ -34,20 +35,20 @@ describe('Tenant Service', (): void => {
...
   })
 
   afterEach(async (): Promise<void> => {
-    await truncateTables(knex, true, dbSchema)
+    await truncateTables(deps, { truncateTenants: true })
   })
 
   afterAll(async (): Promise<void> => {
-    nock.cleanAll()
+    nock.cleanAll() // TODO: rm?
```
mkurapov, line 51: I think we can remove since it seems as though its not used here

BlairCurrey: whoops, yes. meant to include that.

packages/backend/src/tenants/service.ts
BlairCurrey: Ah, of course. updated.

packages/frontend/app/lib/validate.server.ts
```diff
@@ -130,20 +130,32 @@ export const updateWalletAddressSchema = z
...
 
-export const updateTenantSchema = z
+export const updateTenantGeneralSchema = z
+  .object({
+    publicName: z.string().optional(),
+    email: z.string().email().or(z.literal(''))
+  })
+  .merge(uuidSchema)
+
+export const updateTenantIpSchema = z
```
njlie, line 140: nit: `updateTenantIpSchema` -> `updateTenantIdpSchema`

packages/backend/src/tenants/service.ts
```diff
@@ -200,14 +200,12 @@ async function updateOperatorApiSecretFromConfig(
...
-    .first()
 
   if (!tenant) {
     return TenantError.TenantNotFound
   }
   if (tenant.apiSecret !== adminApiSecret) {
-    await Tenant.query(deps.knex)
-      .patch({ apiSecret: adminApiSecret })
-      .where('id', operatorTenantId)
+    await tenant.$query(deps.knex).patch({ apiSecret: adminApiSecret })
```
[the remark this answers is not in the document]
BlairCurrey: We could use `Tenant.query` instead, but I thought this was a bit tighter. Objection handles mutating the object and storing it, and then we set the object objection mutated on the cache.

Alternatively we patch with `Tenant.query` then mutate/form new object ourself, I just felt like that opens a window for divergence. Or, we `patchAndFetch` but I wanted to eliminate that db call.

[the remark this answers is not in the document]
BlairCurrey: For good measure, here is a unit test in objection verifying the mutate behavior

https://github.com/Vincit/objection.js/blob/a1f776a40cc06db118751e87162b095cc3320fd8/tests/integration/patch.js#L511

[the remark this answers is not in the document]
mkurapov: Ah, makes sense, thank you! It wasn't obvious that the `patch` mutated the object in place.

mkurapov (approved):

packages/frontend/app/routes/tenants.$tenantId.tsx
```diff
@@ -237,43 +246,97 @@ export default function ViewTenantPage() {
...
-  const errors: {
-    fieldErrors: ZodFieldErrors<typeof createTenantSchema>
-    message: string[]
+  const actionResponse: {
+    errors: {
+      general: {
+        fieldErrors: ZodFieldErrors<typeof updateTenantGeneralSchema>
+        message: string[]
+      }
+      ip: {
```
njlie, line 255: nit: the idp schema is being dereferenced as "ip" here

BlairCurrey: fixed [f95153b](https://github.com/interledger/rafiki/pull/3328/commits/f95153b128ff1a5b6548447254fe936fee4aeb63)

mkurapov (approved):

PR #3341, opened 2025-03-10
feat(backend): directly use resource server URL to generate URLs for OP

### LINKED ISSUE #3247: Directly use resource server URL to generate URLs for Open Payments resources
## Context
Currently, in order to generate the URLs for Open Payments resources (like incoming payments, quotes, outgoing payments), we generate the URL as such:

```ts
  public getUrl(walletAddress: WalletAddress): string {
    const url = new URL(walletAddress.url)
    return `${url.origin}${Quote.urlPath}/${this.id}`
  }
```

However, this assumes that the wallet address URL origin is equal to the resource server. In a typical implementation, wallet address URL origin = resource server URL, but in the context of Open Payments, it doesn't need to be.

To allow for more flexibility, instead of using wallet address URL origin to `getUrl` on a resouce, we should use the Open Payments resource server URL directly.

## Todos
- [ ] Update the `getUrl` methods on our models to directly use the resource server URL instead of the wallet address URL origin

cozminu:
## Context

fixes #3247 

## Checklist

- [x] Tests added/updated


packages/backend/src/open_payments/payment/incoming/model.ts
```diff
@@ -142,9 +142,8 @@ export class IncomingPayment
     this.receivedAmountValue = amount.value
   }
 
-  public getUrl(walletAddress: WalletAddress): string {
-    const url = new URL(walletAddress.url)
-    return `${url.origin}${IncomingPayment.urlPath}/${this.id}`
+  public getUrl(resourceServerUrl: string): string {
+    return `${resourceServerUrl}${IncomingPayment.urlPath}/${this.id}`
```
mkurapov, line 146: I think would also be good to be careful here if theres an ending slash in the resourceServerUrl (just in case)

mkurapov (commented):
Looks good, just the classic edge case

packages/backend/src/open_payments/payment/incoming/model.ts
cozminu: done to this one and the other 2 (ougoing and quotes)

packages/frontend/package.json
```diff
@@ -26,6 +26,7 @@
     "ilp-packet": "3.1.4-alpha.2",
     "isbot": "^5.1.23",
     "json-canonicalize": "^1.0.6",
+    "next": "^15.2.3",
```
mkurapov, line 29:
```shell
rafiki|main ⇒ pnpm -r why next
Legend: production dependency, optional only, dev only

frontend /Users/max/interledger/rafiki/packages/frontend

dependencies:
@ory/integrations 1.3.1
└── next 15.2.2 peer
```

next is a dependency of @ory/integrations, so instead of adding it here as a direct dependency of the project, we can specify the override for next in `pnpm.overrides` in the root package json, such that we ensure a certain version (in this case, anything above 15.2.3) gets used when resolving dependencies.

mkurapov (approved):

PR #3362, opened 2025-03-30
feat(auth): add event for grant revoked

### LINKED ISSUE #2898: Publish webhook event for grant is revoked
### Context

A requirement from JOPACC (and sensible for all integrators), we need to publish a webhook for when a grant is revoked, such that an integrator, for example, can release funds on hold for a particular user account. This will be the first event that the auth service publishes, so we will need to scaffold the webhook workers in that package as well (publishing to a new ENV variable, WEBHOOK_URL). For now, we just need to publish the grantId in the payload, as well as the time it was revoked.

It is recommended to do this in multiple parts.

dragosp1011:

- Adds webhook workers in Auth service. The webhook url is optional as not all ASE will need to process these events (at least for now). If webhook url is not provided the workers are not started and the events are not saved in db.
- Implements `grant.revoked` event.
- Replaces `trx` with `knex` in some tests where trx was not defined before being used.

## Context

fixes #2898

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/auth/src/accessToken/service.ts
+++ b/packages/auth/src/accessToken/service.ts
@@ -18,7 +18,7 @@ export interface AccessTokenService {
   ): Promise<{ grant: Grant; access: Access[] } | undefined>
   create(grantId: string, trx?: TransactionOrKnex): Promise<AccessToken>
   revoke(id: string, trx?: TransactionOrKnex): Promise<AccessToken | undefined>
-  revokeByGrantId(grantId: string, trx?: TransactionOrKnex): Promise<number>
+  revokeByGrantId(grantId: string, trx?: TransactionOrKnex): Promise<Date>
   rotate(id: string, trx?: TransactionOrKnex): Promise<AccessToken | undefined>
 }
 
@@ -126,12 +126,15 @@ async function revokeByGrantId(
   deps: ServiceDependencies,
   grantId: string,
   trx?: TransactionOrKnex
-): Promise<number> {
-  return await AccessToken.query(trx || deps.knex)
+): Promise<Date> {
+  const revokedAt = new Date()
+  await AccessToken.query(trx || deps.knex)
     .patch({
       revokedAt: new Date()
     })
     .where('grantId', grantId)
+
+  return revokedAt
 }
 async function createAccessToken(
   deps: ServiceDependencies,
--- a/packages/auth/src/accessToken/service.ts
+++ b/packages/auth/src/accessToken/service.ts
@@ -130,7 +130,7 @@ async function revokeByGrantId(
   const revokedAt = new Date()
   await AccessToken.query(trx || deps.knex)
     .patch({
-      revokedAt: new Date()
+      revokedAt
     })
     .where('grantId', grantId)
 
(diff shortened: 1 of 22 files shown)


packages/auth/src/accessToken/service.ts
```diff
@@ -131,12 +131,15 @@ async function revokeByGrantId(
...
-  return await AccessToken.query(trx || deps.knex)
+): Promise<Date> {
+  const revokedAt = new Date()
+  await AccessToken.query(trx || deps.knex)
     .patch({
       revokedAt: new Date()
     })
     .where('grantId', grantId)
+
+  return revokedAt
```
BlairCurrey, line 142: Uses the `revokedAt` variable instead of storing another new `Date`.

```suggestion
): Promise<Date> {
  const revokedAt = new Date()
  await AccessToken.query(trx || deps.knex)
    .patch({
      revokedAt
    })
    .where('grantId', grantId)

  return revokedAt
```

packages/auth/src/app.ts
```diff
@@ -142,6 +144,13 @@ export class App {
       for (let i = 0; i < this.config.databaseCleanupWorkers; i++) {
         process.nextTick(() => this.processDatabaseCleanup())
       }
+
+      // if webhookUrl is not set, webhook events are not saved or processed
+      if (this.config.webhookUrl) {
+        for (let i = 0; i < this.config.webhookWorkers; i++) {
+          process.nextTick(() => this.processWebhook())
+        }
+      }
```
BlairCurrey, line 153: definitely an edge case, but nothing is stopping them from setting a value of 0 for the `webhookWorkers` right? In which case it would just quietly not have any workers (even if url is set). Maybe we handle that somehow? I guess we could check that alongside `webhookUrl`, or maybe error when parsing the config?

dragosp1011: that is correct, but If events are not processed they also should not be stored. I check on saving the webhookUrl as well.

I think a separate env variable (`ENABLE_WEBHOOK`) that will state if the webhook is enabled or not will be best so the decision is more conscientious. All the checks to be depending on it after that.

njlie: Some degree of config parsing might still be necessary, as the scenario where `ENABLE_WEBHOOK` being true and `WEBHOOK_URL` being undefined or `WEBHOOK_WORKERS` being zero could still occur

njlie (approved):

packages/auth/src/webhook/service.ts
```diff
@@ -0,0 +1,237 @@
...
+  pagination?: Pagination
+  filter?: WebhookEventFilter
+  sortOrder?: SortOrder
+}
+
+export interface WebhookService {
+  getEvent(id: string): Promise<WebhookEvent | undefined>
+  getLatestByResourceId(
+    options: WebhookByResourceIdOptions
+  ): Promise<WebhookEvent | undefined>
```
mkurapov, line 31: Since this is unused, we can just remove it

mkurapov:
@dragosp1011 will close this for now - this is something we would like to have in rafiki-v2, however.

PR #3366, opened 2025-04-01
feat(backend): rework fixed send fee calculation

### LINKED ISSUE #3055: Rework Fixed Send Fee Calculation
Change the fixed send fee calculation to happen before creating the quote (`paymentMethodHandlerService.getQuote`) and base on the debit amount (instead of receive amount, as it is currently). This will result in a behavior change (ie different `receiveAmount`) and the `quote/service.test.ts` tests should be updated. Not entirely dependent, but probably best to address after [this PR to add local payments ](https://github.com/interledger/rafiki/pull/2857) is merged. This adds a `debitAmountMinusFees` which is determined during the amount calculations - ensure this field is still populated correctly.

### Context

Currently with fixed send quotes, we get the quote then patch the debit/receive amount based on the fees. For fixed send, we apply the percentage and fixed fee to the `receiveAmount` (after it's been converted by the exchange rate, if any). Then it gets converted by the exchange rate. Given the percentage amount is derived from the already converted `receiveAmount`, applying the exchange rate again messes up the percentage based fee.

### Worked example

| debitAmountValue | basisPointFee | exchangeRate |
|------------------|---------------|--------------|
| 200          |   200 (2%)        | 0.5       |

The current calculation is: 

_Note: initial `receiveAmount` comes from `getQuote`_

$`receiveAmount = 200 \times 0.5`$
$` receiveAmount = receiveAmount - ((receiveAmount \times 0.02) \times 0.5) = 99 `$

The new calculation should be: 

$` 200 - (200 \times 0.02) = 196 `$
_pass into getQuote where 0.5 exchange rate is applied_
$` receiveAmount = `98` `$

### More details on required changes

Basically, we could simply change this:

https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/quote/service.ts#L232

To use debitAmount like this: `const fees = quote.fee?.calculate(quote.debitAmount.value) ?? BigInt(0)`

At time of writing this causes the following tests to fail:

- [2% fee, 0.5 exchange rate, 99n receiveAmount](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/quote/service.test.ts#L659). New, correct `receiveAmount` should be 98n.
- [2% fee, 20 fixed fee,  0.5 exchange rate, 89 receiveAmount](https://github.com/interledger/rafiki/blob/main/packages/backend/src/open_payments/quote/service.test.ts#L660). New, correct `receiveAmount` should be 88n.

This would give us the correct calculation. However, this change is insufficient by itself - we need to move this ahead of creating the quote. We don't need the returned quote to make these calculations so doing it upfront should make the logic a bit clearer.

cozminu:
For fixed debit amounts:
- moved fee calculation before payment quote retrieval (using options.debitAmount)
- subtracted receive amount value from payment quote retrieval
- fee is calculated with the options.debitAmount.value as principal
- added fee back to the final quote to insert in db

For fixed receive amounts:
- moved fee calculation after payment quote retrieval for fixed debit amounts (as quote.debitAmount is needed, rather than the options.debitAmount). This cannot be done before retrieving the payment quote because we don't know the peer's exchange rate in advance.
- fee is calculated with quote.debitAmount.value as principal
- added to the final quote to insert in db

Changed `mockQuote`'s receiveAmount logic a bit because one of the tests were failing:
debitAmountValue | fixedFee | basisPointFee | exchangeRate | expectedReceiveAmountValue
${200n}          | ${20}    | ${200}        | ${0.455}     | ${80n}

Fees: (200 * (200 * 0.0001)) - 20 = 24
Debit - fees: 200 - 24 = 176
Receive: 176 USD * 0.455 = 80.08 XRP (this returned 81 instead of 80 and it should be `Math.floor`'d instead of `Math.ceil`'d because the peer will deliver 80 XRP for 175.82417582 USD)

fixes #3055 

## Checklist

- [ ] Tests added/updated


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -185,10 +189,20 @@ async function createQuote(
...
+            }
+          : undefined
       }
     )
     stopTimerQuote()
 
+    // Calculate fee when receiveAmount is set
+    const fixedReceiveFee = !options.debitAmount
+      ? sendingFee?.calculate(quote.debitAmount.value) ?? 0n
+      : 0n
```
mkurapov, line 204:
```suggestion
    const fixedReceiveFee = options.receiveAmount
      ? sendingFee?.calculate(quote.debitAmount.value) ?? 0n
      : 0n
```

mkurapov: Just to match the comment better

mkurapov: It would also be good to add just a bit in the comment why we need to apply this fee _after_ we got the quote.

PR #3391, opened 2025-04-09
docs: standardize graphql examples

brad-dow:
Each GraphQL example is in a 3-tabbed panel: one for operation, one for arguments, and one for response.

Fixes https://github.com/interledger/rafiki/issues/3090

### DIFF
--- a/packages/documentation/src/content/docs/admin/liquidity/asset-liquidity.mdx
+++ b/packages/documentation/src/content/docs/admin/liquidity/asset-liquidity.mdx
@@ -38,7 +38,7 @@ You can deposit and withdraw asset liquidity through the Rafiki Admin applicatio
     }
   ```
   </TabItem>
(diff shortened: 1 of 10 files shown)


hajjimo (dismissed):

brad-dow:
> @brad-dow I know it wasn't on your list, but can you update the "viewing and revoking grants" page so that both of the arguments have the link to the mutation's input object?

I made this change but the first one was a query, which made it just a little bit different. There's not an input object to point to, so I just linked to the query itself. Also, noticing all of our specs say 'arguments' as well :/ I need to take a GraphQL course

melissahenderson (approved):

melissahenderson:
@brad-dow Yeah, I see that too. I guess we shouldn't be surprised if we have to come back to this in a few months and change them all back to argument 😂

PR #3406, opened 2025-04-19
ci(docker): images should have independent major minor tags

bosbaber:
## Summary
This PR refactors the image building, tagging and pushing github actions so that an appropriate hierarchy of tags are created. 

- Tag independent minor and major docker images on releases
- Refactoring of CI workflow

On releasing of new version v1.1.1
- images created and pushed for the following tags
  + package_name v1.1.1
  + package_name v1.1
  + package_name v1-beta
  
On releasing of new version v1.1.1-beta
- images created and pushed for the following tags
  + package_name v1.1.1-beta
  + package_name v1.1-beta
  + package_name v1-beta
  
  If any of the of the minor or major tags already exist we will simply override them.

## Context
Last week there was a minor bug which caused various issues in testnet. Unfortunately the rafiki charts are currently pinned to a very specific version which meant we had to make changes to both the charts and terraform repositories in order to deploy a new version of the chart to solve the problem.

I had the thought that we should probably pin the appVersion associated with specific Helm charts to the most recent stable *minor* version of our releases. This will allow consumers like testnet to have bug fix patch updates to the most recent minor version. Users that want more control can use the versionOverride feature of the rafiki helm charts to pin to the specific patch version they desire.

There are many environments in which you do not want to pin your application to a specific patch version, but rather to a minor or patch version. Docker images being generated at the moment have exact tags and for this reason many consumers would want to be able to pin to just a minor or major version.

## Checklist
- [ ] Tests added/updated


beniaminmunteanu (dismissed):

.github/actions/image-push/action.yml
```diff
@@ -0,0 +1,67 @@
+name: "Version hierarchy image pusher"
+description: "Will re-tag the docker image to create images for parent versions. For example, if the image is tagged 1.2.3, it will create a tag for 1.2 and 1.0 as well."
+
+inputs:
+  app_name:
+    description: "The name of the application."
+    required: true
+  package:
+    required: true
+  platform_name:
```
mkurapov, line 10: My linter is suggesting to add `description` property to `package` and `platform_name`

.github/actions/parent-versions/action.yml
```diff
@@ -0,0 +1,59 @@
...
+        # build parents
+        minor_label="v${major_num}.${minor_num}"
+        major_label="v${major_num}"
+        if [[ -n "$pre_release" ]]; then
+          minor_label+="-$pre_release"
+          major_label+="-$pre_release"
+        fi
+
+        # emit outputs
+        echo "has_hierarchy=true"     >> $GITHUB_OUTPUT
```
mkurapov, line 57: Do we need `has_hierarchy` if it will always be equal to `generateRelease`? Unless I'm missing something

bosbaber: After finally circling back to this PR your comment prodded me to rethink if the concept of generateRelease even makes sense here. Then that forced me to stare a bug in the face which has now been resolved.

In short, the purpose of this function is to decide if we need an hierarchical version based on the version name. Versions 'nightly' should definitely push, but has no hierarchy. generateRelease makes no sense here so I removed it instead.

I updated functions accordingly and tested pipelines on my private repo.

mkurapov (approved):

PR #3411, opened 2025-04-28
feat(backend): return minSendAmount in quote responses

### LINKED ISSUE #3353: Return minSendAmount in quote responses
### Overview

Currently, if a Quote is requested for a `debitAmount` that is less than 1 unit of the quoted `receiveAmount`, we end up throwing an error:

https://github.com/interledger/rafiki/blob/cbd3a0c90f4853f1d64fe5fb272b2a13347d4eac/packages/backend/src/payment-method/ilp/service.ts#L157-L183

In most situations, this is OK, but this becomes a bit of a challenge for FX transfers, as there is no reliable way to determine what is the minimum debitAmount for a quote that will satisfy at least a minimum one unit of value for the receiving wallet address. This results in having to make a lot of quote requests to figure out the FX.

For example, if a sender’s asset is ZAR and the receiver’s asset is USD, where 1 USD = 20 ZAR:

0.01 ZAR = 0.00… USD (Error)
0.02 ZAR = 0.00… USD (Error)
...
0.20 ZAR = 0.01 USD (OK)

In the example above, the extension will have to make up to 20 quote requests until it reaches an amount that can be represented in 1 unit of USD. 

In order to avoid Open Payments clients having to make these round trips, we can provide a `minSendAmount: { assetScale, assetCode, value }` in the error response that they can subsequently use to create a quote that (should) succeed.

After completing the blocked issues below, we will now be able to add this `minSendAmount` in the `details` to the error object to provide more context about the error.
 

### Blocked by

- [x] https://github.com/interledger/open-payments/issues/573

### LINKED ISSUE #3354: Return proper error object in resource server
### Context

Similar to what was done in #2779 for the Open Payments authorization server, we should also return a proper error object for the resource server, as we are currently only returning a string for an error:

https://github.com/interledger/rafiki/blob/cbd3a0c90f4853f1d64fe5fb272b2a13347d4eac/packages/backend/src/open_payments/route-errors.ts#L35-L47

The error object should be formatted as such:

```
{
        error: {
          code: err.code,
          description: err.message,
          details?: {...}
        }
 }
```

### Blocked by
- [x] https://github.com/interledger/open-payments/issues/573

cozminu:

- solves issue #3353 
- changes the response to be compliant with the new response format

## Context

fixes #3353 
fixes #3354

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/resolvers/quote.ts
+++ b/packages/backend/src/graphql/resolvers/quote.ts
@@ -50,9 +50,9 @@ export const createQuote: MutationResolvers<ApolloContext>['createQuote'] =
       options.receiveAmount = args.input.receiveAmount
     const quoteOrError = await quoteService.create(options)
     if (isQuoteError(quoteOrError)) {
-      throw new GraphQLError(errorToMessage[quoteOrError], {
+      throw new GraphQLError(errorToMessage[quoteOrError.type], {
         extensions: {
-          code: errorToCode[quoteOrError]
+          code: errorToCode[quoteOrError.type]
         }
       })
     } else
--- a/packages/backend/src/graphql/resolvers/quote.ts
+++ b/packages/backend/src/graphql/resolvers/quote.ts
@@ -52,7 +52,8 @@ export const createQuote: MutationResolvers<ApolloContext>['createQuote'] =
     if (isQuoteError(quoteOrError)) {
       throw new GraphQLError(errorToMessage[quoteOrError.type], {
         extensions: {
-          code: errorToCode[quoteOrError.type]
+          code: errorToCode[quoteOrError.type],
+          details: quoteOrError.details
         }
       })
     } else
--- a/packages/backend/src/open_payments/quote/service.ts
+++ b/packages/backend/src/open_payments/quote/service.ts
@@ -224,10 +224,13 @@ async function createQuote(
           }
           const quoteMinSendAmount = err.details.minSendAmount as bigint
           const fixedFee = sendingFee?.fixedFee ?? 0n
+          const feePercentage = (sendingFee?.basisPointFee ?? 0) / 10_000
 
-          details.minSendAmount.value =
-            quoteMinSendAmount +
-            (sendingFee?.calculate(fixedFee + quoteMinSendAmount) ?? 0n)
+          details.minSendAmount.value = BigInt(
+            Math.ceil(
+              Number(quoteMinSendAmount + fixedFee) / (1 - feePercentage)
+            )
+          )
         }
 
         stopTimer()
--- a/packages/backend/src/open_payments/quote/service.ts
+++ b/packages/backend/src/open_payments/quote/service.ts
@@ -223,13 +223,10 @@ async function createQuote(
             }
           }
           const quoteMinSendAmount = err.details.minSendAmount as bigint
-          const fixedFee = sendingFee?.fixedFee ?? 0n
-          const feePercentage = (sendingFee?.basisPointFee ?? 0) / 10_000
 
-          details.minSendAmount.value = BigInt(
-            Math.ceil(
-              Number(quoteMinSendAmount + fixedFee) / (1 - feePercentage)
-            )
+          details.minSendAmount.value = calculateMinSendAmount(
+            quoteMinSendAmount,
+            sendingFee
           )
         }
 
@@ -511,4 +508,23 @@ async function getWalletAddressPage(
     )
   }
   return quotes
+}
+
+function calculateMinSendAmount(
+  quoteMinSendAmount: bigint,
+  sendingFee: Fee | undefined
+): bigint {
+  if (!sendingFee) {
+    return quoteMinSendAmount
+  }
+  const fixedFee = sendingFee.fixedFee ?? 0n
+  // if the fee is 0%, the invertedPercentageFee is 1
+  // if the fee is 100%, the invertedPercentageFee is 0, which is not allowed
+  // We make sure the fee is between 0 and 1
+  const invertedPercentageFee =
+    1 - Math.min(Math.max((sendingFee.basisPointFee ?? 0) / 10_000, 0), 1)
+
+  return BigInt(
+    Math.ceil(Number(quoteMinSendAmount + fixedFee) / invertedPercentageFee)
+  )
 }
--- a/packages/backend/src/open_payments/quote/service.ts
+++ b/packages/backend/src/open_payments/quote/service.ts
@@ -1,7 +1,7 @@
 import { TransactionOrKnex } from 'objection'
 
 import { BaseService } from '../../shared/baseService'
-import { QuoteError, isQuoteError } from './errors'
+import { QuoteError, QuoteErrorType, isQuoteError } from './errors'
 import { Quote } from './model'
 import { Amount } from '../amount'
 import { ReceiverService } from '../receiver/service'
@@ -112,18 +112,18 @@ async function createQuote(
   })
   if (options.debitAmount && options.receiveAmount) {
     stopTimer()
-    return QuoteError.InvalidAmount
+    return new QuoteError(QuoteErrorType.InvalidAmount)
   }
   const walletAddress = await deps.walletAddressService.get(
     options.walletAddressId
   )
   if (!walletAddress) {
     stopTimer()
-    return QuoteError.UnknownWalletAddress
+    return new QuoteError(QuoteErrorType.UnknownWalletAddress)
   }
   if (!walletAddress.isActive) {
     stopTimer()
-    return QuoteError.InactiveWalletAddress
+    return new QuoteError(QuoteErrorType.InactiveWalletAddress)
   }
   if (options.debitAmount) {
     if (
@@ -132,13 +132,13 @@ async function createQuote(
       options.debitAmount.assetScale !== walletAddress.asset.scale
     ) {
       stopTimer()
-      return QuoteError.InvalidAmount
+      return new QuoteError(QuoteErrorType.InvalidAmount)
     }
   }
   if (options.receiveAmount) {
     if (options.receiveAmount.value <= BigInt(0)) {
       stopTimer()
-      return QuoteError.InvalidAmount
+      return new QuoteError(QuoteErrorType.InvalidAmount)
     }
   }
 
@@ -264,7 +264,10 @@ async function createQuote(
       err instanceof PaymentMethodHandlerError &&
       err.code === PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount
     ) {
-      return QuoteError.NonPositiveReceiveAmount
+      return new QuoteError(
+        QuoteErrorType.NonPositiveReceiveAmount,
+        err.details
+      )
     }
 
     deps.logger.error({ err }, 'error creating a quote')
@@ -284,20 +287,20 @@ export async function resolveReceiver(
       { receiver: options.receiver },
       'Could not create quote. Receiver not found'
     )
-    throw QuoteError.InvalidReceiver
+    throw new QuoteError(QuoteErrorType.InvalidReceiver)
   }
   if (options.receiveAmount) {
     if (
       options.receiveAmount.assetScale !== receiver.assetScale ||
       options.receiveAmount.assetCode !== receiver.assetCode
     ) {
-      throw QuoteError.InvalidAmount
+      throw new QuoteError(QuoteErrorType.InvalidAmount)
     }
     if (receiver.incomingAmount && receiver.receivedAmount) {
       const receivingPaymentValue =
         receiver.incomingAmount.value - receiver.receivedAmount.value
       if (receivingPaymentValue < options.receiveAmount.value) {
-        throw QuoteError.InvalidAmount
+        throw new QuoteError(QuoteErrorType.InvalidAmount)
       }
     }
   } else if (!options.debitAmount && !receiver.incomingAmount) {
@@ -308,7 +311,7 @@ export async function resolveReceiver(
       },
       'Could not create quote. debitAmount or incomingAmount required.'
     )
-    throw QuoteError.InvalidReceiver
+    throw new QuoteError(QuoteErrorType.InvalidReceiver)
   }
   return receiver
 }
@@ -339,11 +342,11 @@ function calculateFixedSendQuoteAmounts(
       { fees, estimatedExchangeRate, receiveAmountValue },
       'Negative receive amount when calculating quote amount'
     )
-    throw QuoteError.NonPositiveReceiveAmount
+    throw new QuoteError(QuoteErrorType.NonPositiveReceiveAmount)
   }
 
   if (receiveAmountValue > maxReceiveAmountValue) {
-    throw QuoteError.InvalidAmount
+    throw new QuoteError(QuoteErrorType.InvalidAmount)
   }
 
   deps.logger.debug(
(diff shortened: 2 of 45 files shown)


packages/backend/src/open_payments/route-errors.ts
```diff
@@ -43,7 +43,16 @@ export async function openPaymentsServerErrorMiddleware(
         },
         'Received error when handling Open Payments request'
       )
-      ctx.throw(err.status, err.message)
+
+      ctx.status = err.status
+      ctx.body = {
```
mkurapov, line 48: @oana-lolea 

This is the change for https://github.com/interledger/rafiki/issues/3354, so @cozminu can just make the change as part of this PR.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -184,21 +193,48 @@ async function createQuote(
...
+          const quoteMinSendAmount = err.details.minSendAmount as bigint
+          const fixedFee = sendingFee?.fixedFee ?? 0n
+
+          details.minSendAmount.value =
+            quoteMinSendAmount +
+            (sendingFee?.calculate(fixedFee + quoteMinSendAmount) ?? 0n)
+        }
+
+        stopTimer()
+        return new QuoteError(QuoteErrorCode.NonPositiveReceiveAmount, details)
```
mkurapov, line 234: Usually, throwing would be more relevant to `Error` , but given how we're always returning a value instead of throwing for these service methods, I think it's OK. 

Curious to see if anyone else has a strong preference on this.

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -184,21 +193,48 @@ async function createQuote(
...
+              assetCode: walletAddress.asset.code,
+              assetScale: walletAddress.asset.scale
+            }
+          }
+          const quoteMinSendAmount = err.details.minSendAmount as bigint
+          const fixedFee = sendingFee?.fixedFee ?? 0n
+
+          details.minSendAmount.value =
+            quoteMinSendAmount +
+            (sendingFee?.calculate(fixedFee + quoteMinSendAmount) ?? 0n)
```
mkurapov, line 230: Since `calculate()` already adds the `fixedFee` in the method
```suggestion
            (sendingFee?.calculate(quoteMinSendAmount) ?? 0n)
```

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -113,6 +113,26 @@ async function getQuote(
         options.receiveAmount?.value || options.receiver.incomingAmount?.value
     }
 
+    // Probing quote to get the exchange rate
```
mkurapov, line 116:
```suggestion
    // Probing quote to get the exchange rate even if amountToSend is less than 0 because of fees
```
or something like this, just to explain why it might be 0

mkurapov (commented):
Main idea looks good, just a few questions/suggestions

packages/backend/src/open_payments/route-errors.ts
cozminu: updated comment with ``fixes #3354`

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -184,21 +193,48 @@ async function createQuote(
         description: 'Time to getQuote'
       }
     )
-    const quote = await deps.paymentMethodHandlerService.getQuote(
-      paymentMethod,
-      {
+    let quote: PaymentQuote
+    try {
+      quote = await deps.paymentMethodHandlerService.getQuote(paymentMethod, {
```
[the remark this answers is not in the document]
cozminu: made changes and wrote test

packages/backend/src/rates/util.ts
```diff
@@ -28,7 +28,7 @@ export function convertSource(opts: ConvertSourceOptions): ConvertResults {
   const scaledExchangeRate = opts.exchangeRate * 10 ** scaleDiff
 
   return {
-    amount: BigInt(Math.round(Number(opts.sourceAmount) * scaledExchangeRate)),
+    amount: BigInt(Math.floor(Number(opts.sourceAmount) * scaledExchangeRate)),
```
BlairCurrey, line 31: My intuition is that `floor` would make more sense than `round`. Especially when the opposite direction (`convertDestination`) does `Math.ceil`. I'm trying to figure out the original reasoning. @cozminu can you elaborate on your thought process here?

@mkurapov I vaguely recall talking about the `round` behavior with you when making some related changes here https://github.com/interledger/rafiki/pull/2857. I see this comment you left when originally adding it:

>BigInt divided by BigInt ends up truncating and not rounding the result. I think for currency, rounding up or down to the nearest unit makes more sense (and is standard). We can't really get around using number, since we are dealing with decimals for the exchange rate

https://github.com/interledger/rafiki/pull/1527#discussion_r1245354427

If we do `floor` then we're going to need to update the integration test  that started failing here https://github.com/interledger/rafiki/actions/runs/14999236719/job/42141498858

packages/backend/src/rates/util.test.ts
```diff
@@ -6,10 +6,10 @@ describe('Rates util', () => {
       test.each`
         exchangeRate | sourceAmount | assetScale | expectedResult                                    | description
         ${1.5}       | ${100n}      | ${9}       | ${{ amount: 150n, scaledExchangeRate: 1.5 }}      | ${'exchange rate above 1'}
-        ${1.1602}    | ${12345n}    | ${2}       | ${{ amount: 14323n, scaledExchangeRate: 1.1602 }} | ${'exchange rate above 1 with rounding up'}
+        ${1.1602}    | ${12345n}    | ${2}       | ${{ amount: 14322n, scaledExchangeRate: 1.1602 }} | ${'exchange rate above 1 with rounding up'}
```
BlairCurrey, line 9: small thing, but we should probably change the descriptions here if we go with `Math.floor`. the "with rounding up" and "with rounding down" seem to clarify how the `Math.round` was behaving.

packages/backend/src/rates/util.ts
cozminu: Given a 19:1 conversion ratio for MXN to USD with the same scale of 2:
Wallet A in MXN and Wallet B in USD

If `Math.round` is used and Wallet A sends 29 MXN, then Wallet B will receive 2 USD and this will result in a 14.5:1 conversion ratio for that trade. This means the ASE implementing Rafiki will support the missing 10MXN from the real conversion. 
Sending 29 MXN to 37 MXN will result in a financial loss for the ASE.
Sending 38 MXN will result in a fair trade.
Sending 39 MXN to 47 MXN will result in a financial gain for the ASE.
If there’s no sending fee applied and a user sends minimum required to exploit Math.round (29 MXN to USD) from a wallet to another 1000s of times, it can practically gain money due to favorable conversion rate.

If `Math.floor` is used, Wallet A needs to send >= 38 and <= 56 MXN to convert in 2 USD, removing the chance of a transaction to be a loss for the ASE.

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -113,6 +113,24 @@ async function getQuote(
...
+        amountToSend: 1n,
+        slippage: deps.config.slippage,
+        prices: convertRatesToIlpPrices(rates)
+      })
+      throw new PaymentMethodHandlerError('Received error during ILP quoting', {
+        description: 'Minimum delivery amount of ILP quote is non-positive',
+        retryable: false,
+        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount,
+        details: {
+          minSendAmount: BigInt(ilpQuote.highEstimatedExchangeRate.b.valueOf())
```
mkurapov:
```suggestion
          minSendAmount: Math.ceil(ilpQuote.highEstimatedExchangeRate.reciprocal().valueOf())
```
otherwise, we are only using the denominator

packages/backend/src/graphql/resolvers/quote.ts
```diff
@@ -50,9 +50,9 @@ export const createQuote: MutationResolvers<ApolloContext>['createQuote'] =
       options.receiveAmount = args.input.receiveAmount
     const quoteOrError = await quoteService.create(options)
     if (isQuoteError(quoteOrError)) {
-      throw new GraphQLError(errorToMessage[quoteOrError], {
+      throw new GraphQLError(errorToMessage[quoteOrError.type], {
         extensions: {
-          code: errorToCode[quoteOrError]
+          code: errorToCode[quoteOrError.type]
         }
```
mkurapov: We should also add the `quoteOrError.details` information in the `extensions` object. This way we can have the Admin API be able to read `minSendAmount` as well

packages/backend/src/open_payments/quote/service.ts
```diff
@@ -481,3 +509,22 @@ async function getWalletAddressPage(
...
+function calculateMinSendAmount(
+  quoteMinSendAmount: bigint,
+  sendingFee: Fee | undefined
+): bigint {
+  if (!sendingFee) {
+    return quoteMinSendAmount
+  }
+  const fixedFee = sendingFee.fixedFee ?? 0n
+  // if the fee is 0%, the invertedPercentageFee is 1
+  // if the fee is 100%, the invertedPercentageFee is 0, which is not allowed
```
mkurapov, line 522: "Technically", it's possible that the fee _could_ be more than 100%. Instead of having an `invertedPercentageFee`, I believe we instead can do something like 
`minSendAmount = quoteMinSendAmount + ceil((quoteMinSendAmount + fixedFee) * feePercentage) + fixedFee`

packages/backend/src/payment-method/ilp/service.ts
mkurapov: We should also include returning `minSendAmount` in the `if (ilpQuote.minDeliveryAmount <= BigInt(0))` condition

packages/backend/src/open_payments/quote/service.ts
cozminu: The equation used by me resulted from the way we apply fees in Fee.calculate:
`Fee = (principal * feePercentage) + fixedFee`

I have tried your equation 
`minSendAmount = quoteMinSendAmount + ceil((quoteMinSendAmount + fixedFee) * feePercentage) + fixedFee`
with the following values:
quoteMinSendAmount = 100
fixedFee = 50
feePercentage = 0.2
results: minSendAmount = 180

When calculating quoteMinSendAmount which has the formula of:
`quoteMinSendAmount = minSendAmount - Fee`
`quoteMinSendAmount = minSendAmount - ((principal * feePercentage) + fixedFee)`
100 = 180 - (180 * 0.2 + 50)
100 = 180 - 86 => 100 != 94

Taking more than 100% would work only in the case of fixed receive amount because you can add whatever amount you want to the debit, but not in the case of fixed debit because receive amount can only be lower than debit if it makes sense...

mkurapov: Ah, yes, tripped myself up 😄 the percentage fee of course can't be more than 100%.

mkurapov: Following this now.

`x - (percentageFee * x) - fixedFee = quoteMinSendAmount`

quoteMinSendAmount = 100
fixedFee = 50
feePercentage = 0.2

x - 0.2x - 50 = 100
0.8x - 50 = 100
0.8x = 150
x = 150 / 0.8 
ie `(quoteMinSendAmount + fixedFee) / invertedPercentageFee)`
x = 187.5

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -178,7 +198,10 @@ async function getQuote(
       throw new PaymentMethodHandlerError('Received error during ILP quoting', {
         description: errorDescription,
         retryable: false,
-        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount
+        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount,
+        details: {
+          minSendAmount: BigInt(ilpQuote.highEstimatedExchangeRate.b.valueOf())
```
mkurapov:
```suggestion
          minSendAmount: BigInt(Math.ceil(ilpQuote.highEstimatedExchangeRate.reciprocal().valueOf()))
```

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -113,6 +113,26 @@ async function getQuote(
...
+      })
+      throw new PaymentMethodHandlerError('Received error during ILP quoting', {
+        description: 'Minimum delivery amount of ILP quote is non-positive',
+        retryable: false,
+        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount,
+        details: {
+          minSendAmount: BigInt(
+            Math.ceil(ilpQuote.highEstimatedExchangeRate.reciprocal().valueOf())
+          )
+        }
```
mkurapov, line 132: let's also add this to L174 (`if (ilpQuote.minDeliveryAmount <= BigInt(0))` condition)

packages/backend/src/payment-method/ilp/service.ts
cozminu: oh, right. thank you for spotting! fixed

packages/backend/src/payment-method/ilp/service.ts
cozminu: done

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -150,15 +170,20 @@ async function getQuote(
...
       })
     }
 
     // Because of how it does rounding, the Pay library allows getting a quote for a
     // maxSourceAmount that won't be able to fulfill even a single unit of the receiving asset.
     // e.g. if maxSourceAmount is 4 and the high estimated exchange rate is 0.2, 4 * 0.2 = 0.8
     // where 0.8 < 1, meaning the payment for this quote won't be able to deliver a single unit of value,
-    // even with the most favourable exchange rate. We throw here since we don't want any payments
+    // even with the most favorable exchange rate. We throw here since we don't want any payments
```
mkurapov, line 186: American English autocorrect? 😄

packages/backend/src/payment-method/ilp/service.ts
mkurapov: Let's add a test for this case as well

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -150,15 +170,20 @@ async function getQuote(
       throw new PaymentMethodHandlerError('Received error during ILP quoting', {
         description: 'Minimum delivery amount of ILP quote is non-positive',
         retryable: false,
-        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount
+        code: PaymentMethodHandlerErrorCode.QuoteNonPositiveReceiveAmount,
+        details: {
+          minSendAmount: BigInt(
+            Math.ceil(ilpQuote.highEstimatedExchangeRate.reciprocal().valueOf())
+          )
+        }
```
mkurapov, line 178: Let's add a test for this case as well

cozminu: done

packages/backend/src/payment-method/ilp/service.ts
cozminu: yes.. sorry about that. Installed the British version to avoid this in the future. Btw, do we adhere to a specific english?

mkurapov: No it's totally fine 😄 as long as its (one of) the correct spellings & the comment is understandable. Don't worry about changing anything, it was just my Canadian coming out :)

Issue #3373, opened 2025-04-03
Use `outgoingPaymentGrantSpentAmounts` table during outgoing payment creation


mkurapov:
Currently, when an outgoing payment is being created under a particular grant, we need to make sure that the outgoing payment is within the particular limits of the grant (and the grant's interval).

Instead of doing what we are currently doing: grabbing all of the outgoing payments under that grant and calculating the limits in memory, we can leverage the `outgoingPaymentGrantSpentAmounts` table created in [RAF-1032](https://linear.app/interledger/issue/RAF-1032/create-outgoingpaymentgrantspentamounts-table).

When an outgoing payment is being created we need to:

1. Fetch the existing `outgoingPaymentGrant`, and at the same time, grab the most recent `outgoingPaymentGrantSpentAmounts` row:
   1. When fetching the `outgoingPaymentGrants` we need to use the FOR NO KEY UPDATE lock to prevent any concurrency issues
   2. To fetch the most recent `outgoingPaymentGrantSpentAmounts` row:
      1. `SELECT * FROM “outgoingPaymentGrantSpentAmounts” WHERE grantId = {grantId} ORDER BY createdAt DESC LIMIT 1`.

After we fetch both, in order to make this a backwards compatible change (since we will have existing grants that will *not* end up having any outgoingPaymentGrantSpentAmounts records yet), we will need to do the following (pseudocode):

```javascript
  if (outgoingPaymentGrantRecord && !outgoingPaymentGrantSpentAmountsRecord) {
    // in this case, an existing grant is being requested. We haven't stored any records
    // for it in outgoingPaymentGrantSpentAmountsRecord, so we need to use the old calculation 
    // in `validateGrantAndAddSpentAmountsToPayment` to generate the first values we will insert
    // in the outgoingPaymentGrantSpentAmounts table:

    // 1. upsert outgoingPaymentGrant
    // 2. Determine new cumulative grant amounts using existing validateGrantAndAddSpentAmountsToPayment logic
    // 3. Fail if payment exceeds grant or interval limits
    // 4. insert new outgoingPaymentGrantSpentAmounts record

  } else {
    // 1. upsert outgoingPaymentGrant
    // 2. Determine new cumulative grant amounts (use existing outgoingPaymentGrantSpentAmountsRecord as starting point if present)
    // 3. Fail if payment exceeds grant or interval limits
    // 4. insert new outgoingPaymentGrantSpentAmounts record
  }
```

PR #3412, opened 2025-04-28
feat: calculate grant spent amounts from new table

BlairCurrey:

- changes `create` method on outgoing payment service to use new grant spent amount tables to get the spent amounts. Falls back to existing method of summing all payments for grants that pre-exist this change.

## Context

fixes: #3373 

## Checklist

- [ ] Tests added/updated


packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -239,6 +243,11 @@ async function cancelOutgoingPayment(
   })
 }
 
+// TODO: Basically does all the logic in the validate spent amounts.
+// This felt more natural/easier to implement to me. Than how the task describes splitting
+// the paths in thre create function and only calling validate for the legacy path.
+// Not sure if it introduces any edges cases or jives as well with future issues (failed
+// payments for example).
 async function createOutgoingPayment(
```
BlairCurrey, line 251: Note that this does not implement the logic quite how the issue specified. Had some issues when implementing that and found doing everything in the validate function to be much easier (otherwise I think we have to duplicate a bunch of logic there or chop it up into pieces for reuse - kind of a more tedious refactor either way).

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -723,12 +729,43 @@ describe('OutgoingPaymentService', (): void => {
...
+                receiveAmountScale: debitAmount.assetScale,
+                paymentReceiveAmountValue: adjustedReceiveAmountValue,
+                grantTotalReceiveAmountValue:
+                  adjustedReceiveAmountValue * BigInt(i + 1),
+                intervalDebitAmountValue: null,
+                intervalReceiveAmountValue: null,
+                intervalStart: null,
+                intervalEnd: null,
+                paymentState: 'FUNDING'
+              })
```
BlairCurrey, line 768: the grant spent amounts are kind of an implementation detail. at the end of the day its grantSpent[Receive|Debit]Amounts which matter, which is already tested. IDK, left for now but considered not including this.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -498,20 +505,48 @@ function validateAccessLimits(
...
+  }
+
+  const createdTime = payment.createdAt.getTime()
+  const start = interval.start?.toMillis() ?? -Infinity
+  const end = interval.end?.toMillis() ?? Infinity
+
+  if (createdTime < start) {
+    return 'next'
+  } else if (createdTime >= end) {
+    return 'previous'
```
mkurapov: Double checking this: given an interval [start, end]:
if createdAt < start: would this be in the previous interval? (createdAt equal to start would put us in the current interval)
Conversely, if createdTime > end, we would be in the next interval, right? (maybe if createdAt is equal to end, we count it as being part of the current interval)

packages/backend/src/open_payments/payment/outgoing/errors.ts
```diff
@@ -79,7 +83,9 @@ export const errorToMessage: {
   [OutgoingPaymentError.InactiveWalletAddress]: 'inactive wallet address',
   [OutgoingPaymentError.InvalidAmount]: 'invalid amount',
   [OutgoingPaymentError.NegativeReceiveAmount]: 'negative receive amount',
-  [OutgoingPaymentError.InvalidReceiver]: 'invalid receiver'
+  [OutgoingPaymentError.InvalidReceiver]: 'invalid receiver',
+  [OutgoingPaymentError.InvalidGrantSpentAmountState]:
+    'invalid grant spent amounts'
```
mkurapov: I think we should differentiate between errors served up to an Open Payments client vs those that are just on the Admin API (since this error would not be necessarily useful to an OP client)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: I will revisit this. The refactor of this function in general from returning a boolean to the status. I was thinking I'd need more resolution based off the general logic we discussed but Im not sure I ended up using it.

packages/backend/src/open_payments/payment/outgoing/errors.ts
BlairCurrey: OK I think I've sidestepped this by making it more typesafe here: [934e9f6](https://github.com/interledger/rafiki/pull/3412/commits/934e9f6240f2b65e9b7404a4c67da1e09e7163dd) Yay, [make the invalid state unrepresentable](https://www.awwsmm.com/blog/make-invalid-states-unrepresentable).

Maybe not important anymore, but I'm not exactly following - how would open payments client not care about this but admin api would (given it could happen in either open payments api/admin api)?

It _was_ a bad state (500/internal server error) so it's true that there is nothing the client could really do to recover. But not sure what the difference between the two APIs would be. Again, it's a moot point because it no longer throws the error.

packages/backend/src/open_payments/payment/outgoing/service.ts
mkurapov: >  I was thinking I'd need more resolution based off the general logic we discussed but Im not sure I ended up using it.

I think it will be useful (you will need the three cases) for calculating the spent amounts on outgoing payment failure (RAF-1036), so its good to have. We just need to flip the conditions.

packages/backend/src/open_payments/payment/outgoing/errors.ts
mkurapov: > OK I think I've sidestepped this by making it more typesafe here: 934e9f6 Yay, make the invalid state unrepresentable.
Getting rid of the error entirely works great :) 

My original thinking was to separate `errorToMessage`, and have an `errorToGraphQLMessage` vs `errorToOpenPaymentsMessage` (something like this) such that we dont show "invalid grant spent amounts" to the OP client, and maybe something more generic instead. 
As an aside, we would never end up throwing the `InvalidGrantSpentAmountState` error in the Admin API since there is no grant associated when its a first-party, Admin API payment.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -498,20 +504,32 @@ function validateAccessLimits(
   }
 }
 
-function validatePaymentInterval({
+type IntervalClassification = 'previous' | 'current' | 'next' | 'unrestricted'
+
+function classifyPaymentInterval({
```
mkurapov: Maybe let's have a few unit tests for this fn

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -33,7 +33,8 @@ import {
   OutgoingPaymentState,
```
mkurapov, line 33: I think most of the`validateGrant` tests that check for success and failure cases that have an `interval` amount set. We should also add a test for success and failure cases without an `interval` amount.

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: OK I tried to make this more clear. I switched/prev next. also renamed to `getPaymentIntervalStatus` and removed the `unrestricted` label and just returned null in this case where there is no interval. I think the semantics are much better overall now. [5f9630e](https://github.com/interledger/rafiki/pull/3412/commits/5f9630e5e89b5669c4a0aebcf821e7f2041deccd)

packages/backend/src/open_payments/payment/outgoing/service.test.ts
BlairCurrey: I added `withInterval` as a factor in all the test in `validateGrant` that had a hardcoded interval  [6a11e97](https://github.com/interledger/rafiki/pull/3412/commits/6a11e97c9aed3d36bc7ae2061a1ade6e60f2cecd)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: added [c42fc44](https://github.com/interledger/rafiki/pull/3412/commits/c42fc443404b9a6bdc2f1b64553c6891409fdabc)

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -728,12 +732,12 @@ async function validateGrantAndAddSpentAmountsToPayment(
         intervalSent: {
           assetCode: startingSpendAmounts.debitAmountCode,
           assetScale: startingSpendAmounts.debitAmountScale,
-          value: intervalStatus === 'next' ? 0n : intervalDebitValue
+          value: intervalStatus === 'previous' ? 0n : intervalDebitValue
```
mkurapov:
```suggestion
          value: intervalStatus === IntervalStatus.Previous ? 0n : intervalDebitValue
```
same below

mkurapov (approved):

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -1897,4 +2339,57 @@ describe('OutgoingPaymentService', (): void => {
...
+        payment: { createdAt: new Date('2022-08-15T10:00:00Z') },
+        expected: IntervalStatus.Next
+      },
+      {
+        description: 'returns next when payment is before interval start',
+        interval: '2022-07-01T13:00:00Z/P1M',
+        payment: { createdAt: new Date('2022-06-30T12:59:59Z') },
+        expected: IntervalStatus.Previous
+      }
+    ]
```
mkurapov:
```suggestion
      {
        description: 'returns previous when payment is before interval start',
        interval: '2022-07-01T13:00:00Z/P1M',
        payment: { createdAt: new Date('2022-06-30T12:59:59Z') },
        expected: IntervalStatus.Previous
      }
    ]
```
Ditto above

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -566,48 +574,109 @@ async function validateGrantAndAddSpentAmountsToPayment(
...
-    const asset = await deps.assetService.get(grantPayment.quote.assetId)
-    if (asset) grantPayment.quote.asset = asset
-
-    if (
-      validatePaymentInterval({
-        limits: paymentLimits,
-        payment: grantPayment
+
+  const outgoingPaymentGrantSpentAmounts =
+    OutgoingPaymentGrantSpentAmounts.create({
```
mkurapov, line 631: This just creates the `OutgoingPaymentGrantSpentAmounts` correct? Meaning, no DB call?

mkurapov (approved):
Looks good! Just needs some formatting

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: correct. it gets inserted at the end. I found it was easier to initialize like this type-wise than trying to fenagle something like `ModelObject<OutgoingPaymentGrantSpentAmounts>` ([ModelObject](https://github.com/Vincit/objection.js/blob/a7784ded683c62b6dd857fcfd937b83b72b7a183/typings/objection/index.d.ts#L197)) which largely worked but needed some further massaging in places.

packages/backend/src/open_payments/payment/outgoing/errors.ts
deanmchugh: Updated the `#post` describe block to use `a_string` throughout — `service`, `id_token`, and `roles_token` are all randomised now. In the headers test I extracted `id_token` and `roles_token` to local variables so the assertions still reference the same values. Also applied the same fix to the `let(:platform_tokens)` in `payments_adapter/request_spec.rb` which had the same pattern.

PR #3416, opened 2025-05-01
docs: fixes #3382

### LINKED ISSUE #3382: docs: include more info about clearing/settlement
Rafiki doc feedback:

> Include more details around clearing/settlement. Rafiki is at the clearing layer and is only concerned with clearing. It keeps track of how much liquidity exists between assets, peers, and payments themselves. The settlement method (eg ACH, wire) must be decided upon by peer nodes/by your peering relationships. Within Rafiki, you define the line of credit given to a particular peer (basically, how many payments/ILP packets can be sent/cleared between your instances). Then, it's up to you and your peer to settle. The wallet liquidity itself is managed by the integrator backend. For example, if an incoming payment completes (for a particular wallet address), then the integrator/ASE backend would credit that user account on their own system, however that might look like.

We have most, if not all, of this information already, but it's spread out across multiple pages. Let's see if we can make a new Concepts page for the topic.

melissahenderson:
Adds a page for Clearing and Settlement

### DIFF
--- a/packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
+++ b/packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
@@ -6,13 +6,15 @@ title: Clearing and settlement
 
 When a payment is made over traditional banking rails, the money doesn’t move instantly. First, there are checks to confirm that the money exists and can be transferred. Clearing networks are responsible for exchanging messages between ASEs to facilitate these checks. This process is called clearing. When a payment successfully clears, it means the payer’s ASE has an obligation to the payee’s ASE.
 
-The [Interledger Protocol (ILP)](/overview/concepts/interledger) is not a clearing network. It’s a messaging protocol that shares some similarities with traditional clearing networks.
+The [Interledger Protocol (ILP)](/overview/concepts/interledger) is not a traditional clearing network, but does function in a similar way.
 
-- ASEs must peer with one another before they can send and receive ILP packets. This is comparable to traditional banking, where ASEs must use the same clearing network to transact with one another.
-- ILP packets are akin to the messages exchanged during the clearing process.
-- The exchange of ILP Prepare and Fulfill packets creates obligations between peers. A fulfilled ILP packet is basically a promise to pay.
+- ASEs that implement the protocol must become [peers](/integration/requirements/peers) to transact with one another. This is comparable to traditional banking, where ASEs must use the same clearing network. An ASE can't use Interledger to transact with another ASE unless they have both implemented the protocol and have peered with one another.
+- Peered ASEs exchange ILP packets, which are packets of value that contain transaction information. ILP packets are akin to the messages exchanged during the traditional clearing process.
+- The successful exchange of ILP packets between peers creates obligations between them that must be settled. The receipt of a fulfilled ILP packet is basically a conditional IOU&mdash;a promise to pay&mdash;that affects the financial accounting balances between the peers.
 
-Conceptually, Rafiki sits at the clearing level, but isn't a clearing network. Rafiki uses ILP to [track liquidity](/overview/concepts/accounting) between assets, payments, and peers. An ASE must still connect Rafiki to their existing backend system and internal ledger for authentication, fetching exchange rates, and managing liquidity itself. For example, if an incoming payment completes in Rafiki, the ASE’s backend must credit the recipient’s account on their own system, however that might look.
+You can read more about clearing as it relates to ILP in the [Interledger developer docs](https://interledger.org/developers/rfcs/peering-clearing-settling/).
+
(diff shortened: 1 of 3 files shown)


packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Clearing and settlement
+---
+
```
brad-dow, line 4: I know we did something similar for Open Payments with the TL;DR-esque content. What if we just had a one-liner for "what is clearing and what is settlement"? We don't have to, just an idea if someone came to the page for a quick refresher.

brad-dow (dismissed):

packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
```diff
@@ -0,0 +1,27 @@
...
+## Clearing
+
+When a payment is made over traditional banking rails, the money doesn’t move instantly. First, there are checks to confirm that the money exists and can be transferred. Clearing networks are responsible for exchanging messages between ASEs to facilitate these checks. This process is called clearing. When a payment successfully clears, it means the payer’s ASE has an obligation to the payee’s ASE.
+
+The [Interledger Protocol (ILP)](/overview/concepts/interledger) is not a clearing network. It’s a messaging protocol that shares some similarities with traditional clearing networks.
```
hajjimo, line 9: ILP is not a clearing network, but ASEs implementing the protocol and peering with one another become part of the Interledger network, which seems to function as a clearing network. If the public Interledger network is not a clearing network, then I think we need to add more details on what characteristics are missing that would have qualified it as such.

packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+Conceptually, Rafiki sits at the clearing level, but isn't a clearing network. Rafiki uses ILP to [track liquidity](/overview/concepts/accounting) between assets, payments, and peers. An ASE must still connect Rafiki to their existing backend system and internal ledger for authentication, fetching exchange rates, and managing liquidity itself. For example, if an incoming payment completes in Rafiki, the ASE’s backend must credit the recipient’s account on their own system, however that might look.
+
+In any case, no movement of actual money has occurred yet.
+
+## Settlement
+
+In traditional banking, settlement is the fulfillment of an obligation between ASEs. It turns the promise of payment into a real payment by moving actual money. This occurs over a shared settlement network, such as Fedwire in the United States.
+
+When a payer’s ASE settles with the payee’s ASE, there’s a high chance that the ASE isn’t physically handing over cash. There’s more likely to be an intermediary, like a reserve bank or central bank, that maintains accounts for both ASEs. The intermediary moves funds from one account to another, crediting and debiting the accounts as necessary.
```
hajjimo, line 23: Should we opt for sender and recipient instead of payer and payee, as we did with the OP docs?

packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+The concept of settlement&mdash;finalizing a transaction by exchanging actual funds&mdash;is no different in ILP. However, ILP itself is not a settlement network and cannot exchange funds. A shared settlement network between ASEs, such as Fedwire, is still required. When ASEs agree to [peer with one another](/integration/requirements/peers), one piece they must agree on is which settlement network to use.
```
hajjimo, line 25: ILP not handling settlement is understood, but I wonder if the binary delineation of payments between clearing and settlement might confuse readers. If Interledger is presented as neither handling clearing nor settlement, then an uninitiated ASE might wonder what function it serves. If ILP is only a messaging protocol, is an interface with a formal clearing function still required, or does ILP's similarity with clearing make a completed ILP payment automatically ready for settlement?

packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
melissahenderson: You've made a good point re: characteristics. I need to think about this section some more. It made me remember that the Interledger developer docs say that "clearing is not part of the standard" and also refer to an "Interledger clearing system".

packages/documentation/src/content/docs/overview/concepts/clearing-settlement.mdx
melissahenderson: I thought about that, but I think the audience for the Rafiki docs is different enough from the Open Payments docs that payer/payee is OK. We also only use payer/payee in the Rafiki glossary for ASE, and now this page. It was all over the place in Open Payments.

brad-dow (approved):

PR #3426, opened 2025-05-08
feat(backend): try to match 'wallet address not found' webhook to tenant by prefix

### LINKED ISSUE #3414: [Multi-Tenant] Determine "Wallet Address Not Found" webhook recipient based on wallet address prefix
Currently, the "Wallet Address Not Found" webhook gets set to the Rafiki instance operator's webhook endpoint, as it was initially unclear which tenant's wallet address the request might be referring to.

Since wallet address ranges can now be specified by a tenant, if an unknown wallet address contains the range specified by a tenant, the webhook could be sent to the endpoint of that tenant instead.

njlie:

- On "webhook not found" errors, try to find any tenants whose wallet address prefix matches with the attempted wallet address, then send webhooks to those tenants.

## Context

Fixes #3414.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/wallet_address/service.test.ts
+++ b/packages/backend/src/open_payments/wallet_address/service.test.ts
@@ -59,7 +59,7 @@ describe('Open Payments Wallet Address Service', (): void => {
     await appContainer.shutdown()
   })
 
-  describe('Create or Get Wallet Address3', (): void => {
+  describe('Create or Get Wallet Address', (): void => {
     let tenantId: string
     let options: CreateOptions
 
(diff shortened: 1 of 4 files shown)


packages/backend/src/tenants/settings/service.test.ts
```diff
@@ -371,4 +373,37 @@ describe('TenantSetting Service', (): void => {
       expect(result).toEqual([])
     })
   })
+
+  describe('get settings by value', (): void => {
+    test('can get settings by wallet address prefix setting', async (): Promise<void> => {
```
mkurapov, line 378: we should add a negative test to check that we avoid fetching a tenant if the prefix doesn't match

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -606,6 +606,53 @@ describe('Open Payments Wallet Address Service', (): void => {
...
+                expect.objectContaining({
+                  recipientTenantId: tenant.id,
+                  eventId: walletAddressNotFoundEvents[0].id,
+                  processAt: expect.any(Date)
+                })
+              ]
+            })
+          }
+        )
+      })
```
BlairCurrey, line 655: Oh boy. We've been doing something wrong with `withConfigOverride`. 

This never fails. Throw an error, do `expect(true).toBe(false)`, etc. and you will see it pass.

We need to invoke `withConfigOverride` (and also `await` it).  Like `await withConfigOverride(...)()`.

Outside the scope of this task but we should fix the other usages of `withConfigOverride`.  Alternatively, instead of evoking it, we could change `withConfigOverride` to work like we've been expecting it to.  Just calling the test directly instead of returning a function.

```ts
export async function withConfigOverride(
  getConfig: () => IAppConfig,
  configOverride: Partial<IAppConfig>,
  test: jest.Func
): Promise<void> {
  const config = getConfig()
  const savedConfig = Object.assign({}, config)

  Object.assign(config, configOverride)

  try {
    await test()
  } finally {
    Object.assign(config, savedConfig)
  }
}
```

I did not test this and would not be surprised if this causes some tests to fail (which were previously incorrectly passing). Also note that our current implementation passes in some `arg: any[]` to the test fn but I dont think its actually used anywhere nor do I think we'd even need to use it.

BlairCurrey: Oh shoot, that's not 100% right. Was comparing to another example that was working fine.

it should be 

```ts
      test('creates wallet address not found event for tenant with matching prefix', 
        withConfigOverride(
```

not 

```ts
      test('creates wallet address not found event for tenant with matching prefix', async (): Promise<void> => {
        withConfigOverride(
```

The second argument to the jest `test` should just be `withConfigOverride`. not an async function with `withConfigOverride` in it.

njlie: Thanks for catching this! Glad it's just contained to this PR.

mkurapov (approved):

PR #3439, opened 2025-05-26
Setup for testing static routing

sanducb:

This PR is part of the multi-hop payments research and is meant for experimenting with multi hop payments through static routing

This must not be merged.

## Context

The setup creates 4 instances where instance A is peered with B, B is peered with C, C is peered with D. Payments should be successful from A -> B -> C -> D by using the current Bruno collection.
For now, routes are hardcoded in memory and a complete implementation would require storing these static routes in a database.


packages/backend/src/payment-method/ilp/connector/index.ts
```diff
@@ -73,9 +77,12 @@ export async function createConnectorService({
       tenantSettingService
     },
     compose([
+      //createIlpTimingMiddleware(),
```
sanducb: FYI this will be part of the telemetry logic to measure ILP payments time through multiple hops as well as per hop time in a future commit.

sanducb:
Closing this and reopening [here](https://github.com/interledger/rafiki/pull/3566) from a clean branch.

PR #3440, opened 2025-05-27
feat(auth): add subject to grant

### LINKED ISSUE #3343: Add `subject` field to Authorization Server
The Authorization Server should support `subject` fields in the grant request. It should be similar to the `accesses` in that a separate table manages each grant's `subject` fields.

- [ ] Add `subjects` table to AS database
  - [ ] Has many-to-one relation with grants table
  - [ ] Columns should include `sub_ids` and `sub_id_format`
- [ ] Add logic in grant request endpoint to create associated `subjects` rows 
- [ ] Add logic in token introspection to include the `subjects` access control for the associated grant in the response.

cozminu:

- 

## Context

fixes #3343 

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/localenv/mock-account-servicing-entity/Dockerfile
+++ b/localenv/mock-account-servicing-entity/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-alpine3.20 AS base
+FROM node:20-alpine3.21 AS base
 
 WORKDIR /home/rafiki
 
@@ -29,7 +29,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     --prod \
     | grep -v "cross-device link not permitted\|Falling back to copying packages from store"
 
-FROM base AS builder   
+FROM base AS builder
 
 COPY package.json pnpm-workspace.yaml .npmrc tsconfig.json tsconfig.build.json ./
 COPY localenv/mock-account-servicing-entity ./localenv/mock-account-servicing-entity
@@ -42,7 +42,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     --frozen-lockfile
 RUN pnpm --filter mock-account-servicing-entity build
 
-FROM node:20-alpine3.20 AS runner
+FROM node:20-alpine3.21 AS runner
 
 WORKDIR /home/rafiki
 
--- a/packages/auth/Dockerfile.prod
+++ b/packages/auth/Dockerfile.prod
@@ -44,7 +44,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     --frozen-lockfile
 RUN pnpm --filter auth build
 
-FROM node:20-alpine3.20 AS runner
+FROM node:20-alpine3.21 AS runner
 
 RUN adduser -D rafiki
 
--- a/packages/backend/Dockerfile.prod
+++ b/packages/backend/Dockerfile.prod
@@ -44,7 +44,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     --frozen-lockfile
 RUN pnpm --filter backend build
 
-FROM node:20-alpine3.20 AS runner
+FROM node:20-alpine3.21 AS runner
 
 # Since this is from a fresh image, we need to first create the Rafiki user
 RUN adduser -D rafiki
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -74,7 +74,7 @@
     "@opentelemetry/sdk-node": "^0.52.1",
     "@opentelemetry/sdk-trace-node": "^1.25.1",
     "ajv": "^8.12.0",
-    "axios": "1.12.0",
+    "axios": "^1.12.0",
     "base64url": "^3.0.1",
     "dotenv": "^16.4.7",
     "extensible-error": "^1.0.2",
--- a/packages/frontend/Dockerfile.prod
+++ b/packages/frontend/Dockerfile.prod
@@ -40,7 +40,7 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     --frozen-lockfile
 RUN pnpm --filter frontend build
 
-FROM node:20-alpine3.20 AS runner
+FROM node:20-alpine3.21 AS runner
 RUN adduser -D rafiki
 WORKDIR /home/rafiki
 
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -104,7 +104,7 @@ importers:
         version: 9.0.8
       axios:
         specifier: ^1.12.0
-        version: 1.12.0(debug@4.3.2)
+        version: 1.12.2(debug@4.3.2)
       class-variance-authority:
         specifier: ^0.7.1
         version: 0.7.1
@@ -198,7 +198,7 @@ importers:
         version: 8.12.0
       axios:
         specifier: ^1.12.0
-        version: 1.12.0(debug@4.3.2)
+        version: 1.12.2(debug@4.3.2)
       dotenv:
         specifier: ^16.4.7
         version: 16.4.7
@@ -393,8 +393,8 @@ importers:
         specifier: ^8.12.0
         version: 8.12.0
       axios:
-        specifier: 1.12.0
-        version: 1.12.0(debug@4.3.2)
+        specifier: ^1.12.0
+        version: 1.12.2(debug@4.3.2)
       base64url:
         specifier: ^3.0.1
         version: 3.0.1
@@ -611,7 +611,7 @@ importers:
         version: 2.16.4(typescript@5.8.3)
       axios:
         specifier: ^1.12.0
-        version: 1.12.0(debug@4.3.2)
+        version: 1.12.2(debug@4.3.2)
       class-variance-authority:
         specifier: ^0.7.1
         version: 0.7.1
@@ -730,7 +730,7 @@ importers:
         version: 2.0.2
       axios:
         specifier: ^1.12.0
-        version: 1.12.0(debug@4.3.2)
+        version: 1.12.2(debug@4.3.2)
       pino:
         specifier: ^8.19.0
         version: 8.19.0
@@ -816,7 +816,7 @@ importers:
         version: 7.0.0
       '@koa/bodyparser':
         specifier: ^5.1.1
-        version: 5.1.1(koa@2.15.3)
+        version: 5.1.1(koa@2.16.0)
       '@types/koa':
         specifier: 2.15.0
         version: 2.15.0
@@ -834,7 +834,7 @@ importers:
         version: 1.0.6
       koa:
         specifier: ^2.15.3
-        version: 2.15.3
+        version: 2.16.0
       mock-account-service-lib:
         specifier: workspace:*
         version: link:../../packages/mock-account-service-lib
@@ -1558,7 +1558,7 @@ packages:
       '@babel/core': 7.26.9
       '@babel/helper-compilation-targets': 7.26.5
       '@babel/helper-plugin-utils': 7.26.5
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       lodash.debounce: 4.0.8
       resolve: 1.22.8
     transitivePeerDependencies:
@@ -3010,7 +3010,7 @@ packages:
       '@babel/parser': 7.26.7
       '@babel/template': 7.25.9
       '@babel/types': 7.26.7
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globals: 11.12.0
     transitivePeerDependencies:
       - supports-color
@@ -3024,7 +3024,7 @@ packages:
       '@babel/parser': 7.27.0
       '@babel/template': 7.25.9
       '@babel/types': 7.27.0
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globals: 11.12.0
     transitivePeerDependencies:
       - supports-color
@@ -3038,7 +3038,7 @@ packages:
       '@babel/parser': 7.26.9
       '@babel/template': 7.26.9
       '@babel/types': 7.26.9
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globals: 11.12.0
     transitivePeerDependencies:
       - supports-color
@@ -3925,7 +3925,7 @@ packages:
     engines: {node: ^12.22.0 || ^14.17.0 || >=16.0.0}
     dependencies:
       ajv: 6.12.6
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       espree: 9.6.1
       globals: 13.19.0
       ignore: 5.2.4
@@ -4832,7 +4832,7 @@ packages:
     deprecated: Use @eslint/config-array instead
     dependencies:
       '@humanwhocodes/object-schema': 2.0.3
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       minimatch: 3.1.2
     transitivePeerDependencies:
       - supports-color
@@ -5672,14 +5672,14 @@ packages:
     resolution: {integrity: sha512-Lg3PnLp0QXpxwLIAuuJboLeRaIhrgJjeuh797QADg3xz8wGLugQOS5DpsE8A6i6Adgzf+bacllkKZG3J0tGfDw==}
     dev: true
 
-  /@koa/bodyparser@5.1.1(koa@2.15.3):
+  /@koa/bodyparser@5.1.1(koa@2.16.0):
     resolution: {integrity: sha512-ZBF49xqNVxnmJ+8iXegq+fXPQm9RSX8giNl/aXS5rW1VpNct92wnFbGR/47vfoRJVLARGQ4HVL4WaQ0u8IJVoA==}
     engines: {node: '>= 16'}
     peerDependencies:
       koa: ^2.14.1
     dependencies:
       co-body: 6.1.0
-      koa: 2.15.3
+      koa: 2.16.0
       lodash.merge: 4.6.2
       type-is: 1.6.18
     dev: true
@@ -6410,7 +6410,7 @@ packages:
   /@ory/client@1.9.0:
     resolution: {integrity: sha512-O4a1ijgJtMNIA+ZmWUmCodxX13ID72hOaCB0b9FQGQBzuFgF2x/Yq5D43nrMYZaDtvDvja8J1XIXhUkjz1TDOw==}
     dependencies:
-      axios: 1.12.0(debug@4.3.2)
+      axios: 1.12.2(debug@4.3.2)
     transitivePeerDependencies:
       - debug
     dev: false
@@ -6783,7 +6783,7 @@ packages:
       remark-mdx-frontmatter: 1.1.1
       semver: 7.6.3
       set-cookie-parser: 2.6.0
-      tar-fs: 2.1.1
+      tar-fs: 2.1.4
       tsconfig-paths: 4.1.0
       typescript: 5.4.3
       valibot: 0.41.0(typescript@5.4.3)
@@ -6881,7 +6881,7 @@ packages:
       remark-mdx-frontmatter: 1.1.1
       semver: 7.6.3
       set-cookie-parser: 2.6.0
-      tar-fs: 2.1.1
+      tar-fs: 2.1.4
       tsconfig-paths: 4.1.0
       typescript: 5.8.3
       valibot: 0.41.0(typescript@5.8.3)
@@ -8431,7 +8431,7 @@ packages:
     dependencies:
       '@typescript-eslint/typescript-estree': 5.60.1(typescript@5.8.3)
       '@typescript-eslint/utils': 5.60.1(eslint@8.57.1)(typescript@5.8.3)
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       eslint: 8.57.1
       tsutils: 3.21.0(typescript@5.8.3)
       typescript: 5.8.3
@@ -8451,7 +8451,7 @@ packages:
     dependencies:
       '@typescript-eslint/typescript-estree': 7.5.0(typescript@5.4.3)
       '@typescript-eslint/utils': 7.5.0(eslint@8.57.1)(typescript@5.4.3)
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       eslint: 8.57.1
       ts-api-utils: 1.0.1(typescript@5.4.3)
       typescript: 5.4.3
@@ -8485,7 +8485,7 @@ packages:
     dependencies:
       '@typescript-eslint/types': 5.60.1
       '@typescript-eslint/visitor-keys': 5.60.1
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globby: 11.1.0
       is-glob: 4.0.3
       semver: 7.6.3
@@ -8506,7 +8506,7 @@ packages:
     dependencies:
       '@typescript-eslint/types': 5.62.0
       '@typescript-eslint/visitor-keys': 5.62.0
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globby: 11.1.0
       is-glob: 4.0.3
       semver: 7.6.3
@@ -8527,7 +8527,7 @@ packages:
     dependencies:
       '@typescript-eslint/types': 7.5.0
       '@typescript-eslint/visitor-keys': 7.5.0
-      debug: 4.4.0(supports-color@9.4.0)
+      debug: 4.4.1
       globby: 11.1.0
       is-glob: 4.0.3
       minimatch: 9.0.3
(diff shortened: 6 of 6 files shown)


packages/auth/src/subject/service.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+async function createSubject(
+  deps: ServiceDependencies,
+  grantId: string,
+  subjectRequests: SubjectRequest[],
+  trx?: Transaction
+): Promise<Subject[]> {
+  const subjectRequestsWithGrant = subjectRequests.map((subject) => {
+    return { grantId, subId: subject.id, subIdFormat: subject.format }
```
mkurapov, line 52: We should validate the subIdFormat and subId. For now, format can only be `uri`, and the `id` itself be a URI

packages/auth/src/grant/utils.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  type: 'outgoing-payment',
+  actions: [AccessAction.Create],
+  identifier: 'id'
+}
+
+jest.mock('../access/types', () => ({
+  isIncomingPaymentAccessRequest: (access: AccessRequest) =>
+    access.type === 'incoming-payment',
+  isQuoteAccessRequest: (access: AccessRequest) => access.type === 'quote'
+}))
```
mkurapov: Why do we need this mock?

packages/auth/src/graphql/resolvers/grant.ts
```diff
@@ -89,7 +89,7 @@ export const revokeGrant: MutationResolvers<ApolloContext>['revokeGrant'] =
 export const grantToGraphql = (grant: Grant): SchemaGrant => ({
   id: grant.id,
   client: grant.client,
-  access: grant.access?.map((item) => accessToGraphql(item)),
+  access: grant.access?.map((item) => accessToGraphql(item)) || [],
```
njlie, line 104: I think it might make sense to also include the subject field in the graphql resolver

packages/auth/src/subject/service.test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+      expect(subject.length).toEqual(1)
+      expect(subject[0].grantId).toEqual(grant.id)
+      expect(subject[0].subId).toEqual(subjectRequest.id)
+      expect(subject[0].subIdFormat).toEqual(subjectRequest.format)
+    })
+  })
+
+  describe('getByGrant', (): void => {
+    test('gets access', async () => {
```
njlie: nit: this is getting subject, not access

njlie (commented):
Overall LGTM, just a couple things

packages/auth/src/grant/utils.test.ts
cozminu: canSkipInteraction function uses them and the purpose of the unit test is to test only the unit and not the deps, but I see there are also benefits from using the unmocked functions. removed the mock

packages/auth/src/grant/routes.ts
```diff
@@ -211,6 +218,13 @@ async function createPendingGrant(
     )
   } catch (err) {
     await trx.rollback()
+    if (isGrantError(err)) {
+      throw new GNAPServerRouteError(
+        400,
+        GNAPErrorCode.InvalidRequest,
+        err.message || 'invalid request'
```
mkurapov: We should match GrantErrors to their respective HTTP codes and GNAPErrorCodes. 
This is similar to how @oana-lolea maps [AccessErrors](https://github.com/interledger/rafiki/pull/3449/files#diff-731ce666214e27161ad4407bf81561f9fe71152fa994874082c52119fd73bdfb) to the respective code & error codes in her [PR](https://github.com/interledger/rafiki/pull/3449#discussion_r2142591918).

packages/auth/src/grant/utils.ts
```diff
@@ -10,19 +10,31 @@ export function canSkipInteraction(
...
-    if (!canSkip && (!access.identifier || access.identifier === '')) {
-      throw new Error('identifier required')
-    }
-    return canSkip
-  })
+  const emptySubject = (body.subject?.sub_ids?.length ?? 0) === 0
+  const emptyAccess = (body.access_token?.access?.length ?? 0) === 0
+
+  if (emptySubject && emptyAccess) {
+    throw new Error('subject or access_token required')
```
mkurapov: Should this throw a GrantError instead?

packages/auth/src/graphql/schema.graphql
```diff
@@ -119,6 +121,17 @@ type Access implements Model {
   createdAt: String!
 }
 
+type Subject implements Model {
+  "Unique identifier of the subject object."
+  id: ID!
+  "Wallet address of the subject's account."
+  subId: String!
+  "Format of the subject identifier, which is 'url'."
+  subIdFormat: String!
```
mkurapov:
```suggestion
  "Format of the subject identifier
  subIdFormat: String!
```
Or we specify that it should be `uri`. I lean towards not specify in the schema (if its not an Enum) so the docs don't become out of date with the actual application logic

packages/auth/src/accessToken/routes.ts
```diff
@@ -39,6 +40,7 @@ type ManagementContext = Exclude<TokenHttpSigContext, 'request'> & {
 interface IntrospectBody {
   access_token: string
   access?: AccessItem[]
+  subjects?: SubjectItem[]
```
mkurapov: If we don't use `subjects` anywhere (given that we don't pass it in during the introspection request), then its OK to leave out

packages/auth/src/subject/service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+async function getByGrant(
+  deps: ServiceDependencies,
+  grantId: string,
+  trx?: Transaction
+): Promise<Subject[]> {
+  return Subject.query(trx || deps.knex).where({ grantId })
+}
+
+function validateSubjectRequest(subject: SubjectRequest): void {
```
mkurapov, line 66: We should add tests for these errors

mkurapov (commented):
F

packages/auth/src/grant/utils.ts
cozminu: yes, adapted to throw GrantError

packages/auth/src/subject/service.ts
cozminu: added

packages/auth/src/accessToken/routes.ts
cozminu: removed

packages/auth/src/grant/routes.ts
cozminu: done

packages/auth/src/subject/service.test.ts
cozminu: yes, fixed

packages/auth/src/graphql/resolvers/grant.ts
cozminu: done

packages/auth/src/subject/service.ts
cozminu: done

packages/auth/src/access/service.test.ts
```diff
@@ -19,7 +19,7 @@ describe('Access Service', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
   let accessService: AccessService
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = null as unknown as Knex.Transaction
```
njlie: It might be healthier to actually initialize `trx` in the `beforeAll` block so that the tables get properly truncated when the test is finished, but the editor also doesn't flag a Typescript error. I assume this type casting was done to address the TS error.

```
	...
	trx = await appContainer.knex.transaction()
})
```

packages/auth/src/access/utils.test.ts
```diff
@@ -21,7 +21,7 @@ import { compareRequestAndGrantAccessItems } from './utils'
 describe('Access utilities', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = undefined as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/auth/src/accessToken/routes.test.ts
```diff
@@ -28,7 +28,7 @@ import { GNAPErrorCode } from '../shared/gnapErrors'
 describe('Access Token Routes', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = undefined as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/auth/src/accessToken/service.test.ts
```diff
@@ -24,7 +24,7 @@ import { generateBaseGrant } from '../tests/grant'
 describe('Access Token Service', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = undefined as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/auth/src/grant/service.test.ts
```diff
@@ -29,7 +29,7 @@ describe('Grant Service', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
   let grantService: GrantService
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = null as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/auth/src/grant/utils.test.ts
```diff
@@ -0,0 +1,86 @@
+import { AccessRequest } from '../access/types'
```
njlie, line 1: Thanks for adding these tests! Especially since they cover logic that existed before this PR

packages/auth/src/signature/middleware.test.ts
```diff
@@ -60,7 +60,7 @@ describe('Signature Service', (): void => {
     let grant: Grant
     let interaction: Interaction
     let token: AccessToken
-    let trx: Knex.Transaction
+    const trx: Knex.Transaction = null as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/backend/src/open_payments/payment/outgoing/service.test.ts
```diff
@@ -71,7 +71,7 @@ describe('OutgoingPaymentService', (): void => {
   let receiver: string
   let client: string
   let amtDelivered: bigint
-  let trx: Knex.Transaction
+  const trx: Knex.Transaction = null as unknown as Knex.Transaction
```
njlie: Ditto to prior comment on initializing `trx`.

packages/auth/src/access/service.test.ts
cozminu: Yes, It was a fix to a TS error
```
Variable 'trx' is used before being assigned.ts(2454)
```
Using a transaction is tricky because it can deadlock tests because some queries are done within a transaction and others are not. Managed to fix it, but in another manner. 
```
let trx: KnexOrTransaction

	...
	trx = await appContainer.knex
})
```
Is this ok with you?

packages/auth/src/graphql/resolvers/grant.ts
```diff
@@ -89,7 +91,8 @@ export const revokeGrant: MutationResolvers<ApolloContext>['revokeGrant'] =
...
+  subject: grant.subjects?.map((item) => subjectToGraphql(item)) || [],
```
mkurapov: `grant.subjects` only exists if we add it to the `withGraphFetched` request in the service methods OR we have `subjects` as a separate "child" resolver under grants, and call the `subjectService.getByGrantId` directly. 

An example of this is the `Asset > sendingFee` resolver in `backend`.

packages/auth/src/graphql/schema.graphql
```diff
@@ -95,7 +95,9 @@ type Grant implements Model {
   "Wallet address of the grantee's account."
   client: String!
   "Details of the access provided by the grant."
-  access: [Access!]!
```
mkurapov, line 100: I think we can keep it as `[Access!]!`, if there is no access it can simply be an empty array.

packages/auth/src/grant/routes.ts
```diff
@@ -115,7 +120,7 @@ async function createGrant(
     throw new GNAPServerRouteError(
       400,
       GNAPErrorCode.InvalidRequest,
-      'access identifier required'
+      err instanceof Error ? err.message : ''
```
mkurapov: to avoid empty string, let's provide some generic message instead at least. It's less important for the client, its more so for us to find where to debug it.

packages/auth/src/subject/service.ts
```diff
@@ -0,0 +1,82 @@
...
+  try {
+    if (!subject.id.startsWith('https://')) throw 1
+    new URL(subject.id)
+  } catch {
+    throw new GrantError(
+      GrantErrorCode.InvalidRequest,
+      'subject id must be a valid https url'
+    )
+  }
+  if (subject.format != 'uri') {
```
mkurapov:
```suggestion
  if (subject.format !== 'uri') {
```

packages/auth/src/subject/service.ts
```diff
@@ -0,0 +1,82 @@
...
+  deps: ServiceDependencies,
+  grantId: string,
+  trx?: Transaction
+): Promise<Subject[]> {
+  return Subject.query(trx || deps.knex).where({ grantId })
+}
+
+function validateSubjectRequest(subject: SubjectRequest): void {
+  try {
+    if (!subject.id.startsWith('https://')) throw 1
```
mkurapov: not sure if @njlie agrees, but I think just checking that it is a valid `URL` is sufficient, without checking the `https://`. IMO it can be up to the ASE to determine how strict they want to be with the specifics, like protocol.

packages/auth/src/grant/model.ts
```diff
@@ -155,7 +165,8 @@ export function toOpenPaymentsGrant(
   grant: Grant,
   args: ToOpenPaymentsGrantArgs,
   accessToken: AccessToken,
-  accessItems: Access[]
+  accessItems: Access[],
+  subjectItems: Subject[]
 ): OpenPaymentsGrant {
   return {
     access_token: toOpenPaymentsAccessToken(accessToken, accessItems, {
```
mkurapov: `access_token` could be undefined now, just like subject

packages/auth/src/access/service.test.ts
```diff
@@ -14,12 +14,13 @@ import { IncomingPaymentRequest, OutgoingPaymentRequest } from './types'
...
 import { AccessType, AccessAction } from '@interledger/open-payments'
 import { Access } from './model'
+import { TransactionOrKnex } from 'objection'
 
 describe('Access Service', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
   let accessService: AccessService
-  let trx: Knex.Transaction
+  let trx: TransactionOrKnex
```
mkurapov, line 23: We should remove the `import { Knex } from 'knex'` import

packages/auth/src/interaction/routes.ts
```diff
@@ -148,6 +154,7 @@ async function getGrantDetails(
   ctx.body = {
     grantId: interaction.grant.id,
     access: access.map(toOpenPaymentsAccess),
+    subjects: subjects.map(toOpenPaymentsSubject),
```
mkurapov: there can only be one subject for a grant, but the `sub_ids` can be many. So this can either be a `subject` object with `sub_ids` array, (like the Open Payments spec), or we can directly return `sub_ids` here.

mkurapov: The same applies for the subject updates in the GraphQL schema

packages/auth/src/graphql/resolvers/grant.ts
cozminu: fixed with `withGraphFetched('subjects')` in the service methods

packages/auth/src/subject/service.ts
njlie: For the record, I also agree.

packages/auth/src/grant/service.test.ts
```diff
@@ -311,11 +334,28 @@ describe('Grant Service', (): void => {
       })
     })
 
-    describe('getByIdWithAccess', (): void => {
+    describe('getByIdWithAccessAndSubject', (): void => {
       test('Can fetch a grant by id with access', async () => {
-        const fetchedGrant = await grantService.getByIdWithAccess(grant.id)
+        const grantRequest: GrantRequest = {
+          ...BASE_GRANT_REQUEST,
+          subject: {
```
njlie, line 341: `access` should still be included here so that this service call can properly test that it also retrieves an `access` field along with the grant

packages/auth/src/grant/routes.ts
cozminu: changed to 'unknown error while checking interaction requirement'

packages/auth/src/subject/service.ts
cozminu: fixed

packages/auth/src/grant/model.ts
cozminu: fixed

packages/auth/src/interaction/routes.ts
cozminu: changed to subject object with sub_ids array.
I'm not aware of a subject updates.

packages/auth/src/grant/service.test.ts
cozminu: fixed

packages/auth/src/grant/routes.ts
```diff
@@ -135,7 +135,9 @@ async function createGrant(
     throw new GNAPServerRouteError(
       400,
       GNAPErrorCode.InvalidRequest,
-      'access identifier required'
+      err instanceof Error
+        ? err.message
+        : 'unknown error while checking interaction requirement'
```
mkurapov: Looking at what `canSkipInteraction` throws, we should instead use `err.description`, since that will always be set.

packages/auth/src/grant/errors.ts
```diff
@@ -0,0 +1,49 @@
...
+import { AccessError } from '../access/errors'
+
+export enum GrantErrorCode {
+  InvalidRequest,
+  OnlyOneAccessAmountAllowed
+}
+
+export class GrantError extends Error {
+  code: GrantErrorCode
+  constructor(code: GrantErrorCode, description?: string) {
```
mkurapov:
```suggestion
  constructor(code: GrantErrorCode, description: string) {
```
Looks like `description` is set everywhere already

packages/auth/src/grant/routes.ts
```diff
@@ -161,11 +163,11 @@ async function createApprovedGrant(
...
     await trx.rollback()
-    if (isAccessError(err)) {
+    if (isGrantError(err)) {
       throw new GNAPServerRouteError(
-        errorToHTTPCode[err],
-        errorToGNAPCode[err],
-        err
+        errorToHTTPCode[err.code],
+        errorToGNAPCode[err.code],
+        err.message || 'invalid request'
```
mkurapov:
```suggestion
        err.description || 'invalid request'
```
same as the other comment, we can use `err.description` here, since it will always be set and is most useful. Or rename `GrantError.description` to `GrantError.message`

packages/auth/src/grant/routes.ts
```diff
@@ -175,6 +177,7 @@ async function createApprovedGrant(
     )
   }
   const access = await deps.accessService.getByGrant(grant.id)
+  const subjects = await deps.subjectService.getByGrant(grant.id)
```
mkurapov: Since `subjects` always requires interaction, I think we don't need to fetch `subjects` for the non-interactive grant creation.

mkurapov (commented):
Thinking through this a bit, I think we need to update the response types. 
Since `subject` should only be returned _after_ we have gone through a successful interaction, we should:
- Make sure we do not return `subject` in the initial grant request
- Return `subject` in the grant _continue_ request only


[Section 3.4](https://www.rfc-editor.org/rfc/rfc9635.html#section-3.4) of GNAP:
>  The grant request MUST be in the approved state to return this field in the response.

This will require open-payments-specification changes, but we can work together @cozminu and I can take care of those changes


CC @njlie for a double check :)

packages/auth/src/grant/routes.ts
cozminu: casted `e` as GrantError to fix this.
```
...
} catch (e) {
const err = e as GrantError
throw new GNAPServerRouteError(
  400,
  GNAPErrorCode.InvalidRequest,
  err.message
)
...
```
Other option would be to throw 500 if `e` is not GrantError

packages/auth/src/grant/routes.ts
cozminu: renamed `description` to `message` to fit Error class

packages/auth/src/grant/routes.ts
```diff
@@ -22,19 +22,22 @@ import { InteractionService } from '../interaction/service'
...
 import { generateRouteLogs } from '../shared/utils'
-import { TenantService } from '../tenant/service'
-import { Tenant, isTenantWithIdp } from '../tenant/model'
+import { SubjectService } from '../subject/service'
 import {
   errorToGNAPCode,
   errorToHTTPCode,
-  isAccessError
-} from '../access/errors'
+  GrantError,
```
mkurapov:
```suggestion
  
```

packages/auth/src/grant/model.ts
```diff
@@ -168,19 +178,27 @@ export function toOpenPaymentsGrant(
...
       uri: `${args.authServerUrl}/continue/${grant.continueId}`
-    }
-  }
+    },
+    subject: subjectItems?.length
+      ? {
+          sub_ids: subjectItems.map(toOpenPaymentsSubject)
+        }
+      : undefined
+  } as OpenPaymentsGrant
```
mkurapov, line 202: very minor, but I think for Open Payments we should only return the `subject` key, when it is defined. Same goes for `access_token`

packages/auth/src/grant/service.ts
```diff
@@ -271,6 +287,10 @@ async function create(
     if (!trx) {
       await grantTrx.rollback()
     }
```
mkurapov, line 300:
```suggestion
    await grantTrx.rollback()
```
as `grantTrx` is always defined. Same for L281-283

cozminu: Tests will fail if I do this. Reason being that the transaction starts on a higher level if `trx` is provided and `rollback` or `commit` will be called twice: once here and once on the higher level.

packages/auth/src/grant/model.ts
cozminu: if value is `undefined`, the key will not show up in the response

package.json
```diff
@@ -74,7 +74,8 @@
       "path-to-regexp@>=0.1.7": "^0.1.12",
       "path-to-regexp@>=6.3.0": "^6.3.0",
       "next": "^15.2.3",
-      "form-data": "^4.0.4"
+      "form-data": "^4.0.4",
+      "sha.js@2.4.11": "2.4.12"
```
mkurapov:
```suggestion
      "sha.js": ">=2.4.12"
```
so we always use a version above `2.4.12`

mkurapov (approved):

packages/auth/src/openapi/specs/id-provider.yaml
```diff
@@ -146,6 +146,30 @@ paths:
                 properties:
                   access:
                     $ref: ./auth-server.yaml#/components/schemas/access
+                  subject:
```
mkurapov, line 149: We should make this spec standalone, and move over the correct components (access/subject) into this file, similar to how it was done for `token-introspection`

packages/auth/src/grant/routes.ts
```diff
@@ -353,6 +362,7 @@ async function pollGrantContinuation(
   } else {
     const accessToken = await accessTokenService.create(grant.id)
     const access = await accessService.getByGrant(grant.id)
```
mkurapov, line 364: I think we should only create an `accessToken` if the `access_token` was provided in the initial grant request. We can use length of `access` to verify that, I believe. Otherwise, if the client is just requesting subject information, and not `access_token`, we would return an `access_token` still, eg:

```
{
  "access_token": {
    "access": [],
    "value": "EC6884A68394DAC1E186",
    "manage": "http://localhost:3006/token/965841c8-1671-4174-944d-56778083dfdb",
    "expires_in": 600
  },
  "continue": {
    "access_token": {
      "value": "B785F0F5ED51E08A923E"
    },
    "uri": "http://localhost:3006/continue/b094d601-cd62-43fd-9389-5ed6628c1c1d"
  },
  "subject": {
    "sub_ids": [
      {
        "id": "https://cloud-nine-wallet-backend/accounts/gfranklin",
        "format": "uri"
      }
    ]
  }
}
```

CC @njlie to confirm as well

njlie: Yeah agreed, there's nothing to access, hence no access token.

cozminu: fixed, access_token is no more when the grant doesn't contain access items

PR #3449, opened 2025-06-05
feat: either receiveAmount or debitAmount allowed for outgoing payment limits

### LINKED ISSUE #3337: Update AS spec to ensure correct `limits` during outgoing payment grant creation
## Context

- From prerequisite https://github.com/interledger/open-payments/issues/471:

> Currently, when we create an outgoing payment grant request, we can pass in a limits attribute with debitAmount and/or receiveAmount. For simplicity of the API as well as the simplicity of documenting how these amounts are used, we should update the spec to allow only setting limits.debitAmount OR limits.receiveAmount, not both.

When this change lands, we need to update the open-payments client in the AS to make sure we are correctly validating the incoming request.

## Todos
- [ ] Update open-payments client in AS after issue above is finished

oana-lolea:

- Updated `OutgoingPaymentLimit` to no longer accept both `receiveAmount` and `debitAmount` 
- Added error when both types of amount are provided
- Updated bruno request

## Context

Fixes #3337

## Checklist

- [x] Tests added/updated


packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -518,6 +518,9 @@ function validateAmountAssets(
   payment: OutgoingPayment,
   limits: Limits
 ): boolean {
+  if (limits.debitAmount && limits.receiveAmount) {
+    throw OutgoingPaymentError.OnlyOneAmountAllowed
+  }
```
mkurapov: It is good to check this at the resource server, but also we need to check this at the auth server as well, since that is the place where the grant with the access is actually being created. 

In other words, we should prevent the grant being created in the first place with the incorrect access amounts.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -518,6 +518,9 @@ function validateAmountAssets(
   payment: OutgoingPayment,
   limits: Limits
 ): boolean {
+  if (limits.debitAmount && limits.receiveAmount) {
+    throw OutgoingPaymentError.OnlyOneAmountAllowed
```
mkurapov:
```suggestion
    throw OutgoingPaymentError.OnlyOneGrantAmountAllowed
```

packages/auth/src/access/service.ts
```diff
@@ -63,3 +65,14 @@ async function getByGrant(
...
+function validateLimits(accessRequests: AccessRequest[]) {
+  const areBothLimitsSet =
+    accessRequests.findIndex(
+      (access) => access.limits?.debitAmount && access.limits.receiveAmount
+    ) !== -1
+
+  if (areBothLimitsSet) {
+    throw new Error('Only one of receiveAmount or debitAmount allowed')
+  }
+}
```
mkurapov: Following the stack trace for the error, we would end up sending a 500 - Internal server error back to the client.
Instead, we should reply with a 400  `invalid_request`.

packages/auth/src/grant/routes.ts
```diff
@@ -140,6 +140,9 @@ async function createApprovedGrant(
     await trx.commit()
   } catch (err) {
     await trx.rollback()
+    if (typeof err === 'string') {
+      throw new GNAPServerRouteError(400, GNAPErrorCode.InvalidRequest, err)
+    }
```
mkurapov: I think it would be useful to map errors from the services to the HTTP status & GNAPErrorCodes, similar to how we are doing in the backend's incoming payments for example.

packages/auth/src/access/errors.ts
```diff
@@ -0,0 +1,7 @@
+export enum AccessError {
+  OnlyOneAccessAmountAllowed = 'only one access amount allowed'
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
+export const isAccessError = (o: any): o is AccessError =>
+  Object.values(AccessError).includes(o)
```
mkurapov, line 9: We should also add `errorToHTTPCode` and `errorToGNAPCode` here as well

packages/auth/src/grant/routes.ts
```diff
@@ -190,8 +191,8 @@ async function createPendingGrant(
 
   const trx = await Grant.startTransaction()
 
+  const grant = await createGrantOrThrowError(grantService, body, trx)
```
mkurapov: To simplify, we could probably:
- Only return `Promise<Grant>` in the `grantService.create` (without `GrantError`) and just let the thrown `AccessError` propagate to `grantRoutes.createPendingGrant`
- In the `createPendingGrant` catch block down below, check for `isAccessError(err)`

With this, you'll only end up adding an extra check in the  `createPendingGrant` catch block or the new error without needing to rollback the trx separately

packages/backend/src/open_payments/payment/outgoing/errors.ts
```diff
@@ -49,7 +50,8 @@ export const errorToHTTPCode: {
   [OutgoingPaymentError.InactiveWalletAddress]: 400,
   [OutgoingPaymentError.InvalidAmount]: 400,
   [OutgoingPaymentError.NegativeReceiveAmount]: 400,
-  [OutgoingPaymentError.InvalidReceiver]: 400
+  [OutgoingPaymentError.InvalidReceiver]: 400,
+  [OutgoingPaymentError.OnlyOneGrantAmountAllowed]: 400
```
mkurapov:
```suggestion
  [OutgoingPaymentError.OnlyOneGrantAmountAllowed]: 500
```

Since from the perspective of the client, the input they passed to create outgoing payment could have been OK, it was the auth server handling of the grant request that was incorrect (and a server-side error)

packages/auth/src/grant/service.test.ts
```diff
@@ -267,12 +270,18 @@ describe('Grant Service', (): void => {
         }
 
         const grant1 = await grantService.create(grantRequest)
+        assert.ok(!isAccessError(grant1))
```
mkurapov: Looks like we don't need this now, since we only can return `Grant`

localenv/mock-account-servicing-entity/app/routes/mock-idp.consent.tsx
```diff
@@ -41,8 +41,11 @@ function AuthorizedView({
         </div>
         <div>
           <p>
-            You gave {thirdPartyName} permission to send {currencyDisplayCode}{' '}
-            {amount.toFixed(2)} out of your account.
+            You gave {thirdPartyName} permission to send{' '}
```
mkurapov:
```suggestion
            You gave {thirdPartyName} permission to send {' '}
```

PR #3451, opened 2025-06-05
feat(localenv): add tenanted mock ASE to local environment

### LINKED ISSUE #3446: [Multi-Tenant] Add optional tenanted mock ASE to local environment
Create another Mock ASE that is a tenant of one of the Rafiki instances in the local environment.

It should be created upon inclusion of an option, such as `pnpm localenv:compose:multitenant up`

njlie:

- Adds a tenanted Mock ASE to the local environment under the `cloud-nine` Rafiki instance with its own

## Context

Closes #3446. Or RAF-1064 if you prefer.

This solution feels _quite_ hacky, as it the tenanted Mock ASE has to acquire its own tenant ID after seeding it using the Operator Tenant credentials. Feedback would be appreciated.

This also fixes a bug I found where the asset uniqueness was being enforced without considering the tenant id as a factor for said uniqueness. So assets with the same code and scale would be rejected from creation even if it was being made with a different tenant id from the existing one.

## Checklist

- [ ] Tests added/updated


packages/mock-account-service-lib/src/requesters.ts
```diff
@@ -182,8 +224,11 @@ export async function createPeer(
     input: {
       staticIlpAddress,
       http: {
-        incoming: { authTokens: [`test-${assetCode}`] },
-        outgoing: { endpoint: outgoingEndpoint, authToken: `test-${assetCode}` }
+        incoming: { authTokens: [`test-${assetCode}-${uuid()}`] },
+        outgoing: {
+          endpoint: outgoingEndpoint,
+          authToken: `test-${assetCode}-${uuid()}`
```
mkurapov: The incoming tokens need to match the outgoing tokens for each pair of peers, otherwise we aren't be able to send payments between them. To simplify, we can just add a `token` value to the seed file for each peer, which should be the same for each pair of peers.

We should also specify an asset to peer over.

localenv/cloud-ten-wallet/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+name: c10
+services:
+  cloud-ten-mock-ase:
+    hostname: cloud-ten-wallet
```
BlairCurrey, line 4: love the name 🤣

localenv/mock-account-servicing-entity/app/entry.server.tsx
BlairCurrey, line 1: after seeding the mock ase spits out the operator tenant url:

image

For the cloud-ten tenant it would be nice to print out a url for itself (not the operator tenant). Not exactly sure how to do this. Taking a step back, I think the general problem to solve is making it easy to use the frontend as the non-operator tenant... I followed the link and then went to the tenant tab and got the non-operator secret/id then plugged those in but that felt a bit tedious and that log a bit misleading.

Maybe we can check the `isTenant` config value? Although I'm not sure where the tenant id would come from...

packages/backend/migrations/20241216160130_backfill_tenant_on_assets.js
```diff
@@ -26,5 +28,7 @@ exports.up = function (knex) {
 exports.down = function (knex) {
   return knex.schema.alterTable('assets', (table) => {
     table.dropColumn('tenantId')
+    table.dropUnique(['code', 'scale', 'tenantId'])
```
mkurapov: we should drop unique index before we drop tenantId

mkurapov: I see you've resolved it in the latest commit :+1:

localenv/mock-account-servicing-entity/app/entry.server.tsx
BlairCurrey: Maybe the `tenantId` can be returned from setup from seed? Alternatively, maybe just not logging it if `isTenant`?

packages/mock-account-service-lib/src/seed.ts
```diff
@@ -28,7 +31,25 @@ export async function setupFromSeed(
...
+      seedTenant.publicName,
+      seedTenant.apiSecret,
+      seedTenant.walletAddressPrefix
+    )
+    requesterApolloClient = generateApolloClient({
+      tenantId: createdTenant.id,
+      apiSecret: createdTenant.apiSecret
+    })
+  }
+
```
BlairCurrey: Regarding:

>This solution feels quite hacky, as it the tenanted Mock ASE has to acquire its own tenant ID after seeding it using the Operator Tenant credentials. Feedback would be appreciated.

First, would it make anything easier to add an optional `id` to the create tenant input gql? Maybe makes configuration easier. Would just need to document and validate it's a UUID.

Also, can we run the `setupFromSeed` without being the operator? If not then I initializing this way doesnt seem so offensive...

I guess maybe an alternative could be some optional configuration that seeds the tenant when when initializing the operator (c9 when running the multitenancy script in this case).

Maybe:

- a new `seed-tenant.yaml`. 
- add new env var for the seed-tenant path in c9 docker compose. to be passed in, no defaulting.
- for existing localenv nothing gets passed in 
- the new `localenv:compose:multitenancy` command passes in the `seed-tenant.yml` path to the c9 docker compose
- change seed function to create tenant(s) if provided
- maybe return details from the seeding to log out details in the mock ase?

Just kinda thinking out loud. I think the current changes are nice in that they dont alter the existing local environment.

localenv/mock-account-servicing-entity/app/routes/accounts.create.tsx
```diff
@@ -12,16 +12,28 @@ import { createWallet } from '~/lib/wallet.server'
...
+
+  // Filter out assets that don't belong to the operator if the request returned all tenants' assets
+  return json({
+    assets: !isTenant
+      ? assets.filter(
+          (asset) => asset.node.tenantId === process?.env?.OPERATOR_TENANT_ID
+        )
+      : assets,
+    options
+  })
```
BlairCurrey: not sure if im fully understanding this or not but would it make sense to do this server side? maybe as a new `filter` option in the `listAsset` ?

Could be a separate task.

njlie: Yeah honestly I think this is better off adding a filter that only the operator can use in the graphql query.

mkurapov: Or just `tenantId` like we have for the other list queries (e.g. `outgoingPayments`)

njlie: > Or just `tenantId` like we have for the other list queries (e.g. `outgoingPayments`)

I'll just add this now, so that the current filter doesn't get included when this gets merged.

localenv/mock-account-servicing-entity/app/entry.server.tsx
```diff
@@ -39,20 +42,22 @@ if (!global.__seeded) {
...
+    }
+  )
+    .then((seedResult: { tenantId: string; apiSecret: string } | undefined) => {
       global.__seeded = true
       setTimeout(() => {
         const url = new URL(`http://localhost:${process.env.FRONTEND_PORT}/`)
         const params = new URLSearchParams({
-          tenantId,
-          apiSecret
+          tenantId: seedResult?.tenantId ?? tenantId,
```
BlairCurrey, line 59: small thing, but the log always still shows: `Use this URL to access the frontend with operator tenant credentials`

think we could use `seedResult.tenantId` to omit the `operator` in that message, or just make the message more generic.

mkurapov (commented):
Can we add registering a WEBHOOK_URL for the ASE in the tenant settings on setup?

packages/mock-account-service-lib/src/seed.ts
mkurapov: > First, would it make anything easier to add an optional id to the create tenant input gql? Maybe makes configuration easier. 

Yes, this is a good call, would be quite ergonomic.

PR #3460, opened 2025-06-10
Feat: add paymentMethodProviderService

mkurapov:

- Add ILP_ADDRESS tenantSetting, which by default is the root, operators ilpAddress (ILP_ADDRESS environment variable) concatenated with the tenants id. In the future this can be tenant defined instead.
- Add paymentMethodProviderService that abstracts away payment method generation for incomingPayments
    - Centralized place to generateStreamCredentials (ILP payment method details) where we generate credentials using the configured ilpAddress, which could be the root ilpAddress defined in the environment variable, or looked up using the tenantSettings
    

## Context

Fixes RAF-1069

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/model.test.ts
+++ b/packages/backend/src/open_payments/payment/incoming/model.test.ts
@@ -17,6 +17,7 @@ import {
   IncomingPaymentEventError
 } from './model'
 import { WalletAddress } from '../../wallet_address/model'
+import { OpenPaymentsPaymentMethod } from '../../../payment-method/provider/service'
 
 describe('Models', (): void => {
   let deps: IocContract<AppServices>
@@ -80,16 +81,19 @@ describe('Models', (): void => {
 
     describe('toOpenPaymentsTypeWithMethods', () => {
       test('returns incoming payment with payment methods', async () => {
-        const streamCredentials: IlpStreamCredentials = {
-          ilpAddress: 'test.ilp' as IlpAddress,
-          sharedSecret: Buffer.from('')
-        }
+        const paymentMethods: OpenPaymentsPaymentMethod[] = [
+          {
+            type: 'ilp',
+            ilpAddress: 'test.ilp' as IlpAddress,
+            sharedSecret: ''
+          }
+        ]
 
         expect(
           incomingPayment.toOpenPaymentsTypeWithMethods(
             config.openPaymentsUrl,
             walletAddress,
-            streamCredentials
+            paymentMethods
           )
         ).toEqual({
           id: `${baseUrl}/${Config.operatorTenantId}${IncomingPayment.urlPath}/${incomingPayment.id}`,
@@ -112,67 +116,6 @@ describe('Models', (): void => {
           ]
         })
       })
-
-      test('returns incoming payment with empty methods when stream credentials are undefined', async () => {
-        expect(
-          incomingPayment.toOpenPaymentsTypeWithMethods(
-            config.openPaymentsUrl,
-            walletAddress
-          )
-        ).toEqual({
-          id: `${baseUrl}/${Config.operatorTenantId}${IncomingPayment.urlPath}/${incomingPayment.id}`,
-          walletAddress: walletAddress.address,
-          completed: incomingPayment.completed,
-          receivedAmount: serializeAmount(incomingPayment.receivedAmount),
-          incomingAmount: incomingPayment.incomingAmount
-            ? serializeAmount(incomingPayment.incomingAmount)
-            : undefined,
-          expiresAt: incomingPayment.expiresAt.toISOString(),
-          metadata: incomingPayment.metadata ?? undefined,
-          updatedAt: incomingPayment.updatedAt.toISOString(),
-          createdAt: incomingPayment.createdAt.toISOString(),
-          methods: []
-        })
-      })
-
-      test.each([IncomingPaymentState.Completed, IncomingPaymentState.Expired])(
-        'returns incoming payment with existing methods if payment state is %s',
-        async (paymentState): Promise<void> => {
-          incomingPayment.state = paymentState
-
-          const streamCredentials: IlpStreamCredentials = {
-            ilpAddress: 'test.ilp' as IlpAddress,
-            sharedSecret: Buffer.from('')
-          }
-
-          expect(
-            incomingPayment.toOpenPaymentsTypeWithMethods(
-              config.openPaymentsUrl,
-              walletAddress,
-              streamCredentials
-            )
-          ).toMatchObject({
-            id: `${baseUrl}/${Config.operatorTenantId}${IncomingPayment.urlPath}/${incomingPayment.id}`,
-            walletAddress: walletAddress.address,
-            completed: incomingPayment.completed,
-            receivedAmount: serializeAmount(incomingPayment.receivedAmount),
-            incomingAmount: incomingPayment.incomingAmount
-              ? serializeAmount(incomingPayment.incomingAmount)
-              : undefined,
-            expiresAt: incomingPayment.expiresAt.toISOString(),
-            metadata: incomingPayment.metadata ?? undefined,
-            updatedAt: incomingPayment.updatedAt.toISOString(),
-            createdAt: incomingPayment.createdAt.toISOString(),
-            methods: [
-              expect.objectContaining({
-                type: 'ilp',
-                ilpAddress: streamCredentials.ilpAddress,
-                sharedSecret: expect.any(String)
-              })
-            ]
-          })
-        }
-      )
     })
   })
 
--- a/packages/backend/src/open_payments/payment/incoming/routes.ts
+++ b/packages/backend/src/open_payments/payment/incoming/routes.ts
@@ -111,7 +111,13 @@ async function getIncomingPaymentPrivate(
 
   const streamCredentials = incomingPayment.isExpiredOrComplete()
     ? undefined
-    : deps.streamCredentialsService.get(incomingPayment)
+    : deps.streamCredentialsService.get({
+        paymentTag: incomingPayment.id,
+        asset: {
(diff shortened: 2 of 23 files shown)


packages/backend/src/open_payments/payment/incoming/model.test.ts
```diff
@@ -112,67 +114,6 @@ describe('Models', (): void => {
...
-          expiresAt: incomingPayment.expiresAt.toISOString(),
-          metadata: incomingPayment.metadata ?? undefined,
-          updatedAt: incomingPayment.updatedAt.toISOString(),
-          createdAt: incomingPayment.createdAt.toISOString(),
-          methods: []
-        })
-      })
-
-      test.each([IncomingPaymentState.Completed, IncomingPaymentState.Expired])(
-        'returns incoming payment with existing methods if payment state is %s',
```
mkurapov, line 139: These conditions were not being checked in the model anyway

packages/backend/src/open_payments/payment/incoming/routes.ts
```diff
@@ -109,14 +109,14 @@ async function getIncomingPaymentPrivate(
     )
   }
 
-  const streamCredentials = incomingPayment.isExpiredOrComplete()
-    ? undefined
-    : deps.streamCredentialsService.get(incomingPayment)
+  const paymentMethods = incomingPayment.isExpiredOrComplete()
+    ? []
+    : await deps.paymentMethodProviderService.getPaymentMethods(incomingPayment)
```
mkurapov, line 114: Now, no assumptions in the incoming payment routes about the type of payment methods we return, the routes don't care

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -349,10 +349,15 @@ async function createOutgoingPayment(
...
-          tenantId
+        const ilpPaymentMethod = receiver.paymentMethods.find(
+          (method) => method.type === 'ilp'
         )
+        const peer = ilpPaymentMethod
+          ? await deps.peerService.getByDestinationAddress(
+              ilpPaymentMethod.ilpAddress,
+              tenantId
+            )
+          : undefined
```
mkurapov, line 360: We only need this to store a `peerId` on the outgoingPayment, even though we don't use it for anything

packages/backend/src/open_payments/receiver/model.test.ts
```diff
@@ -124,44 +100,18 @@ describe('Receiver Model', (): void => {
...
-          streamCredentials
+          []
         )
 
       expect(
         () => new Receiver(openPaymentsIncomingPayment, false)
       ).not.toThrow()
     })
-
-    test('throws if stream credentials has invalid ILP address', async () => {
```
mkurapov, line 141: Logic no longer in the receiver's constructor

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -44,19 +39,6 @@ export class Receiver {
...
-    )
-    if (!ilpMethod) {
-      throw new Error('Cannot create receiver from unsupported payment method')
-    }
-    if (!isValidIlpAddress(ilpMethod.ilpAddress)) {
-      throw new Error('Invalid ILP address on ilp payment method')
-    }
-
-    this.ilpAddress = ilpMethod.ilpAddress
-    this.sharedSecret = base64url.toBuffer(ilpMethod.sharedSecret)
```
mkurapov, line 59: Not the concern of the receiver model anymore

packages/backend/src/open_payments/receiver/model.ts
```diff
@@ -100,13 +82,8 @@ export class Receiver {
     return undefined
   }
 
-  public toResolvedPayment(): ResolvedPayment {
-    return {
-      destinationAsset: this.asset,
-      destinationAddress: this.ilpAddress,
-      sharedSecret: this.sharedSecret,
-      requestCounter: Counter.from(0) as Counter
-    }
```
mkurapov, line 109: Decouples ILP stuff from receiver

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -395,6 +398,39 @@ async function pay(
...
+    )
+  }
+
+  return {
+    destinationAddress: ilpPaymentMethod.ilpAddress as IlpAddress,
+    destinationAsset: receiver.asset,
+    sharedSecret: base64url.toBuffer(ilpPaymentMethod.sharedSecret),
+    requestCounter: Pay.Counter.from(0) as Pay.Counter
+  }
+}
```
mkurapov, line 432: The logic that lived previously in the receiver, https://github.com/interledger/rafiki/pull/3460/files#diff-6c83fb3a86095e274a059aa3299a1b2f4ad0560722d661de537e7b8296c34e71

now we just get ilpDetails from the payment methods, when the caller needs it, like in this ilpPaymentService

packages/backend/src/payment-method/ilp/connector/core/controllers/stream.ts
```diff
@@ -12,18 +12,18 @@ export function createStreamController(): ILPMiddleware {
     ctx: ILPContext,
     next: () => Promise<void>
   ): Promise<void> {
-    const { logger, redis, streamServer } = ctx.services
+    const { logger, redis } = ctx.services
     const { request, response } = ctx
 
     if (
       ctx.accounts.outgoing.http ||
-      !streamServer.decodePaymentTag(request.prepare.destination) // XXX mark this earlier in the middleware pipeline
```
mkurapov, line 20: equivalent

packages/backend/src/payment-method/ilp/peer/service.ts
```diff
@@ -373,14 +374,28 @@ async function getPeerByDestinationAddress(
     peerQuery.andWhere('tenantId', tenantId)
   }
 
-  const peer = await peerQuery.first()
+  const peers = await peerQuery
+  const peer = getByLongestPrefixMatch(peers, destinationAddress)
```
mkurapov, line 378: @sanducb since we end up matching to multiple peers with one staticIlpAddress, I ended up needing to resolve the correct one using `getByLongestPrefixMatch`. This has multi-hop implications too likely, so we can potentially come up with a more ergonomic solution for this in the future

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
```diff
@@ -1,20 +1,45 @@
...
 }
+
+export async function getIlpAddressForTenant(ctx: ILPContext): Promise<string> {
+  const tenantId = ctx.state.incomingAccount?.tenantId
+
+  if (tenantId === ctx.services.config.operatorTenantId) {
+    return ctx.services.config.ilpAddress
+  }
+
+  const tenantSettings = await ctx.services.tenantSettingService.get({
```
sanducb, line 55: I think we should cache the tenant ILP addresses as this will be called for each prepare packet. We could leave this as is for now though since we can store it as part of the in memory routing table when the multi-hop feature is merged.

mkurapov: @sanducb agreed, I think we should just cache the whole tenantSettings list -> it changes very rarely

packages/backend/src/payment-method/ilp/peer/service.test.ts
```diff
@@ -431,7 +431,7 @@ describe('Peer Service', (): void => {
       })
 
       await expect(
-        peerService.getByDestinationAddress('test.rafiki', tenantId)
+        peerService.getByDestinationAddress('test.rafiki', tenantId, asset.id)
```
sanducb, line 434: Is `asset.id` necessary here?

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
```diff
@@ -1,20 +1,45 @@
...
+    tenantId,
+    key: TenantSettingKeys.ILP_ADDRESS.name
+  })
+
+  if (!tenantSettings.length) {
```
sanducb: In the case where a tenant exists and reaches this condition which is evaluated as true i.e. tenant does not have it's own ILP address in the tenant settings table, then should we continue and fall back to the operator's address?

packages/backend/src/payment-method/ilp/peer/service.ts
sanducb: I agree that we should find a more ergonomic solution in the future. The only way I can think about improving this right now is making `getPeerByDestinationAddress` the single point where the correct peer address is resolved regardless if we're talking about finding the correct tenanted peer or next hop for routing (or even unify these 2 use cases into routing). Since `getPeerByDestinationAddress` is also used in the [account middleware](https://github.com/interledger/rafiki/blob/b529ae52344720bcfe7ada4a2100fe73ee7996f2/packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts#L107C32-L107C55) for determining the outgoing account, we could get rid entirely of the routing middleware. This is another topic that we can continue after merging this PR, so I would say that we can leave this logic as is.

packages/backend/src/payment-method/ilp/peer/service.test.ts
mkurapov: Not necesary per se (since we would end up returning the first peer), but better in showing how querying for different assetId returns the correct peer

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
mkurapov: @sanducb I ended up adding types to the middleware, and making streamServer undefined if the ilp address for tenant is missing

njlie (approved):
LGTM. Just a +1 from me on caching the tenant settings.

PR #3495, opened 2025-07-04
feat(backend): add tenantId to webhook pagination resolver

njlie:

- Adds a "Tenant" column to various tables in the Admin UI, when looking up tenanted resources.
- Adds resolvers to GraphQL resolver schema that retrieves the tenant for the requested resource.

## Context

Fixes RAF-1073.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/webhook/service.ts
+++ b/packages/backend/src/webhook/service.ts
@@ -296,7 +296,7 @@ async function getWebhookEventsPage(
     query.whereNotIn('type', filter.type.notIn)
   }
 
-  return await query.getPage(pagination, sortOrder).withGraphFetched('tenant')
+  return await query.getPage(pagination, sortOrder)
 }
 
 type FinalizeRecipientsOptions = {


packages/backend/src/graphql/resolvers/asset.ts
```diff
@@ -185,6 +191,27 @@ export const getFees: AssetResolvers<TenantedApolloContext>['fees'] = async (
...
+      throw new GraphQLError(errorToCode[AssetError.UnknownAsset], {
+        extensions: {
+          code: errorToCode[AssetError.UnknownAsset]
+        }
+      })
+
+    const tenantService = await ctx.container.use('tenantService')
+    const tenant = await tenantService.get(asset.tenantId)
+    if (!tenant) return null
+    return tenantToGraphQl(tenant)
```
mkurapov, line 212: I think we can omit populating `settings` in the `tenantToGraphQl`. We do not return `settings` in `tenantService.get(...)`, (which is desired), but more importantly, we already have the 
```
  Tenant: {
    settings: getTenantSettings
  }
```
resolver, which will populate settings if asked for.

packages/documentation/src/content/docs/integration/deployment/docker-compose.mdx
```diff
@@ -128,6 +128,7 @@ rafiki-auth:
       AUTH_PORT: 3006
       INTROSPECTION_PORT: 3007
       INTERACTION_PORT: 3009
+      SERVICE_API_PORT: 3011
```
mkurapov: I think we can remove this (since its only for the service to service call)

packages/frontend/app/routes/webhook-events.tsx
```diff
@@ -129,13 +130,34 @@ export default function WebhookEventsPage() {
...
+                              </span>
+                            ) : (
+                              <span className='text-tealish/80'>
+                                No public name
+                              </span>
+                            )}
+                          </span>
+                          <div className='text-tealish/50 text-xs'>
+                            (ID:{' '}
+                            {truncateUuid(webhook.node.tenant?.id as string)})
```
mkurapov: Can we do this with something like `text-overflow: ellipsis;` instead?

njlie: To be honest, I don't love this either, but for `text-overflow` to work I think we'd need to rework the parent table elements to have overflow behavior that doesn't resort to adding a horizontal scrollbar. Since the scrollbar gives the table unlimited width, it's not possible to trigger any overflow in any child elements.

I suppose accounting for different viewport widths carries too many design implications for us to deal with right now, so I just got rid of this truncation.

packages/documentation/src/content/docs/integration/deployment/docker-compose.mdx
```diff
@@ -142,6 +142,7 @@ rafiki-auth:
       - '3006:3006'
       - '3007:3007'
       - '3009:3009'
+      - '3011:3011'
```
mkurapov: Can remove this as well

packages/backend/src/graphql/schema.graphql
```diff
@@ -993,10 +994,10 @@ type IncomingPayment implements BasePayment & Model {
   metadata: JSONObject
   "The date and time that the incoming payment was created."
   createdAt: String!
-  "The tenant UUID associated with the incoming payment. If not provided, it will be obtained from the signature."
-  tenantId: String
   "The sender's wallet address URL. Applicable only to card payments."
   senderWalletAddress: String
+  "The tenant associated with the incoming payment. If not provided, it will be obtained from the signature."
+  tenant: Tenant
```
mkurapov, line 1000:
```suggestion
  tenant: Tenant!
```
ditto across the rest of the file, since a tenant should be present for all resources

njlie: Making this required makes things a little tricky for the resolver implementation. The generated types will cause the resolvers' typing for paginating expect `tenant`, even though GraphQL should be delegating it to `getWebhookEventTenant` for example.

mkurapov: Yeah, I see this now. Would require a lot of `Omit`s on top of the generated types (in all of the resolvers) to make this work. I think keeping it as is, is sufficient

PR #3510, opened 2025-07-08
Added CardService client in rafiki backend

zeppelin44:

- 

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/card/service.ts
+++ b/packages/backend/src/card/service.ts
@@ -1,0 +1,36 @@
+import { AxiosInstance } from 'axios'
+import { Logger } from 'pino'
+
+const SERVICE_NAME = 'card-service'
+const PAYMENT_FOUNDED_PATH = '/payment-event'
+
+type EventDetails = {
+	requestId: string,
+	outgoingPaymentId: string,
+	resultCode: string
+}
+
+export interface CardService {
+	sendPaymentEvent(eventDetails: EventDetails): Promise<void>
+}
+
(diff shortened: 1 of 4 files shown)


packages/backend/src/card/service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+interface ServiceDependencies {
+  axios: AxiosInstance
+  cardServiceHost: string
+  logger: Logger
+}
+
+export async function createCardService(
+  deps: ServiceDependencies
+): Promise<CardService> {
```
njlie, line 24: We use [this pattern](https://github.com/interledger/rafiki/blob/364101fdc95af5c6b744dcdd3d7652360f881f93/packages/backend/src/open_payments/payment/incoming/service.ts#L79-L85) in order to help disambiguate logs from different services. It helps avoid having to pass in `serviceName` every time a service needs to log something.

packages/backend/src/card/service.ts
```diff
@@ -0,0 +1,54 @@
...
+  axios: AxiosInstance
+  cardServiceHost: string
+  logger: Logger
+}
+
+export async function createCardService(
+  deps_: ServiceDependencies
+): Promise<CardService> {
+  const logger = deps_.logger.child({
+    service: 'card-service'
```
njlie: nit: use PascalCase to match other child logger aliases

`card-service` -> `CardService`

PR #3514, opened 2025-07-08
feat(card-service): card payments table

lengyel-arpad85:


packages/card-service/migrations/20250708070327_card_payments_table.js
```diff
@@ -0,0 +1,31 @@
...
+
+    table.string('cardWalletAddress').notNullable()
+    table.string('incomingPaymentUrl').notNullable()
+
+    table.integer('statusCode').nullable()
+
+    table.uuid('outgoingPaymentId')
+    table.uuid('terminalId')
+
+    table.index('cardWalletAddress')
```
njlie, line 24: `createdAt` and `updatedAt` should also be added in order to match the base model

packages/card-service/src/shared/pagination.ts
```diff
@@ -0,0 +1,68 @@
+import { PaginationArgs } from '@interledger/open-payments'
+
+import { BaseModel, PageInfo, Pagination, SortOrder } from './baseModel'
+
+export function parsePaginationQueryParameters({
```
njlie: I think we can leave out this function and the `PaginationArgs` import that it depends on - this function is meant to parse pagination arguments for OP resources, which this package will not manage. The pagination module in the auth service more closely mirrors the pagination expectations of the card service.

njlie:
There also appears to be an issue with the lockfile that should get fixed before we merge this in.

packages/card-service/src/card-payment/model.ts
```diff
@@ -0,0 +1,16 @@
+import { BaseModel } from '../shared/baseModel'
+
+export class CardPayments extends BaseModel {
```
beniaminmunteanu: I think the model itself should be named `CardPayment`, and just keep the table name in plural 🤔

beniaminmunteanu (approved):

PR #3524, opened 2025-07-08
feat: Integrate Redis client in Card Service for (requestId, posServiceHost) mapping

zeppelin44:


packages/card-service/src/pos-store/service.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+const getPOS = async (
+  deps: StoreDependencies,
+  requestId: string
+): Promise<string> => {
+  const POSHost = await deps.redis.get(requestId)
+  if (!POSHost) {
+    deps.logger.error(`No POS found for requestId: ${requestId}`)
```
njlie: nitpick: With the pino logger we usually pass in an object with relevant data, then the message - for example:

```
deps.logger.error({ requestId }, 'No POS found for requestId')
```

packages/card-service/src/pos-store/service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  return POSHost
+}
+
+const addPOS = async (
+  deps: StoreDependencies,
+  requestId: string,
+  POSHost: string
+) => {
+  await deps.redis.set(requestId, POSHost)
```
njlie: Let's add an expiry as well. Maybe like five minutes or something.

zeppelin44: added 5 minutes for now, maybe we can have it configurable in the future - i dont think anybody will wait 5 minutes at a pos for a response

njlie (commented):
Looks good to me. I would consider merging in the changes from https://github.com/interledger/rafiki/pull/3533 or waiting for it to be merged into the feature branch to avoid having to redo the tests for this feature in another PR.

PR #3526, opened 2025-07-08
feat(pos): pos card services table

lengyel-arpad85:


packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
```diff
@@ -0,0 +1,35 @@
...
+    table.uuid('walletAddressId').notNullable()
+
+    table.string('deviceName').notNullable()
+    table.string('publicKey')
+    table.string('keyId')
+    table.string('algorithm')
+
+    table.string('status').notNullable()
+
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
```
njlie, line 24: Just need to add `updatedAt` to match the base model

njlie (approved):

PR #3528, opened 2025-07-08
feat(pos): merchants services

lengyel-arpad85:


packages/point-of-sale/src/config/app.ts
```diff
@@ -34,5 +34,7 @@ export const Config = {
   env: envString('NODE_ENV', 'development'),
   port: envInt('PORT', 3008),
   trustProxy: envBool('TRUST_PROXY', false),
-  enableManualMigrations: envBool('ENABLE_MANUAl_MIGRATIONS', false)
+  enableManualMigrations: envBool('ENABLE_MANUAl_MIGRATIONS', false),
+  dbSchema: undefined as string | undefined,
+  tigerBeetleClusterId: envInt('TIGERBEETLE_CLUSTER_ID', 0)
```
njlie: The point-of-sale service probably won't be using Tigerbeetle, so we can remove this env var.

packages/point-of-sale/src/tests/tigerbeetle.ts
```diff
@@ -0,0 +1,89 @@
+import { GenericContainer, StartedTestContainer, Wait } from 'testcontainers'
```
njlie: We can delete this, this service probably won't use it

packages/point-of-sale/jest.tigerbeetle-environment.ts
```diff
@@ -0,0 +1,21 @@
+import { startTigerBeetleContainer } from './src/tests/tigerbeetle'
```
njlie: Ditto to other comments regarding removing Tigerbeetle stuff

packages/point-of-sale/jest.env.js
```diff
@@ -0,0 +1,18 @@
+process.env.LOG_LEVEL = 'silent'
+process.env.INSTANCE_NAME = 'Rafiki'
+process.env.KEY_ID = 'myKey'
```
njlie: We probably don't need _all_ of these env vars

packages/point-of-sale/jest.setup.ts
```diff
@@ -0,0 +1,88 @@
+import { knex } from 'knex'
+import { GenericContainer, Wait } from 'testcontainers'
+require('./jest.env') // set environment variables
+
+const POSTGRES_PORT = 5432
+const REDIS_PORT = 6379
```
njlie: `REDIS_PORT` is unused in this service

packages/point-of-sale/jest.setup.ts
```diff
@@ -0,0 +1,88 @@
...
+      const workerDatabaseName = `testing_${i}`
+
+      await db.raw(`DROP DATABASE IF EXISTS ${workerDatabaseName}`)
+      await db.raw(`CREATE DATABASE ${workerDatabaseName} TEMPLATE testing`)
+    }
+
+    global.__BACKEND_KNEX__ = db
+  }
+
+  const setupRedis = async () => {
```
njlie: POS service doesn't use redis atm

packages/point-of-sale/jest.teardown.js
```diff
@@ -0,0 +1,13 @@
+module.exports = async () => {
+  await global.__BACKEND_KNEX__.migrate.rollback(
+    { directory: __dirname + '/migrations' },
+    true
+  )
+  await global.__BACKEND_KNEX__.destroy()
+  if (global.__BACKEND_POSTGRES__) {
+    await global.__BACKEND_POSTGRES__.stop()
+  }
+  if (global.__BACKEND_REDIS__) {
```
njlie: Ditto to previous comments about removing redis stuff

packages/point-of-sale/jest.setup.ts
```diff
@@ -0,0 +1,88 @@
...
+          retries: 5
+        })
+        .withWaitStrategy(Wait.forHealthCheck())
+        .start()
+
+      process.env.DATABASE_URL = `postgresql://postgres:password@localhost:${postgresContainer.getMappedPort(
+        POSTGRES_PORT
+      )}/testing`
+
+      global.__BACKEND_POSTGRES__ = postgresContainer
```
njlie: We should change the global variable to indicate that it's for the POS service instead of the backend

packages/point-of-sale/jest.teardown.js
```diff
@@ -0,0 +1,13 @@
+module.exports = async () => {
+  await global.__BACKEND_KNEX__.migrate.rollback(
```
njlie: Global vars should be namespaced to POS instead of backend here as well.

njlie (approved):

PR #3546, opened 2025-07-10
feat: [RAF-1083][POS Service]: Add GQL Client for Rafiki BE calls for incoming payment operations 

zeppelin44:

- Adding a GQL Client to the POS Service that will be used for incoming payment operations.

## Checklist

- [ ] Tests added/updated


packages/point-of-sale/src/graphql/schema.graphql
```diff
@@ -0,0 +1,1674 @@
+type Query {
```
njlie: Instead of copying the GQL schema here, we can add this package as a target in the `codegen.yml` file in `backend`. Then this will be generated with `pnpm --filter backend generate`.

packages/point-of-sale/jest.env.js
```diff
@@ -1,3 +1,8 @@
 // Jest environment configuration for point-of-sale
 process.env.NODE_ENV = 'test'
 process.env.LOG_LEVEL = process.env.LOG_LEVEL || 'silent'
+
+process.env.TENANT_ID = 'tenant_id'
+process.env.TENANT_SECRET = 'tenant_secret'
+process.env.TENANT_SIGNATURE_VERSION = 'tenant_signature_version'
+process.env.GRAPHQL_URL = 'http://127.0.0.1:3003/graphq'
```
njlie: nit: `graphq` is missing the last 'l'.

PR #3548, opened 2025-07-10
feat(point-of-sale): POS Device service

### LINKED ISSUE #3486: [POS Service] POS Device Service
The POS device service should implement the business logic of the routes exposed by the POS service API, then be added to the Rafiki backend's IoC container. The required features are as follows:

```javascript
// Register a device to a merchant
registerDevice( // May also be called createDevice
  merchantId: string // Retrieved from path params of request
  publicKey: string
  deviceName: string
  walletAddress: string
): POSDevice
```

```
// Retrieve device by key id
getByKeyId(
  keyId: string
): POSDevice
```

```typescript
// Revoke device key
// Soft-delete, changes 'status' field to "REVOKED"
revoke(
  id: string
): POSDevice
```

oana-lolea:

- Created POS Device service
- Changed POS Device to use a walletAddressUrl instead of its id and updated its migration
- Changed POS Device `keyId` to be nonNullable and unique because it will be generated on creation
- Added function that generates `keyId` based on the following structure: 
`pos:<first 6 chars from the deviceName (without whitespaces)><random 6 chars (using uuid)>` 
and added a test for it

## Context

Fixes #3486 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
+++ b/packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
@@ -15,7 +15,7 @@ exports.up = function (knex) {
     table.string('walletAddress').notNullable()
     table.string('deviceName').notNullable()
     table.string('publicKey')
-    table.string('keyId').notNullable()
+    table.string('keyId').notNullable().unique()
     table.string('algorithm')
 
     table.string('status').notNullable()
--- a/packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
+++ b/packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
@@ -12,11 +12,10 @@ exports.up = function (knex) {
       .onDelete('CASCADE')
       .index()
 
-    table.uuid('walletAddressId').notNullable()
-
+    table.string('walletAddress').notNullable()
     table.string('deviceName').notNullable()
     table.string('publicKey')
-    table.string('keyId')
+    table.string('keyId').notNullable()
     table.string('algorithm')
 
     table.string('status').notNullable()
--- a/packages/point-of-sale/src/merchant/devices/service.test.ts
+++ b/packages/point-of-sale/src/merchant/devices/service.test.ts
@@ -1,0 +1,119 @@
+import { IocContract } from '@adonisjs/fold'
+import { AppServices } from '../../app'
+import { Config } from '../../config/app'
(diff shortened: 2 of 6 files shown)


packages/point-of-sale/src/merchant/devices/service.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  revoke(id: string): Promise<PosDevice | PosDeviceError>
+}
+
+export interface CreateOptions {
+  merchantId: string
+  publicKey: string
+  deviceName: string
+  walletAddress: string
+}
```
oana-lolea, line 23: I've just realised that the algorithm is not provided when registering a device. It's supposed to, right? @njlie

packages/point-of-sale/migrations/20250708121812_create_pos_device_table.js
```diff
@@ -12,11 +12,10 @@ exports.up = function (knex) {
       .onDelete('CASCADE')
       .index()
 
-    table.uuid('walletAddressId').notNullable()
-
+    table.string('walletAddress').notNullable()
     table.string('deviceName').notNullable()
     table.string('publicKey')
-    table.string('keyId')
+    table.string('keyId').notNullable()
```
njlie: Would it be possible to give this a unique constraint?

packages/point-of-sale/src/merchant/devices/service.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    const merchant = await merchantService.create('merchant')
+    const createOptions: CreateOptions = {
+      merchantId: merchant.id,
+      publicKey: 'publicKey',
+      deviceName: 'device',
+      walletAddress: 'walletAddress'
+    }
+
+    const device = await posDeviceService.registerDevice(createOptions)
+    assert(!isPosDeviceError(device))
```
njlie, line 126: Let's also verify that the POS device object contains the input values as well.

For the keyId we can test for whether or not it's prefixed with `pos:` and the expected substring of the device name, then accept any string following that

packages/point-of-sale/src/merchant/devices/service.ts
njlie: I never got around to figuring out if a particular algorithm was required by the POS service, or if could be provided by the merchant.

Let's go with allowing the merchant to provide that and I'll make a note about clarifying that later.

njlie (approved):

PR #3555, opened 2025-07-10
feat(point-of-sale): added route for registering a POS device

### LINKED ISSUE #3517: [POS Service] POS Device Registration Route
The merchant service should expose an endpoint for registering POS devices to it. It will then call its POS Service to add an entry to the database.

- [ ] `POST /merchant/:merchantId/devices`
- [ ] Call `posService.create`

```
# POST /merchant/:merchantId/devices
## Request Body
{
  "publicKey": string
  "deviceName": string
  "walletAddress": string
}

## Response Body
### 200 OK
{
    "keyId": "pos:abc123",
    "algorithm": "ecdsa-p256-sha256"
}

### 400 Bad Request
```

oana-lolea:

- Added route for registering a POS Device 

## Context

Fixes #3517 

## Checklist

- [ ] Tests added/updated


packages/point-of-sale/src/merchant/devices/routes.test.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+      })
+      ctx.request.body = CREATE_BODY
+      ctx.params.merchantId = withMerchantId ?? merchantId
+      return ctx
+    }
+  })
+
+  async function createMerchant(name?: string) {
+    return await merchantService.create(name ?? 'Merchant')
```
beniaminmunteanu: Nit:

Multiple use of this method inside of the same test would result in violating the db's unique constraint of the name property.

Not currently an issue, but not having a default value for the name parameter might be safer for future use.

njlie: It's causing issues on my machine when I try to run the tests, I think.

njlie:
I think the test container app needs to have its configured port set to 0 like in the other packages. With multiple tests in the POS Service now it's causing some of them to fail due to port collisions.

packages/point-of-sale/src/merchant/devices/routes.test.ts
oana-lolea: I'll remove it then

njlie (approved):

PR #3560, opened 2025-07-11
feat(pos): device revoke

beniaminmunteanu:

- Nested routing structure for devices and merchants
- Changes in error handling
- Device revoke route

## Context

[Closes RAF-1122](https://linear.app/interledger/issue/RAF-1122/pos-service-revoke-pos-route)

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/point-of-sale/src/merchant/devices/routes.test.ts
+++ b/packages/point-of-sale/src/merchant/devices/routes.test.ts
@@ -7,10 +7,15 @@ import { createTestApp, TestContainer } from '../../tests/app'
 import { createContext } from '../../tests/context'
 import { truncateTables } from '../../tests/tableManager'
 import { MerchantService } from '../service'
-import { CreateBody, createPosDeviceRoutes, PosDeviceRoutes, RegisterDeviceContext, RevokeDeviceContext } from './routes'
+import {
+  CreateBody,
+  createPosDeviceRoutes,
+  PosDeviceRoutes,
+  RegisterDeviceContext,
+  RevokeDeviceContext
+} from './routes'
 import { PosDeviceService } from './service'
 import { faker } from '@faker-js/faker'
-
 
 describe('POS Device Routes', (): void => {
   let deps: IocContract<AppServices>
@@ -95,7 +100,9 @@ describe('POS Device Routes', (): void => {
         deviceId: uuid()
       }
 
-      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow('Device not found')
+      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow(
+        'Device not found'
+      )
     })
 
     test('Returns 404 for device that belongs to different merchant', async (): Promise<void> => {
@@ -123,7 +130,9 @@ describe('POS Device Routes', (): void => {
         deviceId: device.id
       }
 
-      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow('Device not found')
+      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow(
+        'Device not found'
+      )
     })
 
     test('Returns 404 for already deleted device', async (): Promise<void> => {
@@ -152,7 +161,9 @@ describe('POS Device Routes', (): void => {
         deviceId: device.id
       }
 
-      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow('Device not found')
+      await expect(posDeviceRoutes.revoke(ctx)).rejects.toThrow(
+        'Device not found'
+      )
     })
   })
 
@@ -185,7 +196,9 @@ describe('POS Device Routes', (): void => {
       })
       ctx.request.body = CREATE_BODY
       ctx.params.merchantId = uuid()
-      await expect(posDeviceRoutes.register(ctx)).rejects.toThrow('Unknown merchant')
+      await expect(posDeviceRoutes.register(ctx)).rejects.toThrow(
+        'Unknown merchant'
+      )
     })
   })
 })
--- a/packages/point-of-sale/src/merchant/devices/service.ts
+++ b/packages/point-of-sale/src/merchant/devices/service.ts
@@ -14,7 +14,10 @@ export interface PosDeviceService {
 
   revokeAllByMerchantId(merchantId: string): Promise<number>
 
-  getMerchantDevice(merchantId: string, deviceId: string): Promise<PosDevice | undefined>
+  getMerchantDevice(
+    merchantId: string,
+    deviceId: string
+  ): Promise<PosDevice | undefined>
 }
 
 export interface CreateOptions {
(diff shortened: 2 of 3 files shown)


packages/point-of-sale/src/merchant/devices/errors.ts
```diff
@@ -1,8 +1,15 @@
-export enum PosDeviceError {
-  UnknownMerchant = 'UnknownMerchant',
-  UnknownPosDevice = 'UnknownPosDevice'
-}
+export class POSDeviceError extends Error {
```
beniaminmunteanu, line 27: The error handling pattern here was inspired from the backend package.
I think it was a bit specialised for Graphql, which is not the case in the API of this service.

As such, i am opting for error classes, which also implies changing the strategy from returning errors to throwing errors. Thoughts?

packages/point-of-sale/src/merchant/routes.ts
```diff
@@ -46,7 +52,8 @@ export function createMerchantRoutes(
 
   return {
     create: (ctx: CreateMerchantContext) => createMerchant(deps, ctx),
-    delete: (ctx: DeleteMerchantContext) => deleteMerchant(deps, ctx)
+    delete: (ctx: DeleteMerchantContext) => deleteMerchant(deps, ctx),
+    devices: createDeviceRoutes(deps)
```
beniaminmunteanu: nesting devices in merchant routes

packages/point-of-sale/src/merchant/devices/routes.ts
```diff
@@ -0,0 +1,72 @@
...
+      .where({
+        id: deviceId,
+        merchantId: merchantId
+      })
+      .whereNull('deletedAt')
+      .first()
+
+    if (!device) {
+      throw new POSMerchantError(404, 'Device not found')
+    }
```
zeppelin44: I'd say this check should rather be included in the service `revoke` function 
OR
still in the service, but another function for checking the device existence under that merchant (like `deviceExists(deviceId, merchantId)`)

packages/point-of-sale/src/merchant/devices/routes.ts
```diff
@@ -0,0 +1,72 @@
...
+  request: RevokeDeviceRequest
+}
+
+export interface DeviceRoutes {
+  revoke(ctx: RevokeDeviceContext): Promise<void>
+}
+
+export function createDeviceRoutes(deps_: ServiceDependencies): DeviceRoutes {
+  const log = deps_.logger.child({
+    service: 'DeviceRoutes'
```
zeppelin44: Should we use `route` instead of `service` ?

packages/point-of-sale/src/merchant/devices/routes.ts
```diff
@@ -0,0 +1,72 @@
...
+export type RevokeDeviceContext = Exclude<AppContext, 'request'> & {
+  request: RevokeDeviceRequest
+}
+
+export interface DeviceRoutes {
+  revoke(ctx: RevokeDeviceContext): Promise<void>
+}
+
+export function createDeviceRoutes(deps_: ServiceDependencies): DeviceRoutes {
+  const log = deps_.logger.child({
```
zeppelin44: We dont seem to log anything

packages/point-of-sale/src/merchant/devices/errors.ts
zeppelin44: +1 for error classes
OR
have a standardized way of responding from a service, like
`type ServiceResponse<T> = {
   data: T,
   error?: {
      type: string,
      details: string
   }
}
`
so routes/services can check for error after each service call without try/catch blocks. I guess this can be discussed and standardized if agreed so

njlie: Yeah, since GraphQL may have multiple resolvers used in a single query, it's designed so that per-resolver errors may be returned and compiled together before the entire request throws.

The error classes pattern exists in backend for other services, so I think it makes sense to switch here as well.

packages/point-of-sale/src/merchant/devices/routes.ts
njlie: I think `service` is the field name specified in the `ChildLoggerOptions` type for the pino logger.

packages/point-of-sale/src/merchant/devices/routes.ts
oana-lolea: I agree. It could also be a method in `posDeviceService` to find by deviceId.

packages/point-of-sale/src/merchant/devices/service.ts
```diff
@@ -54,10 +54,10 @@ export async function createPosDeviceService({
 async function registerDevice(
   deps: ServiceDependencies,
   { merchantId, publicKey, deviceName, walletAddress, algorithm }: CreateOptions
-): Promise<PosDevice | PosDeviceError> {
+): Promise<PosDevice> {
   const merchant = await Merchant.query(deps.knex).findById(merchantId)
   if (!merchant) {
-    return PosDeviceError.UnknownMerchant
+    throw new POSDeviceError(404, 'Unknown merchant')
```
njlie: I generally think the routes service should be responsible for determining the HTTP response code

beniaminmunteanu: Please have another look, after trying out a couple different approaches, i suggest the current one:

Service, being the one that has the bussiness logic, thows specific errors, so that error messages properly describe the issue.
e.g. `UnknownMerchant` and `UnknownDevice` -> both are 404, but refer to different missing entities.

Routers don't directly throw service errors, but the router's specific error. (`POSDeviceError`)
In the devices and merchants case, I made the router constructor be able to handle a service error, and automatically map the error to the proper status code.

packages/point-of-sale/src/merchant/devices/routes.ts
beniaminmunteanu: logger is not optional in BaseService that this service's interface extends

packages/point-of-sale/src/merchant/devices/service.ts
```diff
@@ -119,6 +123,18 @@ async function revokeAllByMerchantId(
   return revokedCount
 }
 
+async function getMerchantDevice(
+  deps: ServiceDependencies,
+  id: string,
+  merchantId: string
+): Promise<PosDevice | undefined> {
+  const device = await PosDevice.query(deps.knex).findById(id)
```
njlie: This query can be expanded to include the `merchantId`, using `.where()`
```
POSDevice.query(deps.knex).where({ id, merchantId })
```

Then there doesn't have to be a check to see if the merchant id matches.

packages/point-of-sale/src/merchant/devices/routes.test.ts
```diff
@@ -0,0 +1,150 @@
+import { IocContract } from '@adonisjs/fold'
+import assert from 'assert'
```
njlie: nit: unused

njlie (commented):
LGTM, just one suggestion on a service call

njlie (approved):

njlie (approved):

Issue #3455, opened 2025-06-10
[Multi-Tenant} ILP Addresses per Tenant


mkurapov:
@nathan said:

> I'm having a bit of trouble dealing with the asymmetry of peering between a multiple tenants on instance of Rafiki and a single tenant on another instance.
>
> Given a Rafiki instance (A) with two tenants (A-1 and A-2), and a tenant on another instance (B-1), consider the scenario where we are trying to establish a peering relationship between the following:
>
> * A-1 <-> B-1
> * A-2 <-> B-1
>
> Since each peering relationship needs to establish a corresponding pair of incoming/outgoing http tokens, and meet a uniqueness constraint for those tokens, establishing the peering relationship on the B-side gets a little tricky.
>
> * Creating two peers on (B-1) causes a conflict where both entries point to the same ILP address using the same asset.
> * The peers for (A-1) and (A-2) can't use the same incoming token coming from (B-1) because it also creates a conflict.
>
> It might seem apparent to loosen the uniqueness constraint for httpTokens to allow a tenant id to distinguish them further, but I'm pretty sure on the connector level we wouldn't be able to apply this distinction. Thoughts?

PR #3566, opened 2025-07-18
feat(backend): multi-hop payments with static routing

### LINKED ISSUE #3444: Static routing implementation in connector
## Context

As a result of the high level research of routing variations, the best suited solution for our purposes is static routing. Following up on the draft implementation PR [here](https://github.com/interledger/rafiki/pull/3439), there are a few more points to be added:

* If next hop is the destination connector, then it shouldn't make any difference in the routing logic. Currently this case is handled separately.
* Static routes must be specified on peer creation and added to the in memory routing table.
* Wildcard/mask support i.e. being able to route all traffic that has destination in the format `test.*` through a specific next hop.
* (Optional) Each route should have a different weight based on the best exchange rate of all possible next hops. This is different from rate probing (AFAIK) as rate probing happens after the payment path is found. Weights should not be static but rather calculated and updated at some interval.

## To do

Finish implementation, update/add routing tests as well as test this feature on the localenv

sanducb:

- Adds multihop payment support through static routing -> longest prefix match
- Telemetry support for ILP packets processing time per operation i.e. `outgoing-payment`, `incoming-payment`, `routing` or `unknown` (rate probes fall into this category)
- Telemetry support for ILP payment round trip time
- Simplifies determining which tenant is the destination of a payment introduced in [this](https://github.com/interledger/rafiki/pull/3460/files) PR by making it part of the routing logic.

## Context

Closes #3444

### Overall setup and routing logic
The setup creates 3 instances where instance A is peered with B, B is peered with C (please check the setup for exact instance names).
Payments should be successful from A -> B -> C by using the existing Bruno collection. Instances A and C were kept as `cloud-nine` and `happy-life-bank` in order to have minimal changes of the existing setup.

At startup of a Rafiki instance, routes are loaded from the database and stored in the in memory routing table. All subsequent peer updates will also refresh the routing table. For backwards compatibility, if no routes exist then direct peers' address and asset id will be used to populate the routing table.

A routing table entry has the following structure:

| `tenantId:destination` | `next hop` | `asset id` |

where:
- `tenantId` is the tenant id of the caller
- `destination` is the static ILP address of the payment receiver.
- `next hop` is the peer id of the direct peer that will either route or be the destination of the packet
- `asset id` is the asset id of the next hop peer -> this field is mandatory when adding/removing a route but not when querying for the next hop, as one could or could not be interested in what asset the peering relationship has when forwarding the packet.

`tenantId:destination` is called `prefix` in the implementation and is the key of the table. **Longest prefix matching is done against this key.**

The routing logic is now also responsible for resolving the peering asymmetry issue described [here](https://github.com/interledger/rafiki/issues/3455) in a multi-tenanted environment.

## Telemetry

There are 2 key metrics added in this PR:
- `ilp_prepare_packet_processing_ms`: Measures the time it takes to process individual ILP prepare packets through the connector middleware and is a histogram with a label that denotes the operation of the packet (`outgoing_payment`, `incoming_payment`, `routing`, `unknown` -> which includes rate probes). In the ILP metrics Grafana dashboard you can see P50 and P95 percentiles panels for tracking latency.
- `ilp_payment_round_trip_ms`: Measures the round-trip time for completing ILP payment (on the sender side). This one is also a histogram and the average round-trip time can be seen in the dashboard.

## Local testing

Only non-tenanted setup will now have the multi-hop feature with 3 instances. 
Use any of the non-tenanted Open Payments Bruno collections as-is to test this flow.
Run `pnpm localenv:compose:multihop up` to spin it locally.

## Notes

- The reason why `maxPacketAmount` was added to the seed is that if it is not set, then rate probes will lock a big part of a peer's balance during quoting that will only be moved/released it on receiving fulfill/reject. The total amount locked is usually `10^13 + 10^12 + ... + 10^3` (adding each probe packet's decreasing value). This causes issues with payments throughput, as even a few payments could lock the whole balance of a peer. We want to mitigate that by setting `maxPacketAmount` to a reasonable value, such that rate probe packets will not lock any value until they match the expected `maxPacketAmount`. Therefore, expect to see error logs (caused by reject packets) for `AmountTooLargeError` when quoting until rate probes will match the `maxPacketAmount` set. Only `global-bank` needs `maxPacketAmount` set since this is relevant for "receiving" instances when quoting.

### DIFF
--- a/packages/card-service/src/graphql/generated/graphql.ts
+++ b/packages/card-service/src/graphql/generated/graphql.ts
@@ -342,6 +342,8 @@ export type CreatePeerInput = {
   maxPacketAmount?: InputMaybe<Scalars['UInt64']['input']>;
   /** Internal name of the peer. */
   name?: InputMaybe<Scalars['String']['input']>;
+  /** Routes for the peer. */
+  routes?: InputMaybe<Array<Scalars['String']['input']>>;
   /** ILP address of the peer. */
   staticIlpAddress: Scalars['String']['input'];
 };
@@ -1214,6 +1216,8 @@ export type Peer = Model & {
   maxPacketAmount?: Maybe<Scalars['UInt64']['output']>;
   /** Public name for the peer. */
   name?: Maybe<Scalars['String']['output']>;
+  /** Routes for the peer. */
+  routes: Array<Scalars['String']['output']>;
   /** ILP address of the peer. */
   staticIlpAddress: Scalars['String']['output'];
   /** Unique identifier of the tenant associated with the peer. */
@@ -1665,6 +1669,8 @@ export type UpdatePeerInput = {
   maxPacketAmount?: InputMaybe<Scalars['UInt64']['input']>;
   /** New public name for the peer. */
   name?: InputMaybe<Scalars['String']['input']>;
+  /** New routes for the peer. */
+  routes?: InputMaybe<Array<Scalars['String']['input']>>;
   /** New ILP address for the peer. */
   staticIlpAddress?: InputMaybe<Scalars['String']['input']>;
 };
@@ -2607,6 +2613,7 @@ export type PeerResolvers<ContextType = any, ParentType extends ResolversParentT
   liquidityThreshold?: Resolver<Maybe<ResolversTypes['UInt64']>, ParentType, ContextType>;
   maxPacketAmount?: Resolver<Maybe<ResolversTypes['UInt64']>, ParentType, ContextType>;
   name?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
+  routes?: Resolver<Array<ResolversTypes['String']>, ParentType, ContextType>;
   staticIlpAddress?: Resolver<ResolversTypes['String'], ParentType, ContextType>;
   tenantId?: Resolver<ResolversTypes['ID'], ParentType, ContextType>;
   __isTypeOf?: IsTypeOfResolverFn<ParentType, ContextType>;
--- a/packages/point-of-sale/src/graphql/generated/graphql.ts
+++ b/packages/point-of-sale/src/graphql/generated/graphql.ts
@@ -342,6 +342,8 @@ export type CreatePeerInput = {
   maxPacketAmount?: InputMaybe<Scalars['UInt64']['input']>;
   /** Internal name of the peer. */
   name?: InputMaybe<Scalars['String']['input']>;
+  /** Routes for the peer. */
+  routes?: InputMaybe<Array<Scalars['String']['input']>>;
   /** ILP address of the peer. */
   staticIlpAddress: Scalars['String']['input'];
 };
@@ -1214,6 +1216,8 @@ export type Peer = Model & {
   maxPacketAmount?: Maybe<Scalars['UInt64']['output']>;
   /** Public name for the peer. */
   name?: Maybe<Scalars['String']['output']>;
+  /** Routes for the peer. */
+  routes: Array<Scalars['String']['output']>;
   /** ILP address of the peer. */
   staticIlpAddress: Scalars['String']['output'];
   /** Unique identifier of the tenant associated with the peer. */
@@ -1665,6 +1669,8 @@ export type UpdatePeerInput = {
   maxPacketAmount?: InputMaybe<Scalars['UInt64']['input']>;
   /** New public name for the peer. */
   name?: InputMaybe<Scalars['String']['input']>;
+  /** New routes for the peer. */
+  routes?: InputMaybe<Array<Scalars['String']['input']>>;
   /** New ILP address for the peer. */
   staticIlpAddress?: InputMaybe<Scalars['String']['input']>;
 };
@@ -2607,6 +2613,7 @@ export type PeerResolvers<ContextType = any, ParentType extends ResolversParentT
   liquidityThreshold?: Resolver<Maybe<ResolversTypes['UInt64']>, ParentType, ContextType>;
   maxPacketAmount?: Resolver<Maybe<ResolversTypes['UInt64']>, ParentType, ContextType>;
   name?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
+  routes?: Resolver<Array<ResolversTypes['String']>, ParentType, ContextType>;
   staticIlpAddress?: Resolver<ResolversTypes['String'], ParentType, ContextType>;
   tenantId?: Resolver<ResolversTypes['ID'], ParentType, ContextType>;
   __isTypeOf?: IsTypeOfResolverFn<ParentType, ContextType>;


packages/backend/src/accounting/service.ts
```diff
@@ -155,6 +155,76 @@ interface CreateAccountToAccountTransferArgs {
   ): Promise<string[] | TransferError>
 }
 
+// Helper function to identify if an account is a peer account
+function isPeerAccount(account: LiquidityAccount): boolean {
```
sanducb: These logging functions will be removed after reviews. I will leave them here for now so that it's easy to trace peer liquidity before and after transfers.

localenv/cloud-nine-wallet/seed.yml
```diff
@@ -1,30 +1,21 @@
...
-    scale: 0
-    liquidity: 1000000
-    liquidityThreshold: 100000
 peeringAsset: 'USD'
 peers:
-  - initialLiquidity: '10000000'
-    peerUrl: http://happy-life-bank-backend:3002
-    peerIlpAddress: test.happy-life-bank
+  - initialLiquidity: '1000000000000'
+    peerUrl: http://intergalactic-bank-backend:3002
```
njlie: I think this breaks the most basic local environment setup. There might need to be different `docker-compose` files for the multihop environment that point to different seed files, so that `cloud-nine-wallet` can route to `happy-life-bank` with or without any hops in the middle.

packages/backend/src/payment-method/ilp/connector/ilp-routing/test/service.test.ts
```diff
@@ -0,0 +1,491 @@
+import { RouterService } from '../service'
```
njlie: nit: this seems arbitrarily nested inside of a `test` directory

njlie (commented):
Haven't finished reviewing everything, but wanted to share the comments I have already

packages/backend/src/payment-method/ilp/connector/ilp-routing/test/service.test.ts
sanducb: I tried to keep the folder structure of the old routing stuff but since we are not using any of that, I will delete all the unused files in `ilp-routing` and bring routing `service.ts` and `service.test.ts` on the same level.

localenv/cloud-nine-wallet/seed.yml
sanducb: I agree. Current `localenv` setup only makes sense for the purpose of testing the multi-hop logic. I will think about a way of making multi-hop coexist with all the other setups and bring it up here for discussion. Your suggestion [here](https://github.com/interledger/rafiki/pull/3566#issuecomment-3141167710) sounds like a good avenue to explore.

packages/backend/src/payment-method/ilp/peer/service.test.ts
```diff
@@ -374,6 +482,151 @@ describe('Peer Service', (): void => {
         peer
       )
     })
+
+    test('Updates peer routes in router service', async (): Promise<void> => {
```
njlie, line 486: It might be good for the update tests to also ensure that the prior routes were cleared before the new ones are synced with a `.toHaveBeenCalled()` check or something.

BlairCurrey (approved):
Looks good and working as

packages/backend/src/payment-method/ilp/peer/service.ts
```diff
@@ -257,9 +285,21 @@ async function updatePeer(
         }
       }
 
+      const updateData = { ...options }
+
+      if (options.routes !== undefined) {
+        const staticIlpAddress =
+          options.staticIlpAddress ?? existingPeer.staticIlpAddress
+        updateData.routes = [staticIlpAddress, ...options.routes]
+      }
```
mkurapov, line 294: Is the `staticIlpAddress` already added in `syncPeerRoutes`?

mkurapov (commented):
I think deleting a peer will be difficult given the FK constraints + the accounting data. Maybe we don't end up deleting any peers, and instead, create a separate `multihop` docker compose which allows pointing all of the containers to use a separate seed files. In the seed files, the peer which will be skipped for routing will just have "unreachable" ILP address & routes, kind of how you did it for global bank for the cloud nine wallet seed file.

packages/backend/src/payment-method/ilp/peer/service.ts
sanducb: Yes, in `syncPeerRoutes` we already add the `staticIlpAddress` to the routing table. Here we are making sure that we also update the routes db entry with the peer's static address since that is used [here](https://github.com/interledger/rafiki/pull/3566/files/ac0989f11b53c771a06db40617047f78f4bf0cb6) when loading the routes from the db. Redundancy in `syncPeerRoutes` exists to make sure that there is no way to not have the static address in the routing table (even if the peer might be modified to have invalid routes) by taking the address from the peer entity.

packages/backend/src/config/app.ts
```diff
@@ -195,6 +195,7 @@ export const Config = {
   walletAddressRedirectHtmlPage: process.env.WALLET_ADDRESS_REDIRECT_HTML_PAGE,
   localCacheDuration: envInt('LOCAL_CACHE_DURATION_MS', 15_000),
   operatorTenantId: envString('OPERATOR_TENANT_ID'),
+  enableIlpTiming: envBool('ENABLE_ILP_TIMING', true),
```
mkurapov:
```suggestion
  enableIlpTiming: envBool('ENABLE_ILP_TIMING_TELEMETRY', true),
```
I think it's clearer that way

packages/backend/src/graphql/schema.graphql
```diff
@@ -755,6 +759,8 @@ type Peer implements Model {
   liquidity: UInt64
   "The date and time when the peer was created."
   createdAt: String!
+  "Routes for the peer."
+  routes: [String!]
```
mkurapov:
```suggestion
  routes: [String!]!
```

localenv/global-bank/docker-compose.yml
```diff
@@ -0,0 +1,149 @@
...
+      USE_TIGERBEETLE: false
+      AUTH_SERVER_GRANT_URL: ${GLOBAL_BANK_AUTH_SERVER_DOMAIN:-http://global-bank-auth:3006}
+      AUTH_SERVER_INTROSPECTION_URL: http://global-bank-auth:3007
+      AUTH_ADMIN_API_URL: 'http://global-bank-auth:5003/graphql'
+      AUTH_ADMIN_API_SECRET: 'rPoZpe9tVyBNCigm05QDco7WLcYa0xMao7lO5KG1XG4='
+      AUTH_SERVICE_API_URL: 'http://global-bank-auth:5011'
+      ILP_ADDRESS: test.global-bank
+      ILP_CONNECTOR_URL: http://global-bank-backend:5002
+      STREAM_SECRET: BjPXtnd00G2mRQwP/8ZpwyZASOch5sUXT5o0iR5b5wU=
+      API_SECRET: iyIgCprjb9uL8wFckR+pLEkJWMB7FJhgkvqhTQR/964=
```
mkurapov:
```suggestion
      ADMIN_API_SECRET: iyIgCprjb9uL8wFckR+pLEkJWMB7FJhgkvqhTQR/964=
```

PR #3570, opened 2025-07-21
feat(frontend): use tenant wallet address prefix in admin UI

### LINKED ISSUE #3466: [Multi-Tenant] Use tenant's wallet address prefix in "Create Wallet Address" form
The form for creating a wallet address on the Rafiki Admin UI currently uses the operator's wallet address prefix to prepend the input for a newly created wallet address. This field should behave differently based on whether or not it is the operator or another tenant using it.

* If the operator is attempting to create a wallet address:
  * The prefix should reflect the wallet address prefix setting of the tenant that is currently selected in the existing dropdown list.
  * The dropdown list should not list tenants that do not have the wallet address prefix tenant setting configured.
  * A tooltip or subtitle under the field should note that some tenants may not appear if the wallet address prefix tenant setting is not configured.
* If a tenant is attempting to create a wallet address:
  * Use that tenant's wallet address prefix setting to populate the field.
  * If the tenant has not configured a wallet address prefix, display an error and a link to return to the wallet address list page. (Or to go to the tenant management page).

njlie:

- Updates Rafiki Admin UI to use the correct tenant's Wallet Address Prefix in the addon for the wallet address name.

## Context

Closes RAF-1075. Closes #3466.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/frontend/app/routes/wallet-addresses.create.tsx
+++ b/packages/frontend/app/routes/wallet-addresses.create.tsx
@@ -20,23 +20,32 @@ import {
 } from '~/shared/utils'
 import { checkAuthAndRedirect } from '../lib/kratos_checks.server'
 import { type LoaderFunctionArgs } from '@remix-run/node'
-import { whoAmI, loadTenants } from '~/lib/api/tenant.server'
+import { whoAmI, loadTenants, getTenantInfo } from '~/lib/api/tenant.server'
 
 export const loader = async ({ request }: LoaderFunctionArgs) => {
   const cookies = request.headers.get('cookie')
   await checkAuthAndRedirect(request.url, cookies)
 
   const assets = await loadAssets(request)
-  const { isOperator } = await whoAmI(request)
+  const { id, isOperator } = await whoAmI(request)
   let tenants
+  let tenantWAPrefix
   if (isOperator) {
     tenants = await loadTenants(request)
+  } else {
+    const tenant = await getTenantInfo(request, { id })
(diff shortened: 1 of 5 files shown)


packages/frontend/app/routes/wallet-addresses.create.tsx
```diff
@@ -20,23 +20,51 @@ import {
...
-    tenants = await loadTenants(request)
+    const loadedTenants = await loadTenants(request)
+    tenants = loadedTenants.filter(
+      (tenant) => findWASetting(tenant.node.settings) || tenant.node.id === id
+    )
+  } else {
+    const tenant = await getTenantInfo(request, { id })
+    const waPrefixSetting = findWASetting(tenant.settings)
+    if (!waPrefixSetting)
+      throw new Error('Wallet Address Prefix not configured!')
```
mkurapov: This shouldn't be possible, correct? Otherwise, maybe we shouldn't throw and just fallback to the open payments URL?

mkurapov (commented):
Looks good, just one clarification

packages/frontend/app/routes/wallet-addresses.create.tsx
njlie: We don't require this setting on tenant creation. So there's a state where a tenant may exist but not have the wallet address prefix setting configured. I'll change this to fall back on the default OP URL.

mkurapov (approved):

PR #3578, opened 2025-07-28
feat(point-of-sale): service for obtaining walletAddress by its url from Rafiki Backend

### LINKED ISSUE #3559: [POS Service] Service for obtaining walletAddress by its url from Rafiki Backend
Create a request to Rafiki backend to obtain wallet address details, that include its id and cardServiceUrl. The latter will be needed in `card-service-client` to initiate payment to customer's `card-service`.

oana-lolea:

- Added `axios` to package
- Created an axios get request to obtain wallet address details, since we need the url of the `card-service` to initiate the payment 

## Context

Fixes #3559

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/point-of-sale/src/payments/service.ts
+++ b/packages/point-of-sale/src/payments/service.ts
@@ -18,16 +18,19 @@ type ServiceDependencies = {
   axios: AxiosInstance
 }
 
-export type WalletAddress = {
+type OpenPaymentsWalletAddress = {
   id: string
   publicName?: string
   assetCode: string
   assetScale: number
   authServer: string
   resourceServer: string
-  cardService: string
 } & {
   [key: string]: unknown
+}
+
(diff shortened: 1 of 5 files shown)


packages/point-of-sale/src/payments/service.ts
```diff
@@ -9,18 +9,33 @@ import {
...
 
-type PaymentService = {
+export type WalletAddress = {
+  id: string
+  publicName?: string
+  assetCode: string
+  assetScale: number
+  authServer: string
+  resourceServer: string
+  cardService: string
```
njlie: Small thing, the `cardService` probably should be optional here since not every wallet address will support card payments. The POS service's payments service thus will need to be able to determine if the wallet address can support payments and fail the request accordingly, so the typing should support that scenario.

njlie (approved):

PR #3596, opened 2025-08-08
feat(backend): publish webhooks to POS service if applicable

njlie:

- Adds handling for metadata in incoming payments to create an additional webhook to be published to a Rafiki service's POS service on incoming payment completion, if configured.
- Adds `metadata` to the webhook model.

## Context

Fixes RAF-1141.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/resolvers/incoming_payment.test.ts
+++ b/packages/backend/src/graphql/resolvers/incoming_payment.test.ts
@@ -18,10 +18,10 @@ import { v4 as uuid } from 'uuid'
 import { IncomingPaymentService } from '../../open_payments/payment/incoming/service'
 import { AccountingService } from '../../accounting/service'
 import {
-  IncomingPaymentInitiationReason,
   IncomingPayment as IncomingPaymentModel,
   IncomingPaymentState
 } from '../../open_payments/payment/incoming/model'
+import { IncomingPaymentInitiationReason } from '../../open_payments/payment/incoming/types'
 import {
   IncomingPayment,
   IncomingPaymentResponse,
--- a/packages/backend/src/graphql/resolvers/incoming_payment.ts
+++ b/packages/backend/src/graphql/resolvers/incoming_payment.ts
@@ -5,10 +5,8 @@ import {
   IncomingPayment as SchemaIncomingPayment,
   QueryResolvers
 } from '../generated/graphql'
-import {
-  IncomingPayment,
-  IncomingPaymentInitiationReason
-} from '../../open_payments/payment/incoming/model'
+import { IncomingPayment } from '../../open_payments/payment/incoming/model'
+import { IncomingPaymentInitiationReason } from '../../open_payments/payment/incoming/types'
 import {
   isIncomingPaymentError,
   errorToCode,
--- a/packages/backend/src/open_payments/payment/incoming/service.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.ts
@@ -2,9 +2,9 @@ import {
   IncomingPayment,
   IncomingPaymentEvent,
   IncomingPaymentEventType,
-  IncomingPaymentInitiationReason,
   IncomingPaymentState
 } from './model'
+import { IncomingPaymentInitiationReason } from './types'
 import { AccountingService } from '../../../accounting/service'
 import { BaseService } from '../../../shared/baseService'
 import { Knex } from 'knex'
--- a/packages/backend/src/open_payments/payment/incoming/types.ts
+++ b/packages/backend/src/open_payments/payment/incoming/types.ts
@@ -1,0 +1,8 @@
+export enum IncomingPaymentInitiationReason {
+  // The incoming payment was initiated by a card payment.
+  Card = 'CARD',
+  // The incoming payemnt was initiated through Open Payments.
+  OpenPayments = 'OPEN_PAYMENTS',
+  // The incoming payment was initiated by the Admin API.
+  Admin = 'ADMIN'
+}
--- a/packages/backend/src/webhook/service.ts
+++ b/packages/backend/src/webhook/service.ts
@@ -16,7 +16,7 @@ import {
 } from '../tenants/settings/model'
 import { TenantSettingService } from '../tenants/settings/service'
 import { Logger } from 'pino'
-import { IncomingPaymentInitiationReason } from '../open_payments/payment/incoming/model'
+import { IncomingPaymentInitiationReason } from '../open_payments/payment/incoming/types'
 
 // First retry waits 10 seconds
 // Second retry waits 20 (more) seconds
--- a/packages/backend/src/graphql/resolvers/combined_payments.test.ts
+++ b/packages/backend/src/graphql/resolvers/combined_payments.test.ts
@@ -26,7 +26,7 @@ import {
 } from '../../open_payments/payment/combined/model'
 import { getPageTests } from './page.test'
 import { createTenant } from '../../tests/tenant'
-import { IncomingPaymentInitiationReason } from '../../open_payments/payment/incoming/model'
+import { IncomingPaymentInitiationReason } from '../../open_payments/payment/incoming/types'
 
 describe('Payment', (): void => {
   let deps: IocContract<AppServices>
--- a/packages/backend/src/graphql/resolvers/liquidity.test.ts
+++ b/packages/backend/src/graphql/resolvers/liquidity.test.ts
@@ -26,9 +26,9 @@ import {
 } from '../../open_payments/wallet_address/model'
 import {
   IncomingPayment,
-  IncomingPaymentEventType,
-  IncomingPaymentInitiationReason
+  IncomingPaymentEventType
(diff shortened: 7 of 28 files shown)


njlie:
I decided to include this information in the `metadata` of the incoming payment as it felt to small to justify its own model on or linked to the incoming payment model. It also felt metacontextual to OP in general so it seemed like a decent enough fit, but I'm open to a different model for storing this information.

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -372,7 +372,8 @@ async function handleDeactivated(
       tenantId: incomingPayment.tenantId,
       webhooks: finalizeWebhookRecipients(
         [incomingPayment.tenantId],
-        deps.config
+        deps.config,
+        incomingPayment.metadata
```
mkurapov: I think we should have an additional column on the incoming payment, (maybe something like `initiatedBy: card | open-payments | admin` off the top of my head), so its auditable and we don't have to add anything to hide `metadata` fields when the incoming payment is fetched via Open Payments

packages/backend/src/webhook/service.ts
```diff
@@ -295,7 +302,22 @@ export function finalizeWebhookRecipients(
...
   }
 
-  return [...tenantIdSet.values()].map((tenantId) => ({
+  let recipients: Pick<Webhook, 'recipientTenantId' | 'metadata'>[] = [
+    ...tenantIdSet.values()
+  ].map((tenantId) => ({
     recipientTenantId: tenantId
   }))
+
+  if (metadata?.isCardPayment && config.posServiceUrl) {
```
mkurapov: we should do a warn log at if its a card payment but the posServiceUrl is not set

packages/backend/src/webhook/service.ts
```diff
@@ -161,12 +161,18 @@ async function processNextWebhook(
         })
       }
 
-      const settings = await deps_.tenantSettingService.get({
-        tenantId: webhook.recipientTenantId
-      })
-      const formattedSettings = formatSettings(settings)
+      if (webhook.metadata?.sendToPosService) {
+        await sendWebhook(deps, webhook, {
+          webhookUrl: `${deps.config.posServiceUrl}/webhook`
```
mkurapov, line 171: maybe it would be good to have a separate posWebhookServiceUrl (since it can be hosted on a separate service). If you agree, we can turn it into a separate issue

njlie: Captured in https://github.com/interledger/rafiki/issues/3606 / [RAF-1145](https://linear.app/interledger/issue/RAF-1145/rafiki-backend-add-pos-webhook-service-url)

packages/backend/migrations/20250814213541_add_intiated_by_to_incoming_payment.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema.alterTable('incomingPayments', function (table) {
+    table.enum('initiatedBy', ['CARD', 'OPEN_PAYMENTS', 'ADMIN']).notNullable()
```
mkurapov: If the field is not nullable, we must write a migration for it. We either make it nullable, or we set `initiatedBy = OPEN_PAYMENTS` if `client` exists on the incoming payment, `ADMIN` otherwise.

packages/backend/src/config/app.ts
```diff
@@ -203,7 +203,16 @@ export const Config = {
...
+  posServiceUrl: optionalString(envString, 'POS_SERVICE_URL')
+}
+
+function optionalString(envGetter: (envVar: string) => string, envVar: string) {
+  try {
+    return envGetter(envVar)
+  } catch (err) {
+    return undefined
+  }
 }
```
mkurapov: If the function takes in only a string, then we can just use `envString` directly instead of having an `envGetter` param. Or we can keep it like before, calling the function `optional` and its return type is the return type of `envGetter`

packages/point-of-sale/src/payments/service.ts
```diff
@@ -75,7 +75,8 @@ const createIncomingPayment: FnWithDeps<
     variables: {
       walletAddressId,
       incomingAmount,
-      idempotencyKey: v4()
+      idempotencyKey: v4(),
+      isCardPayment: true
```
mkurapov, line 79: we should test that this variable is set

packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -104,7 +107,11 @@ export const createIncomingPayment: MutationResolvers<ForTenantIdContext>['creat
         : new Date(args.input.expiresAt),
       incomingAmount: args.input.incomingAmount,
       metadata: args.input.metadata,
-      tenantId
+      tenantId,
+      initiationReason:
+        ctx.isOperator && args.input.isCardPayment
```
mkurapov, line 117: do we need to check for `isOperator`?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -33,6 +34,7 @@ export interface CreateIncomingPaymentOptions {
   incomingAmount?: Amount
   metadata?: Record<string, unknown>
   tenantId: string
+  initiationReason?: IncomingPaymentInitiationReason
```
```suggestion
  initiationReason: IncomingPaymentInitiationReason
```

packages/backend/src/graphql/resolvers/incoming_payment.ts
njlie: I figure since the admin API is going to be used by tenants and there isn't (for now) a way to extend card payment support on a per-tenant basis, it made sense to restrict this to operator only for now.

packages/backend/migrations/20250814213541_add_intiated_by_to_incoming_payment.js
njlie: When you say migration, are you saying there needs to be another database migration to properly apply the not nullable property, or that the application code needs to be migrated to integrate with that properly?

mkurapov: This:
> are you saying there needs to be another database migration to properly apply the not nullable property

we need to backfill all existing incoming payments to make sure that field is defined for all of the rows, before we can make it non-nullable

packages/backend/src/graphql/resolvers/incoming_payment.ts
mkurapov: Makes sense, let's just add a warn log if its a card payment and the tenant is not an operator

packages/backend/src/graphql/resolvers/incoming_payment.test.ts
```diff
@@ -288,7 +304,125 @@ describe('Incoming Payment Resolver', (): void => {
...
+        const tenant = await createTenant(deps)
+        const tenantWalletAddress = await createWalletAddress(deps, {
+          tenantId: tenant.id
+        })
+        const payment = await createIncomingPayment(deps, {
+          walletAddressId: tenantWalletAddress.id,
+          client,
+          tenantId: tenant.id,
+          initiationReason: IncomingPaymentInitiationReason.Admin
+        })
```
mkurapov: Do we need to create this incoming payment here, before calling the mutation?
(same goes for the test above)

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -372,7 +376,9 @@ async function handleDeactivated(
       tenantId: incomingPayment.tenantId,
       webhooks: finalizeWebhookRecipients(
         [incomingPayment.tenantId],
-        deps.config
+        deps.config,
+        incomingPayment.metadata,
```
mkurapov: Now that we are not using `metadata` for card info, we should instead pass in `initiatedBy`

mkurapov (approved):
Final few comments

packages/backend/src/graphql/resolvers/incoming_payment.test.ts
njlie: Frankly, I copied this from the pattern of other tests in this file but it's not really necessary.

packages/backend/src/open_payments/payment/incoming/types.ts
```diff
@@ -0,0 +1,8 @@
+export enum IncomingPaymentInitiationReason {
```
njlie, line 1: I had to create a types file here because `webhook/service.ts` importing from `incoming/model.ts` was causing a bunch of dependency errors in tests because the Incoming Payment model hadn't been initialized for the test and would break all the exports in the module, even if it wasn't importing the incoming payment model itself.

njlie: I'll capture a code-cleanup issue to move more types from the Incoming payment model into this types file so the code is cleaner, and to keep this PR from not being unnecessarily large.

PR #3600, opened 2025-08-13
fix(backend): made CARD_SERVICE_URL optional

### LINKED ISSUE #3577: [Backend] Make CARD_SERVICE_URL env optional
Not every Rafiki operator will want to enable cards, so it should be able to launch the backend without having the `CARD_SERVICE_URL` env variable configured.

Likewise, we should not display the `cardServiceUrl` under each `walletAddress` when it is fetched using Open Payments.

oana-lolea:

- `CARD_SERVICE_URL` is now optional
- `OpenPaymentsWalletAddress` does not return `cardServiceUrl` if it's not set

## Context

Fixes #3577 
Linear [#1113](https://linear.app/interledger/issue/RAF-1134/backend-make-card-service-url-env-optional)

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/card/service.ts
+++ b/packages/backend/src/card/service.ts
@@ -36,6 +36,14 @@ export async function createCardService(
   }
 }
 
+export async function createNoopCardService(): Promise<CardService> {
+  return {
+    async sendPaymentEvent(_eventDetails) {
+      // do nothing
+    }
+  }
+}
+
 async function sendPaymentEvent(
   deps: ServiceDependencies,
   eventDetails: EventDetails
--- a/packages/backend/src/card/service.ts
+++ b/packages/backend/src/card/service.ts
@@ -36,10 +36,15 @@ export async function createCardService(
   }
 }
 
-export async function createNoopCardService(): Promise<CardService> {
+export async function createNoopCardService(
+  logger: Logger
+): Promise<CardService> {
   return {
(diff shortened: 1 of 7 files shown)


packages/backend/src/index.ts
```diff
@@ -672,13 +672,16 @@ export function initIocContainer(
     })
   })
 
-  container.singleton('cardService', async (deps) => {
-    return createCardService({
-      axios: await deps.use('axios'),
-      logger: await deps.use('logger'),
-      cardServiceUrl: config.cardServiceUrl
+  if (config.cardServiceUrl) {
```
mkurapov: Let's remove this `if` and keep the `cardService` service defined, otherwise we might get unexpected behaviour if a singleton resolves to undefined.

mkurapov: We can do a `NoopCardService` (similar to the `NoopTelemetryService`)

packages/backend/src/card/service.ts
```diff
@@ -36,6 +36,14 @@ export async function createCardService(
   }
 }
 
+export async function createNoopCardService(): Promise<CardService> {
+  return {
+    async sendPaymentEvent(_eventDetails) {
+      // do nothing
```
mkurapov, line 44: Let's have a `warn` log here.

PR #3613, opened 2025-08-21
feat(point-of-sale): handle incoming payment completed webhooks from backend

njlie:

- Adds routes to handle incoming webhooks from a POS service's associated Rafiki backend.
- Adds a request map that the POS service's `/payment` route uses to wait for `incoming_payment.completed` webhooks when creating a card payment.

## Context

Fixes RAF-1140.

## Checklist

- [x] Tests added/updated


packages/point-of-sale/src/payments/routes.ts
```diff
@@ -61,31 +71,57 @@ async function payment(
...
-      incomingPaymentUrl,
+      incomingPaymentUrl: incomingPayment.id,
       date: new Date(),
       signature: body.signature,
       card: body.card
     })
 
+    if (result !== Result.APPROVED) throw new InvalidCardPaymentError(result)
+    const event = await waitForIncomingPaymentEvent(deps.config, deferred)
+    webhookWaitMap.delete(incomingPayment.id)
```
mkurapov, line 94: we should put this in a `finally` block, to make sure we always clear the object from the map

packages/point-of-sale/src/payments/routes.ts
```diff
@@ -61,31 +71,57 @@ async function payment(
...
+    const deferred = new Deferred<WebhookBody>()
+    webhookWaitMap.setWithExpiry(
+      incomingPayment.id,
+      deferred,
+      deps.config.webhookTimeoutMs
+    )
     const result = await deps.cardServiceClient.sendPayment({
       merchantWalletAddress: body.merchantWalletAddress,
-      incomingPaymentUrl,
+      incomingPaymentUrl: incomingPayment.id,
```
mkurapov: The `incomingPayment.id` is a UUID of the incoming payment, we need to  expose the  open payments url of the incoming payment in the mutation response so cardService can pay into it.

localenv/cloud-nine-wallet/dbinit.sql
```diff
@@ -23,7 +23,7 @@ CREATE DATABASE cloud_nine_wallet_point_of_sale;
 ALTER DATABASE cloud_nine_wallet_point_of_sale OWNER TO cloud_nine_wallet_point_of_sale;
 
 CREATE USER happy_life_bank_point_of_sale WITH PASSWORD 'happy_life_bank_point_of_sale';
-CREATE DATABASE happy_life_bank_point_of_sale;
+CREATE DATABASE happy_life_bank_point_ofW_sale;
```
mkurapov: Probably not intended

bruno/collections/Rafiki/environments/Local Playground.bru
```diff
@@ -32,8 +32,8 @@ vars {
   assetScale: 2
   senderTenantId: 438fa74a-fa7d-4317-9ced-dde32ece1787
   RafikiGraphqlHostTenantId: 438fa74a-fa7d-4317-9ced-dde32ece1787
-  senderOpenPaymentsPort: 3000
-  receiverOpenPaymentsPort: 4000
+  senderOpenPaymentsPort: 4000
```
sanducb: Is there a reason for swapping these values?

packages/point-of-sale/src/config/app.ts
```diff
@@ -41,6 +41,9 @@ export const Config = {
   dbSchema: undefined as string | undefined,
   tenantId: envString('TENANT_ID'),
   tenantSecret: envString('TENANT_SECRET'),
-  tenantSignatureVersion: envString('TENANT_SIGNATURE_VERSION'),
-  graphqlUrl: envString('GRAPHQL_URL')
+  tenantSignatureVersion: envInt('TENANT_SIGNATURE_VERSION', 1),
+  graphqlUrl: envString('GRAPHQL_URL'),
+  webhookSignatureVersion: envInt('WEBHOOK_SIGNATURE_VERSION', 1),
+  webhookSignatureSecret: envString('WEBHOOK_SIGNATURE_SECRET'),
+  webhookTimeoutMs: envInt('WEBHOOK_TIMEOUNT_MS', 30000)
```
sanducb: I think there's a typo here: `WEBHOOK_TIMEOUNT_MS` -> `WEBHOOK_TIMEOUT_MS`

bruno/collections/Rafiki/environments/Local Playground.bru
njlie: I think this got caught up in some rebasing I was doing - this change shouldn't have been intended. Thanks for catching!

sanducb (approved):

packages/point-of-sale/src/payments/routes.ts
njlie: I decided to instead augment the timeout error for incoming payment completed webhooks with the incoming payment id, so that it can be handled in the `catch` block and avoid needing to nest too much - the map uses the incoming payment id as a key, which depends on the also-fallible call to create the incoming payment.

PR #3615, opened 2025-08-26
chore(backend, auth): use open-payments-specifications as submodule

### LINKED ISSUE #3608: Use `open-payments-specifications` as a submodule
### Overview

Now that we have the `open-payments-specifications` repository, we should have Rafiki use it as a submodule in order to access the specs.

Currently, in `backend` and `auth`, we use the `OpenAPI` objects exported from the `open-payments` Node client to validate the incoming Open Payments RS & AS requests (this is done using the `createValidatorMiddleware` function). This means that the node client must always be on the latest version of the specs, and must be updated *before* we can support any new Open Payments features in Rafiki. This issue aims to decouple the specs from the client in Rafiki.

In order to do this correctly, we must copy over the specifications during the building the `auth` and `backend` Docker images.

### Notes

Using symlinks is could be an option (ie symlinking from the open-payments-specifications submodule into an `openapi` folder in `auth` and `backend`), but keep in mind that they may work incorrectly when using symlinks in a Windows environment.

### Steps

* Add `open-payments-specifications` repository as a submodule, tagged at version 1.0.3
* Instead of using the `getResourceServerOpenAPI` and`getWalletAddressServerOpenAPI` functions to generate the OpenAPI validators for the Open Payments routes, use the `createOpenAPI` function from the `@interledger/openapi` library to point to a local copy of the OpenAPI specs

sanducb:

This PR adds the open-payments-specifications submodule (v1.0.3), loading specs via createOpenAPI and include them into backend and auth prod images.

## Context

Closes #3608 (RAF-1147)

## Checklist

- [ ] Tests added/updated


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "open-payments-specifications"]
```
mkurapov, line 1: We should add a note in the README (probably both in the root README and the local playground) that we need to initialize the submodule when using this repository. 
Something like this: https://github.com/interledger/open-payments-node?tab=readme-ov-file#local-development-environment

.github/workflows/node-build.yml
```diff
@@ -69,6 +81,9 @@ jobs:
     timeout-minutes: 5
```
mkurapov: for `token-introspection` we need to update how we get the specfications into the project. Currently, in `token-instrospection/package.json` we copy over the specs from the NodeSDK, and then use a symlink into `token-introspection/src/openapi`. Instead, we should use the specs from the submodule

.github/workflows/node-build.yml
```diff
@@ -84,6 +99,9 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
```
mkurapov: we should only recursively fetch for the action jobs/steps that require the specs to be loaded, which should only be `auth` `backend` `integration-test` and `performance-test`

.github/workflows/node-build.yml
```diff
@@ -396,13 +426,19 @@ jobs:
...
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
       - name: Build
         uses: docker/build-push-action@v5
         with:
           push: false
           platforms: ${{ matrix.platform.arch }}
+          context: .
```
mkurapov: Is `context` required here?

packages/token-introspection/src/openapi/token-introspection.yaml
```diff
@@ -68,7 +68,7 @@ paths:
                   type: string
```
mkurapov: Let's keep just one OpenAPI spec - we can remove this one

packages/token-introspection/package.json
```diff
@@ -10,9 +10,7 @@
     "build": "pnpm clean && tsc --build tsconfig.json && pnpm copy-files",
     "clean": "rm -fr dist/",
     "copy-files": "mkdir -p ./dist/openapi/specs && cp -r ./src/openapi/specs/*.yaml ./dist/openapi/specs/",
-    "copy-op-schemas": "cp ./node_modules/@interledger/open-payments/dist/openapi/specs/auth-server.yaml ./node_modules/@interledger/open-payments/dist/openapi/specs/schemas.yaml ./src/openapi/specs/",
     "generate:types": "openapi-typescript ../../openapi/token-introspection.yaml --output src/openapi/generated/types.ts -t",
```
mkurapov:
```suggestion
    "generate:types": "openapi-typescript src/openapi/specs/token-introspection.yaml --output src/openapi/generated/types.ts -t",
```

PR #3623, opened 2025-08-28
chore(backend): add expiry for incoming payments created by pos service

### LINKED ISSUE #3601: [POS Service] Incoming Payment expiry
Whenever an incoming payment is created from the POS service, we should define a reasonable expiry date (`expiresAt` ), such that we have an upper limit as to when the card payment should be completed (and the result is sent back to the POS).

Let's have an optional env variable to set this expiry, with a default value of 10 seconds.

sanducb:

This PR adds expiry for incoming payments created by the POS service

## Context

Closes #3601 

## Checklist

- [ ] Tests added/updated


packages/point-of-sale/src/payments/service.test.ts
```diff
@@ -58,12 +58,13 @@ describe('createPaymentService', () => {
...
       expect.objectContaining({
-        variables: {
+        variables: expect.objectContaining({
           walletAddressId,
           incomingAmount,
           idempotencyKey: expect.any(String),
-          isCardPayment: true
-        }
+          isCardPayment: true,
+          expiresAt: expect.any(String)
```
mkurapov: we can mock Date.now and specify the actual value for `expiresAt` we expect here

packages/point-of-sale/src/payments/service.test.ts
```diff
@@ -33,7 +33,13 @@ describe('createPaymentService', () => {
     jest.clearAllMocks()
   })
 
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
   it('should create an incoming payment and return the incoming payment url (id)', async () => {
+    const now = 0
+    jest.spyOn(Date, 'now').mockReturnValue(now)
```
mkurapov: I t
```suggestion
  afterEach(() => {
    jest.useRealTimers()
  })

it('should create an incoming payment and return the incoming payment url (id)', async () => {
    const now = new Date().getTime()
    jest.useFakeTimers({ now })
```

I believe it is safer to use the built-in Jest timer functions when using fake timers.

mkurapov (approved):

PR #3651, opened 2025-09-17
feat(backend): additional webhook events for outgoing payments

sanducb:

This PR adds new webhook events for outgoing payments.
In order not to have the POS service waiting until the incoming payment expires, we have to reply to the POS service request with the ASEs decision on the outgoing payment. This is done by having backend send new webhooks to the card service: `outgoing_payment.funded` and `outgoing_payment.cancelled`.

## Context

Closes RAF-1156

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/migrations/20250917082624_add_request_id_to_outgoing_payment_card_details.js
+++ b/packages/backend/migrations/20250917082624_add_request_id_to_outgoing_payment_card_details.js
@@ -1,0 +1,37 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  const hasColumn = await knex.schema.hasColumn(
+    'outgoingPaymentCardDetails',
+    'requestId'
+  )
+  if (!hasColumn) {
+    await knex.schema.alterTable(
+      'outgoingPaymentCardDetails',
+      function (table) {
+        table.string('requestId').nullable()
+      }
+    )
+  }
+}
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = async function (knex) {
+  const hasColumn = await knex.schema.hasColumn(
+    'outgoingPaymentCardDetails',
+    'requestId'
+  )
+  if (hasColumn) {
+    await knex.schema.alterTable(
+      'outgoingPaymentCardDetails',
+      function (table) {
+        table.dropColumn('requestId')
+      }
+    )
+  }
+}
--- a/packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
+++ b/packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
@@ -213,15 +213,16 @@ export async function sendWebhookEvent(
     callName: 'OutgoingPaymentLifecycle:sendWebhookEvent'
   })
   // TigerBeetle accounts are only created as the OutgoingPayment is funded.
-  // So default the amountSent and balance to 0 for outgoing payments still in the funding state
-  const amountSent =
-    payment.state === OutgoingPaymentState.Funding
-      ? BigInt(0)
-      : await deps.accountingService.getTotalSent(payment.id)
-  const balance =
-    payment.state === OutgoingPaymentState.Funding
-      ? BigInt(0)
-      : await deps.accountingService.getBalance(payment.id)
+  // So default the amountSent and balance to 0 for outgoing payments still in the funding or cancelled states
+  const isZeroAmountSent =
+    payment.state === OutgoingPaymentState.Funding ||
+    type === OutgoingPaymentEventType.PaymentCancelled
+  const amountSent = isZeroAmountSent
+    ? BigInt(0)
+    : await deps.accountingService.getTotalSent(payment.id)
+  const balance = isZeroAmountSent
+    ? BigInt(0)
+    : await deps.accountingService.getBalance(payment.id)
 
   if (amountSent === undefined || balance === undefined) {
     stopTimer()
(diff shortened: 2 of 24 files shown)


packages/backend/migrations/20250917082624_add_request_id_to_outgoing_payment_card_details.js
```diff
@@ -0,0 +1,37 @@
...
+exports.up = async function (knex) {
+  const hasColumn = await knex.schema.hasColumn(
+    'outgoingPaymentCardDetails',
+    'requestId'
+  )
+  if (!hasColumn) {
+    await knex.schema.alterTable(
+      'outgoingPaymentCardDetails',
+      function (table) {
+        table.string('requestId').nullable()
```
mkurapov: Let's make this non-nullable

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -4,6 +4,7 @@ import { DbErrors } from 'objection-db-errors'
 import { LiquidityAccount } from '../../../accounting/service'
 import { Asset } from '../../../asset/model'
 import { ConnectorAccount } from '../../../payment-method/ilp/connector/core/rafiki'
+import { OutgoingPaymentInitiationReason } from './types'
```
mkurapov: Nit, but we can keep `OutgoingPaymentInitiationReason` here

packages/backend/src/webhook/service.ts
```diff
@@ -308,29 +312,36 @@ export function finalizeWebhookRecipients(
...
-      }
-    ])
-  } else if (
-    initiationReason === IncomingPaymentInitiationReason.Card &&
-    !config.posWebhookServiceUrl
-  ) {
-    logger?.warn('Could not create webhook recipient for point of sale service')
+  if (initiationReason === IncomingPaymentInitiationReason.Card) {
+    recipients = recipients.concat(buildPosRecipient())
   }
```
mkurapov: These new outgoing payment events should be sent only to card service, and not POS service (outgoing payments -> card service, incoming payments -> POS Service).

I think we should also make this function take in an object instead of a list of params for readability

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -213,15 +213,16 @@ export async function sendWebhookEvent(
...
-      ? BigInt(0)
-      : await deps.accountingService.getTotalSent(payment.id)
-  const balance =
-    payment.state === OutgoingPaymentState.Funding
-      ? BigInt(0)
-      : await deps.accountingService.getBalance(payment.id)
+  // So default the amountSent and balance to 0 for outgoing payments still in the funding or cancelled states
+  const isZeroAmountSent =
+    payment.state === OutgoingPaymentState.Funding ||
+    type === OutgoingPaymentEventType.PaymentCancelled
```
mkurapov:
```suggestion
    payment.state === OutgoingPaymentState.Cancelled
```

packages/backend/src/asset/model.ts
```diff
@@ -37,6 +36,7 @@ export class Asset extends BaseModel implements LiquidityAccount {
   ): Promise<Asset> {
     if (this.liquidityThreshold !== null) {
       if (balance <= this.liquidityThreshold) {
+        const { finalizeWebhookRecipients } = await import('../webhook/service')
```
mkurapov: Why do we need the dynamic import here?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -270,6 +276,15 @@ async function cancelOutgoingPayment(
       payment.walletAddressId
     )
 
+    if (payment.initiatedBy === OutgoingPaymentInitiationReason.Card) {
```
mkurapov, line 279: I'm debating whether we keep this conditional, _or_ still store the `webhookEvent` for cancelled & funded, but just don't send it to anyone if it was done for a non-card payment. Thoughts?

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -261,12 +266,23 @@ export enum OutgoingPaymentState {
 }
 
 export enum OutgoingPaymentDepositType {
```
mkurapov, line 268: Technically, neither the cancelled nor the funded events are "deposit" or "funded" type. We can either check whether the `OutgoingPaymentWithdrawType` or `OutgoingPaymentDepositType` are used anywhere and if they are, just make a third category for the cancelled or funded.

packages/backend/src/asset/model.ts
sanducb: We need it because `IncomingPaymentInitiationReason` was causing a cycle in the static imports after I moved it to `model.ts`. (I guess this is why it was in a different file before). So in order to be able to properly use this enum inside webhook service and still have the type in model.ts I had to break the cycle by adding this dynamic import. We can of course keep it in a separate file, but as you suggested [here](https://github.com/interledger/rafiki/pull/3651#discussion_r2360437748) for the outgoing payment equivalent enum, it is cleaner this way. Also, we can hardcode the enum variant (e.g. "CARD") [here](https://github.com/interledger/rafiki/pull/3651/files/e4c9706ab040f486f15ff19ac06db0efac11059c#diff-efa066cc9f163e17bcc35bfaa5a4bc80a60c56789c87f6b2e3dde96f5910013cR367) instead of using the type, but I think that might create confusion on the long run.

packages/backend/src/open_payments/payment/outgoing/service.ts
sanducb: My first thought is that it is confusing to store webhook events that will not be sent, especially if doing so by calling `sendWebhookEvent` here. Is there any reason we would need to store these events (e.g. audit, proofs)? If yes, then I think we should find a way to store this information separately, without touching webhooks logic.

packages/backend/src/webhook/service.ts
```diff
@@ -289,17 +293,25 @@ async function getWebhookEventsPage(
...
     query.whereIn('type', filter.type.in)
   }
+  if (filter?.type?.notIn && filter.type.notIn.length > 0) {
+    query.whereNotIn('type', filter.type.notIn)
+  }
 
   return await query.getPage(pagination, sortOrder)
 }
 
+type FinalizeRecipientsOptions = {
```
mkurapov, line 302: What do you think about having `sendToPosService` and `sendToCardService` parameters, and just let the caller decide when those are true? Then this function doesn't need to know about anything payments related

PR #3662, opened 2025-09-24
fix(backend):  fix webhook resolver filter for OP funded and cancelled events

sanducb:

- This PR fixes resolver filtering for `outgoing_payment.funded` and `outgoing_payment.cancelled` events

## Context

-

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/resolvers/webhooks.ts
+++ b/packages/backend/src/graphql/resolvers/webhooks.ts
@@ -22,9 +22,14 @@ export const getWebhookEvents: QueryResolvers<TenantedApolloContext>['webhookEve
     const { filter, sortOrder, tenantId, ...pagination } = args
     const order = sortOrder === 'ASC' ? SortOrder.Asc : SortOrder.Desc
     const webhookService = await ctx.container.use('webhookService')
-    const filterOrDefaults = filter ?? {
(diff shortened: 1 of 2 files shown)


packages/backend/src/graphql/resolvers/webhooks.ts
```diff
@@ -22,9 +22,14 @@ export const getWebhookEvents: QueryResolvers<TenantedApolloContext>['webhookEve
...
-      type: { notIn: DEFAULT_EXCLUDED_TYPES }
-    }
+    const noFilter =
+      !filter ||
+      (filter?.type?.in &&
+        Array.isArray(filter.type.in) &&
+        filter.type.in.length === 0)
+    const filterOrDefaults = noFilter
+      ? { type: { notIn: DEFAULT_EXCLUDED_TYPES } }
+      : filter
```
mkurapov: I think `noFilter` would be true if we pass in `in: []` but `notIn: ['something']` though, right? Meaning we would end up ignoring the `notIn` filter.

We can either
- Remove the `notIn` filter from the schema (but still have it functional in the service method)
- Do `const filter = { type: { notIn: DEFAULT_EXCLUDED_TYPES, in: <parse the in array> } }`

mkurapov (approved):

PR #3667, opened 2025-09-25
feat(card-service): handle payment cancelled & funded events

sanducb:
This PR handles the `outgoing_payment.funded` and `outgoing_payment.cancelled` events that we are sending from the backend to the card service ([RAF-1156](https://linear.app/interledger/issue/RAF-1156/backend-add-additional-webhook-events-for-outgoing-payments))

## Context

Closes [RAF-1168](https://linear.app/interledger/issue/RAF-1168/card-service-handle-payment-cancelled-and-funded-events)

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/card-service/src/payment/routes.ts
+++ b/packages/card-service/src/payment/routes.ts
@@ -6,7 +6,6 @@ import {
   PaymentCancellationReason,
   PaymentErrorCode,
   PaymentErrorResult,
-  PaymentResolution,
   PaymentResult,
   PaymentResultCode
 } from './types'
@@ -106,7 +105,7 @@ function toPaymentResolution(
   deps: ServiceDependencies,
   body: PaymentEventBody,
   identifiers: PaymentIdentifiers
-): PaymentResolution {
+): PaymentResult {
   const { requestId, outgoingPaymentId } = identifiers
   const cardPaymentFailureReason = body.data.metadata?.cardPaymentFailureReason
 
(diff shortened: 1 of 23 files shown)


packages/card-service/src/payment/routes.ts
```diff
@@ -49,13 +61,130 @@ async function create(deps: ServiceDependencies, ctx: PaymentContext) {
...
+        eventId: body.id,
+        eventType: body.type
+      },
+      'Payment event missing card requestId'
+    )
+    return
   }
+
+  const requestId = identifiers.requestId
+  const deferred = paymentWaitMap.get(requestId)
```
mkurapov, line 83: This isn't related to this PR, but for both card-service and POS service but most of the logic should live in the corresponding `service.ts`, as otherwise, we are coupling the "application layer" with our actual domain code. If made this, say, a GraphQL API, we shouldn't need to change any of the domain logic about how we are handling events and making the payment work. 

In any case, let's get this working and we can do all kinds of polishing later :)

packages/card-service/src/payment/types.ts
```diff
@@ -53,4 +93,4 @@ export type PaymentEventContext = Omit<AppContext, 'request'> & {
   }
 }
 
-export type PaymentResult = PaymentEventBody | void
+export type PaymentResult = PaymentResolution | void
```
mkurapov: do we need `void`?

mkurapov (approved):

packages/card-service/src/payment/routes.ts
sanducb: Created [RAF_1174](https://linear.app/interledger/issue/RAF-1174/cards-and-pos-services-refactor-routes-and-services) for that 👍

PR #3674, opened 2025-09-30
Blair/raf 1036

BlairCurrey:


packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -114,6 +120,8 @@ export async function handleSending(
   }
   stopTimer()
 
+  handleGrantSpentAmounts(deps, payment, receiveAmount)
```
mkurapov:
```suggestion
  await handleGrantSpentAmounts(deps, payment, receiveAmount)
```

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +166,103 @@ function getAdjustedAmounts(
...
+  // TODO: this shouldnt happen. should we error instead?
+  if (!latestSpentAmounts) {
+    deps.logger.warn(
+      { grantId: payment.grantId },
+      'No outgoingPaymentGrantSpentAmounts record found for grantId on payment failure'
+    )
+    return
+  }
+
+  if (latestSpentAmounts) {
```
mkurapov:
```suggestion
  
```
extraneous

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +166,103 @@ function getAdjustedAmounts(
...
+    const reservedDebitAmount = latestSpentAmounts.paymentDebitAmountValue
+    const reservedReceiveAmount = latestSpentAmounts.paymentReceiveAmountValue
+    const settledDebitAmount = await deps.accountingService.getTotalSent(
+      payment.id
+    )
```
mkurapov: If we return the sent/debit amount from the payment method, we can avoid this call to the accounting DB

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -369,7 +369,8 @@ async function createOutgoingPayment(
             client: options.client,
             metadata: options.metadata,
             state: OutgoingPaymentState.Funding,
-            grantId
+            grantId,
+            createdAt: new Date()
```
mkurapov, line 376: This is done automatically

(we have `table.timestamp('createdAt').defaultTo(knex.fn.now())` in the create outgoing payments table migration)

packages/backend/src/payment-method/handler/service.ts
```diff
@@ -30,12 +30,17 @@ export interface PayOptions {
   finalReceiveAmount: bigint
 }
 
+export interface SettledAmounts {
```
mkurapov: I would recommend staying away from "settled". Maybe "payment result" amounts?

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,244 @@ function getAdjustedAmounts(
...
+  // Detect if partial payment
+  // TODO: can partial payments be detected without this grant spent record?
+  // Like just from the payment vs. settledAmounts? If so we could remove the latestPaymentSpentAmounts query
+  const reservedDebitAmount = latestPaymentSpentAmounts.paymentDebitAmountValue
+  const reservedReceiveAmount =
+    latestPaymentSpentAmounts.paymentReceiveAmountValue
+  const debitAmountDifference = reservedDebitAmount - settledAmounts.debit
+  const receiveAmountDifference = reservedReceiveAmount - settledAmounts.receive
+
+  if (debitAmountDifference === 0n && receiveAmountDifference === 0n) return
```
mkurapov: Since debit amount includes fees the full debit amount is _not_ sent across in a payment (ILP or local). As a result, we can just check receiveAmount here, since that always should match.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,244 @@ function getAdjustedAmounts(
...
+    intervalEnd: latestPaymentSpentAmounts.intervalEnd,
+    createdAt: new Date()
+  })
+}
+
+/**
+ * Reverts the grant spent amounts when a payment fails.
+ * Inserts a spent amount record with the reserved amounts adjusted out.
+ */
+async function revertGrantSpentAmounts(
```
mkurapov: Would this function be the same as `handleGrantSpentAmounts`, where the settled amount are 0?

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,244 @@ function getAdjustedAmounts(
...
+    if (!latestIntervalSpentAmounts) {
+      deps.logger.warn(
+        {
+          grantId: payment.grantId,
+          intervalStart: latestPaymentSpentAmounts.intervalStart,
+          intervalEnd: latestPaymentSpentAmounts.intervalEnd
+        },
+        'No outgoingPaymentGrantSpentAmounts record found for grant interval'
+      )
+      return
```
mkurapov: Need to think through this a bit, but I think we would still want to update the `OutgoingPaymentGrantSpentAmounts` instead of returning early

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,244 @@ function getAdjustedAmounts(
...
+
+    newIntervalDebitAmountValue =
+      (latestIntervalSpentAmounts.intervalDebitAmountValue ?? 0n) -
+      debitAmountDifference
+    newIntervalReceiveAmountValue =
+      (latestIntervalSpentAmounts.intervalReceiveAmountValue ?? 0n) -
+      receiveAmountDifference
+  }
+
+  await OutgoingPaymentGrantSpentAmounts.query(deps.knex).insert({
```
mkurapov: would need to add specific `paymentState` here

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: That is a good question... I think maybe so.

In any case, I will look at merging these two functions (`revertGrantSpentAmounts` and `handelGrantSpentAmounts`) because they are pretty much doing the same thing and its probably not a great idea to duplicate the math. Kind of an artifact from when I was handling failures a bit differently.

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: I had to do this so it can be controlled via fake timers in the tests. Otherwise there was a test that failed. It was some business logic checking mocked time vs. real `outgoingPayment.createdAt` (since it is set by postgres and uncontrollable by jest's fake timers).

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,295 @@ function getAdjustedAmounts(
...
+
+  // Get the latest spent amounts record for this specific payment,
+  // not necessarily the latest for this grant
+  const latestPaymentSpentAmounts =
+    await OutgoingPaymentGrantSpentAmounts.query(deps.knex)
+      .where('outgoingPaymentId', payment.id)
+      .orderBy('createdAt', 'desc')
+      .first()
+
+  // TODO: this shouldnt happen. should we error instead?
```
njlie: Would it be sufficient to query for `.whereNotNull('latestPaymentSpentAmounts')` and passing the result through a typeguard so TS picks it up as well?

packages/backend/src/open_payments/payment/outgoing/service.ts
njlie: Would it still work if the test patched the outgoing payment after it was created?

packages/backend/src/payment-method/handler/service.ts
BlairCurrey: changed to `PayResult`

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: Good thought on the patch, but I just tried it and no it doesnt work, because this check that comes right after creating the payment in `outgoingPaymentService.create`:

```ts
          if (
            payment.walletAddressId !== payment.quote.walletAddressId ||
            payment.quote.expiresAt.getTime() <= payment.createdAt.getTime()
          ) {
            throw OutgoingPaymentError.InvalidQuote
          }
```          

Where the quote's `expiresAt` is subject to the mocked time but the payment is not. Trying to patch after its created results in a `InvalidQuote` error.

I think it's probably generally useful to be able to control it from the application level, so it can be mocked like this. Functionally setting it to `new Date()` should be the same as knex.fn.now().

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: Like, as a fallback? Is that your thinking?

I guess maybe that makes sense. This should never happen though so IDK how we'd get the correct interval/grant totals. Since it's totally unexpected I dont think we can just assume the interval/grant amounts were 0 or something. I suppose we can sum them all up like we do for the legacy path on the payment create side.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: added. they should all be complete here.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: >Would this function be the same as handleGrantSpentAmounts, where the settled amount are 0?

I tested this and the answer is *basically*, yes. With some caveats. The amounts it produces are the same. The payment state at the end will be different and it relies on the debit amount calculations which are probably otherwise superfluous as you've noted here: https://github.com/interledger/rafiki/pull/3674/files/6ad2fd9296f52bd6e14143b3da293289d9eb1209#r2444999725 Maybe there is a more straightforward way for simply "reverting" in the case of failure. I will see as I get into de-duping this.

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,295 @@ function getAdjustedAmounts(
   return { maxDebitAmount, maxReceiveAmount }
 }
 
+/**
+ * Compares the final settled amounts with the amounts on hold
+ * and inserts a new OutgoingPaymentGrantSpentAmount record if needed.
+ */
+async function handleGrantSpentAmounts(
```
mkurapov: I realized we also need to call this whenever we call `cancelOutgoingPayment` as well

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,296 @@ function getAdjustedAmounts(
...
+    id: v4(),
+    outgoingPaymentId: payment.id,
+    paymentDebitAmountValue: finalAmounts.debit,
+    intervalDebitAmountValue: newIntervalDebitAmountValue,
+    grantTotalDebitAmountValue: newGrantTotalDebitAmountValue,
+    paymentReceiveAmountValue: finalAmounts.receive,
+    intervalReceiveAmountValue: newIntervalReceiveAmountValue,
+    grantTotalReceiveAmountValue: newGrantTotalReceiveAmountValue,
+    intervalStart: latestPaymentSpentAmounts.intervalStart,
+    intervalEnd: latestPaymentSpentAmounts.intervalEnd,
```
mkurapov: I think we should store the `latestIntervalSpentAmounts`'s `intervalStart` and `intervalEnd` if `latestIntervalSpentAmounts` is defined

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
```diff
@@ -158,6 +164,296 @@ function getAdjustedAmounts(
...
+      const latestIntervalSpentAmounts =
+        await OutgoingPaymentGrantSpentAmounts.query(deps.knex)
+          .where('grantId', payment.grantId)
+          .where('intervalStart', latestPaymentSpentAmounts.intervalStart)
+          .where('intervalEnd', latestPaymentSpentAmounts.intervalEnd)
+          .orderBy('createdAt', 'desc')
+          .first()
+
+      if (!latestIntervalSpentAmounts) {
+        deps.logger.warn(
```
mkurapov:
```suggestion
        deps.logger.error(
```
since i think it would require debugging

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: there are enough differences that I found 2 functions to be better, but i did deduped the shared logic

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: added

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: as discussed in a call, the payment spent amounts' interval and interval ones are equivalent here

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: this stuff got refactored a bit but i implemented this. no longer check/update the debit amount

packages/backend/src/open_payments/payment/outgoing/lifecycle.ts
BlairCurrey: the type issue is kind moot at this point after some refactoring. in terms of the general handling of these cases (which shouldnt happen) they are logging at an error level and returning since we dont necessarily want to fail the worker over it.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -952,3 +959,263 @@ function validateSentAmount(
...
+    receive: newIntervalReceiveAmountValue
+  } = calculateIntervalAmounts(
+    latestPaymentSpentAmounts,
+    latestIntervalSpentAmounts,
+    0n,
+    receiveAmountDifference
+  )
+
+  await OutgoingPaymentGrantSpentAmounts.query(deps.knex).insert({
+    ...latestPaymentSpentAmounts,
```
mkurapov: To be safe, let's not have this spread operator and just define all of the fields in-line, otherwise we might include fields that we don't want to update

packages/backend/src/open_payments/payment/outgoing/lifecycle.test.ts
```diff
@@ -0,0 +1,2311 @@
...
+          expect(processedPaymentId).toBe(payment.id)
+
+          const latestGrantSpentAmounts =
+            await OutgoingPaymentGrantSpentAmounts.query(knex)
+              .where({ grantId: grant.id })
+              .orderBy('createdAt', 'desc')
+              .first()
+
+          // Should have new spent amounts with payment factored out
+          expect(latestGrantSpentAmounts).toMatchObject({
```
mkurapov, line 587: minor, but lets check there are two records that were created here

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -327,6 +328,7 @@ async function pay(
       },
       'ILP payment completed'
     )
+    return { receive: receipt.amountDelivered, debit: finalDebitAmount }
```
mkurapov: I think we should leave a comment here explaining the behaviour why not the receipt.amountSent

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: changed

packages/backend/src/payment-method/ilp/service.ts
BlairCurrey: at this point I think its just easier and more clear to just remove the debit amount from the return type. updated w/ that.

packages/backend/src/open_payments/payment/outgoing/lifecycle.test.ts
BlairCurrey: added a check to make sure they are different records.

PR #3675, opened 2025-09-30
chore(point-of-sale): update pos service response structure

sanducb:

- This PR updates the response structure we return to the POS via `POST /payment`

## Context

Closes #RAF-1175

## Checklist

- [ ] Tests added/updated


packages/point-of-sale/src/payments/routes.ts
```diff
@@ -101,12 +99,19 @@ async function payment(
       }
     )
 
+    if (result === Result.INVALID_SIGNATURE) {
+      webhookWaitMap.delete(incomingPayment.id)
```
mkurapov: We should the cleaning up of the map in a `finally` block

Issue #3453, opened 2025-06-09
docs: update Helm value names in environment variables


hajjimo:
The DevOps team is currently updating the Helm charts.  Consequently, the current Helm value names in the environment variables partials are outdated. Creating this issue to update the relevant sections as new Helm chart(s) are issued.

PR #3679, opened 2025-10-03
docs: 3453/helm env updates

hajjimo:
Addresses #3453 

Map Helm variable names to the updated Helm chart in https://github.com/interledger/charts/tree/main/charts/rafiki/values.yaml

**Required**

- [ ] Used LinkOut component on external links
- [x] Ran Prettier
- [x] Previewed updates in Netlify
- [x] Included a "fixes #" comment

**Conditional**

- [ ] Included code samples where appropriate
- [ ] Updated related READMEs


packages/documentation/src/partials/auth-variables.mdx
```diff
@@ -4,53 +4,55 @@ import { LinkOut } from '@interledger/docs-design-system'
...
+| ------------------------ | --------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `AUTH_DATABASE_URL`      | `config.auth.databaseUrl.value` or `config.auth.databaseUrl.secretKeyRef`                                 | `postgresql://postgres:password@localhost:5432/auth_development` | The URL of the Postgres database storing your Open Payments grant data. Can be provided as a value or secret reference.                                                                                                                                                                                                                                            |
```
BlairCurrey, line 9: I think this `config.auth.databaseUrl.secretKeyRef` should be `config.auth.databaseUrl.secretKeyRef.key` (missing the last `.key`).

I also wondered about just `config.auth.databseUrl` Although thats kinda fuzzy. I think the chart rework breaks some implicit assumptions about the map (namely a 1:1 relation with an explicit helm value). But the new description here is clear that its a reference which I think is the important part.

BlairCurrey: Eh, maybe it doesnt even need to have `.key` and is fine as-is. Using the same reasoning as my musing about just doing `config.auth.databseUrl`. Important part being it comes either from setting here or passing in.

PR #3695, opened 2025-10-09
feat(backend): enforce uniqueness on tenant wallet address prefixes

njlie:

- When a `WALLET_ADDRESS_URL` tenant setting is created/updated, the service now checks that no other settings exist with that same value.

## Context

Closes RAF-1101.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/wallet_address/service.test.ts
+++ b/packages/backend/src/open_payments/wallet_address/service.test.ts
@@ -107,7 +107,7 @@ describe('Open Payments Wallet Address Service', (): void => {
       isOperator | tenantSettingUrl
       ${false}   | ${undefined}
       ${true}    | ${undefined}
-      ${true}    | ${'https://alice.me'}
+      ${true}    | ${`https://alice.me/${uuid()}`}
     `(
       'operator - $isOperator with tenantSettingUrl - $tenantSettingUrl',
       async ({ isOperator, tenantSettingUrl }): Promise<void> => {


packages/backend/src/tenants/model.ts
```diff
@@ -24,6 +24,7 @@ export class Tenant extends BaseModel {
   public apiSecret!: string
   public idpConsentUrl!: string
   public idpSecret!: string
+  public walletAddressPrefix!: string
```
mkurapov: Should `walletAddressPrefix` be non-nullable in the migration if this is required? 
If so, should `walletAddressPrefix` be required when creating a tenant?

packages/backend/src/open_payments/wallet_address/service.ts
```diff
@@ -207,7 +207,7 @@ async function createWalletAddressUrl(
   let finalWalletAddressUrl: string
   if (isValidUrl(options.address)) {
     // in case that client provided full url, verify that it starts with the tenant's URL
-    const walletAddressUrl = new URL(options.address)
+    const walletAddressUrl = new URL(options.address.toLowerCase())
```
mkurapov: should we also lowercase the wallet address in the `relativePath` case as well?
Or just toLowerCase the whole thing before we return from this function

packages/backend/src/open_payments/wallet_address/service.test.ts
```diff
@@ -172,21 +155,23 @@ describe('Open Payments Wallet Address Service', (): void => {
     })
 
     test.each`
-      setting                    | address                        | generated
-      ${'https://alice.me/ilp'}  | ${'https://alice.me/ilp/test'} | ${'https://alice.me/ilp/test'}
-      ${'https://alice.me/ilp'}  | ${'test'}                      | ${'https://alice.me/ilp/test'}
-      ${'https://alice.me/ilp'}  | ${'/test'}                     | ${'https://alice.me/ilp/test'}
-      ${'https://alice.me/ilp/'} | ${'test'}                      | ${'https://alice.me/ilp/test'}
-      ${'https://alice.me/ilp/'} | ${'/test'}                     | ${'https://alice.me/ilp/test'}
+      setting                      | address                          | generated
```
mkurapov:
```suggestion
      prefix                      | address                          | generated
```

packages/backend/src/tenants/service.ts
```diff
@@ -203,11 +231,20 @@ async function updateTenant(
       })
     }
 
+    if (!tenant.walletAddressPrefix && walletAddressPrefix) {
+      tenant = await tenant.$query(trx).patchAndFetch({
+        walletAddressPrefix: walletAddressPrefix.toLowerCase()
+      })
+    } else if (tenant.walletAddressPrefix && walletAddressPrefix) {
+      throw TenantError.InvalidTenantInput
+    }
```
mkurapov: ie, `walletAddressPrefix` cannot be changed from time of creation, correct?

Would it necessarily be a problem if tenants were allowed to change it? Old wallet addresses from the tenant would use the old prefix, and then any new wallet addresses going forward would use the new ones.

packages/frontend/app/routes/tenants.create.tsx
```diff
@@ -48,6 +48,13 @@ export default function CreateTenantPage() {
     placeholder: string
     label: string
   }[] = [
+    {
```
mkurapov: We also need to update the wallet address creation page to fetch `walletAddressPrefix` correctly

packages/backend/src/open_payments/wallet_address/service.ts
njlie: We at least need to lowercase before comparing with the wallet address prefix, since the part of a URL before the path is case-insensitive. But I suppose the more nuanced approach here is to lowercase the origin but keep the casing of the path when both setting a wallet address prefix or comparing the prefix with a wallet address to be created.

packages/backend/src/tenants/model.ts
njlie: Honestly it probably would be easier if it was required. There doesn't seem to be a use case where a tenant can exist but not need to create wallet addresses. I am noticing though, that neither `idpConsentUrl` or `idpSecret` are actually non-nullable in the db schema but are required in the model here, so that should be fixed as well.

packages/backend/src/tenants/service.ts
njlie: If this were a value that wasn't static after being set, then that leads me to consider removing the prefix entirely...

packages/backend/migrations/20250930210751_add_prefix_to_tenant.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+  return knex.schema
+    .alterTable('tenants', (table) => {
+      table.string('walletAddressPrefix').unique()
+    })
+    .then(() => {
+      knex.raw(
```
BlairCurrey:
```suggestion
      return knex.raw(
```

packages/backend/migrations/20250930210751_add_prefix_to_tenant.js
```diff
@@ -0,0 +1,34 @@
...
+        `UPDATE "tenants" SET "walletAddressPrefix" = (SELECT "value" from "tenantSettings" WHERE "tenantId" = "tenants"."id" AND "key" = 'WALLET_ADDRESS_URL')`
+      )
+    })
+    .then(() => {
+      knex.raw(`DELETE "tenantSettings" WHERE "key" = 'WALLET_ADDRESS_URL'`)
```
BlairCurrey:
```suggestion
      return knex.raw(`DELETE "tenantSettings" WHERE "key" = 'WALLET_ADDRESS_URL'`)
```

packages/backend/src/tenants/service.ts
```diff
@@ -256,3 +293,22 @@ async function updateOperatorApiSecretFromConfig(
...
+}
+
+function validateWalletAddressPrefix(prefix: string): boolean {
+  try {
+    new URL(prefix)
+    return true
+  } catch (err) {
+    return false
+  }
+}
```
BlairCurrey: Does this need to accept paths and validate domain as mentioned in the issue?

>The validator for the setting should also accept URL paths as potential inputs as well as fully qualified URLs.

>The validator should expect the configured domain of the Rafiki backend to be included with the input if a fully qualified URL is used.

Either way it might be good to document in the graphql resolver string if its either or fully qualified with an example

packages/backend/src/tenants/service.ts
```diff
@@ -98,18 +101,28 @@ async function createTenant(
...
+
     const tenant = await Tenant.query(trx).insertAndFetch({
       id,
       email,
       publicName,
       apiSecret,
       idpSecret,
-      idpConsentUrl
+      idpConsentUrl,
+      walletAddressPrefix
```
BlairCurrey: Could probably try/catch and return an error when the prefix fails the unique constraint. Would make it friendlier to the client. Currently it just throws the SQL error.

This is what I see in bruno when trying to create a tenant with non unique wallet address:

```json
{
  "errors": [
    {
      "message": "insert into \"tenants\" (\"apiSecret\", \"email\", \"id\", \"idpConsentUrl\", \"idpSecret\", \"walletAddressPrefix\") values ($1, $2, $3, $4, $5, $6) returning \"id\" - duplicate key value violates unique constraint \"tenants_walletaddressprefix_unique\"",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createTenant"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ],
  "data": null
}
```

Perhaps with a new `TenantError.DuplicateWalletAddressPrefix` error? Otherwise IDK if the client has enough info. Could use this test:

```ts
    test('cannot create tenant with duplicate wallet address prefix', async (): Promise<void> => {
      const walletAddressPrefix = `${config.openPaymentsUrl}/something}`
      const createOptions1 = {
        apiSecret: 'test-api-secret',
        publicName: 'test tenant1',
        email: faker.internet.email(),
        idpConsentUrl: faker.internet.url(),
        idpSecret: 'test-idp-secret1',
        walletAddressPrefix
      }
      const createOptions2 = {
        apiSecret: 'test-api-secret',
        publicName: 'test tenant2',
        email: faker.internet.email(),
        idpConsentUrl: faker.internet.url(),
        idpSecret: 'test-idp-secret2',
        walletAddressPrefix
      }

      const tenant1Error = await tenantService.create(createOptions1)
      assert(!isTenantError(tenant1Error))

      const tenant2Error = await tenantService.create(createOptions2)
      assert(isTenantError(tenant2Error))
      expect(tenant2Error).toEqual(TenantError.DuplicateWalletAddressPrefix)
    })
```

packages/backend/src/open_payments/wallet_address/service.ts
njlie: edit: the URL() constructor already lowercases the origin but not the path. I don't know why but the case-insensitive test was failing for some reason and now it's not even after reverting this.

packages/backend/src/tenants/model.ts
njlie: Update on if it wasn't required - if we make this required, we'd need to backfill tenants who were created without this setting configured with some generic value (e.g. the openPaymentsUrl + a uuid). I think this would be okay if we also allow tenants to update the prefix after it's been set, but again, I think allowing an update makes me reconsider even having the prefix at all

PR #3696, opened 2025-10-09
feat(auth): start storing `apiSecret` in `auth` and add tenant signature middleware

### LINKED ISSUE #3692: Start storing `apiSecret` in the `auth` service & authenticate Admin API requests
### Context

Currently, when we create or update tenants, we are not propagating the `apiSecret` to the `auth` service. This means that we are unable to add proper tenant access for the auth admin API ([ROP-10](https://linear.app/interledger/issue/ROP-10/add-tenantedapollocontext-to-auth-service-and-access-control-for)). 

### Todo

- [ ] Start storing `apiSecret` in the `auth` tenants table
- [ ] Propagate the `apiSecret` to the `auth` tenants table during tenant creation and update
- [ ] Like `backend`, add middleware in the Admin API to validate and fetch the correct tenant

mkurapov:

- Add `apiSecret` to tenant table in `auth`
- Seed `apiSecret` from env variable for operator
- Propagate `apiSecret` from `backend` to `auth` when creating new tenants
- On `auth` startup, update `apiSecret` if it has changed
- Add middleware to Admin API to fetch correct tenant from signature
- Update `tableManager` tests to properly handle tenant tests

## Context

Fixes #3692 | RAF-1191

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/auth/src/signature/tenant.test.ts
+++ b/packages/auth/src/signature/tenant.test.ts
@@ -3,7 +3,7 @@ import { faker } from '@faker-js/faker'
 import Redis from 'ioredis'
 import { initIocContainer } from '..'
 import { AppServices, AppContext, TenantedAppContext } from '../app'
-import { Config } from '../config/app'
+import { Config, IAppConfig } from '../config/app'
 import { Tenant } from '../tenant/model'
 import { generateApiSignature } from '../tests/apiSignature'
 import { TestContainer, createTestApp } from '../tests/app'
@@ -19,6 +19,7 @@ describe('Tenant Middlewares', (): void => {
   let deps: IocContract<AppServices>
   let appContainer: TestContainer
   let tenantService: TenantService
+  let config: IAppConfig
   let tenant: Tenant
   let operator: Tenant
   let redis: Redis
@@ -30,6 +31,7 @@ describe('Tenant Middlewares', (): void => {
       ...Config,
       adminApiSecret: operatorApiSecret
     })
+    config = await deps.use('config')
     appContainer = await createTestApp(deps)
     redis = await deps.use('redis')
     tenantService = await deps.use('tenantService')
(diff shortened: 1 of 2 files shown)


packages/backend/src/tenants/service.ts
```diff
@@ -198,6 +199,7 @@ async function updateTenant(
 
     if (idpConsentUrl || idpSecret) {
       await deps.authServiceClient.tenant.update(id, {
+        apiSecret,
```
BlairCurrey, line 202: This is sort of tangential - dont think there is anything wrong here. But I dont think we're actually able to update tenants from the UI. Are we supposed to be able to? 

As-is, it seems like the API secret is only visible to operators. For any tenant (self or non-operators). But its disabled if you're an operator (enabled if you're not, but you cant see it). 

Im not really sure what the intended behavior is supposed to be but I dont see any scenario where you can update from the UI.

njlie: I noticed this too. For the information that the auth service already stores, like the idp secret or consent screen URL, any updates are handled initially by the backend service, then propagated through to the auth service. So I think the apiSecret has this same dynamic.

BlairCurrey: I think its coming back to me... just saw this when trying to update via bruno:

>"Operator cannot update apiSecret over admin api"

So working as intended I think. 

But this begs the question, what happens if we set the env var secret to different things? I believe they would just be different (or perhaps backed would overwrite what auth does itself if it starts up later). Obviously people should not do that but not sure if there is really a graceful way to warn.

packages/auth/src/tenant/routes.test.ts
```diff
@@ -48,6 +48,7 @@ describe('Tenant Routes', (): void => {
     test('Gets a tenant', async (): Promise<void> => {
       const tenant = await Tenant.query().insert({
         id: v4(),
+        apiSecret: v4(),
```
njlie, line 51: This test is missing an assertion that this value was retrieved

packages/auth/src/tenant/routes.test.ts
```diff
@@ -96,6 +97,7 @@ describe('Tenant Routes', (): void => {
     test('Creates a tenant', async (): Promise<void> => {
       const tenantData = {
         id: v4(),
+        apiSecret: v4(),
```
njlie, line 101: Should assert that this value was created

packages/auth/src/signature/tenant.ts
```diff
@@ -0,0 +1,148 @@
+import { canonicalize } from 'json-canonicalize'
+import { AppContext, TenantedAppContext } from '../app'
+import { IAppConfig } from '../config/app'
+import { Tenant } from '../tenant/model'
+import * as crypto from 'crypto'
+import Koa from 'koa'
+
+function getSignatureParts(signature: string) {
```
njlie, line 8: Not relavant to whether or not this gets merged in but I feel as though we are reaching the point where signature verification should be a shared library between the Rafiki services

packages/auth/src/shared/utils.test.ts
```diff
@@ -1,156 +1,6 @@
...
-import { verifyApiSignature, isValidDateString } from './utils'
-import { TestContainer, createTestApp } from '../tests/app'
+import { isValidDateString } from './utils'
 
 describe('utils', (): void => {
-  let deps: IocContract<AppServices>
-  let appContainer: TestContainer
-  let redis: Redis
-
-  describe('admin api signatures', (): void => {
```
njlie, line 17: Some of these tests don't appear to be replaced elsewhere

packages/backend/src/tenants/service.ts
mkurapov: So it looks like the only possibility of updating the `apiSecret` is if the tenant calls the `updateTenant` mutation directly via the API, and not the UI. Do you think that is OK behaviour?

If so, then we can keep this as is, and just propagate the `apiSecret` like the other fields. I also [changed](https://github.com/interledger/rafiki/pull/3696/commits/e39c3433b2b013f819dc342e153b632c38ba3124) the conditional to make sure we get the change across to auth.

packages/auth/src/tenant/routes.test.ts
mkurapov: Actually, do we need this get route in the first place?

packages/auth/src/shared/utils.test.ts
mkurapov: This is now encapsulated here: https://github.com/interledger/rafiki/pull/3696/files#diff-9dc226aa3bdbd465fa663bc9a108df25348ba71dc7982cd1ad178fcdf3524aadR18

packages/auth/src/signature/tenant.ts
mkurapov: Yes, especially now with the cards & POS services

packages/backend/src/tenants/service.ts
BlairCurrey: >So it looks like the only possibility of updating the apiSecret is if the tenant calls the updateTenant mutation directly via the API, and not the UI.

I was able to update in bruno (add in ui, get id, change senderTenantId and backendApiSignatureSecret in bruno config, change id in bruno's "Rafiki Admin API" > "Update tenant" request variables).

~~After I did, I tried to use the admin ui with the new credentials. I saw the and saw some unauthorized errors. Seems like there is an issue there.~~

~~I also saw the same thing when I hastily added in an input for the secret for non-operator tenants.~~

>Do you think that is OK behaviour?

I would expect that they can change it from the UI, if they are allowed to change it at all. Which I think they should.

BlairCurrey: never mind on part of that, but there is something we need to fix if we add a form input for the api secret for non operators. I added one (and the schema validator), submitted, and got unauthorized error. it did update the secret but the browser session didnt seem to udpate. worked fine when i manually cleared and re-entered. There is code that looks like it should be setting on the session but something is wrong there.

Although its interesting that we have that code to update the session... because as-is the code only shows a disabled input for the api secret if you are the operator. You cant update if you are non-operator, nor see it. So the code to update the session on api secret change would just never run? I think we intended to let the non operator update at one point at least.

packages/auth/src/tenant/routes.test.ts
njlie: Don't think so. Anything the auth service knows about the tenant is also known by the backend, which already exposes a way to get a tenant.

packages/auth/src/shared/utils.test.ts
njlie: Does it still test stuff like if the signature is too old to be processed or if it has already been processed?

mkurapov: Ah, good catch. Adding old signature tests

packages/backend/src/tenants/service.ts
mkurapov: I just tested as well and yes, I also need to clear the credentials and re-set them manually in Admin UI before it works.

We can put that as an enhancement but I don't see it being something of high priority to fix.

packages/auth/src/tenant/routes.test.ts
mkurapov: I will do some cleanup in another PR

PR #3697, opened 2025-10-10
feat(auth): add tenant boundaries to auth GraphQL resolvers

mkurapov:

Following up on #3696, this change uses the `TenantedApolloContext` to add tenant boundaries to the 3 resolvers (getGrant, getGrants, revokeGrant) we have in the `auth` admin API

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/auth/src/signature/tenant.test.ts
+++ b/packages/auth/src/signature/tenant.test.ts
@@ -3,7 +3,7 @@ import { faker } from '@faker-js/faker'
 import Redis from 'ioredis'
 import { initIocContainer } from '..'
 import { AppServices, AppContext, TenantedAppContext } from '../app'
-import { Config } from '../config/app'
+import { Config, IAppConfig } from '../config/app'
 import { Tenant } from '../tenant/model'
 import { generateApiSignature } from '../tests/apiSignature'
 import { TestContainer, createTestApp } from '../tests/app'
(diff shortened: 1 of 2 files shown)


packages/auth/src/tests/app.ts
```diff
@@ -96,13 +76,42 @@ export const createTestApp = async (
...
+  await start(container, app)
+
+  const knex = await container.use('knex')
 
   return {
     app,
     port: app.getAuthPort(),
     adminPort: app.getAdminPort(),
     knex,
-    apolloClient: client,
```
mkurapov, line 105: Doesn't need to be on the app always IMO. Resolver test can just call `createApolloClient` as they need

packages/auth/src/index.ts
```diff
@@ -252,6 +252,8 @@ export const gracefulShutdown = async (
   await app.shutdown()
   const knex = await container.use('knex')
   await knex.destroy()
+  const redis = await container.use('redis')
+  redis.disconnect()
```
mkurapov, line 256: Fixes this testing error:

> Jest did not exit one second after the test run has completed.
> 'This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
^C

PR #3706, opened 2025-10-16
feat(backend): add middleware that handles STREAM KYC/additional data frames

sanducb:

- This PR introduces support for sending additional data over ILP prepare packets.  This allows a receiving ASE to reject/fulfill a payment packet based on the validity of the additional data.

## Context

Closes #RAF-1185

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/service.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.ts
@@ -81,10 +81,10 @@ export interface IncomingPaymentService
   ): Promise<IncomingPayment | IncomingPaymentError>
   processPartialPayment(
     id: string,
-    options?: {
-      dataToTransmit?: string
-      partialIncomingPaymentId?: string
-      expiresAt?: Date
+    options: {
+      dataToTransmit: string
+      partialIncomingPaymentId: string
+      expiresAt: Date
     }
   ): Promise<PartialPaymentDecision>
   updatePartialPaymentDecision(
@@ -569,10 +569,10 @@ async function addReceivedAmount(
 async function processPartialPayment(
   deps: ServiceDependencies,
   id: string,
-  options?: {
-    dataToTransmit?: string
-    partialIncomingPaymentId?: string
-    expiresAt?: Date
+  options: {
+    dataToTransmit: string
+    partialIncomingPaymentId: string
+    expiresAt: Date
   }
 ): Promise<PartialPaymentDecision> {
   const { config, knex, redis } = deps
@@ -589,11 +589,11 @@ async function processPartialPayment(
     type: IncomingPaymentEventType.IncomingPaymentPartialPaymentReceived,
     data: {
       ...incomingPayment.toData(0n),
-      partialIncomingPaymentId: options?.partialIncomingPaymentId,
+      partialIncomingPaymentId: options.partialIncomingPaymentId,
       dataToTransmit:
-        options?.dataToTransmit && config.dbEncryptionSecret
+        options.dataToTransmit && config.dbEncryptionSecret
           ? encryptDbData(options.dataToTransmit, config.dbEncryptionSecret)
-          : options?.dataToTransmit
+          : options.dataToTransmit
     },
     tenantId: incomingPayment.tenantId,
     webhooks: finalizeWebhookRecipients(
@@ -608,83 +608,76 @@ async function processPartialPayment(
   })
 
   let decision: PartialPaymentDecision = {}
-  if (options?.partialIncomingPaymentId && options?.expiresAt) {
-    const partialIncomingPaymentId = options.partialIncomingPaymentId
-    const cacheKey = getPartialPaymentDecisionCacheKey(
-      id,
-      partialIncomingPaymentId
-    )
+  const partialIncomingPaymentId = options.partialIncomingPaymentId
+  const cacheKey = getPartialPaymentDecisionCacheKey(
+    id,
+    partialIncomingPaymentId
+  )
 
-    // Bounded polling: wait for decision up to (packet expiry - safetyMs) or maxWaitMs
-    const safetyMs = Number.isFinite(
-      config.partialPaymentDecisionSafetyMarginMs
-    )
-      ? config.partialPaymentDecisionSafetyMarginMs
-      : 100
-    const maxWaitMs = Number.isFinite(config.partialPaymentDecisionMaxWaitMs)
-      ? config.partialPaymentDecisionMaxWaitMs
-      : 1500
+  // Bounded polling: wait for decision up to (packet expiry - safetyMs) or maxWaitMs
+  const safetyMs = Number.isFinite(config.partialPaymentDecisionSafetyMarginMs)
+    ? config.partialPaymentDecisionSafetyMarginMs
+    : 100
+  const maxWaitMs = Number.isFinite(config.partialPaymentDecisionMaxWaitMs)
+    ? config.partialPaymentDecisionMaxWaitMs
+    : 1500
 
-    const now = Date.now()
-    const timeRemaining = Math.max(
-      0,
-      options.expiresAt.getTime() - now - safetyMs
-    )
-    const timeoutMs = Math.min(timeRemaining, maxWaitMs)
-    const pollingFrequencyMs = 50
+  const now = Date.now()
+  const timeRemaining = Math.max(
+    0,
+    options.expiresAt.getTime() - now - safetyMs
+  )
+  const timeoutMs = Math.min(timeRemaining, maxWaitMs)
+  const pollingFrequencyMs = 50
 
-    try {
-      const polledDecision = await poll({
-        request: async (): Promise<PartialPaymentDecision | null> => {
+  try {
+    const polledDecision = await poll({
+      request: async (): Promise<PartialPaymentDecision | null> => {
+        try {
+          const value = await redis.get(cacheKey)
+          if (!value) return null
+
           try {
-            const value = await redis.get(cacheKey)
-            if (!value) return null
-
-            try {
-              const parsed = JSON.parse(value) as PartialPaymentDecision
-              return parsed
-            } catch (parseError) {
-              deps.logger.warn(
-                { parseError, incomingPaymentId: id, cacheKey },
-                'invalid partial payment decision format in cache'
-              )
-            }
-
-            return null
-          } catch (e) {
+            const parsed = JSON.parse(value) as PartialPaymentDecision
+            return parsed
+          } catch (parseError) {
             deps.logger.warn(
-              { e, incomingPaymentId: id },
-              'decision read failed'
+              { parseError, incomingPaymentId: id, cacheKey },
+              'invalid partial payment decision format in cache'
             )
-            return null
           }
-        },
-        stopWhen: (result: PartialPaymentDecision | null) => result !== null,
-        pollingFrequencyMs,
-        timeoutMs
-      })
-      if (polledDecision) {
-        decision = {
-          ...decision,
-          ...polledDecision
-        }
 
-        if (!decision.message && typeof decision.success === 'boolean') {
-          decision.message = decision.success
-            ? 'Additional data approved'
-            : 'Additional data rejected'
+          return null
+        } catch (e) {
+          deps.logger.warn({ e, incomingPaymentId: id }, 'decision read failed')
+          return null
         }
+      },
+      stopWhen: (result: PartialPaymentDecision | null) => result !== null,
+      pollingFrequencyMs,
+      timeoutMs
+    })
+    if (polledDecision) {
+      decision = {
+        ...decision,
+        ...polledDecision
       }
-    } catch (e) {
-      deps.logger.warn(
-        { e, incomingPaymentId: id },
-        'partial payment decision polling timed out or failed'
-      )
-    }
 
-    if (!decision.message) {
-      decision.message = 'No response from ASE'
+      if (!decision.message && typeof decision.success === 'boolean') {
+        decision.message = decision.success
+          ? 'Additional data approved'
+          : 'Additional data rejected'
+      }
     }
+  } catch (e) {
+    deps.logger.warn(
+      { e, incomingPaymentId: id },
+      'partial payment decision polling timed out or failed'
+    )
+  }
+
+  if (!decision.message) {
+    decision.message = 'No response from ASE'
   }
 
   return decision
--- a/packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
+++ b/packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
@@ -39,14 +39,14 @@ export function createPartialPaymentDecisionMiddleware(): ILPMiddleware {
     let message: string | undefined
 
     try {
-      decision = (await ctx.services.incomingPayments.processPartialPayment(
+      decision = await ctx.services.incomingPayments.processPartialPayment(
         incomingPaymentId,
         {
           dataToTransmit: additionalData,
           partialIncomingPaymentId: uuid(),
           expiresAt: prepare.expiresAt
         }
-      )) as PartialPaymentDecision
+      )
 
       if (decision?.success) {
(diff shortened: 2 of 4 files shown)


packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    let reason = 'rejected'
+    try {
+      const parsed = JSON.parse(decision)
+      reason = parsed?.reason || reason
+    } catch (_e) {
+      reason = decision
+    }
+
+    ctx.response.reject = {
```
sanducb: We should throw instead and let the error handling middleware create the reject packet with a `message`.

packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    const expiresAt = ctx.request.prepare.expiresAt
+    const now = Date.now()
+    const timeRemaining = Math.max(0, expiresAt.getTime() - now - safetyMs)
+    const deadline = now + Math.min(timeRemaining, maxWaitMs)
+    const pollIntervalMs = 50
+
+    const readDecision = async (): Promise<string | undefined> => {
+      try {
+        const value = await redis.get(cacheKey)
```
mkurapov: We have a `poll` utility function for this in `utils.ts`

packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
```diff
@@ -0,0 +1,65 @@
...
+    const { config, logger, redis } = ctx.services
+
+    if (!ctx.state.streamDestination || !ctx.state.hasAdditionalData) {
+      await next()
+      return
+    }
+
+    const incomingPaymentId = ctx.state.streamDestination
+    const connectionId = ctx.state.connectionId ?? 'unknown'
+    const cacheKey = `kyc_decision:${incomingPaymentId}:${connectionId}`
```
mkurapov: I think we should incorporate the polling inside the existing incoming payment service that will handle the logic for this. That is where we can also put the code to actually confirm or reject the partial incoming payment.

packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
```diff
@@ -0,0 +1,65 @@
...
+  ): Promise<void> => {
+    const { config, logger, redis } = ctx.services
+
+    if (!ctx.state.streamDestination || !ctx.state.hasAdditionalData) {
+      await next()
+      return
+    }
+
+    const incomingPaymentId = ctx.state.streamDestination
+    const connectionId = ctx.state.connectionId ?? 'unknown'
```
mkurapov: Let's generate our own UUID `partialIncomingPaymentId`

packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
```diff
@@ -0,0 +1,65 @@
+import { Errors } from 'ilp-packet'
+import { ILPContext, ILPMiddleware } from '../rafiki'
+import { StreamState } from './stream-address'
+
+export function createKycDecisionMiddleware(): ILPMiddleware {
```
mkurapov: I think we should not have any references to "kyc" in this code change, but simply partialPaymentDecisionMiddleware or something

mkurapov (commented):
Only focused on the kyc decision middleware here

packages/backend/src/payment-method/ilp/connector/core/middleware/kyc-decision.ts
mkurapov: I think the incoming payment service is fine for this one.

packages/backend/src/payment-method/ilp/connector/index.ts
```diff
@@ -73,13 +74,10 @@ export async function createConnectorService({
...
     },
     compose([
-      // ILP packet processing time (must be first to measure entire chain)
-      createIlpTimingMiddleware(),
-
       // Incoming Rules
       createIncomingErrorHandlerMiddleware(ilpAddress),
       createStreamAddressMiddleware(),
-
+      createKycDecisionMiddleware(),
```
mkurapov: should we move this to be after the `createBalanceMiddleware` ? That way, we will only publish the webhook & start polling for the decision after we were able to created a pending transfer between the peer account to the incoming payment account, confirming we have enough funds to make the transfer anyway.

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -558,24 +564,30 @@ async function addReceivedAmount(
 async function processPartialPayment(
   deps: ServiceDependencies,
   id: string,
-  dataToTransmit?: string
-): Promise<IncomingPayment | IncomingPaymentError> {
-  const { config, knex } = deps
+  options?: {
```
```suggestion
  options: {
```
(since we would only call this function when it is defined)

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -589,7 +601,60 @@ async function processPartialPayment(
...
+    )
+    const timeoutMs = Math.min(timeRemaining, maxWaitMs)
+    const pollingFrequencyMs = 50
+
+    try {
+      decision = await poll({
+        request: async (): Promise<string> => {
+          try {
+            const value = await redis.get(cacheKey)
+            return value ?? ''
```
mkurapov: I think we should expect the `value` to be stringified `{ approved: boolean, reason?: string })`, and we can then just check for `approved` boolean (see https://github.com/interledger/rafiki/issues/3687)

packages/backend/src/payment-method/ilp/connector/index.ts
```diff
@@ -79,7 +80,7 @@ export async function createConnectorService({
       // Incoming Rules
       createIncomingErrorHandlerMiddleware(ilpAddress),
       createStreamAddressMiddleware(),
-
+      createPartialPaymentDecisionMiddleware(),
```
mkurapov: should we move this to be after the `createBalanceMiddleware` ? That way, we will only publish the webhook & start polling for the decision after we were able to created a pending transfer between the peer account to the incoming payment account, confirming we have enough funds to make the transfer anyway.

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -347,7 +347,12 @@ async function pay(
     }
   )
   try {
-    const receipt = await Pay.pay({ plugin, destination, quote })
+    const receipt = await Pay.pay({
+      plugin,
+      destination,
+      quote,
+      appData: Buffer.from('hello kyc')
```
mkurapov: We can use the outgoing payments' dataToTransmit here now I think

mkurapov: We can use the outgoing payments' dataToTransmit here now I think

package.json
```diff
@@ -77,6 +77,8 @@
       "path-to-regexp@>=6.3.0": "^6.3.0",
       "next": "^15.2.3",
       "form-data": "^4.0.4",
+      "@interledger/stream-receiver": "0.3.3-alpha.4",
+      "@interledger/pay": "0.4.0-alpha.10",
```
mkurapov: Let's put these just in the `backend` package (unless I'm mistaken and they are used elsewhere too)

packages/backend/src/payment-method/ilp/service.ts
```diff
@@ -347,7 +347,14 @@ async function pay(
     }
   )
   try {
-    const receipt = await Pay.pay({ plugin, destination, quote })
+    const receipt = await Pay.pay({
+      plugin,
+      destination,
+      quote,
+      appData: outgoingPayment.dataToTransmit
+        ? Buffer.from(outgoingPayment.dataToTransmit, 'utf8')
```
mkurapov: if you use `outgoingPayment.getDataToTransmit()` it will be decrypted, such that the reciever won't have to do that. It is already encrypted via STREAM

packages/backend/src/payment-method/ilp/connector/index.ts
mkurapov: Thoughts on this ^?

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
```diff
@@ -34,8 +47,23 @@ export function createStreamAddressMiddleware(): ILPMiddleware {
...
-    await next()
+    // Decode frames to check for additional data
+    try {
+      if (ctx.state.streamServer && ctx.state.streamDestination) {
+        const replyOrMoney = ctx.state.streamServer.createReply(
+          ctx.request.prepare
+        )
+        let payload: string | undefined
+        if (!isIlpReply(replyOrMoney)) {
+          payload = getAdditionalDataFromReply(replyOrMoney)
```
mkurapov, line 58: Is this the function we won't need anymore if we can get it through the updates to the `stream-receiver` library?

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
```diff
@@ -0,0 +1,65 @@
...
+        { error, incomingPaymentId },
+        'failed to process partial payment'
+      )
+      message = 'Error processing partial payment'
+    }
+    const errorData = Buffer.from(
+      message ?? 'Error processing partial payment',
+      'utf8'
+    )
+    ctx.response.reply = replyOrMoney.finalDecline(errorData)
```
mkurapov, line 70: What are the consequences here if we do not throw (like before)?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -626,16 +705,29 @@ async function updatePartialPaymentDecision(
...
+    logger.warn(
+      {
+        incomingPaymentId: options.id,
+        partialPaymentId: options.partialPaymentId
+      },
+      'redis not configured; cannot update partial payment decision'
+    )
+    return false
+  }
   const cacheKey = `${PARTIAL_PAYMENT_DECISION_PREFIX}:${options.id}:${options.partialPaymentId}`
```
mkurapov: Maybe let's have a separate small function for this so it's consistent in both methods

packages/backend/src/config/app.ts
```diff
@@ -205,6 +205,14 @@ export const Config = {
     'SEND_TENANT_WEBHOOKS_TO_OPERATOR',
     false
   ),
+  partialPaymentDecisionMaxWaitMs: envInt(
```
mkurapov, line 212: Maybe we should have the decision behaviour (send webhook & poll) false by default (even if there are additional data frames on the packet). What do you think? The ASE could enable it smoothly.

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
sanducb: When setting the reply by calling `finalDecline` we respond to the sender with a final F99 error (along with the error message coming from the ASE/Rafiki) and closing frame on the reject. Throwing was not correct IMO since purposefully rejecting a payment shouldn't be treated as an unexpected error i.e. handled by the error middleware.

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
sanducb: We still need it to get through the `dataFrames` array on the `IncomingMoney` reply that was added to `stream-receiver`, unless you are referring to something else.

packages/documentation/src/content/docs/v1-beta/integration/deployment/services/backend-service.mdx
```diff
@@ -106,6 +106,7 @@ The `backend` service exposes a GraphQL [Backend Admin API](/v1-beta/apis/graphq
...
 | `ENABLE_SPSP_PAYMENT_POINTERS`                        | `backend.enable.spspPaymentPointers`                     | `true`                                                                      | When `true`, the SPSP route is enabled.                                                                                                                                                                                                     |
+| `DB_ENCRYPTION_SECRET`                                | _undefined_                                              | _undefined_                                                                 | Base64-encoded 32-byte secret used to encrypt/decrypt transmitted payment data (`dataToTransmit`) stored on incoming/outgoing payment records and events.                                                                                   |
```
mkurapov: We don't need to include it in the v1-beta docs

localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
```diff
@@ -187,6 +188,84 @@ export async function handleIncomingPaymentCompletedExpired(
...
+      }
+
+      const decipher = createDecipheriv(
+        'aes-256-gcm',
+        Uint8Array.from(Buffer.from(dbEncryptionSecret, 'base64')),
+        iv
+      )
+      decipher.setAuthTag(Uint8Array.from(Buffer.from(tag, 'base64')))
+      decipher.update(cipherText, 'base64', 'utf8')
+      decipher.final('utf8')
```
mkurapov: Do you think we should decrypt this data before publishing the webhook?

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -81,20 +81,24 @@ export interface IncomingPaymentService
...
 }
 
 export interface ServiceDependencies extends BaseService {
-  redis: Redis
   knex: TransactionOrKnex
   accountingService: AccountingService
   walletAddressService: WalletAddressService
   assetService: AssetService
   config: IAppConfig
+  redis?: Redis
```
mkurapov: What are the consequences of making this required?

open-payments-specifications
mkurapov, line 1: Do we need to change the specs here? or was it leftover from a merge?

packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts
mkurapov: Ah, I incorrectly thought the `IncomingMoney` reply just returned the StreamData frame, and not the full list of `dataFrames`, meaning we still need to iterate through it as you are doing here :+1:

packages/backend/src/open_payments/payment/incoming/service.ts
sanducb: None, I made it required 👍

open-payments-specifications
sanducb: Leftover from the rebase, fixed it in the latest commit

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    try {
+      decision = (await ctx.services.incomingPayments.processPartialPayment(
+        incomingPaymentId,
+        {
+          dataToTransmit: additionalData,
+          partialIncomingPaymentId: uuid(),
+          expiresAt: prepare.expiresAt
+        }
+      )) as PartialPaymentDecision
```
mkurapov:
```suggestion
      ))
```

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -81,20 +81,24 @@ export interface IncomingPaymentService
...
   processPartialPayment(
     id: string,
-    args: ProcessPartialPaymentArgs
-  ): Promise<IncomingPayment | IncomingPaymentError>
+    options?: {
+      dataToTransmit?: string
+      partialIncomingPaymentId?: string
+      expiresAt?: Date
+    }
+  ): Promise<PartialPaymentDecision>
```
```suggestion
    options: {
      dataToTransmit: string
      partialIncomingPaymentId: string
      expiresAt: Date
    }
  ): Promise<PartialPaymentDecision>
```

PR #3717, opened 2025-10-24
feat(backend, auth): /healthz fails if a database is not connected

njlie:

- Before responding to requests on the `/healthz` endpoint, both the `backend` and `auth` services will try to reach both its postgres database and its redis instance and responding appropriately based on the result.

## Context

Closes RAF-1172.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/auth/src/app.ts
+++ b/packages/auth/src/app.ts
@@ -221,18 +221,28 @@ export class App {
     koa.use(bodyParser())
 
     const redis = await this.container.use('redis')
-    koa.use(async (ctx: { path: string; status: number }): Promise<void> => {
-      if (ctx.path === '/healthz') {
-        const knex = await this.container.use('knex')
-        try {
-          await redis.ping()
-          await knex.raw('SELECT 1')
-          ctx.status = 200
-        } catch (err) {
-          ctx.status = 500
+    koa.use(
+      async (
+        ctx: {
+          path: string
+          status: number
+        },
+        next: Koa.Next
+      ): Promise<void> => {
+        if (ctx.path === '/healthz') {
+          const knex = await this.container.use('knex')
+          try {
+            await redis.ping()
+            await knex.raw('SELECT 1')
+            ctx.status = 200
+          } catch (err) {
+            ctx.status = 500
+          }
+        } else {
+          return next()
         }
       }
-    })
+    )
 
     koa.use(
       async (


packages/backend/src/app.ts
```diff
@@ -306,7 +306,6 @@ export class App {
    */
   public async boot(): Promise<void> {
     this.config = await this.container.use('config')
-    this.logger = await this.container.use('logger')
```
BlairCurrey, line 309: this removal seems to be resulting in the 500 errors on backend/integration/performance tests.

I had assumed the 500 errors were coming from the new healthcheck but as I traced it that didnt make much sense and reverting the `healthz` changes didnt fix the issue.

I guess this was probably removed in error, and the non-null assertion on the logger class property didnt help. We can simply return it.

It's a small thing, but we could probably just assign logger/config in the constructor to bypass the need for the non-null assertion. That would require changing the container singletons to not (unnecessarily?) return promises. Which should be fine. We have this pattern of making them return promises but it seems entirely unnecessary for logger, config, and most others. 

```ts
  logger: Promise<Logger>
  telemetry: Promise<TelemetryService>
  internalRatesService: Promise<RatesService>
  knex: Promise<Knex>
  axios: Promise<AxiosInstance>
  config: Promise<IAppConfig>
  httpTokenService: Promise<HttpTokenService>
  ... etc
 ```
 
Then that makes it looks like maybe we can even just move the entire boot function to the constructor, since it no longer needs to be async. That would also change some timing though (currently we do constructor, then other stuff, then boot) but I dont think that should matter. Guess it's a little can of worms and we should at least just return the line here.

Comment (njlie) [reply packages/backend/src/app.ts:309]: Yeah for now I'll just reintroduce the deleted line.

File: packages/auth/src/app.ts [hunk 2]
```diff
@@ -268,8 +268,16 @@ export class App {
 
     const router = new Router<DefaultState, AppContext>()
     router.use(bodyParser())
-    router.get('/healthz', (ctx: AppContext): void => {
-      ctx.status = 200
+    router.get('/healthz', async (ctx: AppContext): Promise<void> => {
```
mkurapov, line 295: let's add this route for the Admin API as well (like in `backend`)

BlairCurrey (approved):

packages/auth/src/app.ts
```diff
@@ -220,6 +220,32 @@ export class App {
...
+          const knex = await this.container.use('knex')
+          try {
+            await redis.ping()
+            await knex.raw('SELECT 1')
+            ctx.status = 200
+          } catch (err) {
+            ctx.status = 500
+          }
+        } else if (ctx.path !== '/graphql') {
+          ctx.status = 404
```
mkurapov: Can remove this, as its handled by the next middleware

Issue #3375, opened 2025-04-04
Add Open Payments route for getting grant spent amounts


mkurapov:
Now that we have implemented a new method for getting grant spent amounts in [RAF-1031](https://linear.app/interledger/issue/RAF-1031/implement-new-grant-spent-amount-calculation), we can leverage this to expose a new `GET /outgoing-payment-grant` route in the resource server. 

We will leverage the GNAP access token provided in the header to determine the current grant requested.<br><br>**Route**

`GET /outgoing-payment-grant`<br><br>**Description**

Returns the spent amounts for the current grant corresponding to the presented GNAP access token. If a grant was created with an interval, the amounts returned are for the current interval.

**Response**

```
{
    "spentReceiveAmount": { }
    "spentDebitAmount": { }
}
```

**Errors**

* 403 - Forbidden (if token presented does not match with a grant containing `outgoing-payment` type with `create` access)

PR #3756, opened 2025-11-21
feat: add /GET outgoing-payment-grant route

BlairCurrey:

- adds new `/GET outgoing-payment-grant` route
- changes introspection to only require matching identifier if it's provided (from required if grant has identifier). This is important because the outgoing payment create grant is created with the wallet address url as the identifier (coming from the create request body), but we don't have the wallet address for this new route.

## Context

fixes https://github.com/interledger/rafiki/issues/3375

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/auth/middleware.ts
+++ b/packages/backend/src/open_payments/auth/middleware.ts
@@ -55,12 +55,12 @@ function contextToRequestLike(ctx: HttpSigContext): RequestLike {
 
 function toOpenPaymentsAccess(
   type: AccessType,
-  action?: RequestAction,
+  action: RequestAction,
   identifier?: string
 ): AccessItem {
   return {
     type: type,
-    actions: action ? [action] : [],
+    actions: [action],
     identifier
   } as AccessItem
 }


packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -64,7 +65,9 @@ function toOpenPaymentsAccess(
   } as AccessItem
 }
 
-export function createTokenIntrospectionMiddleware({
+export function createTokenIntrospectionMiddleware<
+  T extends IntrospectionContext = WalletAddressUrlContext
```
BlairCurrey: Can pass in the type, defaulting to the previous type (`WalletAddressUrlContext`). Existing routes are unchanged, but this new outgoing payment spent amount which uses token introspection and doesnt have the wallet address url can pass in it's own context without the url.

njlie: Perhaps it might be overly granular to do so, but for the sake of comprehensiveness a test affirming that it can accept a context without a `walletAddressUrl` could be nice to have.

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -1209,6 +1219,94 @@ export async function updateGrantSpentAmounts(
...
+  // If no records exist, return zero amounts
+  if (!latestGrantSpentAmounts) {
+    return {
+      spentDebitAmount: null,
+      spentReceiveAmount: null
+    }
+  }
+
+  // If an interval is specified, try to get the latest record for it
+  if (limits?.interval) {
```
njlie: Small optimization, but this block doesn't seem to depend on having `latestGrantSpentAmounts`, so performing the read for that could be done after this (and potentially avoided if it returns) block, right?

packages/auth/src/access/utils.ts
```diff
@@ -57,6 +57,7 @@ export function compareRequestAndGrantAccessItems(
 
   if (
     grantAccessIdentifier &&
+    requestAccessIdentifier &&
```
mkurapov, line 60: Let's add a test for this

packages/backend/src/open_payments/payment/outgoing/routes.ts
```diff
@@ -71,6 +80,33 @@ async function getOutgoingPayment(
...
+export interface GrantContext extends IntrospectionContext {
+  grant: { id: string; limits?: Limits }
+}
+
+async function getOutgoingPaymentGrantSpentAmounts(
+  deps: ServiceDependencies,
+  ctx: GrantContext
+): Promise<void> {
+  if (!ctx.grant.id) {
+    throw new OpenPaymentsServerRouteError(403, 'Could not find grant')
```
mkurapov: Let's log here instead, and return a generic message to the client: `Invalid Grant`

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -1209,6 +1219,94 @@ export async function updateGrantSpentAmounts(
...
+    const now = new Date()
+    const currentInterval = getInterval(limits.interval, now)
+
+    if (currentInterval?.start && currentInterval?.end) {
+      const currentIntervalSpentAmounts =
```
mkurapov: We can probably check that the `currentInterval` is within the interval of `latestGrantSpentAmounts` to avoid the additional `OutgoingPaymentGrantSpentAmounts.query`

packages/backend/src/open_payments/payment/outgoing/routes.ts
BlairCurrey: this ended up not being necessary at all - grant.id is not an optional property and the introspection middleware will always validate that its a real grant id.

packages/auth/src/access/utils.ts
BlairCurrey: added

packages/backend/src/open_payments/auth/middleware.ts
BlairCurrey: added [2cd5f0f](https://github.com/interledger/rafiki/pull/3756/commits/2cd5f0f3d334b206caf6bfcecbbfc763ebfd4920)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: That's good. Changed.

(Also refactored to make it a little less verbose)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: I took Nathan's suggestion which means we always just query once. I think if we got the latest and compared we would still need to refetch if it wasnt an interval match https://github.com/interledger/rafiki/pull/3756#discussion_r2566043293

mkurapov: I think we can always fetch the latest grant spent amounts row, and if `currentInterval` is after the interval of the latest grant spent amounts row, we just return 0 spent. As a result, we don't need to fetch `currentIntervalSpentAmounts` explicitly

packages/backend/src/app.ts
```diff
@@ -698,6 +701,19 @@ export class App {
       outgoingPaymentRoutes.get
     )
 
+    // GET /outgoing-payment-grant
+    // Get grant spent amounts (scoped to interval, if any) from grant
+    // with outgoing payment create access
+    router.get(
+      '/:tenantId/outgoing-payment-grant',
+      createTokenIntrospectionMiddleware<IntrospectionContext>({
+        // Expects token used for outgoing payment payment creation
```
mkurapov:
```suggestion
        // Expects outgoing payment token with any action
```

mkurapov:
```suggestion
        // Expects outgoing payment token with any action
```

packages/backend/src/app.ts
```diff
@@ -698,6 +701,19 @@ export class App {
       outgoingPaymentRoutes.get
     )
 
+    // GET /outgoing-payment-grant
+    // Get grant spent amounts (scoped to interval, if any) from grant
+    // with outgoing payment create access
```
mkurapov, line 707:
```suggestion
    // with outgoing payment access
```

packages/backend/src/open_payments/auth/middleware.ts
mkurapov: Just a thought, do you think it would be more explicit/defensive if we have an `ignoreIdentifier` arg for the middleware (default false)?
Then if `ignoreIdentifier` is false, and there is no `walletAddressUrl` we throw

mkurapov (approved):

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: updated

packages/backend/src/app.ts
BlairCurrey: I had forgotten about this decision (make it work for all outgoing payment grants). However, this now seems kind of non sensical. My expectation was that this would allow us to get the grant amounts... but there wont be any unless its has type `create`. You cant spend money against a grant read-only grant.

Currently if its not create, we show a 403 error. Instead if we accepted any action, it would show `null` spent amounts. I guess I could see a case for both but feel like the current 403 makes sense. Its probably not important (and may even be misleading) to show null spent amounts for such a grant - suggests maybe there could be or will be amounts in the future.

packages/backend/src/open_payments/auth/middleware.ts
BlairCurrey: I went further in explicitness and made it a new middleware and extracted the shared behavior out into a function. The original introspection middleware is now the same as before.

This avoids:
- using a generic ctx type and checking if wallet address is defined or not
- `ignoreIdentifier` and maybe erroring at runtime
- potentially making action type optional and setting grant on context for all outgoing payment access types (instead of just create), depending on if we make it "work" with all action types https://github.com/interledger/rafiki/pull/3756#discussion_r2593874235

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -1209,6 +1219,108 @@ export async function updateGrantSpentAmounts(
...
+    }
+  }
+
+  // Interval is specified - determine the current interval
+  const now = new Date()
+  let currentInterval
+
+  try {
+    currentInterval = getInterval(limits.interval, now)
+    if (!currentInterval?.start || !currentInterval?.end) {
```
mkurapov, line 1268: should we also add a warn log if this conditional happens?

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -1209,6 +1219,108 @@ export async function updateGrantSpentAmounts(
...
+  // Check if the latest record is from the current interval
+  const recordIntervalStart = latestGrantSpentAmounts.intervalStart
+  const recordIntervalEnd = latestGrantSpentAmounts.intervalEnd
+
+  if (
+    recordIntervalStart &&
+    recordIntervalEnd &&
+    recordIntervalStart.getTime() ===
+      currentInterval.start.toJSDate().getTime() &&
+    recordIntervalEnd.getTime() === currentInterval.end.toJSDate().getTime()
```
mkurapov: If you want, we can use luxon library directly and do something like `Interval.from(recordInterval).equals(currentInterval)`

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -166,6 +132,86 @@ export function createTokenIntrospectionMiddleware({
   }
 }
 
+export function createOutgoingPayemntGrantTokenIntrospectionMiddleware() {
```
mkurapov:
```suggestion
export function createOutgoingPaymentGrantTokenIntrospectionMiddleware() {
```

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: updated to that. also added a warning if there is no start or end (would expect to have both here)

packages/backend/src/open_payments/payment/outgoing/service.ts
BlairCurrey: yeah probably so. would be unexpected.

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -166,6 +132,87 @@ export function createTokenIntrospectionMiddleware({
...
+      throw err
+    }
+
+    await next()
+  }
+}
+
+async function introspect(
+  ctx: IntrospectionContext,
+  accessItem: { type: AccessType; action?: RequestAction; identifier?: string }
```
mkurapov:
```suggestion
  accessItem: { type: AccessType; action: RequestAction; identifier?: string }
```

packages/backend/src/open_payments/auth/middleware.ts
mkurapov: Also would be good to write a few simple tests for this middleware

BlairCurrey: added several

packages/backend/src/open_payments/auth/middleware.ts
```diff
@@ -54,12 +55,12 @@ function contextToRequestLike(ctx: HttpSigContext): RequestLike {
 
 function toOpenPaymentsAccess(
   type: AccessType,
-  action: RequestAction,
+  action?: RequestAction,
```
mkurapov:
```suggestion
  action: RequestAction,
```

BlairCurrey: committed locally along with another required change in the body of the function

Issue #3369, opened 2025-04-03
Implement new grant spent amount calculation


mkurapov:
Whenever an outgoing payment is being created, we need to calculate whether the payment is within the limits of grant (and within the grant's interval, if applicable). To improve and make this calculation more performant, we can store cumulative amounts (individual payment amounts, interval amounts, and total grant amounts) that have been spent under a particular grant, in a new `outgoingPaymentGrantSpentAmounts` table. This allows us to look up how much has been spent under a grant simply by fetching the most recent `outgoingPaymentGrantSpentAmounts` row for the grant. This issue groups sub tasks necessary to achieve this.

PR #3757, opened 2025-11-21
feat: grant spent amounts

BlairCurrey:
Parent branch for grant spent amounts calculation rework. Most of the code was already reviewed, but I did add something here that was not captured in the other issues. And that is getting the grant spent amounts in this edge case:

1. amount has been spent against grant
2. new rafiki version is deployed with this method for calculating spent amounts
3. `GET `/outgoing-payment-grant` is called (no spent amounts, but amount was spent against the grant)

In this case we are summing all historical payments (added that in this PR). We do the same thing on the outgoing payment create side so I shared the logic.

fixes https://github.com/interledger/rafiki/issues/3369
linear https://linear.app/interledger/issue/RAF-1031/implement-new-grant-spent-amount-calculation

### DIFF
--- a/.github/workflows/node-build.yml
+++ b/.github/workflows/node-build.yml
@@ -19,26 +19,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/workflows/rafiki/env-setup
-
-      # --- DEBUG: tool versions ---
-      - name: Debug Prettier + Node versions
-        run: |
-          node --version
-          pnpm --version
-          pnpm exec prettier --version
-
-      # --- DEBUG: show prettier config resolution ---
-      - name: Debug Prettier config
-        run: |
-          pnpm exec prettier --debug-check packages/documentation/src/content/docs/admin/admin-user-guide.mdx
-
-      # --- DEBUG: show actual diff ---
-      - name: Debug Prettier diff
-        run: |
-          pnpm exec prettier packages/documentation/src/content/docs/admin/admin-user-guide.mdx > /tmp/prettier.out
-          diff -u packages/documentation/src/content/docs/admin/admin-user-guide.mdx /tmp/prettier.out || true
-
-
       - run: pnpm checks
 
   backend:


packages/backend/src/open_payments/auth/middleware.test.ts
```diff
@@ -868,3 +871,498 @@ describe('HTTP Signature Middleware', (): void => {
...
+      }
+      mockTokenIntrospectionClient.introspect.mockResolvedValue(tokenInfo)
+
+      const middleware =
+        createOutgoingPaymentGrantTokenIntrospectionMiddleware()
+      const next = jest.fn()
+
+      await middleware(ctx, next)
+
+      expect(mockTokenIntrospectionClient.introspect).toHaveBeenCalledWith({
```
mkurapov, line 1111: let's also validate that next was called heere

mkurapov (approved):
After we have updated the NodeJS client, we should also update an integration test to verify grant spent amounts as well

packages/backend/migrations/20250409173617_add_outgoing_payment_grant_spent_amounts_table.js
```diff
@@ -0,0 +1,41 @@
...
+      table.bigInteger('intervalDebitAmountValue').nullable()
+      table.bigInteger('grantTotalDebitAmountValue').notNullable()
+      table.string('paymentState').notNullable()
+      table.timestamp('intervalStart').nullable()
+      table.timestamp('intervalEnd').nullable()
+      table.timestamp('createdAt').defaultTo(knex.fn.now())
+    })
+    .then(() => {
+      return knex.raw(
+        'CREATE INDEX outgoingPaymentGrantSpentAmounts_grantId_createdAt_desc_idx ON "outgoingPaymentGrantSpentAmounts" ("grantId", "createdAt" DESC)'
```
BlairCurrey, line 30: I think I'll leave it for now.... but I was thinking about changing this:

```ts
'CREATE INDEX outgoingPaymentGrantSpentAmounts_grantId_createdAt_desc_idx ON "outgoingPaymentGrantSpentAmounts" ("grantId", "createdAt" DESC)'
```

To include the interval fields: 

```ts
knex.raw('CREATE INDEX ... ON "outgoingPaymentGrantSpentAmounts" ("grantId", "intervalStart", "intervalEnd", "createdAt" DESC)')`
```

because we look it up like this in the worker when processing the payment:

```ts
await OutgoingPaymentGrantSpentAmounts.query(deps.knex)
          .where('grantId', grantId)
          .where('intervalStart', latestPaymentSpentAmounts.intervalStart)
          .where('intervalEnd', latestPaymentSpentAmounts.intervalEnd)
          .orderBy('createdAt', 'desc')
          .first()
```

Dont have great intuition on tradeoffs. I guess it would slow writes some amount in all cases and speed up reads some amount in a subset of cases (when we lookup by interval - only in worker I think). Probably no db storage space difference since a created_at index is going to already lead to a unique index entry per row. Overall I guess it feels like a bad idea. I guess we probably read a bit more (read and write on create, read and write in worker, read in `GET /outgoing-payment-grant`) but we only lookup by interval in one place.

sanducb (approved):

PR #3761, opened 2025-11-28
docs: Add ESLint and Prettier config to documentation package

JonathanMatthey:
#### Description:
Adds ESLint and Prettier configuration to packages/documentation, mirrored from paymentpointers.org:
eslint.config.js with TypeScript, Astro, and Prettier integration

- .prettierrc and .prettierignore
- format and lint scripts in package.json
- Formatted all existing files to pass lint

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ pnpm format
 ```
 
 This command will:
+
 - Format all files using Prettier
 - Automatically fix ESLint issues where possible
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,8 +11,9 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "lint": "eslint --max-warnings=0 --fix .",
+    "lint:docs": "pnpm -C ./packages/documentation lint",
```
Anca2022, line 14: 🔥 I learned something new, thanks!

packages/documentation/.github/workflows/ci.yml
Anca2022, line 1: I think this file is now redundant, since formatting is already checked in the `node_build` workflow.

PR #3766, opened 2025-12-03
feat(backend): store data to be transmitted to receiver in outgoing payments

njlie:

- Adds two environment variables, the `DB_ENCRYPTION_SECRET` and the `DB_ENCRYPTION_IV`, to be used in the symmetric encryption of sensitive columns in the database
- Adds a `dataToTransmit` field to the outgoing payment model
- Adds a `dataToTransmit` field to the `depositOutgoingPaymentLiquidity` GraphQL resolver

## Context

Fixes RAF-1182 and fixes RAF-1179.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/schema.graphql
+++ b/packages/backend/src/graphql/schema.graphql
@@ -605,7 +605,7 @@ input DepositOutgoingPaymentLiquidityInput {
   "Unique key to ensure duplicate or retried requests are processed only once. For more information, refer to [idempotency](https://rafiki.dev/apis/graphql/admin-api-overview/#idempotency)."
   idempotencyKey: String!
   "Data to be encrypted and sent to the receiver."
-  senderData: String
+  dataToTransmit: String
 }
 
 input CreateIncomingPaymentWithdrawalInput {
--- a/packages/backend/src/open_payments/payment/outgoing/model.ts
+++ b/packages/backend/src/open_payments/payment/outgoing/model.ts
@@ -134,11 +134,11 @@ export class OutgoingPayment
 
   public tenantId!: string
 
-  public senderData?: string
-  public getSenderData(key?: string): string | null {
-    if (!this.senderData) return null
-    if (!key) return this.senderData
-    const { tag, cipherText, iv } = JSON.parse(this.senderData)
+  public dataToTransmit?: string
+  public getDataToTransmit(key?: string): string | null {
+    if (!this.dataToTransmit) return null
+    if (!key) return this.dataToTransmit
+    const { tag, cipherText, iv } = JSON.parse(this.dataToTransmit)
 
     const decipher = createDecipheriv(
       'aes-256-gcm',
@@ -146,10 +146,10 @@ export class OutgoingPayment
       iv
     )
     decipher.setAuthTag(Uint8Array.from(Buffer.from(tag, 'base64')))
-    let decryptedSenderData = decipher.update(cipherText, 'base64', 'utf8')
-    decryptedSenderData += decipher.final('utf8')
+    let decryptedDataToTransmit = decipher.update(cipherText, 'base64', 'utf8')
+    decryptedDataToTransmit += decipher.final('utf8')
 
-    return decryptedSenderData
+    return decryptedDataToTransmit
   }
 
   static get relationMappings() {
--- a/packages/backend/src/open_payments/payment/outgoing/service.ts
+++ b/packages/backend/src/open_payments/payment/outgoing/service.ts
@@ -707,12 +707,18 @@ export interface FundOutgoingPaymentOptions {
   tenantId: string
   amount: bigint
   transferId: string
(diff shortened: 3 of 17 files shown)


packages/backend/src/graphql/schema.graphql
```diff
@@ -1092,6 +1094,8 @@ type OutgoingPayment implements BasePayment & Model {
   grantId: String
   "Tenant ID of the outgoing payment."
   tenantId: String
+  "Data to be transmitted to receiver."
+  dataToTransmit: String
```
sanducb: Why do we need `dataToTransmit` here?

packages/backend/src/graphql/schema.graphql
```diff
@@ -604,6 +604,8 @@ input DepositOutgoingPaymentLiquidityInput {
   outgoingPaymentId: String!
   "Unique key to ensure duplicate or retried requests are processed only once. For more information, refer to [idempotency](https://rafiki.dev/apis/graphql/admin-api-overview/#idempotency)."
   idempotencyKey: String!
+  "Data to be encrypted and sent to the receiver."
+  dataToTransmit: String
```
sanducb: Even though `dataToTransmit` is explicit enough for ourselves to understand it, I think that integrators might benefit from having it named something like `senderData` or `senderDataToTransmit` because of "sender's data" being a common phrase in a "payments" context. WDYT? CC @mkurapov

localenv/cloud-nine-wallet/docker-compose.yml
```diff
@@ -115,6 +115,8 @@ services:
       OPERATOR_TENANT_ID: 438fa74a-fa7d-4317-9ced-dde32ece1787
       CARD_SERVICE_URL: 'http://cloud-nine-wallet-card-service:3007'
       CARD_WEBHOOK_SERVICE_URL: 'http://cloud-nine-wallet-card-service:3007/webhook'
+      DB_ENCRYPTION_SECRET: 'zO9KogehJECHReHgQr+ZWGkmgOD4AYa4ksUxALSwgM8='
+      DB_ENCRYPTION_IV: 'e9jyNk0CKajCgI93Ga2v23R/1wGZ2lO339QRaOFgxHM='
```
mkurapov: Should we instead generated a unique IV every time the data is signed and store it together with the data? Otherwise, I think we would get the same ciphertext given the same input/ `dataToTransmit`

packages/backend/src/graphql/schema.graphql
mkurapov: I think we can indeed remove this from this resolver (I don't think the ASE needs to look it up necessarily).

packages/backend/src/graphql/schema.graphql
njlie: If "sender data" is an existing concept in payments, then I think `senderData` would work here as the field name.

mkurapov: @sanducb good q & I see your point but in the spirit of keeping it more generic, `dataToTransmit` works better I think. Even though it will usually be the sender's data (or more specifically, the sending customer's data) that is sent, the ASE can choose to pass in some other payment metadata in there

packages/backend/src/open_payments/payment/outgoing/service.ts
```diff
@@ -756,7 +764,14 @@ async function fundPayment(
     if (error) {
       return error
     }
-    await payment.$query(trx).patch({ state: OutgoingPaymentState.Sending })
+
+    await payment.$query(trx).patch({
+      state: OutgoingPaymentState.Sending,
+      dataToTransmit:
+        deps.config.dbEncryptionSecret && dataToTransmit
+          ? encryptDbData(dataToTransmit, deps.config.dbEncryptionSecret)
```
mkurapov, line 772: I actually quite like this, gives ASE option to encrypt if they choose to do so

packages/backend/src/open_payments/payment/outgoing/model.ts
```diff
@@ -133,6 +134,24 @@ export class OutgoingPayment
 
   public tenantId!: string
 
+  public dataToTransmit?: string
+  public getDataToTransmit(key?: string): string | null {
+    if (!this.dataToTransmit) return null
+    if (!key) return this.dataToTransmit
+    const { tag, cipherText, iv } = JSON.parse(this.dataToTransmit)
```
mkurapov, line 141: Key rotation might be a bit tricky here, since we could end up encrypting with one key, change the key, try to decrypt with the old one and fail. We can focus on this on a follow-up PR.
(maybe we allow to configure a list of keys instead?)

njlie: Captured in [RAF-1207](https://linear.app/interledger/issue/RAF-1207/handle-key-rotation-for-existing-data).

PR #3772, opened 2025-12-09
feat(backend): add handlePartialPayment for incoming payments

njlie:

- Adds a `handlePartialPayment` service call to the incoming payment service, to be called when encrypted data is transmitted through an ILP stream.

## Context

Fixes RAF-1183 and fixes RAF-1184.

## Checklist

- [x] Tests added/updated


localenv/mock-account-servicing-entity/app/lib/webhooks.server.ts
mkurapov:
```suggestion
          dataToTransmit: acc.name
```

Since we have it already anyway

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -549,6 +555,43 @@ async function addReceivedAmount(
...
+  const incomingPayment = await IncomingPayment.query(knex)
+    .findById(id)
+    .withGraphFetched('asset')
+  if (!incomingPayment) return IncomingPaymentError.UnknownPayment
+
+  await IncomingPaymentEvent.query(knex).insertGraph({
+    incomingPaymentId: incomingPayment.id,
+    type: IncomingPaymentEventType.IncomingPaymentPartialPaymentReceived,
+    data: {
+      ...incomingPayment.toData(0n),
```
mkurapov, line 574: Here, we need to add a new property to the data that specifies the amount that was only sent in the "partial" payment, so the receiving ASE can make a decision whether to action it or not. The `receivedAmount` should remain unchanged, ie, it will include the cumulative amount received thus far for the incoming payment.

I will encapsulate this in a separate issue.

packages/backend/src/graphql/schema.graphql
mkurapov: If we keep the DB encryption secret as optional, this can just be: "Data to transmit to the recipient during the payment"

packages/backend/src/graphql/resolvers/liquidity.test.ts
mkurapov: I think instead of this, we can just assert that depositSpy was called with dataToTransmit (since the resolver doesn't really need to know about how it gets used in the method itself)

packages/backend/src/open_payments/payment/outgoing/service.ts
mkurapov, line 773: Do you think the encryption should be an optional feature? I think this might be a good call, as it allows the integrator to decide whether they want it or not.

njlie: Yeah, I figured leaving it open was the best possibility - I definitely can see cases where an integrator might not want it, and therefore not have to deal with another piece of the environment to configure.

sanducb (approved):
Looks good!

mkurapov (approved):

Issue #3638, opened 2025-09-10
docs: build out Integration > Requirements > Assets


brad-dow:
Currently, [the Assets page](https://rafiki.dev/integration/requirements/assets/) only shows how to create an asset. We need to expand this page by creating sections for updating an asset and deleting an asset. Both of these sections should contain the multi-tabbed interface for operation, variables, and response.

PR #3774, opened 2025-12-10
docs: update tenants, assets, and admin user guide

brad-dow:
Fixes https://github.com/interledger/rafiki/issues/3638

**Description of changes**

- Updated assets.mdx with sections for updating and deleting assets 
- Updated tenants.mdx to include more contextual information before each operation (add, update, delete) 
- Updated the admin user guide to specify read-only values when editing 

Overall, improvements made for multi-tenancy consistency and cohesion across pages

**Required**

- [x] Used LinkOut component on external links
- [x] Ran Prettier
- [x] Previewed updates in Netlify
- [ ] Included a "fixes #" comment

**Conditional**

- [ ] Included code samples where appropriate
- [ ] Updated related READMEs

### DIFF
--- a/packages/documentation/src/content/docs/integration/requirements/assets.mdx
+++ b/packages/documentation/src/content/docs/integration/requirements/assets.mdx
@@ -15,7 +15,7 @@ An asset represents an item of value that can be transferred via the Interledger
 
 Operators can create, view, edit, and delete both their own assets and those that belong to any tenant.
 
-Tenants can view, edit, and delete only their own assets. They cannot create assets.
+Tenants can create, view, edit, and delete their own assets only.
 
 ## Add an asset
 
@@ -83,9 +83,9 @@ The `asset` object in the response will include the `tenantId` of the tenant to 
 
 ## Update an asset
 
-Once an asset has been created, you may need to adjust its operational threshold values. You can use the `updateAsset` mutation to modify the `withdrawalThreshold` and `liquidityThreshold`.
+Once an asset has been created, you may need to adjust its operational threshold values. You can use the `updateAsset` mutation to modify the `liquidityThreshold`.
 
-These thresholds influence when Rafiki triggers low‑liquidity notifications but do not directly deposit or withdraw asset liquidity.
+This threshold determines when Rafiki triggers low‑liquidity notifications or webhooks, but does not directly deposit or withdraw asset liquidity.
 
 For information about adding or removing liquidity, see [Asset liquidity](/admin/liquidity/asset-liquidity).
 
@@ -160,17 +160,6 @@ The `asset` object in the response will include the `tenantId` of the tenant to 
 Deleting an asset removes it from active use within a Rafiki instance.
 
 You can only delete an asset if it's not associated with any peers or wallet addresses. Rafiki prevents deleting any asset that's still in use.
-
-:::danger
-Deleting an asset is permanent and cannot be reversed.
-
-You can only delete an asset if:
-
-- No peers reference the asset
-- No wallet addresses are associated with it
-
-If the asset is still in use, the backend prevents deletion and returns an error.
-:::
 
 <Tabs>
   <TabItem label="Operation">
--- a/packages/documentation/src/content/docs/admin/admin-user-guide.mdx
+++ b/packages/documentation/src/content/docs/admin/admin-user-guide.mdx
@@ -215,7 +215,9 @@ The Assets page allows you to manage assets in your Rafiki instance, including v
 On this page, all configured assets appear in a table where you can view the asset ID, the asset code, the scale, and the withdrawal threshold. The list of assets shows only those for the current tenant, as determined by your API credentials.
 
 :::note
-Operators can create, edit, and delete assets for any tenant. Tenants can only view and manage their own assets.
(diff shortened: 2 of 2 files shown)


packages/documentation/src/content/docs/integration/requirements/assets.mdx
```diff
@@ -81,3 +83,152 @@ The `asset` object in the response will include the `tenantId` of the tenant to
...
+          tenantId
+        }
+      }
+    }
+    ```
+  </TabItem>
+
+  <TabItem label="Variables">
+    The input object for the update operation requires the asset `id`.  
+    Only the threshold fields are editable; all other fields are immutable.
```
melissahenderson: Is this supposed to start on a new line? It's OK if it's intentional.

packages/documentation/src/content/docs/integration/requirements/assets.mdx
```diff
@@ -81,3 +83,152 @@ The `asset` object in the response will include the `tenantId` of the tenant to
...
+  </TabItem>
+</Tabs>
+
+---
+
+## Delete an asset
+
+Deleting an asset removes it from active use within a Rafiki instance.
+
+This operation succeeds only if the asset is **not** referenced by any peers or wallet addresses.
```
melissahenderson: I'm not sure what is meant by "referenced". I did a _super_ quick search and I don't see "reference" being used anywhere else (other than in the list a bit below) in this context. If that's the right word to use, then that's fine.

packages/documentation/src/content/docs/integration/requirements/tenants.mdx
```diff
@@ -111,6 +115,10 @@ After you create a tenant, securely communicate the tenant `id` and `apiSecret`
 
 ## Update a tenant
 
+Use the `updateTenant` mutation to modify an existing tenant’s configuration.
+
+Tenants can update their own profile details such as the public name, contact email, or identity provider URLs. Operators can update any tenant’s information except for the `apiSecret`, which cannot be changed through this mutation.
```
melissahenderson: Is there some other mutation that lets them change the api secret, or can it just not be changed without deleting and recreating the tenant?

packages/documentation/src/content/docs/integration/requirements/assets.mdx
brad-dow: This verbiage came from a deeper dive into how deletes actually work, but it doesn't sound very natural. Basically, that delete mutation checks (at the database level) if any peer or wallet address record is using an asset's `id` as a key. 

I have some similar content in the live version of the admin user guide - https://rafiki.dev/admin/admin-user-guide/#delete-asset

"Reference" is used twice here and "associated" is used once. Let me find a smoother way to match similar phrasing - maybe something like "...if the asset is not associated with any peers..."

packages/documentation/src/content/docs/integration/requirements/assets.mdx
brad-dow: Not intentional...good catch

packages/documentation/src/content/docs/integration/requirements/assets.mdx
```diff
@@ -13,15 +13,14 @@ An asset represents an item of value that can be transferred via the Interledger
...
 
-- Operators can create assets for any tenant
-- Operators can view any asset
-- Operators can edit and delete assets that belong to any tenant
-- Tenants can only view their own assets
-- Tenants can only edit and delete their own assets
-- Tenants cannot create assets
+Operators can create, view, edit, and delete both their own assets and those that belong to any tenant.
+
+Tenants can view, edit, and delete only their own assets. They cannot create assets.
```
mkurapov: Tenants can also create their own assets

packages/documentation/src/content/docs/integration/requirements/assets.mdx
```diff
@@ -81,3 +80,150 @@ The `asset` object in the response will include the `tenantId` of the tenant to
 
   </TabItem>
 </Tabs>
+
+## Update an asset
+
+Once an asset has been created, you may need to adjust its operational threshold values. You can use the `updateAsset` mutation to modify the `withdrawalThreshold` and `liquidityThreshold`.
```
mkurapov: Can we de-emphasize `withdrawalThreshold` here? 😅 Its referring to old behaviour for web monetization for wallet addresses. But `liquidityThreshold` is good to keep, this is indeed after which we would send the liquidity low notifications/webhooks

packages/documentation/src/content/docs/integration/requirements/assets.mdx
```diff
@@ -81,3 +80,150 @@ The `asset` object in the response will include the `tenantId` of the tenant to
...
+
+---
+
+## Delete an asset
+
+Deleting an asset removes it from active use within a Rafiki instance.
+
+You can only delete an asset if it's not associated with any peers or wallet addresses. Rafiki prevents deleting any asset that's still in use.
+
+:::danger
```
mkurapov: Do you think we need the danger here? Given that it not possible to delete an asset with any existing references, I think admins shouldn't worry about breaking something :)

PR #3805, opened 2026-01-21
fix: internal server error

### LINKED ISSUE #3804: [BUG] Interval Server Error when making Grant Payment Across Versions (v2.2.0-beta -> main)
I am seeing a bug while trying to test grant spent amount calculations in this scenario:

1. on last release, start a grant and make some payments against it.
2. upgrade to latest rafiki with new grant spent amount calc method
3. verify you can get the correct spent amounts (should use legacy calc method)
4. make an additional payment against the same grant
5. very the new spent amounts are correct (should use the new calc method, should have a record in new spent amount table)

The new outgoing payment create on the latest Rafiki commit fails with an error about not finding the route. This appears to have something to do with the static routing. In theory the route details should be persisted then loaded on start, and fallback to `peer.staticIlpAddress` when there are none (this case), but it still doesnt find a route when we look for it in the connector.

Full dump triggered by bruno create request:

```json
{"level":20,"time":1768507066406,"pid":30,"hostname":"cloud-nine-wallet-backend","service":"RouterService","destination":"test.happy-life-bank.-yZ6e3T4e_FLNQ3gz9yqPlnYtCgpX0ewmFUwelccW68TMI9_RC3Jg0JaMm2COvdgYy-FCJjAr937fGNZuM9LkuSdwzWyXyU","tenantId":"438fa74a-fa7d-4317-9ced-dde32ece1787","msg":"no static route found"}
{"level":50,"time":1768507066406,"pid":30,"hostname":"cloud-nine-wallet-backend","service":"ConnectorService","destinationAddress":"test.happy-life-bank.-yZ6e3T4e_FLNQ3gz9yqPlnYtCgpX0ewmFUwelccW68TMI9_RC3Jg0JaMm2COvdgYy-FCJjAr937fGNZuM9LkuSdwzWyXyU","msg":"unknown destination account"}
{"level":30,"time":1768507066406,"pid":30,"hostname":"cloud-nine-wallet-backend","service":"ConnectorService","err":{"type":"UnreachableError","message":"unknown destination account","stack":"Error: unknown destination account\n    at account (/home/rafiki/packages/backend/src/payment-method/ilp/connector/core/middleware/account.ts:126:13)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at /home/rafiki/packages/backend/src/payment-method/ilp/connector/core/middleware/stream-address.ts:38:5\n    at ilpTiming (/home/rafiki/packages/backend/src/payment-method/ilp/connector/core/middleware/ilp-timing.ts:46:7)\n    at Rafiki.handleIlpData (/home/rafiki/packages/backend/src/payment-method/ilp/connector/core/rafiki.ts:171:5)\n    at RateProbe.sendRequest (/home/rafiki/node_modules/.pnpm/@interledger+pay@0.4.0-alpha.9/node_modules/@interledger/pay/src/request.ts:114:32)","ilpErrorCode":"F02"},"msg":"Error thrown in incoming pipeline"}
{"level":50,"time":1768507076970,"pid":30,"hostname":"cloud-nine-wallet-backend","service":"IlpPaymentService","err":"RateProbeFailed","msg":"Received error during ILP quoting"}
InternalServerError: Internal Server Error
    at Object.throw (/home/rafiki/node_modules/.pnpm/koa@2.16.0/node_modules/koa/lib/context.js:97:11)
    at openPaymentsServerErrorMiddleware (/home/rafiki/packages/backend/src/open_payments/route-errors.ts:105:14)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at processTimers (node:internal/timers:516:9)
```

See: `"no static route found"` in `RouterService`.

## Reproduction steps:

- increase grant expiry from 10m so it doesnt expire often while testing
- checkout last release: `v2.2.0-beta`
- initialize clean rafiki: `rafiki` (alias on my machine to whipe volumes, build, start containers)
- start Bruno webmonetization example. stop after 2 payments
- spin down rafiki
- checkout latest rafiki: `main`
- start rafiki again: `pnpm localenv:compose up --build`
- attempt to make the 3rd payment from Bruno and see internal server error

BlairCurrey:

- not properly falling back to `staticIlpAddress` when setting routes

## Context

fixes #3804 

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -65,6 +65,7 @@ import { AuthServiceClient } from './auth-service-client/client'
 import { createTenantSettingService } from './tenants/settings/service'
 import { createPaymentMethodProviderService } from './payment-method/provider/service'
 import { createRouterService } from './payment-method/ilp/connector/ilp-routing/service'
+import { loadRoutesFromDatabase } from './shared/utils'
 
 BigInt.prototype.toJSON = function () {
   return this.toString()
@@ -665,34 +666,6 @@ export const gracefulShutdown = async (
 
   const telemetry = await container.use('telemetry')
   telemetry.shutdown()
-}
-
-const loadRoutesFromDatabase = async (
-  container: IocContract<AppServices>
-): Promise<void> => {
-  const peerService = await container.use('peerService')
-  const routerService = await container.use('routerService')
-  const logger = await container.use('logger')
-
-  const peers = await peerService.getPage()
-  logger.info(
-    { peerCount: peers.length },
-    'loading static routes from database'
-  )
-
-  for (const peer of peers) {
-    // If no routes are set, we use the static address of our peers as the only routes
-    const routes =
-      peer.routes && peer.routes.length ? peer.routes : [peer.staticIlpAddress]
-    for (const route of routes) {
-      await routerService.addStaticRoute(
-        route,
-        peer.id,
-        peer.tenantId,
-        peer.assetId
-      )
-    }
-  }
 }
 
 export const start = async (
--- a/packages/backend/src/shared/utils.test.ts
+++ b/packages/backend/src/shared/utils.test.ts
@@ -11,7 +11,8 @@ import {
   sleep,
   getTenantFromApiSignature,
   ensureTrailingSlash,
-  encryptDbData
+  encryptDbData,
+  loadRoutesFromDatabase
 } from './utils'
 import { AppServices, AppContext } from '../app'
 import { TestContainer, createTestApp } from '../tests/app'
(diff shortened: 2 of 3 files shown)


packages/backend/src/index.ts
```diff
@@ -682,7 +682,8 @@ const loadRoutesFromDatabase = async (
 
   for (const peer of peers) {
     // If no routes are set, we use the static address of our peers as the only routes
-    const routes = peer.routes || [peer.staticIlpAddress]
```
BlairCurrey, line 685: was resolving to `[]` when `peer.routes` was `[]`, should return the fallback

mkurapov: Nice find, why did it fail intermittently though? 

And do we have a test case for this?

BlairCurrey: I wouldn’t say it failed intermittently. I noticed this and made a similar change earlier on in my debugging but was still having issues (dont recall if it was the same error, or different error, but creating the payment still didn’t work). I had some other provisional changes - perhaps there were issues with those. But creating the payment was never working when doing this upgrade. 

Although I just had a thought - Why was it working when NOT doing the version upgrade though? Maybe this is what you mean. Im just tracing through the code but I’m pretty sure this is because the route gets set by the create peer mutation during seeding. Whereas the seeding fails when re-upping since the peer already exists - doesn't create a new peer, just works with what was loaded on app start (which was nothing). And the initial seeding didn't set it because that version didn't contain that logic.

I added tests for this function. To support doing so i moved to the `shared/utils` package. I also considered a new `index.test.ts` but that felt like overkill, or moving to the router service with a name that highlights the side effect. Something like `routerService.hydrateStaticRouteDependency()`. I think utils is fine and simpler though.

mkurapov: OK, I think I managed to clarify it. Basically yes, its due to the seeding failing, even though the seed script should be handling that gracefully:

1. On `v2.2.0-beta` the seed file for cloud nine contains the peer: `staticIlpAddress: test.happy-life-bank`.
2. Then we check out `main` & spin up the environment
3. Then `loadRoutesFromDatabase` gets called on startup, which sets the `routes` (erroneously) to `[]`.
4. The seed file on `main` for cloud nine contains the peer: `staticIlpAddress: test.happy-life-bank, routes: [test.happy-life-bank]`. And I see this error: "duplicate peer found for same ILP address and asset. If seeding has already completed this can probably be ignored". Which is because in the seed script, before we attempt to create or update the peer we check the peer's existance via:
```ts
    const existingPeer = await getPeerByAddressAndAsset(
      peerStaticIlpAddress,
      assets[peeringAsset].id
    )
``` 
but because `getPeerByAddressAndAsset` resolver ends up grabbing the peer from the static routes, we can't fetch the peer. Then the error happens because we try creating the peer again instead of just updating it.

BlairCurrey: Yes, exactly.

And we have the code to handle this exact situation (falling back to `staticIlpAddress`), it just had a bug.

Issue #3690, opened 2025-10-07
docs: Create guide for signing requests to the Admin API


brad-dow:
This issue comes from Max's feedback on the initial multi-tenancy docs update:

> I think we also need a guide on how to sign requests to the Admin API, basically doing the inverse of the signature verification:

```
  const timestamp = Date.now()
  const version = process.env.SIGNATURE_VERSION

  const { query, variables, operationName } = request
  const formattedRequest = {
    variables,
    operationName,
    query: print(query)
  }

  const payload = `${timestamp}.${canonicalize(formattedRequest)}`
  const hmac = createHmac(
    'sha256',
    process.env.API_SECRET
  )
  hmac.update(payload)
  const digest = hmac.digest('hex')

  headers['signature'] = `t=${timestamp}, v${version}=${digest}`
  headers['tenant-id'] = process.env.OPERATOR_TENANT_ID
```

PR #3807, opened 2026-01-23
docs: add guide for signing Admin API requests

brad-dow:
**Description of changes**

Fixes https://github.com/interledger/rafiki/issues/3690

This update adds a new "Signing Admin API requests" guide that explains how a client can generate and include HMAC SHA-256 signatures when calling the Backend Admin API.

The guide includes the following:

- Step-by-step instructions for creating valid request signatures
- JavaScript code example provided by Max
- Configuration reference table for relevant environment variables
- A brief explanation of how Rafiki validates these signed requests
- Cross-links to/from "Verify webhook signatures" page in case the reader ended up in the wrong, but similar content

**Required**

- [x] Used LinkOut component on external links
- [x] Ran Prettier
- [ ] Previewed updates in Netlify
- [x] Included a "fixes #" comment

**Conditional**

- [x] Included code samples where appropriate
- [ ] Updated related READMEs

### DIFF
--- a/packages/documentation/src/content/docs/apis/graphql/signing-admin-api-requests.mdx
+++ b/packages/documentation/src/content/docs/apis/graphql/signing-admin-api-requests.mdx
@@ -13,7 +13,7 @@ This signature authenticates the caller and ensures the request body has not bee
 
 Each request is HMAC‑signed with SHA‑256 using the tenant’s or operator’s API secret. This signature ensures that Rafiki can verify where the request originated and confirm that the payload data matches what was originally signed.
 
-All signed requests include two headers: the `signature` header and the `tenant-Id` header.
+All signed requests include two headers: the `signature` header and the `tenant-id` header.
 
 ### `signature` header
 
@@ -103,7 +103,7 @@ const GET_ASSET = gql`
 `
 
 const requestBody = {
-  query: print(GET_ASSET),  // converts `DocumentNode` to string
+  query: print(GET_ASSET), // converts `DocumentNode` to string
   variables: { id: 'asset-id-here' },
   operationName: 'GetAsset'
 }


melissahenderson (dismissed):

packages/documentation/src/content/docs/apis/graphql/signing-admin-api-requests.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+Rafiki requires every request to the Backend Admin API to include a valid cryptographic signature.
+
+This signature authenticates the caller and ensures the request body has not been altered in transit.
+
+## Signature overview
+
+Each request is HMAC‑signed with SHA‑256 using the tenant’s or operator’s API secret. This signature ensures that Rafiki can verify where the request originated and confirm that the payload data matches what was originally signed.
+
+All signed requests include two headers: the `signature` header and the `tenant-Id` header.
```
BlairCurrey:
```suggestion
All signed requests include two headers: the `signature` header and the `tenant-id` header.
```

packages/documentation/src/content/docs/apis/graphql/signing-admin-api-requests.mdx
```diff
@@ -0,0 +1,130 @@
...
+  query: print(query)
+}
+
+const payload = `${timestamp}.${canonicalize(formattedRequest)}`
+const hmac = createHmac('sha256', process.env.ADMIN_API_SECRET)
+hmac.update(payload)
+const digest = hmac.digest('hex')
+
+headers['signature'] = `t=${timestamp}, v${version}=${digest}`
+headers['tenant-id'] = process.env.OPERATOR_TENANT_ID
```
BlairCurrey, line 119: @mkurapov thoughts on making what the request looks like more explicit? I have some familiarity with our implementation (which this comes from) but I still had to stop and think about what `formattedRequest` really is, other than "some GQL request in whatever form we need here". What I find unclear currently:

- `request` is not defined anywhere. We can kinda reason about what we destructure off of it but we dont know exactly. Is query a string, object, some special gql class instance? (it's the latter)
- `print` is not defined anywhere

In addition to making it more explicit, it might be useful to use less libraries. This js implementation is an example that should illustrate the concept in general. Someone might need to do this in go, rust etc. I figure the less magic happening from third party libraries the better. And for that matter, someone in js might not be using the same gql libraries.

What I would probably suggest is the more explicit one with less 3rd party libraries:

```js
import { createHmac } from "crypto";
import { canonicalize } from "json-canonicalize";

const timestamp = Date.now();
const version = process.env.ADMIN_API_SIGNATURE_VERSION;

const requestBody = {
  query: `
    query GetAsset($id: String!) {
      asset(id: $id) {
        id
        code
        scale
      }
    }
  `,
  variables: { id: "asset-id-here" },
  operationName: "GetAsset",
};

// Canonicalize ensures both client and server produce identical JSON strings
// by sorting object keys deterministically and normalizing whitespace.
const payload = `${timestamp}.${canonicalize(requestBody)}`;                                                                     

const hmac = createHmac('sha256', process.env.ADMIN_API_SECRET)
hmac.update(payload)
const digest = hmac.digest('hex')

headers['signature'] = `t=${timestamp}, v${version}=${digest}`
headers['tenant-id'] = process.env.OPERATOR_TENANT_ID
```

Or the more explicit version of what we already have with print and the `gql` tagged function:

```js
import { createHmac } from 'crypto'
import { canonicalize } from 'json-canonicalize'
import { gql } from '@apollo/client'
import { print } from 'graphql/language/printer'

const timestamp = Date.now()
const version = process.env.ADMIN_API_SIGNATURE_VERSION

const GET_ASSET = gql`
  query GetAsset($id: String!) {
    asset(id: $id) {
      id
      code
      scale
    }
  }
`

const requestBody = {
  query: print(GET_ASSET),  // converts `DocumentNode` to string
  variables: { id: 'asset-id-here' },
  operationName: 'GetAsset'
}

// Canonicalize ensures both client and server produce identical JSON strings
// by sorting object keys deterministically and normalizing whitespace.
const payload = `${timestamp}.${canonicalize(requestBody)}`
const hmac = createHmac('sha256', process.env.ADMIN_API_SECRET)
hmac.update(payload)
const digest = hmac.digest('hex')

headers['signature'] = `t=${timestamp}, v${version}=${digest}`
headers['tenant-id'] = process.env.OPERATOR_TENANT_ID
```

mkurapov: Good point @BlairCurrey! I like the most explicit version (#2), easiest for integrators to copy-paste 👍

PR #3809, opened 2026-01-23
test: provisional spent amount assertions

BlairCurrey:

- updates node sdk across rafiki (and updated grant types where necessary)
- adds assertions to integration tests for spent amounts

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/grant/service.test.ts
+++ b/packages/backend/src/open_payments/grant/service.test.ts
@@ -14,6 +14,7 @@ import {
   AccessType,
   AccessAction,
   AuthenticatedClient,
+  isFinalizedGrantWithAccessToken,
   mockGrant,
   mockAccessToken,
   mockPendingGrant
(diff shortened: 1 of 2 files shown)


packages/backend/src/open_payments/payment/outgoing/routes.test.ts
```diff
@@ -402,10 +402,7 @@ describe('Outgoing Payment Routes', (): void => {
         grantId,
         limits: undefined
       })
-      expect(ctx.body).toEqual({
-        spentDebitAmount: null,
-        spentReceiveAmount: null
-      })
+      expect(ctx.body).toEqual({})
```
BlairCurrey: I had to make this change to satisfy the open api validation. This wasn't intended. The spec does not require the `spentDebitAmount`/`spentReceiveAmount` but they cannot be present and null, as I originally intended. 

Quick recap: these are null (or undefined now) because we can't determine the asset/scale without any amounts (otherwise they could be 0). 

I prefer the original null properties personally. Just a matter of if that's the better pattern and worth the additional spec change.

packages/auth/package.json
```diff
@@ -16,8 +16,6 @@
     "test:sincemain": "pnpm test -- --changedSince=main",
     "test:sincemain:cov": "pnpm test:sincemain --coverage",
     "prepack": "pnpm build",
-    "postinstall": "pnpm copy-op-schemas",
-    "copy-op-schemas": "cp ./node_modules/@interledger/open-payments/dist/openapi/specs/auth-server.yaml ./node_modules/@interledger/open-payments/dist/openapi/specs/schemas.yaml ./src/openapi/specs/",
```
BlairCurrey, line 20: leftover from before submodules. shouldnt need this anymore I believe.

njlie (approved):

mkurapov (approved):

PR #3819, opened 2026-02-03
feat(backend): add confirmPartialIncomingPayment resolver

njlie:

- Adds the `confirmPartialIncomingPayment` graphql resolver

## Context

Fixes RAF-1186.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/graphql/resolvers/incoming_payment.ts
+++ b/packages/backend/src/graphql/resolvers/incoming_payment.ts
@@ -296,6 +296,8 @@ export const getIncomingPaymentTenant: IncomingPaymentResolvers<TenantedApolloCo
     return tenantToGraphQl(tenant)
   }
 
+const PARTIAL_PAYMENT_DECISION_PREFIX = 'partial_payment_decision'
+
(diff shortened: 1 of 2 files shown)


packages/backend/src/graphql/resolvers/incoming_payment.ts
```diff
@@ -293,6 +296,72 @@ export const getIncomingPaymentTenant: IncomingPaymentResolvers<TenantedApolloCo
...
+        errorToMessage[IncomingPaymentError.InvalidState],
+        {
+          extensions: {
+            code: errorToCode[IncomingPaymentError.InvalidState]
+          }
+        }
+      )
+
+    const redis = await ctx.container.use('redis')
+    const cacheKey = `kyc_decision:${input.incomingPaymentId}:${input.partialIncomingPaymentId}`
```
sanducb: I think we could rename this to `ase_decision` instead of `kyc_decision` to make it more general. WDYT?

njlie: Yeah I think that fits better, actually. I suppose we could make it even _more_ general and call the prefix `payment_decision` or `partial_payment_decision` since we don't need to know _where_ the decision came from, just _what_ the decision is.

Issue #3830, opened 2026-02-11
Support JWK in Auth Server


BlairCurrey:
- [ ] Data layer: migration to make client optional, new jsonb jwk field. update objection model.
- [ ] Service layer: if grant is not interactive, can create with client OR jwk (not both, not neither). client (wallet address url) is required for consent interaction. validate json schema.
- [ ] Middleware: update signature middleware to use JWK if available instead of assuming wallet address on client
- [ ] Token introspection endpoint: `grantToTokenInfo` should return client or jwk

PR #3842, opened 2026-02-21
feat: support jwk directed identity in auth server

BlairCurrey:
fixes https://github.com/interledger/rafiki/issues/3830

The backend build failures will be resolved in another issue (all ci checks should pass in the final PR to integrate both backend/auth changes).

### DIFF
--- a/packages/auth/src/grant/service.test.ts
+++ b/packages/auth/src/grant/service.test.ts
@@ -237,7 +237,7 @@ describe('Grant Service', (): void => {
         const testJwk = generateTestJwk()
         const grantRequest: CreateGrantInput = {
           jwk: testJwk,
-          ...BASE_GRANT_REQUEST,
+          interact: BASE_GRANT_REQUEST.interact,
           access_token: {
             access: [
               {
--- a/packages/auth/src/grant/utils.test.ts
+++ b/packages/auth/src/grant/utils.test.ts
@@ -1,7 +1,7 @@
 import { AccessRequest } from '../access/types'
 import { IAppConfig } from '../config/app'
 import { CreateGrantInput } from './service'
-import { canSkipInteraction, parseRawClientField } from './utils'
+import { canSkipInteraction, parseRawClientField, RawClientField } from './utils'
 import { AccessAction } from '@interledger/open-payments'
 import { JWK } from 'token-introspection'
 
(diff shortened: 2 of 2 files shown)


packages/auth/src/accessToken/routes.test.ts
```diff
@@ -366,10 +367,99 @@ describe('Access Token Routes', (): void => {
       expect(ctx.body).toMatchObject({
         active: true,
         access: [],
-        client: grant.client,
+        client: { walletAddress: grant.client },
         grant: grant.id
       })
     })
+
+    test('Throws 500 error when introspecting token for grant with no client identity', async (): Promise<void> => {
```
BlairCurrey, line 375: arguably dont need this test since it purely defensive but included for comprehensiveness. shouldnt happen because we are validating grant has client or jwk on write side but no way to make that invalid state completely unrepresentable.

packages/auth/src/signature/middleware.test.ts
```diff
@@ -408,6 +471,84 @@ describe('Signature Service', (): void => {
       })
     })
 
+    test('grant continuation middleware fails if grant has no client identity', async (): Promise<void> => {
```
BlairCurrey, line 474: same as this https://github.com/interledger/rafiki/pull/3842/changes#r2837087665

packages/auth/src/grant/routes.ts
BlairCurrey, line 1: Not about my changes specifically, but just an observation. I found this route was doing a lot of work that should maybe be handled in the service layer. Right now both the services and the route layer care about things like what kind of client identifier (jwk or wallet address), if its interactive, etc. Lots of orchestration.

njlie: We can capture this in an issue that can live in the backlog for now. Might even be worth a "good first issue" tag?

packages/auth/src/signature/middleware.ts
```diff
@@ -100,7 +112,18 @@ export async function grantContinueHttpsigMiddleware(
...
+  try {
+    identity = getGrantClientIdentity(grant)
+  } catch {
+    throw new GNAPServerRouteError(
+      500,
+      GNAPErrorCode.RequestDenied,
+      'internal server error'
+    )
+  }
+
```
mkurapov, line 125: Thinking out loud: would an Open Payments client call the continue route after getting a non-interactive grant using a JWK? In general, what does it mean to continue a non-interactive grant?

njlie: _Technically_ it is possible to continue a non-interactive grant, but that does not happen in Open Payments. In general in GNAP, continuing an non-interactive grant means that the Authorization Server was expecting some other non-interactive step to be completed by the client before continuing a grant. Perhaps we could remove this.

PR #3895, opened 2026-04-14
fix: fix cron job for monthly sync discussion workflow

oana-lolea:

- Run job each Wednesday
- Checking if it's the first Wednesday of the month. If not, we skip creating the disscution.

## Context

As GitHub actions uses an older standard for `cron`, it does not support setting the first Wednesday of every month.
So I set the `cron` to start on every Wednesday, and as a first step of the workflow we check if it's the first Wednesday of that month. If not, we skip the workflow.

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/.github/workflows/create_monthly_sync.yml
+++ b/.github/workflows/create_monthly_sync.yml
@@ -1,9 +1,9 @@
 name: Create Monthly Sync discussion
 
-# Each Wednesday of every month at 12AM UTC
+# First Wednesday of every month at 12AM UTC
 on:
     schedule:
-      - cron: "0 0 * * 3"
+        - cron: "0 0 1-7 * 3"
 
 jobs:
     create-discussion:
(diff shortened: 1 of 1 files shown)


.github/workflows/create_monthly_sync.yml
```diff
@@ -1,58 +1,66 @@
 name: Create Monthly Sync discussion
 
-# First Wednesday of every month at 12AM UTC
+# Each Wednesday of every month at 12AM UTC
 on:
     schedule:
-        - cron: "0 0 1-7 * 3"
+      - cron: "0 0 * * 3"
```
oana-lolea: Scheduled the job to run on every Wednesday, so that below we can check if it's the first Wednesday.

.github/workflows/create_monthly_sync.yml
```diff
@@ -3,14 +3,23 @@ name: Create Monthly Sync discussion
...
         steps:
+            - name: Check if first Wednesday
+              id: check
+              run: |
+                DOM=$(date +%d)
+                if [ "$DOM" -gt 7 ]; then
+                  echo "Not the first Wednesday of the month, skipping"
+                  exit 0
+                fi
+
```
oana-lolea, line 22: Skipping if it's not the first Wednesday of the month

sabineschaller (approved):
Looks good, thank you!

PR #3888, opened 2026-04-06
feat(backend, mock-ase): encrypted data exchange

sanducb:
## Changes proposed in this pull request

This PR adds support for sending encrypted additional data over ILP on which the receiving ASE can action upon i.e. approve/reject a specific payment (e.g. if AML/KYC checks were not successful).

## Context

Closes Rafiki Encrypted Data Exchange [project](https://linear.app/interledger/project/rafiki-encrypted-data-exchange-c1cf420c7e5c/issues)

## Checklist

- [ ] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [ ] Make sure that all checks pass
- [ ] Bruno collection updated (if necessary)
- [ ] Documentation issue created with `user-docs` label (if necessary)
- [ ] OpenAPI specs updated (if necessary)

PR #3911, opened 2026-04-23
chore: minor updates for encrypted data exchange feature

mkurapov:

- Renamed `dataToTransmit` to `dataFromSender` when dealing with exchanged data on recipient side
- Updating [how we handle](https://github.com/interledger/rafiki/pull/3911#discussion_r3132996236) timed out response from ASE 
- Updated guide for testing encrypted data exchange

## Context

Related to RAF-1221

## Checklist

- [x] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/service.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.ts
@@ -82,7 +82,7 @@ export interface IncomingPaymentService
   processPartialPayment(
     id: string,
     options: {
-      dataToTransmit: string
+      dataFromSender: string
       partialIncomingPaymentId: string
       expiresAt: Date
     }
@@ -570,7 +570,7 @@ async function processPartialPayment(
   deps: ServiceDependencies,
   id: string,
   options: {
-    dataToTransmit: string
+    dataFromSender: string
     partialIncomingPaymentId: string
     expiresAt: Date
   }
@@ -590,10 +590,10 @@ async function processPartialPayment(
     data: {
       ...incomingPayment.toData(0n),
       partialIncomingPaymentId: options.partialIncomingPaymentId,
-      dataToTransmit:
-        options.dataToTransmit && config.dbEncryptionSecret
-          ? encryptDbData(options.dataToTransmit, config.dbEncryptionSecret)
-          : options.dataToTransmit
+      dataFromSender:
+        options.dataFromSender && config.dbEncryptionSecret
+          ? encryptDbData(options.dataFromSender, config.dbEncryptionSecret)
+          : options.dataFromSender
     },
     tenantId: incomingPayment.tenantId,
     webhooks: finalizeWebhookRecipients(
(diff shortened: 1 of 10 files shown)


packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -82,7 +82,7 @@ export interface IncomingPaymentService
   processPartialPayment(
     id: string,
     options: {
-      dataToTransmit: string
+      dataFromSender: string
```
mkurapov, line 85: Thought it made sense to rename `dataToTransmit` to `dataFromSender` when dealing with this data on the recepient's side

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -693,12 +693,12 @@ function getPartialPaymentDecisionCacheKey(
 }
 
 interface PartialPaymentDecision {
-  success?: boolean
+  success: boolean
```
mkurapov, line 696: See https://github.com/interledger/rafiki/pull/3911/changes#r3132996236

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
```diff
@@ -35,20 +35,20 @@ export function createPartialPaymentDecisionMiddleware(): ILPMiddleware {
...
         incomingPaymentId,
         {
-          dataToTransmit: additionalData,
+          dataFromSender: additionalData,
           partialIncomingPaymentId: uuid(),
           expiresAt: prepare.expiresAt
         }
       )
 
-      if (decision?.success !== false) {
```
mkurapov, line 51: I made `decision.success` a required field. While testing I noticed that because we were not returning `success` in the `processPartialPayment` function when we time out waiting for a response from the ASE, we actually ended up completing the payment correctly here, since decision `decision?.success !== false` was truthy

mkurapov:
One general comment @njlie @sanducb, should we hide `dataFromSender` field when we are resolving the webhook events in the Admin API/UI?


Screenshot 2026-04-23 at 20 42 17

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
mkurapov: Actually, based on original discussion https://github.com/interledger/rafiki/pull/3888#discussion_r3094213725, I see why we had this before. I'm trying to reason through this. Does it make sense to approve the payment if we don't get a response from the ASE? @sanducb

sanducb:
> One general comment @njlie @sanducb, should we hide dataFromSender field when we are resolving the webhook events in the Admin API/UI?

I think for now we can leave it as is and modify it in the future based on any external requirements

packages/backend/src/payment-method/ilp/connector/core/middleware/partial-payment-decision.ts
sanducb: Thanks for adding the decision success default to false, that is more readable 👍 I think we should only allow packets if `decision.success` is true like we currently have it

sanducb:
Looks good, thank you @mkurapov! 
I think it's ready to merge once [this](https://github.com/interledgerjs/interledgerjs/pull/494) Pay library fix gets merged, the package published and we bump the version

sanducb (approved):

PR #3919, opened 2026-05-06
feat: prep for release v2.4.0-beta

mkurapov:

- Address axios vulnerabilities (CVE-2026-42033, CVE-2026-42035, CVE-2026-42043, CVE-2026-42264)
- Generate docs for updated GraphQL documentation

## Context

Fixes ROP-36

## Checklist

- [ ] Tests added/updated


BlairCurrey (approved):
Looks ok just had some non-blocking questions.

Why are the docs regenerated? If there are content differences I think this means the Spanish ones are out of date - is that accounted for anywhere (task to update?)

Whats the reason for doing the axios override? I realize this override pre-exists the PR and we are just updating it - I guess I'm asking more for my general knowledge. Is there a transitive dependency or something we are trying to control?

mkurapov:
> Why are the docs regenerated? If there are content differences I think this means the Spanish ones are out of date - is that accounted for anywhere (task to update?)

They do not automatically get built from the GraphQL schemas, so I built them manually. I will updating the links for Spanish one in a separate issue.

> Whats the reason for doing the axios override? 

This is to fix the vulnerabilities listed in the description, making sure that the minimum version across all of the packages in the repo is an updated, non-vulnerable one via `overrides`

BlairCurrey:
> This is to fix the vulnerabilities listed in the description, making sure that the minimum version across all of the packages in the repo is an updated, non-vulnerable one via `overrides`

What I'm wondering is why it's an override and not just bumping the individual package versions.

My understanding of overrides was that they were for forcing versions of packages we can't directly control ourselves. And ideally temporarily. But maybe we are intentionally using it to ensure the same version is used across packages? Im not sure that's a good idea. Pnpm has a catalog feature that can handle this and in their docs for it they directly state this would be misusing overrides: https://github.com/pnpm/rfcs/blob/main/text/0001-catalogs.md#comparison-to-overridesresolutions

> Using this feature should be considered a hack in most cases, something that is done temporarily while waiting for a bug to be fixed, or to avoid excessive duplication caused by an overly strict meta-dependency specifier.

The original override was `"axios@>=0.8.1 <0.28.0": ">=0.28.0"`. This preserves the option to use pre `<0.28.0`. If we are mandating everything is >= some version then I dont see the point of the override vs. pinning everything above the vulnerability. Locally it appears that pinning all our packages directly resolves the audit warning (using `pnpm audit`, not trivy mind you).

Frankly IDK if we even need catalogs. We can just maintain the version directly across pacakges imo. We can still easily upgrade simultaneously with `pnpm up -r axios`.

mkurapov:
@BlairCurrey good point - for some reason I disregarded that we have it as a direct dependency, even though we do in a few of the packages.

We should to do both: upgrade/pin the version across the dependencies, and also keep the override because there a few dependencies, mostly dev that have axios as a dependency themselves.

Catalogs might be a great option for us in the rafiki-v2 repo, given the amount of packages we have.

mkurapov:
Made PR for upgrading and pinning axios version ☝️

PR #3940, opened 2026-06-10
feat(backend): soft delete expired incoming payments with no funds

takundachirema:

- Add deletedAt to the incoming payments model
- In the IncomingPaymentService.handleExpired method, set deletedAt field instead of deleting the record and set the state to expired and processAt to null so that it is not processed by another worker.
- Updated IncomingPaymentData type so that IncomingPaymentEvents have all metadata for WebHook

## Context

Fixes RAF-1224

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/payment/incoming/service.ts
+++ b/packages/backend/src/open_payments/payment/incoming/service.ts
@@ -155,9 +155,9 @@ async function updateIncomingPayment(
   deps: ServiceDependencies,
   options: UpdateOptions
 ): Promise<IncomingPayment | IncomingPaymentError> {
-  const incomingPayment = await IncomingPayment.query(
-    deps.knex
-  ).patchAndFetchById(options.id, options)
+  const incomingPayment = await IncomingPayment.query(deps.knex)
+    .patchAndFetchById(options.id, options)
+    .whereNull('deletedAt')
   if (incomingPayment) {
     const asset = await deps.assetService.get(incomingPayment.assetId)
     if (asset) incomingPayment.asset = asset
@@ -372,7 +372,10 @@ async function handleExpired(
       processAt: new Date()
     })
   } else {
-    deps.logger.debug({ amountReceived }, 'deleting expired incoming payment')
+    deps.logger.debug(
+      { amountReceived },
+      'soft deleting expired incoming payment'
+    )
     await incomingPayment.$query(deps.knex).patch({
       state: IncomingPaymentState.Expired,
       deletedAt: new Date(),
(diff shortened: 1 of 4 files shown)


packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -371,7 +373,11 @@ async function handleExpired(
     })
   } else {
     deps.logger.debug({ amountReceived }, 'deleting expired incoming payment')
```
```suggestion
    deps.logger.debug({ amountReceived }, 'soft deleting expired incoming payment')
```

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -745,7 +751,7 @@ async function getPage(
 ): Promise<IncomingPayment[]> {
   const { filter, pagination, sortOrder, tenantId, walletAddressId, client } =
     options ?? {}
-  const query = IncomingPayment.query(deps.knex)
+  const query = IncomingPayment.query(deps.knex).whereNull('deletedAt')
```
mkurapov, line 758: Let's add a small test case for this in the `getPage -> filters` test section

packages/backend/src/open_payments/payment/incoming/service.ts
```diff
@@ -134,7 +134,9 @@ async function getIncomingPayment(
   deps: ServiceDependencies,
   options: GetOptions
 ): Promise<IncomingPayment | undefined> {
-  const incomingPayment = await IncomingPayment.query(deps.knex).get(options)
+  const incomingPayment = await IncomingPayment.query(deps.knex)
+    .get(options)
+    .whereNull('deletedAt')
```
mkurapov, line 139: Let's also add this `deletedAt` query in `updateIncomingPayment` and `completeIncomingPayment` functions as well (to make sure we don't end up updating or completing an already deleted incoming payment)

mkurapov (commented):
Looking good, just two comments

packages/backend/src/open_payments/payment/incoming/service.ts
takundachirema: I had added a getPage test to the 'Soft deletes an expired incoming payment with no money' test case, starting on line 817. Would you like me to put that in getPage -> filters test section instead, or I can add another test?

mkurapov: Ideally we should have these tests a bit more self-contained and scoped to a particular function in the future, e.g. `describe('complete incoming payment') -> test('returns error for deleted incoming payment')` but this is sufficient and is testing everything that we want

mkurapov (approved):

PR #3944, opened 2026-06-23
fix(backend): Improve exchange rate handling

takundachirema:

- Removed logic in the assets service that checks for exchange rate URL config
- Added rates errors.ts and updated rates service.ts to use them
  - The error codes will distinguish between missing exchange rate URL and network or other errors when fetching exchange rates
- Updated ilp payment-methods service so that it does not throw when sender and receiver wallet asset codes are the same
- Added error handling for rate errors in quote service.ts and ilp payment handler service.ts
  - Instead of throwing a 500 error with no details, a detailed message will be shown with the cause of quote error being the exchange rate service
- Added error handling for rate errors in local payment handler service.ts
  - Check for RatesError type in local payment handler when doing currency conversion

## Context

Fixes ROP-52

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/backend/src/open_payments/payment/outgoing/errors.ts
+++ b/packages/backend/src/open_payments/payment/outgoing/errors.ts
@@ -19,7 +19,8 @@ export enum OutgoingPaymentError {
   NegativeReceiveAmount = 'NegativeReceiveAmount',
   InvalidReceiver = 'InvalidReceiver',
   InvalidInterval = 'InvalidInterval',
-  OnlyOneGrantAmountAllowed = 'OnlyOneGrantAmountAllowed'
+  OnlyOneGrantAmountAllowed = 'OnlyOneGrantAmountAllowed',
+  CouldNotFetchRates = 'CouldNotFetchRates'
 }
 
 export const quoteErrorToOutgoingPaymentError: Record<
@@ -33,7 +34,8 @@ export const quoteErrorToOutgoingPaymentError: Record<
   [QuoteErrorCode.InactiveWalletAddress]:
     OutgoingPaymentError.InactiveWalletAddress,
   [QuoteErrorCode.NonPositiveReceiveAmount]:
-    OutgoingPaymentError.NegativeReceiveAmount
+    OutgoingPaymentError.NegativeReceiveAmount,
+  [QuoteErrorCode.CouldNotFetchRates]: OutgoingPaymentError.CouldNotFetchRates
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
@@ -54,7 +56,8 @@ export const errorToHTTPCode: {
   [OutgoingPaymentError.NegativeReceiveAmount]: 400,
   [OutgoingPaymentError.InvalidReceiver]: 400,
   [OutgoingPaymentError.InvalidInterval]: 500,
-  [OutgoingPaymentError.OnlyOneGrantAmountAllowed]: 500
+  [OutgoingPaymentError.OnlyOneGrantAmountAllowed]: 500,
+  [OutgoingPaymentError.CouldNotFetchRates]: 500
 }
 
 export const errorToCode: {
@@ -72,7 +75,9 @@ export const errorToCode: {
   [OutgoingPaymentError.InvalidReceiver]: GraphQLErrorCode.BadUserInput,
   [OutgoingPaymentError.InvalidInterval]: GraphQLErrorCode.InternalServerError,
   [OutgoingPaymentError.OnlyOneGrantAmountAllowed]:
-    GraphQLErrorCode.BadUserInput
+    GraphQLErrorCode.BadUserInput,
+  [OutgoingPaymentError.CouldNotFetchRates]:
+    GraphQLErrorCode.InternalServerError
 }
 
 export const errorToMessage: {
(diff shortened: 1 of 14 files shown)


packages/backend/src/open_payments/quote/service.ts
```diff
@@ -218,30 +218,43 @@ async function createQuote(
...
         }
 
-        stopTimer()
-        return new QuoteError(QuoteErrorCode.NonPositiveReceiveAmount, details)
+        if (err.code === PaymentMethodHandlerErrorCode.CouldNotGetRates) {
+          stopTimer()
+          return new QuoteError(QuoteErrorCode.CouldNotFetchRates, {
+            description: err.description,
+            details: err.details
+          })
```
mkurapov: The second argument, `details` get bubbled up to the Open Payments client. I would say since we don't want to show any internal message other than 'could not fetch rates' for this error, we can simply do: 
```suggestion
          return new QuoteError(QuoteErrorCode.CouldNotFetchRates)
```

Having the error logged in the rates service internally is sufficient for this

packages/backend/src/open_payments/payment/outgoing/errors.ts
```diff
@@ -90,7 +95,8 @@ export const errorToMessage: {
   [OutgoingPaymentError.InvalidReceiver]: 'invalid receiver',
   [OutgoingPaymentError.InvalidInterval]: 'invalid interval',
   [OutgoingPaymentError.OnlyOneGrantAmountAllowed]:
-    'only one of receiveAmount or debitAmount allowed'
+    'only one of receiveAmount or debitAmount allowed',
+  [OutgoingPaymentError.CouldNotFetchRates]: 'could not fetch exchange rates'
```
mkurapov: Because this is the error that gets sent to the OP client, we could say something like "exchange rate conversion currently not supported for these assets".

packages/backend/src/open_payments/quote/service.ts
takundachirema: Oh yes, I see your point. Can I update the details so that they are client friendly and they don't show server side info which is not meant for the client? This also links to you second comment.

packages/backend/src/open_payments/payment/outgoing/errors.ts
takundachirema: I was just thinking it might be that the reason for the exchange rates not being fetched is something other than the exchange URL not being set, for example a network error. I could change the details so that when there is no exchange rate URL it will say 'Not supported', otherwise it'll say some other error which is client friendly without showing server info.

mkurapov: Yes, as long as we have a good client message without specifying any internal workings, that works!

packages/backend/src/open_payments/quote/service.test.ts
```diff
@@ -584,6 +584,40 @@ describe('QuoteService', (): void => {
       })
     })
 
+    test('succeeds when exchange rates service fails but receiver wallet asset code matches sender', async (): Promise<void> => {
```
mkurapov: maybe this tests belongs in `packages/backend/src/payment-method/ilp/service.test.ts` test since that is where the actual logic lies

packages/backend/src/open_payments/quote/service.ts
mkurapov:
```suggestion
          return new QuoteError(QuoteErrorCode.CouldNotFetchRates)
```
from previous review, let's not return these details object back to the client. I think the only error we should do that for is the `QuoteNonPositiveReceiveAmount` as we are currently doing.

packages/backend/src/open_payments/quote/errors.ts
```diff
@@ -49,5 +52,6 @@ export const errorToMessage: {
   [QuoteErrorCode.InvalidAmount]: 'invalid amount',
   [QuoteErrorCode.InvalidReceiver]: 'invalid receiver',
   [QuoteErrorCode.InactiveWalletAddress]: 'inactive wallet address',
-  [QuoteErrorCode.NonPositiveReceiveAmount]: 'non-positive receive amount'
+  [QuoteErrorCode.NonPositiveReceiveAmount]: 'non-positive receive amount',
+  [QuoteErrorCode.CouldNotFetchRates]: 'could not fetch exchange rates'
```
mkurapov: Related to the comments from the previous review:
since this error message (and the outgoing payment [one](https://github.com/interledger/rafiki/pull/3944/changes#diff-b57056a22f35d41fb8d7faa6cb1a085ae2af3f7af4f6d13f552ad665b7b1f3d5R99)) is the one that gets sent to the end Open Payments client, we could say something like "could not convert between assets", since that is a bit more clearer IMO. 

No changes required to the RatesErrors, they are good for internal logging :+1:

takundachirema: Done 👍

packages/backend/src/open_payments/quote/service.test.ts
takundachirema: Done thanks, missed that.

packages/backend/src/open_payments/quote/service.ts
takundachirema: Done 👍

mkurapov (approved):

PR #3957, opened 2026-08-03
fix(backend): reduce test flakiness in Redis-backed and concurrency tests

cozminu:
## Summary
- Give each Jest worker its own Redis logical database (by `JEST_WORKER_ID`) so parallel workers don't stomp on each other's state via `flushall`
- Bump Redis's database count to 64 since the default 16 isn't enough on high core count machines
- Fix a flaky assertion in the concurrent asset creation test that assumed a specific request would win the idempotency lock race

## Test plan
- [x] Ran backend test suite locally with parallel workers


mkurapov (approved):
Thanks for fixing this, Cozmin!

PR #3961, opened 2026-08-06
fix(frontend): multiple a11y fixes

sidvishnoi:
## Changes proposed in this pull request

Multiple fixes observed in Rafiki Admin Accessibility Report. 
Review commit-by-commit.

## Context

https://docs.google.com/document/d/17svFl6B1g_189D9kt4NAccPt3pZxRmcJxJNOUMwjcsM/edit

## Checklist

- [ ] Related issues linked using `fixes #number`
- [ ] Tests added/updated
- [x] Make sure that all checks pass
- [ ] Bruno collection updated (if necessary)
- [ ] Documentation issue created with `user-docs` label (if necessary)
- [ ] OpenAPI specs updated (if necessary)

PR #3965, opened 2026-08-11
fix(frontend): address many color contrast a11y issues

sidvishnoi:

Address color contrast issues (most of them).

- RadixUI orange theme does not pass WCAG guidelines in general. They use APCA (https://github.com/radix-ui/colors/issues/41#issuecomment-1927892117; that better algorithm will be part of WCAG3 later).
   - To pass compliance, change base theme from `orange` to `amber`, so we can pass most color issues in least effort (and not define own palette right now).
- Update some badges to use `outline` variant. The `soft` doesn't pass WCAG contrast guidelines with amber either.
- For some buttons, using `classic` variant - that uses shadows and adds contrast.

I could not think of a better style for Next/Prev pagination links - left them as is for now.

## Context

Follow up of https://github.com/interledger/rafiki/pull/3961

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/frontend/app/components/ApiCredentialsForm.tsx
+++ b/packages/frontend/app/components/ApiCredentialsForm.tsx
@@ -54,7 +54,12 @@ export const ApiCredentialsForm = ({
         <Form method='post' action='/api/set-credentials' className='space-y-4'>
           <p className='text-green-600 mb-4'>✓ API credentials configured</p>
           <input hidden readOnly name='intent' value='clear' />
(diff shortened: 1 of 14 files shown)


sidvishnoi:
Example changes:

| Before | After |
| --- | --- |
| image | image |
| image | image |

JoblersTune:
Can we keep the accent color the same for the payment filter
image
Same for the events filter for webhooks

JoblersTune (approved):

mkurapov (approved):

PR #3967, opened 2026-08-11
fix(frontend/Snackbar): announce correctly, pause timer on interaction

sidvishnoi:

Multiple a11y fixes to Snackbar:
- Add accessible role and make sure it's reliably announced
- Ensure hide timer could be paused if user hovers (or focuses) over notification (it should stay, not disappear immediately if user is interacting). Also, increase timeout from 2s to 5s.
- Move Snackbar to start in tab-order, so one doesn't have to tab till end to interact with it (by that time, notification goes away)

In future, we can get rid of `SnackbarAnnouncer` when `ariaNotify` is better supported.

## Context

Follow up of https://github.com/interledger/rafiki/pull/3961

## Checklist

- [ ] Tests added/updated

### DIFF
--- a/packages/frontend/app/components/Snackbar.tsx
+++ b/packages/frontend/app/components/Snackbar.tsx
@@ -1,7 +1,7 @@
 import { Transition } from '@headlessui/react'
 import { cx } from 'class-variance-authority'
 import type { FC } from 'react'
-import { Fragment, useEffect } from 'react'
+import { Fragment, useEffect, useRef } from 'react'
 import { type Message } from '~/lib/message.server'
 import { CheckCircleSolid, XIcon, XCircleSolid } from './icons'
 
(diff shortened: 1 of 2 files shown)


packages/frontend/app/components/Snackbar.tsx
```diff
@@ -19,22 +19,50 @@ interface SnackbarProps {
...
+// Live regions are only reliably announced by screen readers when the browser
+// already has the node registered and then content its changes.
+const SnackbarAnnouncer: FC<Pick<SnackbarProps, 'message' | 'show'>> = ({
+  message,
+  show = false
+}) => (
+  <div
+    role={message?.type === 'error' ? 'alert' : 'status'}
+    aria-live={message?.type === 'error' ? 'assertive' : 'polite'}
+    aria-atomic='true'
```
sidvishnoi, line 38: This is technically redundant, but practically useful. With Firefox in particular.

packages/frontend/app/root.tsx
```diff
@@ -134,6 +134,13 @@ export default function App() {
           >
             Skip to main content
           </a>
+          <Snackbar
+            id='snackbar'
+            onClose={() => setSnackbarOpen(false)}
+            show={snackbarOpen}
+            message={message}
+            dismissAfter={200000}
```
JoblersTune: This is now 200 seconds, right? Surely 20 seconds is sufficient? This is just informational.

sidvishnoi: Oops, leftover from testing.

sidvishnoi:
```suggestion
            dismissAfter={2000}
```

sidvishnoi: 2000 was default. Maybe should be 5s?

JoblersTune: I think 5 seconds would be best

packages/frontend/app/root.tsx
```diff
@@ -134,6 +134,13 @@ export default function App() {
           >
             Skip to main content
           </a>
+          <Snackbar
+            id='snackbar'
+            onClose={() => setSnackbarOpen(false)}
+            show={snackbarOpen}
+            message={message}
+            dismissAfter={2000}
```
sidvishnoi:
```suggestion
            dismissAfter={5000}
```

JoblersTune (approved):

PR #3974, opened 2026-09-04
chore(deps): upgrade documentation dependencies for Astro 7 (DOCSDEV-75)

Infi-Knight:
## Summary

Move `packages/documentation` to Astro 7 and Starlight 0.42, part of the DOCSDEV-75 cycle.

**Only `packages/documentation/` changes, plus the lockfile.** 

### 1. Markdown pipeline moves to `unified()`

Astro 7 replaces the remark and rehype pipeline with Satteri, a Rust processor. Satteri cannot
run remark or rehype plugins. This package needs four of them, so the config now wraps them in
the `unified()` escape hatch from `@astrojs/markdown-remark`. 

### 2. Remove an unused `Badge` import

`signing-admin-api-requests.mdx` imported `Badge` from the design system. **The design system
has never exported `Badge`.** Astro 6 tolerated the missing export. Astro 7 fails the build with
`[MISSING_EXPORT]`. `Badge` was never used in the file, so the import is gone.

### 3. Remove six dead `client:load` directives

Two on `MermaidWrapper` and four on `Tooltip`, across both locales and both doc versions. Both
are Astro components, so a hydration directive does nothing on them. All six only produced a
build warning.

This is not an Astro 7 regression. Astro 6.3.8 emits the identical warning.

**The rendered HTML does not change.** I rebuilt and diffed all six pages. Two are byte
identical. The other four differ only in the random UUID that `Tooltip.astro` generates on
every build, and they are byte identical once that UUID is normalized.

### 4. Turn on the accessibility rules

`eslint-plugin-astro` 3 exposes the `jsx-a11y` rule set, so `eslint.config.js` now extends
`flat/jsx-a11y-recommended`.

**Result: 34 rules active, 0 findings.** The rules are confirmed to run against `.astro` files,
so the zero is a real result.

## Verification

Node 24.18.1, pnpm 10.33.0.

| Check | Result |
| --- | --- |
| `pnpm --filter documentation run build:docs` | Passes. 174 pages |
| `starlight-links-validator` | All internal links valid |
| `pnpm lint:docs` | Passes, `--max-warnings=0` |
| `pnpm checks` at root | Passes |
| `pnpm --filter documentation run build:docs:graphql` | Passes. spectaql stays at 3.0.9 |
| Heading anchors and table of contents | Work. 28 wrapped anchors on the accounting page |
| Math through rehype-mathjax | Renders in both locales |
| `/llms.txt`, `/llms-full.txt`, `/llms-small.txt` | All generate |
| Pagefind search | Works |
| **All 35 Mermaid diagrams** | **35 of 35 render, zero syntax errors** |
| Mermaid view and download buttons | Both work |
| Header, PageSidebar and Footer overrides | All three render. `Astro.locals.starlightRoute` still works |
| `expressiveCode` border overrides | Identical to production |
| Version switcher and `v1-beta` pages | Work. The sidebar stays scoped to the version |
| Both locales | Work |
| Console errors | Zero across 8 pages |
| `compressHTML` default change to `'jsx'` | **No visible effect.** Rendered text is identical to production on 3 pages, so the option is left at its new default |

## One thing to know about the lockfile

`packages/frontend` and `localenv/mock-account-servicing-entity` each changed exactly one line.
The `@remix-run/dev@2.17.5` peer resolution key gained `(lightningcss@1.33.0)`. Astro 7 and
Vite 8 pull `lightningcss` into the workspace, and `vite@6.4.1` declares it an optional peer,
so pnpm folds it into the key.

- **The installed version of `@remix-run/dev` does not change.** It stays 2.17.5.
- Vite uses lightningcss only when `css.transformer: 'lightningcss'` is set. No config in
  either package sets it.

This leak cannot be avoided without adding root `pnpm.overrides`, which is out of scope here.

## Test plan

- [ ] CI passes
- [ ] Deploy preview renders, including diagrams, both locales and both doc versions


Infi-Knight:
## A note on the red CI checks, for the reviewer

`gh pr checks` reports 14 failures on this PR. **That count is misleading, because it counts a
cancelled job as a failure.** The run itself tells a different story, and it matches `main`.

| Run | Real scan failures | Cancelled by fail-fast | Succeeded |
| --- | --- | --- | --- |
| This PR, run `33909303497` | 2 | 18 | 0 |
| `main`, run `33805877155` | 2 | 18 | 0 |

**The pattern is identical to `main`.** Which two matrix legs lose the race before fail-fast
cancels the rest is not deterministic, so the named jobs differ between runs.
`Docker Grype Scan (linux/amd64, amd64, point-of-sale)` fails in both.

I read the failing log instead of guessing. The findings are Alpine base image CVEs in
`libcrypto3` and `libssl3`, version `3.5.7-r0` with the fix in `3.5.8-r0`, plus the
`brace-expansion` npm package. The scans run with `--fail-on high --only-fixed`, so any
unpatched base image fails them.

**This PR touches no Dockerfile and no runtime dependency of any scanned package.** The scanned
images are auth, backend, frontend, card-service and point-of-sale. The documentation package is
not containerized.

### What did pass

- **`prerequisite`**, which runs `pnpm checks` and `pnpm lint:docs`
- **The Netlify deploy preview.** This matters most here, because **no GitHub Actions job runs
  `astro build` for the docs package.** The deploy preview is the real build gate.

I verified the deploy preview itself, not only the build status:

- **All 35 Mermaid diagrams render**, across 8 pages, both locales and both doc versions
- The Mermaid view and download buttons both work
- The `PageSidebar` override still trims the table of contents on the 3 pages it targets, and
  leaves other pages alone
- Pagefind search works
- Zero console errors across 8 pages
- **Rendered text is identical to production on 5 pages**, so the `compressHTML` default change
  to `'jsx'` has no visible effect

One thing worth your attention is in the description: the lockfile changed one line for
`packages/frontend` and `localenv/mock-account-servicing-entity`. It is a peer resolution key
gaining `lightningcss`, and the installed `@remix-run/dev` version does not move. I proved it
inert by running `typecheck` and `build` on both packages. `lightningcss` does not appear in the
scan findings either.

Anca2022:
Same issue you already flagged on WM:
image

This PR also needs to wait for `interledger/docs-design-system` v0.14.2 before merging please 😄

Anca2022 (dismissed):
Just need to update `interledger/docs-design-system` to v0.14.2 for the stylistic fix you already flagged. 
The rest looks good! 😁

mkurapov (approved):

